PR #12, opened 2018-08-30
Changed default csv directory permissions to 750 from 777.

nevermindewe:
csv directory should probably not be world-writable.

Next steps are to make this configurable with the plugin.


baallan:
Good catch. I have developed a patch that will make this configurable for the next release. It will be submitted to the ogc gitlab instance, then migrate to github.

baallan:
This has been fixed in ogc MR 1019,1020.

PR #41, opened 2019-04-17
New Lustre client plugin

morrone:
This lustre client plugin is expected to work with at least lustre 2.8 and
lustre 2.10.  In lustre 2.12 the gathered stats are no longer retrievable
without enhanced privileges, so this will not work with lustre 2.12.0.

This plugin takes a different approach than the existing lustre plugin.
It focusses on the llite stats file and just a handful of additional
single-file values.

This plugin works without any configuration (and, in fact, ignores
any configuration).  It simply creates metric sets for any found
lustre client stats.

It supports creating metric sets dynamically as new lustre file
systems are mounted.  It would also remove old metric sets for
lustre file systems that have been unmounted, but that functionality
is disabled until ldms_set_delete() is fixed in a later ovis v4
release.


baallan:
Before we could cherry pick this back to ogc for v3, we need some changes. I'll list the ones that jump out at me without testing first. There may be others with runtime testing.

a) add in the ldms/configure.ac an --enable-llnl-lustre210 option and conditional build controls in llnl/Makefile.am (like: if ENABLE_LLNL_EDAC).
b) it needs to support the standard options for producer, instance, schema, jobid, componentid.
c) If I run this sampler at 1Hz, do I really need to rescan the input directory for every sample? or can I get away with rescanning for new ones every N seconds and simply don't update the set data for entries that disappear? N would be a configuration parameter with some more meaningful name than N; maybe 'rescan'.
d) what happens if /proc gets unmounted? are the  error codes checked on all the places they should be and trapped in that case? [yeah, it sounds silly, but it happens in production].

And it would be helpful if you could post a short slice of the data that gets produced (csv store).

morrone:
Thanks for the review!

> a) add in the ldms/configure.ac an --enable-llnl-lustre210 option and conditional build controls in llnl/Makefile.am (like: if ENABLE_LLNL_EDAC).

I can, but what is the use case for that?  There are no external dependencies in the plugins, so they can always be built.  configure options like that usually exist to control building of things that might have external dependencies, and therefore might build in some situations, but not in others.

> b) it needs to support the standard options for producer, instance, schema, jobid, componentid.

Not all of them make sense for this plugin, since the plugin is more dynamic than most.  

Producer I could certainly add as an option.  The default behavior of using the hostname will probably be fine for almost everyone, but it wouldn't hurt to allow the user to specify it.

Instance name is dynamically generated, so I don't know how that could be employed very sanely in this plugin.  "instance" might be interpreted as a prefix to be applied to the instance name that the plugin generates, but then the semantics would be different than in most of the other plugins.  I'm not sure that would be a good idea.

The schema name is intentionally the same everywhere.  It always describes the same data format, so the schema name is always the same.  I'm not sure that I see a use case for changing the schema name in this method of use.

Instance name is dynamically generated in this plugin, so I don't know how that could be employed very sanely.  "instance" might be interpreted as a prefix to be applied to the instance name that the plugin generates, but then the semantics would be different than in most of the other plugins.  I'm not sure that would be a good idea.

The jobid and componentid options seem specific to certain schemas rather than built in values like the metric set instance name.  Since my schema doesn't have them, does it make sense for this plugin to parse those options?  Must my schema contain them?

jobid doesn't seem like something people would generally use with this plugin.  I am expecting that the ldmsd and plugin would be long-lived through many jobs running on the system, so it is not clear to me why a statically assigned jobid at configuration time is useful.  Can you explain the use case there?

Ditto component id.  What is a "component" in respect to a lustre client metric set?

> c) If I run this sampler at 1Hz, do I really need to rescan the input directory for every sample? or can I get away with rescanning for new ones every N seconds and simply don't update the set data for entries that disappear? N would be a configuration parameter with some more meaningful name than N; maybe 'rescan'.

Yep, that is actually why I broke the refresh() function out cleanly, in case we wanted to do that at some point.  The compute time at 1Hz is inconsequential, so I figured it was premature optimization.  Is that going to be required for initial landing?

> d) what happens if /proc gets unmounted? are the error codes checked on all the places they should be and trapped in that case? [yeah, it sounds silly, but it happens in production.

It is my intent that it handle that just fine.  If you see something that I missed, let me know.

morrone:
Here's am example of the data produced from ldlm_ls:

```
  METADATA --------
    Producer Name : opal186
    Instance Name : opal186/lquake-ffff94f4c92fb000
      Schema Name : lustre_llite_general
             Size : 1944
     Metric Count : 38
               GN : 4
  DATA ------------
        Timestamp : Wed Apr 17 14:50:13 2019 -0700 [722717us]
         Duration : [0.000048s]
       Consistent : TRUE
             Size : 328
               GN : 120
  -----------------
M char[]     hostname                                   "opal186"
M char[]     fs_name                                    "lquake"
M char[]     llite                                      "lquake-ffff94f4c92fb000"
D u64        stats.dirty_pages_hits                     0
D u64        stats.dirty_pages_misses                   0
D u64        stats.read_bytes.sum                       1873435335
D u64        stats.write_bytes.sum                      25730754372
D u64        stats.brw_read.sum                         0
D u64        stats.brw_write.sum                        0
D u64        stats.ioctl                                1
D u64        stats.open                                 272400
D u64        stats.close                                272400
D u64        stats.mmap                                 0
D u64        stats.page_fault                           0
D u64        stats.page_mkwrite                         0
D u64        stats.seek                                 0
D u64        stats.fsync                                0
D u64        stats.readdir                              32328
D u64        stats.setattr                              258508
D u64        stats.truncate                             0
D u64        stats.flock                                0
D u64        stats.getattr                              46494868
D u64        stats.create                               0
D u64        stats.link                                 0
D u64        stats.unlink                               0
D u64        stats.symlink                              0
D u64        stats.mkdir                                2235
D u64        stats.rmdir                                0
D u64        stats.mknod                                0
D u64        stats.rename                               1
D u64        stats.statfs                               154
D u64        stats.alloc_inode                          277177
D u64        stats.setxattr                             1
D u64        stats.getxattr                             47017170
D u64        stats.getxattr_hits                        46625381
D u64        stats.listxattr                            0
D u64        stats.removexattr                          256276
D u64        stats.inode_permission                     2484808
```

morrone:
@baallan , I updated this pull request with some of the changes that you requested.

There is now a configuration option.  I called it "llnl-lustre".  This plugin is not intended to be for only lustre 2.10, so I didn't want to have the version in the configure string.

I added all of the options that you requested, although two of them I ignore. :)  I'm not sure what to do with "schema" (which is static) or "instance" (which is dynamically generated).

I didn't change the directory scan.  The dvs_sampler works the same way, so there is at least precedence for this approach.  If it really bothers you, I'll bow to your desire and add a refresh interval tomorrow.

Speaking of dvs_sampler, it looks like it too will accept an "instance" option and then silently ignores it.  So it doesn't look like I am alone in thinking it doesn't make sense as a required option.

dvs_sampler does allow overriding the schema name.  I don't quite see the rationale in letting users change the name of a fixed schema, but I'll do it if required.

Thanks!

baallan:
The project has adopted a guideline that autoconf have a default disable flag for new plugins until wide testing has been completed, even for plugins that just parse text files. This gets us around build bugs in actual code or in automake definitions, should such appear.

I can relent on the version in the plugin name, at least for now.

schema gets passed through (or should) to the set registration. It allows users who don't like lustre_client_general to rename the schema to something simpler and it's a bit of future-proofing for ldms v5 which will be allowing multiple instances of the same sampler (how that future would benefit your particular sampler isn't yet obvious).

dvs_sampler might be commonly deployed at a sampling frequency of 1/minute (I don't know this to be true). I know lustre sampling at 1/second is what we want to do at snl. It may be that rescanning /proc is blindingly fast and my suggestion is a premature optimization. Have you any testing done running the sampler at 10Hz and seeing no problems/delays? For how many lustre mount points? What if there is a mount point per user?

dvs and our local ib switch sampler both ignore instance, so there is precedence for it as you say. As long as there is a man page documenting how the sampler overrides the instance naming convention, we're good.

Unless you have some future larger schema in mind, I would drop the stats prefix from the metrics.
Are the brw* stats actually used, or are they leftovers that haven't been dropped from the report format provided by lustre?

component_id is an arbitrary integer id, typically mapped reversibly to hostname by a site-specifiic convention. It should be included in all metric sets, like producer is also user definable. It's fine, but most commonly redundant, to include host separately in your metrics.

jobid is key to all our job level analysis on clusters. In v3 plugins, it is supported in all plugins (or nearly so) by using the LJI_ macros from ldms_jobid.h as seen in meminfo.c. In ldms v4, using the sampler_base takes care of this. We have found it unreliable at best to try to do joins of slurmdb data with ldms data after the fact.

morrone:
I've made most of the changes that you talk about (and you were probably just responding to my first message anyway), so the conversation that follows in this comment is just because I am interested and shouldn't have any bearing on the landing of this patch.

> schema gets passed through (or should) to the set registration. It allows users who don't like lustre_client_general to rename the schema to something simpler and it's a bit of future-proofing for ldms v5 which will be allowing multiple instances of the same sampler (how that future would benefit your particular sampler isn't yet obvious).

I'm unclear on the future proofing argument.  Having multiple instances of the same sampler on a node would seem to have no bearing on the schema name.  If the schema represents the same data, and for for a plugin like mine it would, the schema name absolutely should be the same in both instances.  The instances will have their individual metrics sets differentiated by having different "instance" names in the metric sets.  Correct?  I should think that the schema name should only ever be different when the format of the data in the metric set is different (i.e. when the actual schema is really differerent).

> component_id is an arbitrary integer id, typically mapped reversibly to hostname by a site-specifiic convention. It should be included in all metric sets, like producer is also user definable. It's fine, but most commonly redundant, to include host separately in your metrics.

Yes, I was definitely getting confused about the sematics there.  Is component_id maybe a byproduct of history?  Maybe from a time before strings were supported?  When you say that the hostname field that I added is redundant, did you mean with component_id, or with producer?  Because if you meant with component_id, then I think you have that backwards. :)   For a site like ours, making up site-wide unique integer identifiers for absolutely every node that mounts lustre is a level of administrative complexity that I think our sysadmins would balk at.  And it would be entirely redundant with hostnames, which are already unique site-wide and instantly recognizable to humans without needing a look-up chart.  And if you mean that hostname is redundant with producer, then isn't component_id also redundant with producer?  That is the line of thought that made me start to doubt that I could trust the producer field, or maybe that producer doesn't appear in the csv data like the rest of the things in my schema or something.

So yeah, if producer will appear in each line of the csv data, and at least at our site we can always set it to hostname, then my hostname field in the schema can be dropped.

> jobid is key to all our job level analysis on clusters. In v3 plugins, it is supported in all plugins (or nearly so) by using the LJI_ macros from ldms_jobid.h as seen in meminfo.c. In ldms v4, using the sampler_base takes care of this. We have found it unreliable at best to try to do joins of slurmdb data with ldms data after the fact.

I've added the jobid, so just just another reminder that this comment is just for discussion purposes, and doesn't have any bearing on the landing of the patch.  I have found that the LJI macros and the sampler_base both make too many assumptions about the way the plugin operates to be easily usable.  I am wondering why this is the best approach too.  Why should every plugin implement the same data in every single metric set?

Would it perhaps be easier to have a single "jobid" plugin that records the jobid?  That way the data is available in the same place as the rest of the LDMS data, and there is no need to go back to the slurmdb database (or equivalent).

If the jobid absolutely needs to be in every single metric set, then perhaps the jobid field should come precreated in all metric sets when ldms_set_new() is called?

morrone:
OK, I think these are the only two outstanding issues:

> dvs_sampler might be commonly deployed at a sampling frequency of 1/minute (I don't know this to be true). I know lustre sampling at 1/second is what we want to do at snl. It may be that rescanning /proc is blindingly fast and my suggestion is a premature optimization. Have you any testing done running the sampler at 10Hz and seeing no problems/delays? For how many lustre mount points? What if there is a mount point per user?

Yes, I think those will be absolutely fine.  I don't think the single directory listing is going to produce nearly as much load as the actual stat files' reading as the rate and number of filesystems are scaled up.  But I can see that this is cause for concern, so I'll just add a separate refresh rate in the interest of getting this landed quickly.

> Unless you have some future larger schema in mind, I would drop the stats prefix from the metrics.
Are the brw* stats actually used, or are they leftovers that haven't been dropped from the report format provided by lustre?

Yes, expanding the schema in the future is what I had in mind.  I copied the approach of the existing lustre plugins, in which the prefix like "stats." provides a hint to where the data originated in lustre (the llite.*.stats file).  I tried to reduce the collected client data to a saner minimum, but we may need to add more fields in the future.  The ost/mdt plugins that I have in the wings work simularly, and they do have some extra fields from outside of the .stats file.  So, for instance, the ost plugin will have fields with the "stats." prefix for fields that come from the stats file.  The ost plugin has fields like "stats.read_bytes" alongside "kbytesfree".  The latter comes from the "kbytesfree" file, not fro the stats file.

But I'm not entirely wed to doing this the way that the the old lustre ldms plugins did it.  If it needs to change to land, I certainly will do so.  Let me know.

And yes, every field that I added in my plugin still exists in the Lustre code in at least some of the versions that this plugin is intended to support.

morrone:
@baallan, OK, I think that I addressed all of the issues that you raised in this update.  I added a configurable rescan interval for scanning the llite directory (default 30 seconds).  I ditched the "stats." prefix.  And I changed the schema name to a shorter "lustre_client" for good measure. :)

Current output example:

```
  METADATA --------
    Producer Name : opal2
    Instance Name : opal2/lquake-ffff8f103a61e000
      Schema Name : lustre_client
             Size : 1864
     Metric Count : 40
               GN : 5
  DATA ------------
        Timestamp : Fri Apr 19 14:50:54 2019 -0700 [82052us]
         Duration : [0.000018s]
       Consistent : TRUE
             Size : 336
               GN : 43
  -----------------
M char[]     hostname                                   "opal2"
M char[]     fs_name                                    "lquake"
M char[]     llite                                      "lquake-ffff8f103a61e000"
M u64        component_id                               0
D u64        job_id                                     0
D u64        dirty_pages_hits                           0
D u64        dirty_pages_misses                         0
D u64        read_bytes.sum                             0
D u64        write_bytes.sum                            0
D u64        brw_read.sum                               0
D u64        brw_write.sum                              0
D u64        ioctl                                      0
D u64        open                                       0
D u64        close                                      0
D u64        mmap                                       0
D u64        page_fault                                 0
D u64        page_mkwrite                               0
D u64        seek                                       0
D u64        fsync                                      0
D u64        readdir                                    0
D u64        setattr                                    3
D u64        truncate                                   0
D u64        flock                                      0
D u64        getattr                                    6
D u64        create                                     0
D u64        link                                       0
D u64        unlink                                     0
D u64        symlink                                    0
D u64        mkdir                                      0
D u64        rmdir                                      0
D u64        mknod                                      0
D u64        rename                                     0
D u64        statfs                                     3
D u64        alloc_inode                                1
D u64        setxattr                                   0
D u64        getxattr                                   12
D u64        getxattr_hits                              11
D u64        listxattr                                  0
D u64        removexattr                                0
D u64        inode_permission                           0
```

baallan:
I checked out the current branch and there are a few warnings when compiled -Wall that need to be fixed. (Ignore warnings related to LJI macros so long as jobid values get stored correctly).

I don't understand how

lustre_client.c:                if (llite_general_schema_init() < 0) {
lustre_client_general.c:int llite_general_schema_init(const char *producer_name)
lustre_client_general.c:        log_fn(LDMSD_LDEBUG, SAMP" llite_general_schema_init()\n");
lustre_client_general.h:int llite_general_schema_init();

is managing to compile without warnings, unless there are some hideous k&r c behaviors allowed through. maybe compile -pedantic?

The schema name from config needs to be passed through somehow.

Is it possible for you to test the dynamic behavior of this plugin (make a lustre item go away and come back) under valgrind? We need to be sure there are no wild pointers, leaks other than the expected ones related to sets, etc related to this plugin.

morrone:
Yes, I agree that is odd, and needs to be cleaned up.  It didn't effect operation though.  -Wall warnings have been fixed.

tom95858:
> > Unless you have some future larger schema in mind, I would drop the stats prefix from the metrics.
> > Are the brw* stats actually used, or are they leftovers that haven't been dropped from the report format provided by lustre?
> 
> Yes, expanding the schema in the future is what I had in mind.

Then you really should change the new name. Otherwise it will be very difficult to use this data subsequently for analysis when there is a mix of data with the same schema name, but different content.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+#include "ldms_jobid.h"
+#include "lustre_client.h"
+#include "lustre_client_general.h"
+
+#define _GNU_SOURCE
+
+#define LLITE_PATH "/proc/fs/lustre/llite"
+#define OSD_SEARCH_PATH "/proc/fs/lustre"
+
+ldmsd_msg_log_f log_fn;
```
baallan: this cannot be global. it needs to be passed to supporting library code via arg list if the sampler is split into library code.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+static char producer_name[LDMS_PRODUCER_NAME_MAX];
+static uint64_t component_id;
+static time_t last_refresh;
+static time_t refresh_interval = 30;
+#ifdef ENABLE_JOBID
+bool with_jobid = true; /* for LJI */
```
baallan: this needs to be LJI_GLOBALS, not manually declared.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+}
+
+static const char *usage(struct ldmsd_plugin *self)
+{
+        log_fn(LDMSD_LDEBUG, SAMP" usage() called\n");
+	return  "config name=" SAMP " [producer=<prod_name>] [instance=<inst_name>] [component_id=<compid>] [schema=<sname>] [rescan_sec=<rsec>] [with_jobid=<jid>]\n"
+		"    <prod_name>  The producer name (default: the hostname)\n"
+		"    <inst_name>  The instance name (ignored, dynamically generated)\n"
+		"    <compid>     Optional unique number identifier (default: zero)\n"
+		"    <sname>      Optional schema name (ignored, static schema name)\n"
```
baallan: must not be ignored. allowing user tuning schema name would help address Tom's comment if the schema evolves or has optional parts in the future.

ldms/src/sampler/llnl/lustre_client.h
```diff
@@ -0,0 +1,66 @@
...
+ */
+#ifndef __LUSTRE_LLITE_H
+#define __LUSTRE_LLITE_H
+
+#include "ldms.h"
+#include "ldmsd.h"
+
+#define SAMP "lustre_client"
+
+extern ldmsd_msg_log_f log_fn;
```
baallan: these externs need to go away with the globals they represent.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+        struct llite_data *llite;
+        char path_tmp[PATH_MAX]; /* TODO: move large stack allocation to heap */
+        char *state;
+
+        log_fn(LDMSD_LDEBUG, SAMP" llite_create() %s from %s\n",
+               llite_name, basedir);
+        llite = calloc(1, sizeof(*llite));
+        if (llite == NULL)
+                goto out1;
+        llite->name = strdup(llite_name);
```
baallan: leaks in llite_destroy?

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+        if (llite->name == NULL)
+                goto out2;
+        snprintf(path_tmp, PATH_MAX, "%s/%s", basedir, llite_name);
+        llite->path = strdup(path_tmp);
+        if (llite->path == NULL)
+                goto out3;
+        snprintf(path_tmp, PATH_MAX, "%s/stats", llite->path);
+        llite->stats_path = strdup(path_tmp);
```
baallan: leaks

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+        if (llite->stats_path == NULL)
+                goto out4;
+        llite->fs_name = strdup(llite_name);
```
baallan: leaks

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,360 @@
...
+        log_fn(LDMSD_LDEBUG, SAMP" llite_create() %s from %s\n",
+               llite_name, basedir);
+        llite = calloc(1, sizeof(*llite));
+        if (llite == NULL)
+                goto out1;
+        llite->name = strdup(llite_name);
+        if (llite->name == NULL)
+                goto out2;
+        snprintf(path_tmp, PATH_MAX, "%s/%s", basedir, llite_name);
+        llite->path = strdup(path_tmp);
```
baallan: leaks?

baallan:
A quick spin with valgrind using the test in the pull request to morrone/ovis lustre_plugins_v3 turned up some leaks, as noted in my comments on the source code here.

ldms/src/sampler/llnl/lustre_client_general.c
```diff
@@ -0,0 +1,276 @@
...
+#include "ldms.h"
+#include "ldmsd.h"
+#include "ldms_jobid.h"
+#include "lustre_client.h"
+#include "lustre_client_general.h"
+
+static ldms_schema_t llite_general_schema;
+
+/* all of these will be prefixed with "stats." in the schema */
+static char *llite_stats_uint64_t_entries[] = {
```
baallan: It would eliminate thousands of ldms_metric_by_name calls if you change ..._entries to
struct entry { const char *name, int idx; };
static struct entry llite_stats_uint64_t_entries[] = {
 { "dirty_pages_hits", -1},
 ...
};
and then at metric registration, set the idx field to the return code from ldms_schema_metric_add, and then at metric setting, just use the idx value.

ldms/src/sampler/llnl/lustre_client_general.c
```diff
@@ -0,0 +1,276 @@
...
+        rc = ldms_schema_meta_array_add(sch, "fs_name", LDMS_V_CHAR_ARRAY, 64);
+        if (rc < 0)
+                goto err2;
+        rc = ldms_schema_meta_array_add(sch, "llite", LDMS_V_CHAR_ARRAY, 64);
+        if (rc < 0)
+                goto err2;
+	rc = ldms_schema_meta_add(sch, "component_id", LDMS_V_U64);
+        if (rc < 0)
+                goto err2;
+	rc = LJI_ADD_JOBID(sch);
```
baallan: save the rc value from LJI_ADD_JOBID somewhere and use it rather than ldms_metric_by_name.

morrone:
> Then you really should change the new name. Otherwise it will be very difficult to use this data subsequently for analysis when there is a mix of data with the same schema name, but different content.

Absolutely.  When the schema is change in the future, the schema name will be changed as well.  What new name would I be changing right now, though?  Did you want a "_v1" added to the name or something?

ldms/src/sampler/llnl/lustre_client.c
morrone: Why?  This is a plugin, which presumably is dlopen()ed by ldmsd, right?  So unless ldmsd dlsym()s the plugin's log_fn, it doesn't effect the larger program, right?  You say, into "library code".  This plugin is not split into library code, it is merely broken up into multiple files for organizational purposes.

ldms/src/sampler/llnl/lustre_client.c
morrone: I'd like to, but the jobid macros make some assumptions that make them difficult to use with a plugin that employs multiple files.

ldms/src/sampler/llnl/lustre_client.c
morrone: This would seem to be unrelated to Tom's comment.  Tom's comment was about when the plugin author changes the schema.  When that happens, the plugin author is responsible for changing the schema name.  Understood.  Pushing that responsibility onto users to change the schema name when the plugin author makes a schema change seems undesirable.  How many users can really be expected to figure out which schema changes have happened in a a schema that hasn't changed names, and then use a configuration option to change the schema name themselves?  Seems like an odd design choice.

But this one is an easy change, so even if I don't see what use it has, I'll add this tomorrow.

ldms/src/sampler/llnl/lustre_client.c
morrone: Thanks!  I'll fix these tomorrow.

ldms/src/sampler/llnl/lustre_client.c
baallan: ldmsd is intended to be used in a variety of exotic situations, including fully static linking/embedded. So we need to keep the symbol scoping clean, which in this case is not at all hard: just pass log_fn to the init function in your support file and store it in a static variable of that file as well. pass with_jobid through the *..._sample function signature.

morrone: Ah, I see.

ldms/src/sampler/llnl/lustre_client_general.c
morrone: That doesn't quite solve the problem though.  Be aware that the list of values in the lustre stats file is not a static list of values.  I can't predict which will be there and which will not.  I might make an assumption of order, but I think it is best to avoid doing that as well.

With the proposed change,  I still need to do a linear walk on the array to find the index number each time I see it in the lustre stats file.  That seems no better than simply using ldms_metric_by_name().  

If we ever thought the linear walk was a real issue, then it would seem to me that implementing a hashed lookup under ldms_metric_by_name() would be the way to do it, rather than making each plugin implement the same thing.

morrone:
You know whats funny?  The existing lustre plugins use unchangeable schema names, _and_ they have dynamically generated schemas that will be different on each node (if I understand correctly).  But now the new plugin with static schemas will have a changeable schema name. :)

ldms/src/sampler/llnl/lustre_client_general.c
baallan: I'm not clear on why the list of metrics is not static. The *plugin reported* list must be static-- that's why we have a schema. You can define the convention for handling missing metrics as found in the lustre data (0, -1, etc; we generally favor 0 and/or setting a flag elsewhere in the data set if something is expected but missing. Alternatively, if the content for a given lustre installation may be any subset of the list you have in the code depending on the exact lustre configuration and version, you can define the schema by those metrics present when you start the sampler. If this is the case, append a checksum of the sorted metric names collectable to the user provided schema name before schema creation-- this will ensure there is never a schema abiguity. 
You can avoid linear searches using bsearch and defining your entries list array in sorted order and use -1 for the entries not seen by schema creation time.

baallan:
The existing lustre plugins take an argument (list of mount points) which fixes the schema. Missing mount points report 0 values. The admin can define the old lustre plugin schema name (e.g. lustre_f_g). But you are correct that the lustre (and sysclassib and procstat) plugins have long been a thorn in our side wrt schema.

morrone:
@baallan, I don't think those are memory leaks.  My best guess is that you shutdown ldmsd without first calling stop() and term() for the new plugin.  It looks like ldmsd fails to call term() on its own on exit (which seems like a bug to me), so yes, it will look like the plugin leaked memory because it was not given a chance to clean up.

When term() is called, I see no leaked memory from this plugin under valgrind.

ldms/src/sampler/llnl/lustre_client_general.c
morrone: The metric contents are static, but the data from /proc is not.  In the metric I statically include all _possible_ values that might be seen in the "stats" file in /proc.  Lustre only starts printing out a stat once someone has actually _done_ that operation at least once.  This means that the various values only gradually start to appear as the mount is used.

Only including the metrics found at plugin start up time would be very, very wrong, because it is very likely that _none_ of the values will show anything at start up time.  And the the number of values in the stats file will almost certainly grow over time, so we should really just have all of them listed in the ldms metric set at start up time.

This kind of sucks for monitoring purposes.  We would probably prefer that lustre it just print out a line with a zero value instead of not printing the line at all.  In newer lustre proc files, such as "job_stats", they employ a new json format and they _do_ always include values of zero rather than printing nothing.  So they seem to have learned, but not updated old stats to work that way.

The long and short of it is that when I parse the lustre stats file, I need to look up the index number by the name I find in the stats file.  I can't make an simple numerical transformation from the line number in the stats file to the index number in the metric set.

morrone:
OK, there is a new set of patches available.  Major changes are:

- Everything combined into one file to satisfy the LJI_* macros.
- No non-static globals
- Schema name can be changed by user
- Included baallan's patches

morrone:
I just pushed another update that avoids calling ldms_metric_by_name() in the main loop, instead using an rbt for faster index lookup.

The job_id index is also stored in global to avoid lookup.

I think that addresses everything requested.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+static char producer_name[LDMS_PRODUCER_NAME_MAX];
+static char schema_name[LDMS_SET_NAME_MAX];
+static uint64_t component_id;
+static time_t last_refresh;
+static time_t refresh_interval = 30;
+static int job_id_index;
+LJI_GLOBALS;
+
+/* red-black tree root for llites */
+static struct rbt llite_tree;
```
baallan: speed comment: use of rbt here is log(n_mounts) string compares, as far as I can tell. If n is always small, or searches are infrequent, we don't care.  If search of llite_tree is frequent, lib/src/coll/idx.h may be a better choice (o(1)).

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+/*******************************************************************/
+/* ldms_metrict_by_name() employs a linear lookup and is verboten for
+ * use by plugins in the main sample() loop.  This section section
+ * implemants a faster lookup using the rbt structure, and offers
+ * alt_ldms_metric_by_name() as a limited alternative to the original.
+ * Someday, perhaps ldms_metric_by_name() will be improved and this
+ * section can be removed.
+ */
+/* red-black tree root for metric index lookup table */
+static struct rbt metric_lookup_tree;
```
baallan: speed comment: use of rbt here is log(n_metrics) string compares, afaict. trading memory for speed with lib/src/coll/idx.h may be a better choice (o(1)). I can live with this version going to production and the improvement being done soonish.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+        if (rbn == NULL) {
+                return -1;
+        }
+        entry = container_of(rbn, struct metric_lookup_entry, node);
+
+        return entry->index;
+}
+/*******************************************************************/
+
+static int llite_general_schema_is_initialized()
```
baallan: This function is unneeded. if we fix config() to create the schema. If config() cannot create the schema, then it returns an error and the plugin will never be sampled, so sample never needs to check for schema existence.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+        /* ldms_set_publish(set); */
+
+        return set;
+}
+
+static void llite_stats_sample(const char *stats_path,
+                                   ldms_set_t general_metric_set)
+{
+        FILE *sf;
+        char buf[512];
```
baallan: If you can put a reasonable upper bound on the size of the 'stats' file (4k seems likely), it could be faster to read the entire file in a single read rather than multiple fgets queries. The kernel security features of late to mitigate TLB attacks added overhead. Unfortunately, there is no convenient string equivalent of fgets, so I can live with this version going to production and the improvement being done whenever we have a global solution for all our samplers that use fgets.

ldms/src/sampler/llnl/lustre_client.c
baallan: leaks no more than the current  lustre sampler for which term is also never called in normal use.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+	if (value != NULL) {
+		log_fn(LDMSD_LWARNING, SAMP ": ignoring option \"instance=%s\"\n",
+                       value);
+	}
+
+	value = av_value(avl, "schema");
+	if (value != NULL) {
+                snprintf(schema_name, sizeof(schema_name), "%s", value);
+	}
+
```
baallan: llite_general_schema_init should be called here and the function return an error if it fails. them sample() is simpler.

baallan: I would like to see config() accept option blacklist=file_name which is a file containing a single column of metric names. Any metric name appearing in the file is omitted from the schema (and in the metric name index lookup, retain a negative number and ignore them at sample time when found in stats file). But again, I'm fine with not adding this feature in the initial version of the plugin.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+                llite = container_of(rbn, struct llite_data, llite_tree_node);
+                llite_general_sample(llite->name, llite->stats_path,
+                                   llite->general_metric_set);
+        }
+}
+
+static int sample(struct ldmsd_sampler *self)
+{
+        log_fn(LDMSD_LDEBUG, SAMP" sample() called\n");
+        if (llite_general_schema_is_initialized() < 0) {
```
baallan: this check can be omitted if config() is right.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+        rbt_init(&new_llite_tree, string_comparator);
+
+        /* Make sure we have llite_data objects in the new_llite_tree for
+           each currently existing directory.  We can find the objects
+           cached in the global llite_tree (in which case we move them
+           from llite_tree to new_llite_tree), or they can be newly allocated
+           here. */
+
+        dir = opendir(LLITE_PATH);
+        if (dir == NULL) {
```
baallan: This warning should be issued exactly once before a successful sampling occurs; it needs a static guard variable. The guard can be reset to it's initial state (warning message allowed) after a successful sample. In the case where /proc disappears, we do not want the log flooded with redundant warnings.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+out1:
+        fclose(sf);
+
+        return;
+}
+
+static void llite_general_sample(const char *llite_name, const char *stats_path,
+                                 ldms_set_t general_metric_set)
+{
+        log_fn(LDMSD_LDEBUG, SAMP" llite_general_sample() %s\n",
```
baallan: delete this call, unless you really want it and add a debug=<0/1> option to the sampler and put if (debug) around the logging call.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+{
+        ldms_schema_t sch;
+        int rc;
+        int i;
+
+        log_fn(LDMSD_LDEBUG, SAMP" llite_general_schema_init()\n");
+        sch = ldms_schema_new(schema_name);
+        if (sch == NULL)
+                goto err1;
+        rc = ldms_schema_meta_array_add(sch, "hostname", LDMS_V_CHAR_ARRAY, 64);
```
baallan: 64 here should be replaced with the appropriate c macro defining the maximum host name limit. HOST_NAME_MAX is probably what you want.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+        if (rc < 0)
+                goto err2;
+        rc = ldms_schema_meta_array_add(sch, "fs_name", LDMS_V_CHAR_ARRAY, 64);
```
baallan: Does lustre define an fs_name size limit with a macro? we prefer them to magic numbers. PATH_MAX or NAME_MAX if lustre headers don't have something.

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,641 @@
...
+        if (rc < 0)
+                goto err2;
+        rc = ldms_schema_meta_array_add(sch, "llite", LDMS_V_CHAR_ARRAY, 64);
```
baallan: need the appropriate macro instead of a magic 64 here also.

baallan:
@morrone  I retested and the functionality overall looks good enough. I closed the comments attached to the source about memory leaks; please review the other comments. The config() function needs a minor revision and the logging and missing macro constants usage has some nits. There are several 'could be better' type comments which you could revise to address or you can defer to the version of the plugin for ldms v4.

ldms/src/sampler/llnl/lustre_client.c
morrone: config() is not broken.  This plugin is explicitly designed to require no configuration.  ldmsd does not require that config() be called before the plugin is started, and this plugin does not require that config() be called to work. It will operate just fine for many use cases by running on defaults.

And because ldmsd does not enforce that "config" is executed before "start" for a plugin, if I were to change the implementation as you suggest, I would just have to create a new "config_was_called()" check anyway to enforce that config() is called before sample().

I'm not suggesting that ldmsd should change.  I think it is entirely reasonable, and in fact desirable, to be able to start a plugin without needing to config() it if the plugin can reasonably operate in that mode.  Anything that reduces the configuration burden on sysadmins is desireable.

But the current implementation is intentional.

ldms/src/sampler/llnl/lustre_client.c
morrone: fgets() doesn't necessarily imply an underlying read() each time it is used.  It depends on the buffering method selected.  Perhaps one could use setvbuf() if this became enough of a concern?

ldms/src/sampler/llnl/lustre_client.c
morrone: Done.

ldms/src/sampler/llnl/lustre_client.c
morrone: Done.

ldms/src/sampler/llnl/lustre_client.c
morrone: The logging call almost certainly does exactly that check, I hope.  It must do something like "If current debug level < LDMSD_LDEBUG,  do nothing", right???

ldms/src/sampler/llnl/lustre_client.c
morrone: Done.

ldms/src/sampler/llnl/lustre_client.c
morrone: See other comment on this subject.

ldms/src/sampler/llnl/lustre_client.c
morrone: Done.

ldms/src/sampler/llnl/lustre_client.c
morrone: Fine.  I switched it to using coll/idx.h.  (would be nice to have some documentation in there)

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,654 @@
...
+ *
+ *      Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ *
+ *      Redistributions in binary form must reproduce the above
+ *      copyright notice, this list of conditions and the following
+ *      disclaimer in the documentation and/or other materials provided
+ *      with the distribution.
+ *
+ *      Neither the name of Sandia nor the names of any contributors may
```
tom95858: This should probably say "Neither the name of  Lawrence Livermore ... ", not Sandia..

ldms/src/sampler/llnl/lustre_client.c
tom95858: This will lead to a schema with the same name having different values. As it is, the contents are consistent. As it is, the schema seems short and sweet. Why mess with it?

ldms/src/sampler/llnl/lustre_client.c
```diff
@@ -0,0 +1,654 @@
...
+
+static ldms_schema_t llite_general_schema;
+
+static char *llite_stats_uint64_t_entries[] = {
+        "dirty_pages_hits",
+        "dirty_pages_misses",
+        "read_bytes.sum",
+        "write_bytes.sum",
+        "brw_read.sum",
+        "brw_write.sum",
```
tom95858: Does the .sum mean that the value is cumulative? Are all of these values sums, or are the other values rates?

ldms/src/sampler/llnl/lustre_client.c
baallan: @valleydlr  owes us some verbiage here more generally. apparently there was some concern with snl legal.

ldms/src/sampler/llnl/lustre_client.c
baallan: Always calling config is 'standard practice' in v3 and is enforced in v4 via sampler_base::base_config
config values for producer and instance are always required, So config is effectively always required. conveniently, producer is in most cases (including lustre_client)  going to be given the same value as hostname {this differs for things like the router identifiers, though}.
There is no configuration burden on sysadmins at this level of detail/defaults if they use the systemd scripts shipped with ldms and based on libgenders.

ldms/src/sampler/llnl/lustre_client.c
baallan: The log call does check, but for a sampler that is coded correctly we do not want any per-sample debug logging unless it is 'opt-in'. It makes seeing things in the debug logs hard by default.

ldms/src/sampler/llnl/lustre_client.c
baallan: With the other requested feature (blacklist capability) the schema will not be all identical, so need to be nameable.

ldms/src/sampler/llnl/Plugin_lustre_client.man
```diff
@@ -0,0 +1,76 @@
...
+.PP
+Within ldmsd_controller or a configuration file:
+.nf
+load name=lustre_client
+config name=lustre_client
+start name=lustre_client interval=1000000
+.fi
+
+.SH SEE ALSO
+ldmsd(8), ldms_quickstart(7), ldmsd_controller(8), ldms_sampler_base(7)
```
baallan: need to delete the reference to ldms_sampler_base(7) for v3

ldms/src/sampler/llnl/Plugin_lustre_client.man
```diff
@@ -0,0 +1,76 @@
+.\" Manpage for Plugin_lustre_client
+.\" Contact ovis-help@ca.sandia.gov to correct errors or typos.
+.TH man 7 "18 Feb 2018" "v4" "LDMS Plugin lustre_client man page"
```
baallan: v4 here needs to be v3

ldms/src/sampler/llnl/Makefile.am
```diff
@@ -22,3 +22,12 @@ libedac_la_CFLAGS = $(AM_CFLAGS) -I$(srcdir)/$(SAMPLER)
 pkglib_LTLIBRARIES += libedac.la
 dist_man7_MANS += Plugin_edac.man
 endif
+
+# lustre client plugin
+if ENABLE_LLNL_LUSTRE
+liblustre_client_la_SOURCES = lustre_client.c
+liblustre_client_la_LIBADD = $(COMMON_LIBADD) $(JOBID_LIBFLAGS)
+liblustre_client_la_CFLAGS = $(AM_CFLAGS) -Wall -I$(srcdir)/$(SAMPLER)
```
baallan: take -Wall out of makefile.am and use 
CFLAGS=-Wall ./configure ...
instead

ldms/src/sampler/llnl/Plugin_lustre_client.man
```diff
@@ -0,0 +1,76 @@
...
+.br
+This MUST be lustre_client.
+.TP
+producer=<pname>
+.br
+The producer string value.  Default: hostname.
+.TP
+instance=<set_name>
+.br
+The name of the metric set.  This will be ignored, and the individual metric set instances will have dynamically generated instance names.
```
baallan: a more flexible alternative here would be instead of ignoring instance to define
instance=<set_prefix>
"the prefix of the metric sets. Complete set names will be set_prefix/<device>."
But I'm fine with it either way for the initial merge.

ldms/src/sampler/llnl/lustre_client.c
morrone: If people want to use debug level logging, and they want to not see those messages from a particular plugin, they can simply not use that plugin while debugging, right?  Or if you really want to have different  log levels for different components, then it wouldn't be too difficult to add that functionality at logging library level rather than repeating it in each plugin.

ldms/src/sampler/llnl/lustre_client.c
morrone: I'm not really planning to add blacklists at this point.

ldms/src/sampler/llnl/lustre_client.c
morrone: Use of "sampler_base::base_config" seems to be a voluntary policy enforcement rather than a code enforcement.  I think that the ldmsd plugin code really needs to add a flag to track whether or not a plugin has config() called (and also returned a success return code) before it allows "start" to be issued for that plugin.  If that is really the semantics, then check and enforce them so that we don't need to write a bunch of the same checks over and over again in every plugin.

Offering a library that does those checks isn't the same.  In fact, I already have issues with base_config not being designed in a way that works well with my plugins, so I didn't use them for v4.  But the plugin config/start semantics should still be enforced correctly by ldmsd.

ldms/src/sampler/llnl/lustre_client.c
baallan: Please delete the DEBUG lines from the sample and llites_refresh. 
v3 is primarily in maintenance mode and per-plugin logging behaviors are not on the drawing board for it or for v4 (near release). v5 is just getting started internally. We don't use a separate logging library with the feature you suggest, but we  can discuss separately an extension to the logging API if you want to contribute one that trivially enables per-plugin control.

ldms/src/sampler/llnl/lustre_client.c
baallan: I made a github pr covering the license issue.

ldms/src/sampler/llnl/Plugin_lustre_client.man
baallan: this is covered in my pr 2.

ldms/src/sampler/llnl/lustre_client.c
baallan: If we get the rest of the nits sorted, I will contribute blacklist in a subsequent patch. it's good to have schema name from the start handled as expected for all other plugins.

morrone:
I have the plugins (server ones too) building externally against the TOSS ovis ldms packages.  I think in hindsight I agree that is the better approach than landing this in the upstream tree.  While this plugin might be close-ish to landing, I don't think that we'll want to make quite a few of the changes to the server plugins that will be required by upstream rules.  Our use requirements and schema design approaches just don't seem to be close enough aligned.  So with that in mind, I am going to retract this pull request.

Thank you for all of the feedback!

PR #70, opened 2019-11-06
Add ovis-test/test.h to dist so canonical tarball is buildable.

morrone:


lib/src/Makefile.am
```diff
@@ -1,4 +1,4 @@
-nobase_include_HEADERS = ovis-test/test.h ovis-lib-config.h
```
narategithub: I think `ovis-lib-config.h` is generated by `configure` script and should not be distributed. I think adding `ovis-test/test.h` to the `EXTRA_DIST` list is more preferable.

morrone:
@narategithub , good call.  I fixed that and refresh the PR.

morrone:
Landed by cherry-pick.

Issue #69, opened 2019-11-05
ldms packaging plan


morrone:
The [wiki describes](https://github.com/ovis-hpc/ovis/wiki/DirChanges-v3-to-v4) how in ovis v3, LDMS, baler, and sos were all under one project, but as of ovis v4, LDMS, Baler, and SOS are independent projects.

What impact (if any) is this planned to have on the approach to packaging?  Will LDMS move to having package names that start with "ldms", or is the project sticking with ldms packages being named "ovis"?

Whatever the plan, I think it would be very good to start uploading canonical tarballs of each release to github.  The tarball would be the result of running "make dist" on the project.

baallan:
We don't expect major changes to packaging v4 other than sos and baler would be in their own rpms and ldms (and lib) will continue in the ovis-* rpm names. The packaging scripts themselves will be moving to ovis-hpc/distribution from baallan/distribution as soon as I have finished up the non-genders cray rpm specs for distribution/cle*.

There is no canonical tar ball because the result of make dist depends on configure which depends heavily on platform.

morrone:
There really shouldn't _be_ "packaging scripts".  I should just be able to write a sane .spec file that will build the ldms canonical tarball.

I'll drop the "v4" part of this ticket.  Lets just discuss where we want to go with the build system of ldms.

I think it would be a good idea to move ldms into its own repository like sos and baler.

ldms should have its own, clean, single build system (no recursive calls into sub-project autogen/configure, etc.).

Can we move in that direction in a reasonable time frame?

baallan:
Eliminating the separation of ovis/lib/configure.ac from ovis/ldms/configure.ac would be a fine start.
There seems to be some historically based aversion to this simplification, but I've never managed to find out where it originates. 
To the best of my knowledge, there is no other project that depends on ovis/lib. For example, baler and sos have their own support 'classes' which in some cases are clones (or divergent clones) of ovis/lib classes.

morrone:
OK, great to know!  I was afraid that there needed to be an ovis-lib project that the other projects could build against.  But if ovis/lib can just be relocated under ovis/ldms that makes things even easier.

baallan:
I think without much loss of sanity we could easily move ovis/lib/src/* to ovis/ldms/src/*, but maybe there's an alternate product roadmap for ovis/lib that I've missed. This would cut the amount of exported configuration trivia for 3rd party use in half.

morrone:
Dropping this ticket.

[...]

PR #72, opened 2019-11-07
add ldms/src/contrib and config args to use it.

baallan:
Our autoconf scripts never touch anything in lib,ldms/src/contrib/*
directly. The hooks fire the builds of contrib plugins after the main build is complete.


baallan:
The user can enable contrib plugins with --enable-contrib --with-contrib=sampler/tutorial_sampler,my_plugin,store/store_tutorial.

baallan:
@morrone So you would be more in favor of solution (1) in
https://github.com/ovis-hpc/ovis/wiki/Proposal-3.
I think I prefer (1) also. 

In some sense, #72 is about having our cake and eating it too: the ovis source tree can have a defined (if ugly, as you say) way to drop in complete plugin projects from hardware vendors.
This enables the project to maintain a "batteries included" option from a single repository without the core framework developers taking on refactoring/maintenance of some potentially very ugly configure scripts(E.g. mpi or cuda dependent plugins) when we want to accept 3rd party plugins to the battery collection.

Each example plugin directory under the ovis/ldms/contrib directory individually looks a great deal like your llnl/ldms-plugins-llnl (full autotooling)
 
Even if you would never want your plugins dropped in the tree in this fashion, us having to develop and maintain that build hook ensures that we actually do export in the installation tree all the information a 3rd party package might need to ensure build compatibility.

For me, the holy grail of fixing our build system is to eliminate recursive automake and recursive configure entirely, but we don't have a spare developer to throw at maintaining such a scheme. That would certainly make the build faster and make it easier to drop in single-file plugins, but it still wouldn't solve the "intersection of all possible plugins in configure.ac" issue.

morrone:
>So you would be more in favor of solution (1) in

Yes, I would very much prefer (1) to (2).

But what I had originally suggested many months ago was to simply add a new "contrib" subdirectory for plugins.  It would work almost exactly the same as the current plugins.  No independent recursive build system.  Perhaps just split the ldsm/src/sampler directory into ldms/src/sampler/official and ldms/src/sampler/contrib.  The "contrib" plugins would all default to not being built, but otherwise no added complexity to the build system.  Call that option (3).

But since I figured out on my own how to do option (1) with ldms-plugins-llnl, now I think that is probably the prefered approach over all.  We should probably just document the external APIs and clean up the installed libraries so others can do the same more easily.

> Even if you would never want your plugins dropped in the tree in this fashion, us having to develop and maintain that build hook ensures that we actually do export in the installation tree all the information a 3rd party package might need to ensure build compatibility.

Complex hooks in-tree are really not a great way to go about doing that.  It would be better to have a sample project external to the tree and just build that as part of a testing process (coordinated at a level above/outside the ldms tree).  An automated build farm or CI system could make that easy.  Or you can script it externally.  Lots of ways to skin that cat that do not require complicating the build system.

> For me, the holy grail of fixing our build system is to eliminate recursive automake and recursive configure entirely, but we don't have a spare developer to throw at maintaining such a scheme.

For me, that is less of a holy grail than it is a nearer-term requirement if we are going to continue to use ldms.  There are two areas that are really making ldms far more effort to work with than is really necessary: build/packaging and configuration (or three areas, if you consider build and packaging separate).  I would advocate pausing developer effort on other tasks until those things are in better shape.  And I'll try to do my part as well.

I'm dipping my toes into the water on how to improve things now.  Some of my PRs (#71, #75, #76) are a result of beginning to investigate how to make that happen.  They are early cleanup that will make the greater build/packaging solution easier. It is also the motivation behind my question in issue #69.

baallan:
Happy to see other toes in the water.

Option (3) is what we had starting as far back as ldms v2: sampler/$someoneelse was contributed. In a specific case llnl/ contributed edac. The project leads have been generally dissatisfied with the result -- vendors forked instead of contributing back. Granted forks may have been due to API problems or the fact of 'closed development' inside a private repository, rather than build integration issues. 

Example projects external to the tree tend to get out of sync quickly. This I have seen on other projects where the core team was not able to focus on 3rd party extensibility at all times.

One of the big insanities detected during the 'contrib' exercise is that the installed ovis header structure is not followed inside the ovis C code. This means any dependent build has a big mess of include flags so that unadorned '#include "sampler_base.h"' will be resolved correctly.

See also: https://github.com/baallan/distribution/blob/master/HISTORY.md for why there is a repo for packaging scripts and why they are a bit complicated. Normally this level of tribal build knowledge is hidden inside semi-private vendor packaging script repositories or regression setups.   

If you think building and packaging is hard now, try maintaining and testing ldmsd on pre-general-availability platforms with a single configure script if it picks up dependencies on mpi, cuda, etc. 
The following existing plugins would all need to become separate little projects:
transport: zap/rdma, zap/$cray-name-here
sampler: lustre*,  edac, cray*, sysclassib, opa2, syspapi, job_info, slurm, msr_interlagos, ...
store: sos
you get the idea. That's a lot of extra spec files, a lot of extra configure files.

If we go with a "clean core, no batteries, spec files are easy" approach to ldmsd, then we really have to add a staffer somewhere whose full time job is nothing but tracking spec files and testing for all the supported environments and (separately) all the supported plugins with non-trivial platform dependencies.  Because ultimately what customers want delivered is "batteries included".

morrone:
>One of the big insanities detected during the 'contrib' exercise is that the installed ovis header structure is not followed inside the ovis C code. This means any dependent build has a big mess of include flags so that unadorned '#include "sampler_base.h"' will be resolved correctly.

Yes!  I would definitely like to see the headers better organized within the project to allow internal plugins to use the same "#include <ldms/sampler/sampler_base.h>" include style as the plugin would use if it was pulled out of the tree.  That is totally fixable.

>If you think building and packaging is hard now, try maintaining and testing ldmsd on pre-general-availability platforms with a single configure script if it picks up dependencies on mpi, cuda, etc.

I really can't imagine how the added complexity of multiple configure scripts would make that situation _easier_.  I have had to maintain software and test is on man pre-general availability platforms, and I really don't have any idea why the number of configure scripts would come into play at all.

> See also: https://github.com/baallan/distribution/blob/master/HISTORY.md for why there is a repo for packaging scripts and why they are a bit complicated.

Well, that is interesting, but I think that some of the answers are not quite accurate.  Take this answer:

"Significant subsystems of OVIS are excluded depending on the target platform. No single target platform can compile all of the available options together."

The point of the one dist tarball is that is includes _all_ of the source and a "configure" script.  It is through the configure script that the build is adjusted to apply to the target platform through a combination of automatic tests and user input (configure command line options).  It is not at all necessary to have different tarballs for different platforms.  The talk about different sets of autotools on different platforms is addressed _by_ creating the "make dist" tarball.  You can agree on a set of platforms where "make dist" can be run, and then that resulting tarball can run on _all_ platforms.  Perhaps not all platformas have a new enough autotools package to generate the make-dist tarball, but that is probably a rare enough condition that it is acceptable.  I've certainly done that before.  I make-dist on a comptiable architecture, and then use the tarball on the target build platform.

The part about double configure on RPMs is fixable too with a small change in approach, which will lead to a great deal of reduced complexity in the build system.  The rpm spec file is designed to site_outside_ the package and conduct the package's build and installation.  When you try to pass all of the configure scripts options in to an on-the-fly build spec file which in turn regurgitates the same configure options again when it is run, _that_ is when you get into trouble.  I've seen how this approach plays out on other projects.  It is best to avoid this entirely. If there is a .spec.in file in a project that generates a .spec file, is should have very, very, very little that is actually substituted by autoconf.  Pretty much just version number substitution, and little else.  This is a detailed topic, but I've been down this path many times, and I am confident that ldms's build and package complexity can be reduced significantly.  I'm not aware of any special inherent complexity in ldms that wasn't also in present in other non-trivial projects I've worked with and solved with simpler solutions.

> The following existing plugins would all need to become separate little projects
[cut]
>you get the idea

No, I'm sorry, I don't understand the argument there.  Why would all of those existing plugins suddenly need to be separate projects?

>If we go with a "clean core, no batteries, spec files are easy" approach to ldmsd, then we really have to add a staffer somewhere whose full time job is nothing but tracking spec files and testing for all the supported environments and (separately) all the supported plugins with non-trivial platform dependencies. Because ultimately what customers want delivered is "batteries included".

I really don't think that is the way to look at it.  Your customers are suffering with a very difficult to use build and packaging system _now_.  Adding complexity is not the same as adding batteries.

baallan:
The result of make dist in the current repository depends on configure options because of Makefile.am constructions such as
if $FEATURE_ENABLED
SUBDIRS += feature
endif
or
if $plugin_enabled
plugin_stuff
endif

A contributed makefile.am refactorization that uses configure to disable features in the build without excluding them from the make dist rules would go a long way toward eliminating divergence in the output of make dist across platforms. DIST_SUBDIRS was not used originally but probably should be now.

baallan:
" Your customers are suffering with a very difficult to use build "

@morrone Are you telling me production customers in TOSS environments are doing something other than ./firerpms (which is pretty simple) now? Who are these people and what are they doing with koji instead of that?

I haven't had any (other) production people report issues with this approach to canning the knowledge which is normally buried in something like koji.

morrone:
>The result of make dist in the current repository depends on configure options because of Makefile.am constructions such as

I understand. That can be fixed. In fact I already have a fix for most of those in PR #83, in the commit titled "Fix conditional SUBDIRS usage for make dist".

> Are you telling me production customers in TOSS environments are doing something other than ./firerpms (which is pretty simple) now?

We're using it for v3, but we would very much like to eliminate firerpms from our workflow.

oceandlr:
Re LDMS User's Group discussion on 11/11/2019:
I need the following functionality, in short order (i.e., in time for SC).
* A place for contributors to put sampler and store plugins in the LDMS source tree that are not yet officially supported. Not in separate, externally maintained repos. This will enable others to have access to them while under development.
* For anyone using the plugins, optional configure and build of these plugins as part of the LDMS build. Not a two step configure and/or build. This enables ease of config and build.
* Ease of use for the developer of the plugins wrt changes to Makefiles, configure.ac's, and configuration flags both for building and configuring while in the contrib area. Minimal disruption when proposing moving them from the contrib area to the main line supported area (in source). 

Can we do something that a) in short order satisfies this need, b) satisfies the desire to not do a temporary solution that we change later, and c) moves us toward the desired overall solution (don't have to solve all the header changes etc, make-dist as desired etc)?

morrone:
Yes, I think we can solve that, and without introducing complications that will impact the build system cleanup effort.  I will try to get an example in a PR today.

morrone:
@oceandlr , I started looking at the tree and it looks like you already created a samplers/contrib directory back in March.  I was planning nearly the exact same thing. Can you explain what is lacking in that implementation?

Does it need documentation so that people know how to use that directory?

Do we need an "example" module that does the bare minimum, to make it even easier to get started?

valleydlr:
@morrone Nothing was lacking from that implementation other than an example. The concern was that the top level Makefile.am, configure.ac have to be modified. We will move the contrib directory up (peer of sampler and store) and add the tutorial example. Additionally we will also accept this pull request. This will provide both avenues for contribution and kick the can down the road as they are orthogonal approaches. Appropriate documentation will be included as README's in each "contrib" directory (they will obviously not both be named contrib).

baallan:
I will be simplifying this request to:
-  exclude the tutorial contrib examples @oceandlr  will be putting in simpler form under ldms/contrib
-  rename the configure switches controlling the build of external projects that have been dropped in the tree at ldms/src/third.
- update the concerned documentation files.

This will provide (a) a complete example for independent external projects that we can include in our direct testing, (b) a hook for contributors that want to use something other than automake as their build system, (c) the stub for anyone collecting a "batteries included" ovis source by dropping git submodules (or the equivalent) into ovis/ldms/src/third.

Those preferring to contribute in the "dependent on changing the ovis configure.ac" way, will be able to follow the tutorial examples from @oceandlr.

morrone:
If contributors want to use something other than automake, the it simply shouldn't be in the ldms project.  The ldms project uses automake.  For those that want a different build system, they they should just build externally.

Modifying configure.ac and Makefile.am is something that anyone that has used any autotools project will be able to do.  And if they have never done it before, then this is a learned skilled that will likely prove useful in the future, because it is common with thousands of other projects.

I really think we need to avoid introducing more completely independant build systems into the same source tree.

morrone:
@valleydlr, so I would say that it is completely unreasonable to expect to be able to land something into the ldms project without editing configure.ac and/or Makefile.am.  So if we just drop that unreasonable requirement, we could basically meet the remaining requirements by just bumping the "contrib" directory up one level and improving documentation.  Is that correct?

oceandlr:
@morrone @valleydlr @baallan @oceandlr committed the simple contrib in #88

baallan:
@oceandlr This has been:
* simplified to merely my_plugin/ (no tutorials)
* moved to ldms/src/third-plugins/ only (no lib/src/contrib, assuming we will unify lib/ and ldms/)
* renamed my config options enable-contrib/with-contrib down to just --enable-third-plugins=list per a comment from @morrone .
* better documented.

morrone:
I thought maybe we were going with the "option 1" instead? As seen in PR #88?

baallan:
I suspect, as there is no conflict between the two, we will be going with both, even if (2) serves only as documentation and testing. But it's up to the *dlrs.

morrone:
If a decision hasn't been made, then maybe I should expand on my argument.

I see option 1 in PR #88 as the "Keep It Simple" approach to a contrib directory. This is a method that is simple and instantly familiar to anyone that has worked with autotools (automake/autoconf/libtool) in the past.  And since it is probably safe to say that autotools are the most used build system tools in all of Linuxdom, thats a lot of people that will find this approachable. For anyone that hasn't used autotools before, the barrier can still be low to adding contrib plugins with a little bit of documentation (and that knowledge is knowledge that will aid them on other autotools projects, not just ldms).

Option 2 in PR #72 is the "Not Invented Here" solution.  It seems like nifty solution in search of a problem. It replicates the functionality we get from #88, with a whole lot more complexity, and in a way that is totally unique to the ovis project. This is something that developers new to ovis would need to learn just for ovis. It is not something that anyone would expect to find, and it is not knowledge that is reusable in other projects, like PR #88.

Further, PR #72 complicates the dist tarball.  Instead of a normal straight tarball, the canonical tarball would be a tarball of tarballs.  Again, this is a one-off solution that only applies only to ovis. That is going to be surprising to developers used to working with, and packaging, projects that are built with autotools. So one could argue that it violates the "Principle of Least Astonishment".

The latest suggestion is that PR #72 could serve only as "documentation and testing". I would argue that in order for that to be true, we would have to never let actual plugins land in the tree under the directory it creates. But if we do that, then it really isn't serving a useful purpose. If people want to maintain a plugin external to the ldms repo, then this is not the approach that people are likely to want to use to build those external packages. And if someone requests that, they should probably be directed to use the more common idiom: build and install ovis, then build and install the externally packaged plugin. This is the approach taken by https://github.com/LLNL/ldms-plugins-llnl.  It is the much more common approach to building something external that has a dependency on another package.

If people wish to have their plugin landed in the upstream ldms tree, they can use the PR #88. So again, that doesn't seem to leave #72 with a use case sufficient to justify the added complexity and one-off, non-standard approach it adds to ldms's build system.

As for testing, I am unclear on how #72 would be good for testing. I imagine there are simpler, more common approaches to any testing needs. For instance, if I am testing and iterating on a plugin with a completely separate build system (e.g. ldms-plugins-llnl), there is no way that I want to have ldms recursively control my build system.  That is unnecessary complexity. By keeping ldms-plugins-llnl completely out of the ldms tree, I don't need to get the ldms source code at all. I can simply compile and iterate on _just my plugin_, targeted against the lmds headers and libraries that are installed by .rpm packages.  When I _also_ want to test against a newer ldms, I still have the freedom to install that newer ldms and target my plugin builds against that. It is difficult to imagine a use case where it is truly advantageous to have the plugin independently packaged (in terms of its build system, like autotools), but also built recursively by ldms.

If one desires the the plugin source to be in the ldms tree and built by ldms, one can use the #88 approach with ease.

tom95858:
This has been cherry-picked into OVIS-4

PR #76, opened 2019-11-08
Rename libsampler_base to libldmsd_sampler_base

morrone:
Rename libsampler_base to libldmsd_sampler_base to reduce installed namespace polution.

Part of issue #71.


tom95858:
I know this sounds like a knit, but could you please rename this to libldmsd_sampler_base instead so that it's clear that this is a daemon library?

baallan:
This PR will need to be revised after PR 72 is merged, since all the contrib/ samplers are dependent on the base lib.

morrone:
> I know this sounds like a nit, but could you please rename this to libldmsd_sampler_base instead so that it's clear that this is a daemon library?

Not at all, that is helpful! It is not always clear to me from the current directory structure where the line is drawn between what is considered the ldms "service" and what is considered the "daemon".  The sampler/sampler_base.[ch] files are in a directory parallel to the "ldmsd" directory rather than under it, so I guessed incorrectly that plugins were considered part of the "service".

You name change request will help to make that a little more clear.

PR #78, opened 2019-11-09
Move pedigree files

morrone:
Change the pedigree scripts into text files.  Move the pedigree
files' locations to docdir rather than bindir.

It looks like ovis/lib's doc build was broken, so that is tweaked
as well to allow the ovis lib pedigree file to go in the ovis-lib
doc directory.

Part of issue #71.

### DIFF
--- a/automake/pedigree.am
+++ b/automake/pedigree.am
@@ -1,24 +1,21 @@
-install-exec-local:
+install-pedigree:
 	mkdir -p `dirname $(PEDIGREE)` ; \
 	if [ -f $(PEDIGREE) ]; then \
 		rm -f $(PEDIGREE); \
 	fi ; \
-	echo '#!/bin/bash' > $(PEDIGREE) ; \
-	echo echo commit-id: "$(OVIS_GIT_LONG)" >> $(PEDIGREE) ; \
-	echo echo package: "$(PACKAGE_STRING)" >> $(PEDIGREE) ; \
-	echo echo questions to: "$(PACKAGE_BUGREPORT)" >> $(PEDIGREE) ; \
-	echo echo source location: "$(abs_top_srcdir)" >> $(PEDIGREE) ; \
-	echo echo build location: "$(abs_top_builddir)" >> $(PEDIGREE) ; \
-	echo echo compile hostname: "$$HOSTNAME" >> $(PEDIGREE) ; \
-	echo echo compile cpu: "`uname -p`" >> $(PEDIGREE) ; \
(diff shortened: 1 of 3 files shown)


baallan:
ldms-pedigree is installed as a program so that it is easily found by users and robots. It has a bunch of debugging info content that for some reason (long escaping me) we didn't want baked into the binary. A similar bunch of information was part of the binary in ldms v2. So moving ldms-pedigree off to a .txt in doc or lib directory breaks the 'easily found' requirement.
Since it is named starting with ldms-, i'm not clear why it's part of issue 71.

morrone:
One of the pedigree scripts is named "lib-pedegree", which is the connection to issue #71.  I was just cleaning up both at the same time.  And I don't see why doc is at all hard to find.  And I don't see why this is anything other than an extremely niche use, so the very minor effort of looking in docs for those niche uses seems preferable to me than doing something less than standard in the bin directory.

For the vast majority of users, a properly formed package will provide all needed pedegree information.

baallan:
I agree that lib-pedigree is a rather inauspicious name. If we did away with the two levels of configure (lib/configure.ac, ldms/configure.ac got merged up to ovis/configure.ac) then lib-pedigree would go away.  We would have only ldms-pedigree.

The doc is hard to find because it's under a version number named directory in share (or any other random place depending on configure arguments and the target OS)  instead of in bin/.

The use of pedigree is not niche at all, in our experience supporting new users and new platforms; in most of those cases, there have been source-based deployments long before there are packages. In initial builds, the doc options are often disabled entirely.

morrone:
How about I remove lib-pedigree now?  While they are technically different build systems, they are in the same git repo in one same tree, and (almost?) always built as one. Maybe ldms-pedigree is sufficient then?

Alternatively I could just change lib-pedigree to "ovis-lib-pedigree".

baallan:
I would prefer a PR that consolidates configure.ac, lib/configure.ac and ldms/configure.ac (and the attendant m4/ subdirs) all into the top level configure.ac. Failing that, I have no objection in the interim to the ovis-lib-pedigree rename. Other developer opinions may vary.

morrone:
> I would prefer a PR that consolidates configure.ac, lib/configure.ac and ldms/configure.ac (and the attendant m4/ subdirs) all into the top level configure.ac

I am not opposed to that.  I had been avoiding that larger change on the assumption that change would be unlikely to go into v4. I have been focusing on things that I'm hoping can be used in a v4 release, because we would like to move to that in the near term.

PR #79, opened 2019-11-09
Rename libparse_stat to libldmsd_parse_stat

morrone:
Rename libparse_stat to libldmsd_parse_stat

Part of #71

### DIFF
--- a/ldms/src/sampler/dstat/Makefile.am
+++ b/ldms/src/sampler/dstat/Makefile.am
@@ -1,5 +1,6 @@
 pkglib_LTLIBRARIES =
 lib_LTLIBRARIES =
+noinst_LTLIBRARIES =
 check_PROGRAMS =
 dist_man7_MANS =
 
@@ -14,7 +15,7 @@ COMMON_LIBADD = $(CORE)/libldms.la libparse_stat.la \
 	@LDFLAGS_GETTIME@ -lovis_util -lcoll
 
 if ENABLE_DSTAT
-lib_LTLIBRARIES += libparse_stat.la
+noinst_LTLIBRARIES += libparse_stat.la
 libparse_stat_la_SOURCES = parse_stat.c parse_stat.h
 
 libdstat_la_SOURCES = dstat.c


baallan:
This is not a convenience library.
libparse_stat is intended for use by more than one sampler. renaming it ldms_parse_stat rather than parse_stat is the appropriate fix. The libdstat plugin is dependent on libparse_stat; the parse functions are not copied into the dstat plugin binary.

morrone:
> This is not a convenience library.
> libparse_stat is intended for use by more than one sampler. 

Those two statements seem unrelated to me.  Could you please elaborate? A convenience library fully supports (if not encourages) use by more than one sampler.

morrone:
You probably mean that you want those libraries to be dynamically linked to to parse_stat?

baallan:
I don't mean to say "I want those libraries to be dynamically linked". I'm telling you libparse_stat is dynamically linked by plugins, so it needs to be installed. Run ldd and nm -u on an installed ovis-ldms/libdstat.so and you will see.

morrone:
It will work as I've done it as well. Then it doesn't need to be installed in libdir, where we put things when we desire to share the code _outside_ of the project rather than inside the project.

You sound attached to the current approach though, so I'll go with a rename.

morrone:
Actually, I'm not entirely sure if I was clear: my approach works because the convenience library is statically linked into the plugin, and therefore eliminates a dynamic library at run time.

Sure, plugins would be slightly larger than they need to be, but I think that is reasonable for small support libraries like this.

But in the interest of expediency, I will still submit a patch that just does a rename.

morrone:
I updated the PR to do a rename rather that changing it to a convenience library.

tom95858:
This library is only used by one plugin. I would inclined to not make this a library at all and just link it statically with the plugin that uses it. If in the future another user comes alone we can make a library and either ship it or bind it non-shared.

morrone:
> This library is only used by one plugin. I would inclined to not make this a library at all and just link it statically with the plugin that uses it. If in the future another user comes alone we can make a library and either ship it or bind it non-shared.

That was my original approach in this PR, so I at least am on board with that.

morrone:
OK, the PR has been updated back to the original noinst library/statically linked approach.

tom95858:
This has been cherry-picked into OVIS-4

PR #82, opened 2019-11-10
Relocate json libraries from lib to ldms

morrone:
Move the json and json/json_parser libaries from lib to ldms and
change the libraries from installed libraries to noinst
convenience libraries.  Likewise the associated headers changed
to stay in-tree.The json_test binary is changed from sbin_PROGRAMS
to check_PROGRAMS.

This is in support of issue #71, to clean up the installed
namespace footprint of ldms. It is also aligned with the desire to
eliminate recursive autoconf infrastructure.


tom95858:
This library is being deprecated in favor of json_util.

tom95858:
I think @nichamon has a change somewhere that gets rid of this library altogether. I would like to rename the json_util library though once this is done to something like you suggest. Also, do we have an end-game for all of this packaging wrt G/A of 4.3.3? Can @morrone be at the LDMS UG meeting tomorrow to discuss?

morrone:
Both json_util and the third party json_parser one are moved and made internal-only in this pull request.

Getting rid of the third/json_parser in the future would be great! But is there any reason not to land this in the mean time?  If anything it should be slightly easier to remove after this commit.

Yes, I will try to call in to the meeting on Monday.

tom95858:
You're right, there's no reason we can't just take this and remove the deprecated library as another change.

morrone:
nichamon:get-rid-of-3rd-party-json-lib-v4 in PR #91 is now available, so I rebased this PR onto that as requested.

tom95858:
Hi @morrone, I cherry picked and build tested this change and got the error below.

make[5]: Entering directory `/home/tom/work/ovis-pkg/ovis/build-test/ldms/src/core'
  CC       ldms.lo
  CC       ldms_xprt.lo
../../../../ldms/src/core/ldms_xprt.c:68:28: fatal error: json/json_util.h: No such file or directory
 #include <json/json_util.h>

There are internal (ldmsd), and external (in the works)  applications as well as plugins (slurm) that use this library. So I am inclined to leave the library where it is, rename it libldmsd_json and bind everything shared. Let me know what you think.

tom95858:
@morrone ping....

PR #83, opened 2019-11-11
Various fixes for things missing from the dist tarball

morrone:


ldms/src/store/Makefile.am
```diff
@@ -25,26 +25,31 @@ pkglib_LTLIBRARIES += libstore_amqp.la
 endif
 
 if ENABLE_KOKKOS
-SUBDIRS += kokkos
+MAYBE_KOKKOS = kokkos
```
baallan: @morrone Is this transformation supposed to move 'make dist' toward being configure option independent? don't we need dist_subdirs for that instead? Can we get rid of MAYBE_* instead of extending it in our .am files?

morrone: Yes, this is the approach officially recommened the automake (libtool?) manual.  Automake is smart enough to understand that the subdirectory is conditional when this idiom is used.  It will correctly include it in the dist, and also correctly conditionally build.

No, you don't need DIST_SUBDIRS.  This

oceandlr:
@morrone @nichamon @narategithub
Looks like a problem here:

ldms/src/sampler/contrib/ipmireader/Makefile.am:
EXTRA_DIST += libipmisensors.la
EXTRA_DIST += Plugin_ipmisensors.man

I think this is supposed to be:
EXTRA_DIST += Plugin_ipmireader.man
EXTRA_DIST += Plugin_ipmisensors.man

If so this will need to be fixed and it is causing a conflict in #88

morrone:
@oceandlr, yes that was my mistake, sorry.  I have pushed PR #96 with that fix, and I also rebased your PR#88 commits onto it for you in penance.

morrone:
Landed by cherry-pick.

ldms/src/store/Makefile.am
narategithub: I know that this is already been merged, but just want to be clear. This change is only for Automake coding style right? (I'm guessing that it is from here https://www.gnu.org/software/automake/manual/html_node/Subdirectories-with-AM_005fCONDITIONAL.html#Subdirectories-with-AM_005fCONDITIONAL). The code before the change already distribute the sub directory even though the condition is not true.

morrone: @narategithub, I can't recall about this directory in particular, but there were ones that were fixed by this by having this change.  So it is probably best to be consistent.

narategithub: I totally agree that code should be consistent. I think we consistently use conditional `SUBDIRS+=..` and the sub directories still get distributed even with the feature disabled. The `MAYBE_` introduced the inconsistency.

morrone: Previously not all subdirs were being included in the dist tarball.  That is what led me down this path. If you have an alternate explanation for why that was and how to fix it, I'm certainly willing to listen.

narategithub: Could you please point out which subdirs have the problem? I'm quite sure that it is not kokkos as I just checked out the old branch (without this patch) and ran bare configure and make dist and I saw kokkos sources there.

I'll be also making a list of all non-distributed files in another issue soon, so that we can discuss on which we intentionally left out and which are bugs :)

morrone: @narategithub, well crud, I think you are right. I think I was confused by both kokkos.h and kokkos_sampler.c missing from the dist tarball.  But kokkos.h is just missing from _SOURCES and kokkos_sampler.c may be dead code.  And then I think I changed the others for consistency?  I don't know.  But I'm not seeing it needed anymore either.  Sorry!

Yes, it is great to start a new bug that lists the remaining missing files.  I asked about missing man pages in #101, but there are a number of other files that are still unaccounted for.

narategithub: That's alright :) I too have too many oops that I lost count :sweat_smile: . Thanks for submitting patches though, the `EXTRA_DIST` stuff you put in is a real missing files.

PR #84, opened 2019-11-11
Fix mount point truncation in Cray DVS Sampler

tom95858:
The buffer used to read the mount point was too small and would get truncated.


oceandlr:
@narategithub @tom95858 @valleydlr @nichamon 
Since this has been pulled into master, can we close this issue?

PR #85, opened 2019-11-11
Move ovis_util test_* programs from bin_ to noinst_

morrone:


baallan:
This change will break 'make check' by default because some of the programs have expected fail exits currently.  We don't for the project as a whole yet have a process that leads to 'make check' or 'make installcheck' lead to a nice stream of PASS/XFAIL messages in the occasionally seen autotools way.

morrone:
"make check" isn't _running_ the programs, just as "make" didn't run those test programs when they were installed in the bindir.  So why would it break anything?

The only change in procedure is that when you want to run whatever testing scripts you have, run "make check" in the project to get the test programs built instead of just "make".

If "noinst_PROGRAMS" would make you happier than "check_PROGRAMS", I am certainly willing to shift to that.

baallan:
I suspect installcheck and manual (without 'make check') scenarios can be satisfied with these as shifted to noinst_PROGRAMS.

morrone:
OK, I changed the PR to use "noinst" instead of "check".

PR #86, opened 2019-11-12
Add ifdef check for new EHWPOISON errno

tom95858:
The older errno.h header files do not have this errno. Protect this definition with an ifdef to support the build.


baallan:
+1

oceandlr:
@narategithub @tom95858 @valleydlr 
Since this is now pulled into master, can we close this?

PR #88, opened 2019-11-13
Sampler Store tutorial code and new contrib

oceandlr:
New directories ldms/contrib/sampler and ldms/contrib/store. 
Moved previous ldms/sampler/contrib with ipmi_reader.
Added tutorial sampler and store.


oceandlr:
@morrone @baallan @brandt This is the EZ contrib related to #72

ldms/src/contrib/sampler/tutorial/Makefile.am
```diff
@@ -0,0 +1,29 @@
...
+
+
+if ENABLE_TUTORIAL_SAMPLER
+libtutorial_sampler_la_SOURCES = tutorial_sampler.c $(BASE_SRC)
+libtutorial_sampler_la_LIBADD = $(COMMON_LIBADD)
+libtutorial_sampler_la_CFLAGS = $(AM_CFLAGS) 
+pkglib_LTLIBRARIES += libtutorial_sampler.la
+dist_man7_MANS += Plugin_tutorial_sampler.man
+
+endif
```
morrone: We will want an "EXTRA_DIST = Plugin_tutorial_sampler.man" here, after the "endif".  This is an unfortunate workaround for a known bug in automake.  For things like _SOURCES, automake is smart enough to add the files to EXTRA_DIST automatically.  Sadly it was forgotten to do the same for _MANS.  I don't believe that even the dist_ prefix helps when the _MANS line appears inside of the conditional.

ldms/src/contrib/sampler/Makefile.am
```diff
@@ -0,0 +1,16 @@
+ACLOCAL_AMFLAGS = -I m4
+SUBDIRS = .
+lib_LTLIBRARIES =
+pkglib_LTLIBRARIES =
+
+if ENABLE_IPMIREADER
```
morrone: I would like to encourage folks to start using the following pattern for conditional subdirs:

if COND_OPT
  MAYBE_OPT = opt
endif
SUBDIRS += $(MAYBE_OPT)

So this one would be:

if ENABLE_IPMIREADER
  MAYBE_IPMIREADER = ipmireader
endif
SUBDIRS += $(MAYBE_IPMIREADER)

When this pattern is followed, automake is able to set DIST_SUBDIRS on its own automatically. This is the recommended pattern from the automake manual:

https://www.gnu.org/software/automake/manual/html_node/Subdirectories-with-AM_005fCONDITIONAL.html#Subdirectories-with-AM_005fCONDITIONAL

ldms/src/contrib/store/tutorial/Makefile.am
```diff
@@ -0,0 +1,28 @@
...
+ldmsstoreincludedir = $(includedir)/ldms
+ldmsstoreinclude_HEADERS = 
+
+if ENABLE_TUTORIAL_STORE
+libstore_tutorial_la_SOURCES = store_tutorial.c
+libstore_tutorial_la_CFLAGS = $(AM_CFLAGS)
+libstore_tutorial_la_LIBADD = $(STORE_LIBADD)
+pkglib_LTLIBRARIES += libstore_tutorial.la
+dist_man7_MANS += Plugin_store_tutorial.man
+endif
```
morrone: EXTRA_DIST for Plugin_store_tutorial.man outside of the conditional.

ldms/src/contrib/store/Makefile.am
```diff
@@ -10,6 +10,6 @@ do_subst = @LDMS_SUBST_RULE@
 	$(do_subst) < $< > $@
 	chmod 755 $@
 
-if ENABLE_IPMIREADER
-SUBDIRS += ipmireader
+if ENABLE_TUTORIAL_STORE
+SUBDIRS += tutorial
 endif
```
morrone: Use new "MAYBE" pattern.

morrone:
I don't think I marked all of the spots that need the "MAYBE" conditional pattern or EXTRA_DIST for conditional man pages. It would be good to check all the files if you refresh this.

oceandlr:
@morrone @baallan I think these look straightforward and I can get them in tonight 11/14 before they get pulled into the test/integration branch

ldms/src/contrib/sampler/tutorial/Makefile.am
oceandlr: @morrone I don't understand why this would go after the endif. If the "if" isn't met, there wouldn't be this additional file.

ldms/src/contrib/store/Makefile.am
oceandlr: I have changed these in the files related to contrib. I have held off doing this in other Makefile.am's since I wanted to localize the change. I can do these all in another change if that is the generally accepted route. @morrone @baallan @tom95858 @narategithub

ldms/src/contrib/sampler/tutorial/Makefile.am
morrone: We want all source files to appear in the "make dist" tarball regardless of whether the part was selected for creation at build and install time.  For instance, even though the _SOURCES right above this are inside the "if", those files _are_ included in the "make dist" tarball even when the "if" is not met.  autoconf should be doing the same with _MANS, but it has a bug and we need to help it.  EXTRA_DIST will accomplish that.

ldms/src/contrib/store/Makefile.am
morrone: Right, I was only suggesting you change the ones introduced by your commit, not all of the ones in the source tree.  I have a PR already out there to fix up most of the other locations in the tree.

ldms/src/contrib/sampler/tutorial/Makefile.am
oceandlr: Since its outside of the endif, if the user specifies a file that does not exist (e.g., EXTRA_DIST = FOOBAR.man), the make dist will fail. Are we ok with this? How much are we requiring in contrib plugins vs main line plugins? config is sane, but we (main line developers) aren't guaranteeing the plugin will actually work? @morrone @baallan @tom95858 @narategithub @valleydlr @nichamon

ldms/src/contrib/sampler/Makefile.am
oceandlr: updated commit to include this @morrone

ldms/src/contrib/store/Makefile.am
oceandlr: updated in commit

oceandlr:
@oceandlr is closing this, since it was rebased and pulled in in #96

PR #90, opened 2019-11-13
Get rid of the 3rd-party JSON lib dependency

nichamon:
ldmsctl relied on a 3rd-party JSON lib. The change modified ldmsctl so
that it uses OVIS json library in lib/src/json.


nichamon:
I created the pull request for the wrong branch.

PR #92, opened 2019-11-14
ldmsd: plugn_sets_handler returns a corrupted message if no plugin sets

nichamon:


oceandlr:
@nichamon @narategithub @tom95858 @valleydlr 
Since this has been pulled into master, can we close this issue?

PR #96, opened 2019-11-15
Contrib rebase

morrone:
This fixes a bug in EXTRA_DIST from my earlier commit, and includes a rebase of PR #88 onto it.


oceandlr:
Rebase penance? Nice! 

@tom95858 @narategithub @valleydlr we will need the fix part of this for the current master, but not necessarily the contrib part (what was #88). Do we want to get this all in before SC? Or do we want to split out only the fix part that we need?

@morrone

morrone:
Yes, I'm happy to split it back up again, if that is what is decided.

valleydlr:
@morrone @oceandlr @narategithub @tom95858 Lets take #96 as is prior to SC as it has no impact on the core.

oceandlr:
@morrone @narategithub @tom95858 
@oceandlr merged this.

PR #107, opened 2019-11-19
Remove the 'fabric' transport from the slurm notifier man page

nichamon:


tom95858:
Let's leave this in. The feature exists, but is experimental.

PR #109, opened 2019-11-21
Add version string in ldmsd for binary string dump check

narategithub:
`ldmsd -V` shows all the version numbers that we need, but `ldms-test`
has a use case that needs to determine the version of the binary in the
environment that could not execute that binary. This patch will make
"LDMSD_VERSION #.#.#" presents in `ldmsd` and could be check with
`strings /opt/ovis/sbin/ldmsd | grep LDMSD_VERSION`.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -116,6 +116,9 @@ pthread_mutex_t log_lock = PTHREAD_MUTEX_INITIALIZER;
 size_t max_mem_size;
 char *max_mem_sz_str;
 
+/* NOTE: For determining version by dumping binary string */
+char *_VERSION_STR_ = "LDMSD_VERSION " OVIS_LDMS_VERSION;
+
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 I built with `-Wall -Werror` and tested with `ldms-test/agg_test`

tom95858:
Did you confirm that the linker does not discard the string given that it's not used? You should make this value stays around if you compile -O3.

narategithub:
Let me make sure with '-O3'. Will let you know later.

narategithub:
Turing on '-O3 -Wall -Werror' uncovers more use-before-init errors :( I'll post the patch in a bit.

narategithub:
@tom95858 the string has survived `-O3`.

Please also see #113 for more compilation warning fix that somehow don't show up with `-O0`.

narategithub:
@tom95858 p.s. The `github/master` + this patch + #113 has been tested with `ldms-test/papi_store_test`.

narategithub:
FYI, current master + this patch (#109) + #113 also passed all tests in `ldms-test`.

Issue #108, opened 2019-11-20
Files missing from dist tar ball


narategithub:
These are the files in the source tree not included in the dist tar ball. (See script that generate this list at the end). Let's discuss what to do with each of these (e.g. remove from source tree if they are abandoned, include into dist, or intentionally not distribute).

```
-0_AVOID_GITHUB_ZIP_FILES/README.md
-automake/pedigree.am
-.gitignore
-.gitmodules
-helper-scripts/autogen.sh
-helper-scripts/configure.ac
-helper-scripts/Makefile.am
-helper-scripts/scripts/ldmsd_sampler_services
-helper-scripts/scripts/Makefile.am
-helper-scripts/scripts/ovis_head_services
-helper-scripts/scripts/ovis_services
-kernel/kldms/ghal.c
-kernel/kldms/ghal.h
-kernel/kldms/.gitignore
-kernel/kldms/gpcd_kapi.h
-kernel/kldms/gpcd_lib.h
-kernel/kldms/gpcd_pub.h
-kernel/kldms/gpcdr.c
-kernel/kldms/gpcdr_pub.h
-kernel/kldms/Kbuild
-kernel/kldms/Kconfig
-kernel/kldms/kgpcdr.c
-kernel/kldms/kldms.c
-kernel/kldms/kldms.h
-kernel/kldms/kldms_ibm_pwr.c
-kernel/kldms/kldms_vmstat.c
-kernel/kldms/Makefile
-ldms/config/.gitignore
-ldms/etc/init.d/ldmsd_aggregator.in
-ldms/etc/init.d/ldmsd_sampler.in
-ldms/etc/init.d/Makefile.am
-ldms/etc/logrotate.d/ldmsd_aggregator
-ldms/etc/logrotate.d/ldmsd_sampler
-ldms/etc/logrotate.d/Makefile.am
-ldms/examples/ldmsctl/flatfile.src
-ldms/examples/ldmsd-rotate-log.sh
-ldms/examples/set_array/.gitignore
-ldms/examples/set_array/set_array_agg.cfg
-ldms/examples/set_array/set_array_sampler.cfg
-ldms/examples/set_array/start_agg.sh
-ldms/examples/set_array/start_sampler.sh
-ldms/examples/update-auto-interval/agg.conf
-ldms/examples/update-auto-interval/agg-disable-auto-interval.conf
-ldms/examples/update-auto-interval/env.sh
-ldms/examples/update-auto-interval/samplerd.conf
[...]

PR #112, opened 2019-11-21
merge configures to top (complete)

baallan:
I have tested this on rhel7 (dist, distclean, distcheck, uninstall, and build from dist tar file).
It should also be applicable to 4.3.3, but isn't strictly necessary.
If there are not quick objections from other's testing, it should be merged.

Note: re issue #108, other than a few missing extra_dist that are clearly useful (which I fixed), the items listed by narate are missing in dist because they are not built or installed or documented. Fixing these is out of scope for just merging configure.ac.

This consolidates the 3 configure.ac files into one and fixes collateral damage such as redundant installed files like ovis-lib-* vs ovis-ldms-* and  between formerly two pkglibdirs

There are no C level algorithmic or API changes. There are obviously changes that will simplify (and require change) in downstream packaging tools. Other than doc tree and python lib, there are no changes affecting installed library locations if in the old version both lib and ldms were configured with the same pkglibdir name.
  
Updating a small number of include (ovis-ldms-config.h) or import statements (ovis_lib->ovis_ldms) detectable with build, grep or trivial testing is included.  
```
OVIS_*_LIB*DIR and OVIS_*_INCDIR have been replaced with common
 OVIS_LIB_ABS and OVIS_INCLUDE_ABS in all makefile.am to guarantee
 consistency even when system-installed ldms exists.
```
Refactoring/merging of complex bits, like the doxygen trees of lib, ldms is not included, as this should probably be considered only after we consolidate lib/src with ldms/src and reorganize (shrink) the number of tiny support libraries.


baallan:
I anticipate testing this on toss3 and the OS for bluewaters.
@valleydlr can you test (or delegate to someone other than me) testing for CLE6?
@morrone please review.
@tom95858 can you tag someone at ogc to run your tests, particularly related to python?

Makefile.am
```diff
@@ -76,3 +69,8 @@ TAG.txt:
 		TAG="NO_GIT"; \
 	fi ; \
 	echo "$$TAG" > TAG.txt
+
+W=@OVIS_INCLUDE_REL@
+X=@OVIS_INCLUDE_ABS@
+Y=@OVIS_LIB_REL@
+Z=@OVIS_LIB_ABS@
```
morrone: Were these stray testing lines that need to be removed now?

baallan: Thanks for the catches. I repushed with all the suggested fixes, except helper-scripts which has always been an orphan; documented that.

baallan:
I've replaced OVIS_EXEC_SCRIPTS with ac_config_files almost as suggested (cleaner version) and fixed some lib dependency bugs that broke parallel build. More platform testing and verification of make dist will be useful. I have not verified all the suggested reductions to extra-dist work as expected yet.

baallan:
I  added the other extradist changes from @morrone after more testing.

Unrelated to the @morrone patch, a comparison of the dist tar content to the original source reveals the following differences, which I will look into resolving (what is missing? what is extra? what is just cruft?)

MISSING in dist, or from .in files (verify) or cruft that should be elsewhere
README.md
helper-scripts/*
kernel/kldms/*
etc/*
ldms/src/ldmsd/ldmsd_upload.c
ldms/src/sampler/lustre/fnv_hash.h, llite1_keys.txt
ldms/src/sampler/papi/tests
ldms/src/sampler/shm/mpi_profiler/wrap/wrap.py
ldms/src/sampler/shm/shm_util/Makefile
ldms/src/sampler/spank
ldms/src/sampler/syspapi/test
ldms/src/sampler/switchx_eth.c
ldms/src/swig/create_set_test.py
ldms/src/swig/pytest.py
ldms/src/test/failover-*
lib/bin/src/map_cmd.c ,(makefile.am?)
lib/etc/ovis/log-rotate.in
lib/python/
lib/src/coll/map.h,map_priv.h,libmap.c
lib/src/ovis_util/rmaninfo.[ch]
packaging/*
util/aries
util/mmr_reader
util/cray_specific
util/failover
util/gemini
util/msr
util/sample_baler_scripts
util/sample_init_scripts/genders/examples/genders.shaun
util/sample_init_scripts/genders/systemd/etc/sysconfig/ldms.d/ldmsd.*.conf
util/sample_init_scripts/genders/sysv/etc/sysconfig/ldms.d* (symlink?)
util/sample_init_scripts/genders/test

Extra in dist tree:
ldms/config/*
lib/config/*
lib/src/ovis-lib-configvars?

morrone:
FYI, there is a list of the missing dist files in #108.  It doesn't address the extras that shouldn't be there (which includes some .in files).

configure.ac
```diff
@@ -1,41 +1,628 @@
 dnl Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.63)
-AC_INIT(ovis, 4.3.3, ovis-help@sandia.gov)
+AC_INIT(ovis-ldms, 4.3.3, ovis-help@sandia.gov)
```
morrone: Do we want to rename the package in this PR?  This means that the dist tarball will be named "ovis-ldms-X.X.X.tar.gz" int the future. Are packages expected to rename as well?

baallan: i would prefer to leave that for another time. i’m in the midst of cleaning up and adding spec files that handle the codes as software collections without double full configure.

morrone: What I mean is that this PR _does_ change the package name, and perhaps that should have been left to some other time.  If this PR introduces a name change, then we probably need to discuss the implications now.  If that was accidental and the PR will be updated to not change the name, then we can postpone that discussion for the future.

baallan: Ah, ok. yes, the merge of ovis-lib, ovis-ldms, and ovis all to ovis-ldms (in terms of autoconf package declarations) is to preserve all the existing third-party init scripting uses of the ldms pkglibdir location /usr/lib64/ovis-ldms where plugins are found. If we want to change that pkglibdir, I think that is something that should only occur in v5.

morrone: I see. OK, that is probably fine then for v4.  It is probably too much trouble to introduce a "alternatepkglibdir" everywhere.  We'll just deal with the dist tarball being misnamed for now.

baallan: One could argue that ovis-ldms is the correct name for our github (ovis-hpc) ldms product. Ovis is a project with a group of products (ldms, baler, sos, several others historically) and if anyone forks ldms, then those here remain ovis-ldms, ovis-sos, etc and someone may choose to release/package xyz-ldms. Obviously, we'd prefer everything developed is contributed to the community version instead of forks appearing.

baallan:
@tom95858 @narategithub @oceandlr So is master on github now pre-v5? Or is some future magic planned such that version 4.100.x will become 4.4 at the next 4 release?

@morrone @tom95858 Should we be landing the config cleanup pull request outstanding on OVIS-4 branch to facilitate llnl/toss packaging, or only on master, or both?

Due to the force push on master, there's now a ton of conflicts to my PR.

tom95858:
> @tom95858 @narategithub @oceandlr So is master on github now pre-v5? Or is some future magic planned such that version 4.100.x will become 4.4 at the next 4 release?

master is ovis-next. There is  no version specified yet.

> 
> @morrone @tom95858 Should we be landing the config cleanup pull request outstanding on OVIS-4 branch to facilitate llnl/toss packaging, or only on master, or both?

It should go on both:

 OVIS-4 is the v4.x.y development branch

master is the ovis-next development branch


> 
> Due to the force push on master, there's now a ton of conflicts to my PR.

master now points to a completely different rev history, i.e. ovis-next

This pull request should probably be abandoned and resubmitted as two separate pull requests, one on OVIS-4, and another (probably very different) on master

For OVIS-4, what I had been doing with pull requests that had conflicts was to cherry-pick the commits from the pull request into OVIS-4, fixing the conflicts, pushing the result to OVIS-4, and then closing the pull request.

baallan:
replaced with PR 130

PR #130, opened 2020-01-23
merge configures to top for v4

baallan:
This consolidates the 3 configure.ac files into one and fixes collateral damage such as redundant installed files like ovis-lib-* vs ovis-ldms-* and between formerly two pkglibdirs. M4 code is consolidated to the top.

There are no C level algorithmic or API changes. There are obviously changes that will simplify (and require change) in downstream packaging tools. Other than doc tree and python lib, there are no changes affecting installed library locations if in the old version both lib and ldms were configured with the same pkglibdir name.

Updating a small number of include (ovis-ldms-config.h) or import statements (ovis_lib->ovis_ldms) detectable with build, grep or trivial testing is included.
Change OVIS_LIB_HAVE_AUTH to OVIS_LDMS_HAVE_AUTH in any c conditionals.

 OVIS_*_LIB*DIR and OVIS_*_INCDIR have been replaced with common
 OVIS_LIB_ABS and OVIS_INCLUDE_ABS in all makefile.am to guarantee
 consistency even when system-installed ldms exists.

Refactoring/merging of complex bits, like the doxygen trees of lib, ldms is not included, as this should probably be considered only after we consolidate lib/src with ldms/src and reorganize (shrink) the number of tiny support libraries.

The helper-script configure.ac and makefile has never been tied to the top; perhaps these scripts should be relocated or eliminated since they make invalid assumptions (relative to autotools) about sysconfdir. In any case, helper-scripts is omitted from this refactoring.

For libraries built in the same directory, using the form libxyz.la for lib adds instead of -lxyz, so that local dependencies are correctly computed automatically.


baallan:
@tom95858 is ogc satisfied with this or do you still have testing in progress?

tom95858:
Your contrib patch broke the packaging build for LBL. I think this patch sits on top of that. I'll pull this in and then rebase Eric's fix on top of that.

PR #132, opened 2020-01-27
Markdown link fix

narategithub:
TLDR; Make internal hyper links in Markdown files work on GitHub, GitLab
and in Doxygen-genarated HTML.

Historically, Doxygen does not understand relative document location in
the Markdown source file. It understands only a path relative to the
doxygen program working directory (which is the top-dir of the project).
GitLab on the other hand understands both document relative paths and
top-dir relative paths. As such, all internal document links in the
Markdown files are relative to the top-dir.

Unfortunately, GitHub only understands relative paths, but not top-dir
relative paths. This breaks document browsing from `README.md` and
`ldms/README.md` on GitHub.

This patch address the problem as follows:
  - Change all internal hyperlink references to be relative to its
    current file.
  - Apply a Doxygen input filter script to expand the relative paths to
    top-dir relative paths.

### DIFF
--- a/ldms/doc/mdlinkfilter.py
+++ b/ldms/doc/mdlinkfilter.py
@@ -1,0 +1,64 @@
+#!/usr/bin/python
+
+import os
+import io
+import re
+import sys
+
+if __name__ != '__main__':
+    raise Exception("This is not a module")
+
+execfile(os.getenv('PYTHONSTARTUP', '/dev/null'))
+
+# inp = sys.stdin.read()
+path = sys.argv[1]
+DIR = os.path.realpath(os.path.dirname(path))
+TOP = os.path.realpath(sys.path[0] + '/../../')
+inp = open(path).read()
+INP = inp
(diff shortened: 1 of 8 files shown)


narategithub:
@valleydlr @tom95858 

Right now the internal document hyper links in Markdown files (e.g. https://github.com/ovis-hpc/ovis/blob/master/ldms/README.md) are broken on GitHub. This patch fix the issue. I have tested that this now works on GitHub, GitLab and Doxygen-generated HTML.

ldms/doc/mdlinkfilter.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python
```
baallan: @narategithub can you update this to "#! /usr/bin/env python3" so that mdlinkfilter will work on both rhel7 and rhel8?
may be  a few py3 syntax changes needed to do this.

PR #136, opened 2020-02-04
zap_sock: return wrong maximum message size

nichamon:


nichamon:
@tom95858  I didn't use `getsockopt` because there is no socket file descriptors when the `max_msg` value is assigned in `zap_transport_get`. Moreover, the `zap_max_msg` API directly returns the value of `zap->max_msg`.

baallan:
@tom95858 do we need the zap_sock patch for v4 also? We're about to start large scale testing and need to know.

tom95858:
Please confirm that zap_max_msg matches the socket buffer size for the transports that use a socket. IOW if the transport reports zap_max_msg of 1M, then setsockopt should be used to set the sockbuf size to match when it is created in connect, and to resize it as part of accept handling. We should also confirm that this size matches the recv buf size for the RDMA transport.

lib/src/zap/sock/zap_sock.c
```diff
@@ -2030,8 +2030,8 @@ zap_err_t zap_transport_get(zap_t *pz, zap_log_fn_t log_fn,
 	if (!z)
 		goto err;
 
-	/* max_msg is unused (since RDMA) ... */
-	z->max_msg = (1024 * 1024) - sizeof(struct sock_msg_hdr);
+	/* max_msg is used only by the send/receive operations */
+	z->max_msg = (1024 * 1024) - sizeof(struct sock_msg_sendrecv);
```
tom95858: Whatever we set this too, i.e. 1M - sizeof(struct sock_msg_sendrecv), the sockbuf size should be set to 1M.

nichamon: I just pushed more changes. With the changes, the send & receive sockbuf sizes are set to 1M in connect and when handling a connect request.

PR #137, opened 2020-02-04
json_util copies a va_list before it is to be reused in jbuf_append_va.

nichamon:


baallan:
It appears this also applies to OVIS-4.

tom95858:
I think the description is misleading. I think you mean to say that the va_list _must_ be copied if it is to be reused in subsequent calls to jbuf_append_va.

baallan:
@tom95858 please cherry pick this back to the OVIS-4 branch if it applies.

PR #140, opened 2020-02-06
fix missing ptr deref in string operation.

baallan:
fixes potential filesingle config parser array overrun.


baallan:
I see someone already fixed this in ldmsd-samplers on master, but it didn't get mentioned.

PR #142, opened 2020-02-06
zap fabric post_wr rc v4

baallan:
the case is not expected to occur ever, but returns random stack if not fixed.


baallan:
@tom95858 @narategithub is there a better rc value or log message if the unexpected happens?
Or is this case expected but merely a no-op? A quick scan of zap_fabric.c made me think this is an unexpected case.

tom95858:
This is fixed in commit f12867fbb30fc2c31d43a433f493221974c03b1a

PR #143, opened 2020-02-06
Json util new uninit return in v4

baallan:
This patch returns null instead of random stack, and handles the build with -DNDEBUG.

The same problems exist in master; this should be ported to master.

### DIFF
--- a/lib/src/json/json_util.c
+++ b/lib/src/json/json_util.c
@@ -215,7 +215,7 @@ json_entity_t json_entity_new(enum json_value_e type, ...)
 	double d;
 	char *s;
 	va_list ap;
-	json_entity_t e, name, value;
+	json_entity_t e = NULL, name, value;
 
 	va_start(ap, type);
 	switch (type) {


lib/src/json/json_util.c
```diff
@@ -259,9 +259,11 @@ json_entity_t json_entity_new(enum json_value_e type, ...)
 		e = json_dict_new();
 		break;
 	case JSON_NULL_VALUE:
+		e = NULL;
```
tom95858: I would prefer if we init 'e' at the top of the function to shut up the static analysis tool, and remove this line and the break below so it falls through to the assert

lib/src/json/json_util.c
```diff
@@ -259,9 +259,11 @@ json_entity_t json_entity_new(enum json_value_e type, ...)
...
 		break;
 	default:
 		assert(0 == "Invalid entity type");
+		e = NULL;
```
tom95858: If we init at the top, this doesn't need to be here.

lib/src/json/json_util.c
baallan: are we sure that logically json_null_value never occurs at this point in the code even from malformed input?

baallan:
I've redone this to init e=null and leave the rest alone.

tom95858:
I think the correct fix is to initialize like you've done and then in teh JSON_NULL_VALUE case

baallan:
I suggest we merge this patch and someone more familiar with testing json_util can make the suggested improvement in null_value handling.

tom95858:
If you want to fix the code. Fix the code. I see no value in quieting a static analysis tool that introduces new bugs.

tom95858:
OK, this has been refactored as follows and pushed to OVIS-4

Fixed the uninitialized return value, implemented the JSON_NULL_VALUE entity type,
and removed the unused 'err' label.

PR #144, opened 2020-02-07
zap_rdma returns a wrong max_msg size.

nichamon:


nichamon:
@tom95858 I have tested this on cygnus by sending application messages that used up the max message length.

PR #145, opened 2020-02-07
zap_ugni returns a wrong max_msg size.

nichamon:


nichamon:
@tom95858 I marked this as a draft because currently I have no way to compile and test the code.

PR #148, opened 2020-02-10
json_util copies a va_list before it is reused in jbuf_append_va.

nichamon:


nichamon:
@tom95858 This patch is a compliment to #137 but for OVIS-4.

PR #152, opened 2020-02-14
Zap sock returns a wrong maximum message size

nichamon:


nichamon:
This is a complement of #136 for OVIS-4.

PR #153, opened 2020-02-14
Zap_ugni returns a wrong maximum message size

nichamon:


nichamon:
The patch is the same as #145 but for OVIS-4.

nichamon:
I haven't tested whether the code can be compiled or not.

PR #146, opened 2020-02-07
zap_fabric returns a wrong max_msg size

nichamon:

PR #154, opened 2020-02-14
Zap_fabric returns a wrong maximum message size

nichamon:


nichamon:
The patch is the same as #146 but for OVIS-4.

nichamon:
I haven't compiled the patch yet.

PR #155, opened 2020-02-14
zap_rdma returns a wrong maximum message size.

nichamon:


nichamon:
This is the same as #144 but for OVIS-4.

PR #174, opened 2020-02-27
add missing error message when edac sampler disables itself

baallan:


ldms/src/sampler/llnl/edac.c
```diff
@@ -390,15 +390,20 @@ static int sample(struct ldmsd_sampler *self)
...
 		if (myFile == NULL)
 		{
-			msglog(LDMSD_LERROR, SAMP ": failed to open file\n");
+			msglog(LDMSD_LERROR, SAMP ": failed to open file %s\n",
+					command[i]);
 			rc = EINVAL;
 			edac_valid=0;
 			goto out;
 		}
+		errno = 0;
```
tom95858: Why are you setting errno to zero here?

ldms/src/sampler/llnl/edac.c
```diff
@@ -390,15 +390,20 @@ static int sample(struct ldmsd_sampler *self)
...
 		s = fgets(lineBuffer, sizeof(lineBuffer), myFile);
+		int en = errno;
```
tom95858: I'm a bit stumped by this paradigm. Why not just use errno instead of assigning it to a local?

baallan: Can one reasonably assume that msglog arg setup is not going to mess with errno before the argument at the position of en is used? If so, then you are correct that en is unnecessary and is in this case just seat-belt and suspenders.

ldms/src/sampler/llnl/edac.c
baallan: that's a good question. it should be higher up since (at least per man) only sscanf and fopen use errno rather than fgets. will revisit all the error reports here since they are mushing all cases in to einval.

baallan:
@tom95858 this has been revised after carefully perusing the man pages.

PR #177, opened 2020-02-27
The -a option at command-line doesn't work.

nichamon:


ldms/src/ldmsd/ldmsd.c
```diff
@@ -1571,6 +1571,20 @@ ldmsd_listen_t ldmsd_listen_new(char *xprt, char *port, char *host, char *auth)
 	return NULL;
 }
 
+static int __set_default_auth()
+{
+	ldmsd_auth_t auth_dom;
+	int rc = 0;
+	auth_dom = ldmsd_auth_new_with_auth(DEFAULT_AUTH, auth_name, auth_opt,
```
tom95858: Why isn't this just called ldmsd_auth_new? What does the extra _with_auth mean? I don't see an ldmsd_auth_new interface anywhere that would collide.

Also, in my checkout of this code, DEFAULT_AUTH == " _DEFAULT_AUTH_ ". This looks like a configuration issue of some kind, where we're not getting a reasonable default defined by configure.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -2020,6 +2034,9 @@ int main(int argc, char *argv[])
 	if (do_kernel && publish_kernel(setfile))
 		cleanup(3, "start kernel sampler failed");
 
+	if (__set_default_auth())
```
tom95858: I don't understand this. The 'a' option sets auth_name correctly. It's just not used in the call to ldmsd_listen_new on line 2039. Why not just call ldmsd_auth_new with auth_name and auth_opt instead of NULL, NULL?

Also, I don't see this fixing the issue where -a is _not_ specified and DEFAULT_AUTH is garbage.

ldms/src/ldmsd/ldmsd.c
nichamon: @narategithub: please respond to @tom95858 's questions.

ldms/src/ldmsd/ldmsd.c
nichamon: > The 'a' option sets auth_name correctly. It's just not used in the call to ldmsd_listen_new on line 2039. Why not just call ldmsd_auth_new with auth_name and auth_opt instead of NULL, NULL?

__set_default_auth actually creates an auth configuration object. The name of the object is " _DEFAULT_AUTH_ ", the authentication plugin is 'auth_name', and 'auth_opt' is the attribute value list of the auth plugin. 

`ldmsd_listen_t ldmsd_listen_new(char *xprt, char *port, char *host, char *auth);`

The auth argument to ldmsd_listen_new() is the configuration object name, not the auth plugin name, so I cannot pass auth_name to the call.

Also I need to wait until all -A options (auth options) are parsed before I can create the default auth config object. Therefore, the call to __set_default_auth could be anywhere from Line 1794 and before the -x options are processed. I put it here because it is right before the -x options are processed.

The name __set_default_auth is misleading I think. '__create_default_auth' may be better.

@narategithub intentionally added the prefix and suffix spaces to prevent name-collision with user-specified auth config objects. " _DEFAULT_AUTH_ " is a special auth config object name. Any listening endpoint and producers that users don't specify an auth config object will use the " _DEFAULT_AUTH_ " configuration object.

> Also, I don't see this fixing the issue where -a is not specified and DEFAULT_AUTH is garbage.

DEFAULT_AUTH value is " _DEFAULT_AUTH_ " which is the name of the default auth config object and `auth_name` is initialized to "none". If -a is not specified, when `__set_default_auth` is called the auth config object " _DEFAULT_AUTH_ " will use the "none" auth plugin.

Moreover, if the `auth` argument to `ldmsd_listen_new()` function is NULL, the function will use the " _DEFAULT_AUTH_ " config object automatically. Anyway I think it is less confusing if I pass DEFAULT_AUTH when I call `ldmsd_listen_new` here.

ldms/src/ldmsd/ldmsd.c
narategithub: I went with `ldms_auth_new_with_auth()` to follow the other config objects signatures (e.g. prdcr_new_with_auth()). Our current use case does not have other credentials creating the auth object. So, I'm fine with refactoring this to just `ldms_auth_new()`.

The `DEFAULT_AUTH` is the key (name) of the default auth object. I use `" DEFAULT_AUTH "` so that it is guaranteed to not collide with user-specified names. The `-a` CLI option will set the `plugin` of the `DEFAULT_AUTH`, and the `-A` CLI options will set the auth options of the `DEFAULT_AUTH`.

nichamon:
@tom95858 I made some minor changes to the patch for ease of read. 
* Rename `__set_default_auth` to `__create_default_auth`.
* Pass `DEFAULT_AUTH` to `ldmsd_listen_new()` when LDMSD processes -x.
* Remove the unused code in `listen_handler()`.

PR #181, opened 2020-03-03
Put `job_set` reference when deleting the sampler

narategithub:
The `job_set` handle was not properly put down when the sampler instance
is deleted.


baallan:
please cherry pick/backport to OVIS-4 if the same bug exists there.

narategithub:
This is only in v5.

PR #182, opened 2020-03-03
Change the `samp->delete_set()` interface to use set name.

narategithub:
The sole real use case of `samp->delete_set()` is in `app_sampler`.
The way it did before the patch was that when it received a task exit
event, it constructed a set name associated with the task and obtained
the `ldms_set_t` handle by `ldms_set_by_name()` (as `samp->delete_set()`
requires a handle to delete). Since this handle is different than the
one in the `samp->set_list` (created by `samp->create_set()`),
`samp->delete_set()` already has to iteratively compare the name of the
given `ldms_set_t` against those in the `samp->set_list` (please see
commit-id: 0f9f9dcb4). So, `samp->delete_set()` with set name makes more
sense from this use case.

Also, with -D_REF_TRACK_ compilation flag to track the correct reference
usage, we found out that all `ldms_set_t` were automatically removed
from the `set->local_rbd_list` in `ldms_set_delete()`. Hence, the use
case of

```C
s = ldms_set_by_name(name);
ldms_set_delete(s);
ldms_set_put(s);
```

in app_sampler will remove `s` twice from the list (in ldms_set_delete()
and in ldmse_set_put()), leading to memory corruption.

This patch is a work around to the above mentioned problem and is a
short-term solution. A real solution regarding LDMS set reference will
be handled separately.

### DIFF
--- a/ldms/src/ldmsd-samplers/app_sampler/app_sampler.c
+++ b/ldms/src/ldmsd-samplers/app_sampler/app_sampler.c
@@ -1223,7 +1223,6 @@ int __handle_task_init(app_sampler_inst_t inst, json_entity_t data)
 int __handle_task_exit(app_sampler_inst_t inst, json_entity_t data)
 {
 	ldmsd_sampler_type_t samp = LDMSD_SAMPLER(inst);
-	ldms_set_t set;
 	int len;
 	json_entity_t job_id;
 	json_entity_t task_pid;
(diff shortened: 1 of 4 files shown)


baallan:
Will the anticipated 'real solution' be applicable to v4, or an incompatible change? And same question for this patch.

narategithub:
The `samp->delete_set()` only existed in v5.

narategithub:
Also, we don't do `ldms_set_put()` after `ldms_set_delete()` in v4. However, after `ldms_set_delete()` all set handles became invalid in all threads. I believe codes in v4 is aware of this.

I think in v5, the way it should be is that the reference should not be invalidate by others, and should be properly put down by the owner (the one that asked for the reference).

PR #185, opened 2020-03-04
Fix ldmsd default auth processing

narategithub:
The default authentication object was set up after the configuration
file has done processing, but `prdcr_add` in the config file that did
not specify `auth` needed the default auth object to create the `prdcr`.
Hence, the default auth object (and its attributes) should be set up
during config file processing, when we knew that `set CLI` options are
done. If no config file is given, the default auth is set up in the
`ldmsd_init()` as usual.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -1370,6 +1370,23 @@ void handle_pidfile_banner()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


baallan:
is this new to v5, or something that applies in some form to v4 also?

narategithub:
This is new in v5

PR #188, opened 2020-03-11
ldms_ls: a typo when allocating memory & cause string overflow at print

nichamon:


nichamon:
I think this is the cause of the bizarre ldms_ls behavior that sometimes it returns 'the set doesn't exist' or 'segfault' when a set instance name is given.

@valleydlr has locally applied the fix on Voltrino and the symptom went away.

PR #198, opened 2020-03-20
fixes for ncsa-snl demo

baallan:
This pr has several independent fixes to enable the snl-ncsa ui pipeline. each commit is independent, except the two modifications to the function_csv store.


baallan:
@gentile this has been run through valgrind memory and thread checkers using the included testing scripts.

PR #209, opened 2020-03-30
Fix store_amqp build error due to amqp interface change

narategithub:


narategithub:
store_amqp in v5 does not have this issue (addressed when it was ported to v5).

PR #210, opened 2020-04-01
Remove OVIS_GIT_SHORT macro

narategithub:
Remove OVIS_GIT_SHORT macro and TAG.txt. No one is using it.


narategithub:
NOTE: This has been tested with `make dist` and install from the dist tarball. The OVIS_GIT_LONG in ovis-ldms-config.h is also inspected.

PR #211, opened 2020-04-01
sharing build test script for v4+ubuntu18

baallan:


tom95858:
The text says v4+ubuntu18, but the file is named make-all-ubuntu-12. Is one or the other a typo?

baallan:
not a typo. the updated script file was created back in the days of U12. you may be right that it's time for a rename...

PR #214, opened 2020-04-09
fix redundant job set warnings cluttering the logs

baallan:
This appends a warnings field to the base sampler structure and uses it to suppress repeated spew of 5 warnings to the log related to jobinfo insufficiency. A warning is issued once and then not issued again unless it is first resolved and then reappears.


tom95858:
I think this looks great. Thanks @baallan...

PR #223, opened 2020-04-23
add plugins_rpath

baallan:
This adds the plugins_rpath (a generalization of the configure.ac soslibdir variable) to configure and environment scripts. Plugins_rpath can be added to the runtime LD_LIBRARY_PATH to allow resolution of any off-path libraries used by ldms plugins, such as libpapi and libsos.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,6 +52,7 @@ AC_C_CONST
 AC_C_FLEXIBLE_ARRAY_MEMBER
 
 OVIS_PKGLIBDIR
+
 dnl change sharedstatedir default
 test "$sharedstatedir" = '${prefix}/com' && sharedstatedir='${prefix}/var/lib'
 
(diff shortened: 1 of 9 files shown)


configure.ac
```diff
@@ -472,8 +473,10 @@ if test -n "$CHECK_SOS"; then
 fi
 if test -n "$SOS_LIB64DIR"; then
 	soslibdir="$SOS_LIB64DIR"
+	plugins_rpath="$plugins_rpath:$soslibdir"
```
tom95858, line 476: nit: Can't we just move this line to outside the if since it's the same line of code on both sides of the if/else.

baallan: I was in a hurry; it should have been the 4 line change now provided instead of two. The revision stops append of empty values to plugins_rpath.

tom95858:
@baallan is this ready to go?

baallan:
@tom95858 yes please.

PR #224, opened 2020-04-23
fix missing link variable for papi-sampler.

baallan:
makefile missing variable added


baallan:
@tom95858 or @narategithub please review and merge.

PR #229, opened 2020-04-28
switch to strtok_r from strtok in procinterrupts sampler

baallan:
fix thread safety problem


tom95858:
@baallan are you seeing an error with this? If this code path is actually multi-threaded it has much bigger problems than strtok.

tom95858:
Ah, I think I get it. This code isn't multi-threaded, but there are other threads in ldmsd using strtok...

baallan:
@tom95858 yes, there should be no daemon/plugin threads using strtok. We have quite a few instances of main using strtok in option handling which should be ok until something forks/loads a threaded plugin.

PR #237, opened 2020-05-05
fix oom in ldms_ls

baallan:


tom95858:
Hi @baallan, what is the goal here? The command is not multi-threaded, this is not a daemon. Cleaning up memory allocated before failure is code without purpose. This patch simply adds a bunch of code that doesn't actually do anything. Please advise.

baallan:
Isn't the underlying libldms and xprt multithreaded? We have no reason to assume that localtime isn't in use somewhere that might conflict with the print callback, or won't become so.

tom95858:
ldms_ls does not receive asynchronous data from peers. It is not a daemon receiving requests from peers; so yes, we have reason to believe that there is no thread toiling away using localtime and if it were, it would be using the thread-safe version!! The underlying issue here is that we continue to add code to a release vehicle that adds regression risk (however small) and is not fixing any symptomatic problem.

baallan:
@tom95858 reverted the localtime call change.

PR #244, opened 2020-05-05
fix bugs in ldmsd_stream.c v4

baallan:
missing double free, missing null ptr chk, and redundant typedef,

### DIFF
--- a/ldms/src/ldmsd/ldmsd_stream.c
+++ b/ldms/src/ldmsd/ldmsd_stream.c
@@ -26,19 +26,19 @@ static int s_cmp(void *tree_key, const void *key)
 	return strcmp((char *)tree_key, (const char *)key);
 }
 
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -156,6 +155,8 @@ int ldmsd_stream_publish(ldms_t xprt,
 	ldmsd_req_attr_t first_attr, attr, next_attr;
 	int rc;
 	size_t this_rec;
+	if (!data_len)
+		return 0;
```
tom95858, line 159: Hi @ballan, seems like this should be EINVAL?

baallan: @tom95858  isn't the logical result of sending 0 bytes on a stream a successful no-op? seems like the function did what it was told.

PR #251, opened 2020-05-06
fix json parser nits

baallan:
json_util.h: ifdef sometimes redundant macro
json_util.c: fix leak from incorrect oom check; clear unused vars.


tom95858:
Hi @baallan, I'm going to close this. Changes coming in ovis_json will resolve this and the yy-prefix on the symbols.

PR #255, opened 2020-05-06
v4 eliminate magic number buffer sizing and theoretical buffer overflow

baallan:
This commit groups 3 files with only magic number induced problem fixes.


tom95858:
This does not fix any problem afaict. It is purely a stylistic change. I do agree that it is nice to make it easy to change the size of these static buffer allocations with one line of code. But LBUFSZ is no less magical than 255. In fact it is less so, because you have to look up what LBUFSZ means.

PR #257, opened 2020-05-06
v4 fix bugs in perfevent.c

baallan:
thread unsafe use-after-free
missing oom check
memory leaks
improperly unset return codes
missing include
missing read error check
remove unused vars

### DIFF
--- a/ldms/src/sampler/perfevent.c
+++ b/ldms/src/sampler/perfevent.c
@@ -57,6 +57,7 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <sys/errno.h>
+#include <sys/ioctl.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


ldms/src/sampler/perfevent.c
```diff
@@ -511,13 +519,21 @@ static int sample(struct ldmsd_sampler *self)
...
-		int event_index = (read_result / sizeof(long long)) - 1; /* start from the last event added to the list */
+		if (read_result < 0) {
+			free(data);
+			if (!readerrlogged) {
+				msglog(LDMSD_LERROR, "perfevent: read event failed.\n");
+				readerrlogged = 1;
+			}
+			break;
+		}
+		/* int event_index = (read_result / sizeof(long long)) - 1;  start from the last event added to the list? */
```
tom95858, line 536: not sure why this is left here in a comment. are you not certain of this code? has it been tested? are any of these changes symptomatic?

baallan: left it just in case someone recognized it as desired debugging code. Six weeks without comment seems to indicate it is not, so I removed it just now.

PR #263, opened 2020-05-06
WIP: review apparent v4 leaks request

baallan:
The files in this commit all have comments containing the string 'leak'
marking places with apparent leaks as noted. Seeking input from the experts.

@tom95858 @narategithub @nichamon since all of these are in/near the control path, please comment whether these apparent leaks are somehow actually not or what a good fix is in each case.


tom95858:
Hi @baallan, this is not a code change to be reviewed and approved.

PR #268, opened 2020-05-19
add base_auth_parse to allow samplers to use standard uid/gid/perm

baallan:
Parsing and checking uid/gid/perm is complicated enough that every
sampler should be able to use it without cloning code or being forced to use the
base schema jobid/compid/appid logic. This factors out the security
config parsing into a function and then calls it from the base_config
function.

I have a network sampler coming that needs this.

### DIFF
--- a/ldms/src/sampler/sampler_base.c
+++ b/ldms/src/sampler/sampler_base.c
@@ -233,51 +233,10 @@ base_data_t base_config(struct attr_value_list *avl,
 	else
 		base->job_set_name = strdup(job_set_name);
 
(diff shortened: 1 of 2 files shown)


tom95858:
@baallan please move this pull request to master. Refactoring all the samplers to use this is not appropriate for a maintenance release.

baallan:
@tom95858 there is no change for any existing sampler for this. The sampler base_config calls the new function where the code it wraps used to sit.

baallan:
@tom95858 this PR is required to support the ibnet sampler PR.

PR #273, opened 2020-06-03
distclean-local and nola added

Snell1224:


morrone:
I think that you are probably going to need to cleanup your patch history on this one and resubmit the pull request. Look into the "git rebase -i" command to help with that. There should be just one commit with the "nola" change, and just one commit with the "distclean" change.

PR #278, opened 2020-06-08
Include error messages in sampler_base ooms.

baallan:
report errors from sampler-base when failures occur, often due to out of set memory.

### DIFF
--- a/ldms/src/sampler/sampler_base.c
+++ b/ldms/src/sampler/sampler_base.c
@@ -331,9 +331,14 @@ ldms_set_t base_set_new(base_data_t base)
 {
 	int rc;
 	base->missing_warned = 0;
+	errno = 0;
 	base->set = ldms_set_new(base->instance_name, base->schema);
-	if (!base->set)
+	if (!base->set) {
+		const char *serr = ovis_strerror(errno);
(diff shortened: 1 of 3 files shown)


ldms/src/sampler/sampler_base.c
```diff
@@ -331,9 +331,17 @@ ldms_set_t base_set_new(base_data_t base)
 {
 	int rc;
 	base->missing_warned = 0;
+	errno = 0;
 	base->set = ldms_set_new(base->instance_name, base->schema);
-	if (!base->set)
+	if (!base->set) {
+		size_t sz = 128;
+		char s[sz];
```
tom95858, line 338: I'm not sure how this works at all. Has this been compile tested?

baallan: this is c99 dynamic array (alloca wrapper syntax). It's equivalent to the
#define X 128
s[X]

It ran (and was needed) as part of dev/test of Omar's lustre sampler update during the stria DST.

baallan:
@tom95858 Unless there is some violent objection to ovis_strerror, I will also make a commit to get rid of all other strerror_r calls in ldms.

baallan:
@tom95858  is there a problem merging this in its revised form?

PR #279, opened 2020-06-08
V4 lustre 2.12 fixes

baallan:
this combines Omar's updates for lustre2_client and lnet_stats with fixes found during stria testing.


baallan:
@aazizo @valleydlr @tom95858 lustre 2.12 should be working now with the old format samplers.

PR #281, opened 2020-06-08
Add nola v4

Snell1224:


baallan:
@tom95858 if someone could review and merge Sara's build rules patch, it would make our jenkins tests start passing with the main branch. Looks good to me (we did test it on voltrino).

PR #285, opened 2020-06-19
configurable ib mad network sampler plugin ibnet

baallan:
A number of supporting utilities are included:
* ldms_ibnet_schema_name provides the schema name that will be used for a given set of plugin configuration parameters. This is handy for configuring storage policies.

* ldms-ibnet-sampler-gen provides configuration files that spread the load of sampling
the ports of the input fat tree fabric description over the nodes listed in the input samplers file. It also generates optionally a lid/hostname mapping file useful for labeling data during analysis. It also generates optionally visualizations of the network hardware topology (not the routing tables) using graphviz.

* ldms-gen-lidfile.sh dumps a lid/guid/port mapping usable for small network systems with only one sampler.


baallan:
@oceandlr @valleydlr @tom95858 the sampler used in the stria test 2 weeks ago, plus docs, utilities, is available for merge with OVIS-4. it is disabled in build by default.

PR #288, opened 2020-06-23
Added new option for python configuration

Snell1224:
"ENABLE_LIB_PYMAP" added to control cython check for sos and ldms installation without python plugins/support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -219,6 +219,7 @@ OPTION_DEFAULT_ENABLE([core], [ENABLE_CORE])
 OPTION_DEFAULT_ENABLE([mmap], [ENABLE_MMAP])
 OPTION_DEFAULT_ENABLE([perf], [ENABLE_PERF])
 OPTION_DEFAULT_DISABLE([yaml], [ENABLE_YAML])
+OPTION_DEFAULT_DISABLE([lib-pymap], [ENABLE_LIB_PYMAP])
 
 dnl Options for store
 OPTION_DEFAULT_ENABLE([store], [ENABLE_STORE])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -523,7 +524,9 @@ fi
 
 dnl swig stuff
 if test -z "$ENABLE_SOS_TRUE"; then
-	which cython >/dev/null 2>&1 || AC_MSG_ERROR("cython not found")
```
tom95858, line 526: I think this is a left-over from when sos used to be a subproject of ovis. I don't think this check is needed anymore.

configure.ac
```diff
@@ -523,7 +524,9 @@ fi
...
+	if test -z "$ENABLE_LIB_PYMAP_TRUE"; then
+		which cython >/dev/null 2>&1 || AC_MSG_ERROR("cython not found")
+	fi 
 	AX_PYTHON_DEVEL()
```
tom95858, line 530: So this should all be unnecessary.

baallan:
@tom95858 is Nick's ovis/lib/python/Map.pyx something that is part of sos now?
the cython result is used only for this file. If we delete ovis/lib/python, Sara's patch would be to get rid of the cython dependency instead.

configure.ac
tom95858: All of the Python3 support, which is being back ported from master to OVIS-4 requires Cython.

tom95858: Hi @baallan , all of the Python3 support, which is being back ported from master to OVIS-4 requires Cython.

baallan: so ldms use of cython will be independent of sos? In which case the backport will be adding configure tests for cython? If yes to both, Sara can revise this to just delete the cython check. Then when the backport is merged things will continue to be fine.

tom95858:
The Map API was added to allow one to create an id mapping from component_id to some other id name space. The intent was so that you could have simple distance computations like id_0 - id_1 that would be meaningful in different domains, e.g. network vs. facilities. You can disable Map.pyx  in the short term and the build won't require Cython. Can you please provide insight into why this is an issue in the 1st place?

baallan:
in current v4, if sos is enabled it then tests for cython (which is of course not there). this makes the build regression tests fail if any plugin sos related is enabled.

Snell1224:
@tom95858 I've removed the "LIB_PYMAP" option that was added from the first commit along with the cython check line since the build won't need it for the short-term. I haven't been able to squash these commits into one on this pull request so would it be better to start a new pull request?

configure.ac
tom95858: Yes, the Python API that was formerly generated using SWIG has been re-implemented in Cython for Python3.

PR #290, opened 2020-06-24
Ldms csv export sos fix

bschwal:
Fixed ldms-export-csv-sos script to be reflect the compound indices SOS-4 has by default.
Also added the --kind-file option to pass in a user-generated kind file to the script.

### DIFF
--- a/ldms/scripts/ldms-csv-export-sos
+++ b/ldms/scripts/ldms-csv-export-sos
@@ -482,6 +482,11 @@ def get_filenames(fn, args):
             schemafile = strip_gz(schemafile)
             mapfile = fn.replace("." + x, ".MAPSOS." + x)
             mapfile = strip_gz(mapfile)
+    if args.kind_file:
+        if os.path.isfile(kind) and args.kind_file != kind:
+            if args.verbose:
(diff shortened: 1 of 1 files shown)


ldms/scripts/ldms-csv-export-sos
```diff
@@ -693,7 +698,7 @@ def parse_schema(fn):
         a = []
         for i in sch["attrs"]:
             if i["type"] in metric_types:
-            a.append(i["name"])
+            	a.append(["name"])
```
baallan, line 701: @bschwal please verify that 4-space and not tab indents are used everywhere in this python script.
also, at 701 was converting i["name"] to ["name"] intentional? what does that fix?

bschwal: @baallan tab changes made accordingly, the "name" change was not intentional.

baallan: assuming you've run it after the space changes, looks good to merge.

bschwal: I have with success.

baallan:
@tom95858 this is ready to merge. It's part of the ingest chain for our grafana servers.

PR #299, opened 2020-07-06
ldmsd: APIs to convert time from string to integer (us) and vice versa

nichamon:


baallan:
@nichamon please adjusting the time conversions to use uint64_t instead of the ambiguous (and potentially too small) long type.

PR #303, opened 2020-07-08
Update sampler 'variable' to actually work as planned.

baallan:
This sampler periodically redefines its sets and schemas, reusing names.
This works, but the current aggregation never sees the new definitions of recycled names.
Symptom of this is that the store only gets the first cycle of data in each schema.
Is there something about unpublish needed to convince downstream to drop old definitions?

to run the sampler, do install and then
bin/ldms-static-test.sh variable

### DIFF
--- a/ldms/scripts/examples/variable
+++ b/ldms/scripts/examples/variable
@@ -1,13 +1,37 @@
 export plugname=variable
 portbase=61072
-LDMSD -p prolog.sampler 1 2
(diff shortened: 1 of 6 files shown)


baallan:
@tom95858 @valleydlr I made a pull request with the updated 'variable' sampler I emailed about earlier.
The error described is on vanilla redhat 7/sock transport (vm, not cluster).
A collateral effect of the commit is to move it to a subdirectory and add a man page to document it.

PR #305, opened 2020-07-09
Change all python scripts to explicitly call python2

morrone:
RHEL8 rpm build scripts throw an error if it finds scripts using
just "python" on the #! line. They want us to now explicitly call
either "python2" or "python3".


baallan:
I think we're in the process of everything switching to python3.

morrone:
The switch to python3 is on the master branch, to the best of my understanding. On the OVIS-4 branch we are still at python2. I need to build OVIS-4 on RHEL8 now.

nick-enoent:
I'm in the process of porting OVIS-4 to python3

morrone:
@nick-enoent 

OK. Please make sure that you update all the scripts to explicitly call "python3".

PR #308, opened 2020-07-09
Fix bug in hello_stream build that breaks "make dist"

morrone:
The referenced hello_sampler.h file does not exist.


oceandlr:
Ack! Thanks! @morrone

oceandlr:
@tom95858  @valleydlr  -- we should take this in.

PR #309, opened 2020-07-10
Variable subdir v4

Snell1224:
add enable option for "variable" sub-directory


morrone:
I think that you may have forgotten to include some files in this PR. At least, this adds ldms/src/sampler/variable/Makefile to AC_CONFIG_FILES, but there is no included Makefile.am that would allow that directive to work.

If this is intended to move the variable.c file into a subdirectory, I think that probably should happen in the same PR.

If you would like any tips for testing before pushing a PR, or using git to determine whether all of your changes are captured in commits, I'm happy to chat about that here.

morrone:
My bad! I missed that 3ac18f8f just landed.

PR #312, opened 2020-07-15
ovis event: the worker state stuck in FLUSHING if ev_flush is called.

nichamon:
With the change the worker state is reset back from FLUSHING to RUNNING
after all events have been flushed.


nichamon:
@tom95858 I found the bug by investigating the cause that an updater cannot be started after it has been stopped.

nichamon:
@tom95858 The problem with restarting Updaters is that when ev_flush() is called and the worker state is changed to FLUSHING and it stuck in that state forever. When Updaters are restarted, ev_post() returned with EBUSY because the worker is in the FLUSHING state. The change resets the worker state back to RUNNING after it flushes all outstanding events.

Issue #310, opened 2020-07-11
hello_stream build issues


morrone:
The new hello_stream sampler (on the OVIS-4) has some oddities in the build system.

First, there is a perl script named "parser.pl" that is installed under sysconfdir (/etc/sysconfig with some default installs). I do not think that sysconfdir is the correct place for a script like this. Additionally, anything that we install needs less generic naming than "parser.pl".

Next, the package installs config file into:

```$(sysconfdir)/ldms/hello_sampler_util```

Everything else installs into ${sysconfdir)/ldms.d, not $(sysconfdir)/ldms. That is probably a bug?

Also, we should probably try to avoid the use of custom install-data-local/uninstall-local hooks for simple file installs. I think something like this:

```
STREAM_FILES = \
       $(srcdir)/hello_stream1.conf \
       $(srcdir)/hello_stream_store.conf \
       ${srcdir}/README


HELLO_CONF_DIR = $(DESTDIR)$(sysconfdir)/ldms/hello_sampler_util/stream_configs
install-data-local: $(STREAM_FILES)
       $(MKDIR_P) $(HELLO_CONF_DIR)
       $(INSTALL_DATA) $(STREAM_FILES) $(HELLO_CONF_DIR)/


uninstall-local:
       rm -rf $(STREAM_DIR)
```

can be done something like this instead:

```

helloconfigsdir = $(sysconfdir)/ldms.d/hello_sampler_util/stream_configs
helloconfigs_DATA = hello_stream1.conf \
                                  hello_stream_store.conf \
                                  README
```

oceandlr:
Hi @morrone 

Right now the streams utilities are in use case testing, and intended to be supported in V4. The goal was to get some of the support utilities in an encapsulated place, consistent with similar utilities. In this case, the utilities for the hello_stream should be in encapsulated in <InstallDir>/etc/ldms/hello_sampler_util which is at the same level as <InstallDir>/etc/ldms/aries_mmr_set_configs.  The utilities include this parser as well as example configs for the use of streams. 

Are you saying this is not happening?

morrone:
@oceandlr,yeah, sorry, you are correct that it winds up in /etc/ldms/hello_sampler_util when using the default system prefix.
[...]

PR #314, opened 2020-07-15
move some examples files from /etc to /share

oceandlr:
Re issue #310 

Moved aries_mmr_set_configs and hello_sampler_util from XXX/etc to XXX/share/doc/ovis-ldms-X.Y.Z/examples

### DIFF
--- a/ldms/src/sampler/aries_mmr/aries_mmr_set_configs/Makefile.am
+++ b/ldms/src/sampler/aries_mmr/aries_mmr_set_configs/Makefile.am
@@ -57,7 +57,9 @@ MMR_FILES = \
 	$(srcdir)/metric_set_rtr_3_4_c \
 	$(srcdir)/metric_set_rtr_3_4_s
 
-MMR_DIR = $(DESTDIR)$(sysconfdir)/ldms/aries_mmr_set_configs
+
+OPV=-$(PACKAGE_VERSION)
+MMR_DIR = $(docdir)$(OPV)/examples/aries_mmr_set_configs
 install-data-local: $(MMR_FILES)
 	$(MKDIR_P) $(MMR_DIR)
 	$(INSTALL_DATA) $(MMR_FILES) $(MMR_DIR)/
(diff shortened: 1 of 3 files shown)


oceandlr:
BEFORE accepting this:
@valleydlr -- are you ok with the move of aries_mmr_set_configs? Do you use it in place?
@tom95858 @morrone -- in #310, @morrone was against the install-data-local ... I have kept it as is. If we want this change then there are other places that should change as well.
@tom95858 -- will a move at this point break with any standard policy?

oceandlr:
@baallan I based this off of XXX/ovis/ldms/src/third-plugins/my_plugin/Makefile.am and ovis/ldms/src/third-plugins/my_plugin/scripts/Makefile.am. Did I misread those?

BTW, it looked to me at the top level configure that $(docdir) gets defined, so it is redefined in those third-plugin places. This is why I used $(datadocdir) as my local name. Am in interpreting this correctly, and if so, should we do similarly here in third-plugins?

baallan:
@oceandlr the third-plugin subdirs are standalone projects in themselves, so things differ. The naming I suggest follows the rule $top/ldms/Makefile.am:install-data-local

oceandlr:
updated to address @baallan requested change (sorry if that should go in the thread elsewhere)
Please all still see the 2nd comment for review considerations...

baallan:
@oceandlr https://www.gnu.org/software/automake/manual/html_node/Standard-Directory-Variables.html
is the baseline supported by automake. The use of PACKAGE_VERSION instead of just PACKAGE is a common extension used to avoid conflicts when multiple major versions may be installed on the same system.

tom95858:
@oceandlr, @baallan: is everyone happy with this change?

oceandlr:
@valleydlr should weigh in on the move since I don't know if he is expecting the Aries config files to be in a particular place.
Also do you @tom95858 have any option on the install-data-local that @morrone was against? However, I don't know what the alternative would be and if we do want to change it then we should change elsewhere, so maybe that should just be a different change, if we want to go that route.

morrone:
> @valleydlr should weigh in on the move since I don't know if he is expecting the Aries config files to be in a particular place.
> Also do you @tom95858 have any option on the install-data-local that @morrone was against? However, I don't know what the alternative would be and if we do want to change it then we should change elsewhere, so maybe that should just be a different change, if we want to go that route.

I don't have a problem with cleaning that up some time later. FYI, I documented the alternative to using custom install-data-local hooks in issue #310.

valleydlr:
I have no problem with this. The aries configs can be copied from one place just as well as another.

oceandlr:
ok @tom95858, ready to go

PR #315, opened 2020-07-15
added libpapi_sampler.so to blacklist

Snell1224:
added "libpapi_sampler.so" to blacklist

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -1006,6 +1006,7 @@ void ldmsd_mm_status(enum ldmsd_loglevel level, const char *prefix)
 }
 
 const char * blacklist[] = {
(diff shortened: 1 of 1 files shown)


morrone:
Can you please elaborate on why we should do this?

baallan:
@morrone There are sampler plugins in v4  (including papi_sampler) which use the gnu linker dlopen ctor/dtor conventions to do initializations which are not valid (cause seg fault) unless there is a fully started ldmsd. The plugin api changes in v5 appear to have resolved this for the offending plugins in v4. These bad (mostly experimental) plugins cannot be dlopened to get their usage strings and then dlclosed, as is needed to support the 'ldmsd -u' command.

I believe that all the code in the v4 ctor blocks should have been created as call-once in the config() or/and sample method of the blacklisted plugins.

PR #316, opened 2020-07-17
Introduce create-release.yaml GitHub workflow

morrone:
This introduces a new "Create Release" GitHub workflow, which automates
the process of creating a release on GitHub any time a new tag is pushed
to the repo.

The workflow triggers any time a new tag is pushed to the repo that
matches the pattern 'OVIS-4*'.

First the workflow builds the cannonical tarball.

Next the workflow generates a new release on the
https://github.com/ovis-hpc/ovis/releases page. The release notes
are taken from the body of the annotated tag.

Finally, the cannonical tarball is uploaded to the Assets of the
newly created release.

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -1,0 +1,48 @@
+name: Create Release
+
+on:
+  push:
+    tags:
+      - 'OVIS-4*'
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-18.04
+
+    steps:
+    - run: sudo apt install gettext
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


baallan:
It appears this worked, as the releases page has a new item at the top for alpha.0

morrone:
> It appears this worked, as the releases page has a new item at the top for alpha.0

@baallan, I do not think that the timing worked out for that. This probably got pushed to the repo before the tag was pushed. The item for the alpha.0 tag in the releases page is the normal one that appears for any tag. If this had triggered, we would see a full release with attached tarball, and we would see the action that made it happen under the "Actions" tab.

PR #321, opened 2020-07-20
test for sampler fail handling

baallan:
when a sampler fails to sample in v4, it is stopped. this PR provides a test of this failure sequence.


baallan:
@tom95858 the error dump on ldmsd exit i mentioned in today's phone call appears to be related to sampler handling rather than rdma. this pr reproduces the symptom even when sock is in use.

PR #342, opened 2020-07-26
python3 configure script

narategithub:
- Modify autoconf configure script (top-level, lib/ and ldms/) to use
  /usr/bin/python3 by default.
- Also check that the python is version 3.6 or greater.


tom95858:
Yes, sure, combining them is certainly simpler

PR #344, opened 2020-07-28
Remove an unused global static variable from ldmsd_stream.c

narategithub:


narategithub:
I'm merging this trivial fix now as OVIS-4 build failed if `-Wall -Werror` options are supplied.

PR #345, opened 2020-07-28
Refactor the ldmsd_stream_publish test program to really the publish API

nichamon:
Without the change, the ldmsd_stream_publish test program doesn't call
the ldmsd_stream_publish API which is the API used by any ldmsd stream
application. The change makes the test program call ldmsd_stream_publish
to really test the API and the underlying logic.


nichamon:
@tom95858  The change has been tested by the test script in the ldms-test repo.

nichamon:
@tom95858 After I discussed with Narate, I will close this pull request. The ldmsd_stream_publish test program will be added to the ldms-test repo instead of modifying the one in the OVIS repo, so we could avoid unnecessary change in the source tree.

PR #348, opened 2020-07-30
zap_sock - enable EPOLLOUT only when necessary

narategithub:
Enable EPOLLOUT only when write() returns EBUSY or EWOULDBLOCK.

This has passed `zap_test` and all test cases in `ldms-test` repo.


narategithub:
@tom95858 v5 code is a bit different. But, I'll apply something similar to it as well.

tom95858:
This change makes me nervous without sufficient scalability testing. Let's push this to 4.3.5/6 when we can get more miles on the scalability testing.

PR #353, opened 2020-08-02
Remove unused vars from tutorial store and sampler

narategithub:


narategithub:
@oceandlr I'm merging this in as OVIS-4 won't build with `-Wall -Werror` when tutorial store and sampler are enabled.

oceandlr:
Thanks. Maybe someone can update the build check to use those flags
I don't know how it currently works (maybe @morrone @baallan @narategithub can tell me where to look)

baallan:
@oceandlr Commonly we use -Wall (enable a common set of warnings) but not Werror, which converts all warnings into errors. LDMS has almost never been 'warning clean' though most of the plugins are. gcc (and certain language committees) are fond of including warnings about stuff that is stylistic rather than affecting generated code. 

As a project we do mostly prefer to eradicate unused variables.

oceandlr:
Ok. I was just wondering if putting it in the check would help discovery of such variables, however I see there would be other issues. I will make sure I check with -Wall in the future before putting in a request, to keep this from happening again.

thanks!

morrone:
@baallan , @oceandlr  -Wall of course turns on a whole class of warnings, but we could certainly introduce automating build testing that uses a smaller set of warnings. For instance, we could used -Wunused-variable. Or we could use -Wunused and get a catch all of the unused warnings.

I am happy to assist with this. With github actions/workflows, the main limitation is that the Linux VMs on github (GitHub-hosted runners) are only Ubuntu. We would need to work out a reasonable set of ldms configure options that will work with standard packages available under Ubuntu 18.04 (or 20.04 soon). But there is probably a good deal that we can do with that.

Maybe we should start a new issue to discuss the expansion of automated testing?

PR #350, opened 2020-07-31
Remove the logic that supports multiple lookup on a set

nichamon:

PR #356, opened 2020-08-03
ldms_set_delete checks if the set exists in the id tree before removing it.

nichamon:


nichamon:
@tom95858  I have tested that with 2 producers pointing to the same set and 2-level aggregation. I tested the following scenarios.
The 2 sampler daemons were killed one-by-one.
The 2 sampler daemons were killed at the same time.
The 1st-level aggregator was killed while the 2 sampler daemons and the 2nd-level aggregator were running.
The 2 sampler daemons were killed and then the 2nd-level aggregator was killed.

There were no crashes.

 I have also tested this with the patch that removes the logic supporting multiple lookups (Pull request #350) with the same scenarios.

PR #359, opened 2020-08-04
zap_sock - only enable EPOLLOUT when necessary

narategithub:
This passed `zap_test` and all test cases in `ldms-test` repository.


narategithub:
This is for v5.

PR #363, opened 2020-08-11
ldms_xprt: send_req_cancel_push() returns a random value on success

nichamon:
The rc variable wasn't initialized so the function might return a random
number on success.


morrone:
A good git commit message should describe what the commit _does_. In this case, the commit message is more of a bug report. This is a tiny patch, so a short message is certainly acceptable. But a better commit message might be:

```Initialize return code in ldms_xprt.c:send_req_cancel_push()```

morrone:
Here's a good guide for git commit messages that I think we should strive to follow:

https://chris.beams.io/posts/git-commit/#seven-rules

PR #364, opened 2020-08-12
Fix linking flags

nichamon:
Replace the linking flags of ovis libraries from -l to *.la. Otherwise, ovis cannot be built on Ubuntu.

### DIFF
--- a/ldms/src/ldmsd-samplers/appinfo/Makefile.am
+++ b/ldms/src/ldmsd-samplers/appinfo/Makefile.am
@@ -7,7 +7,10 @@ libappinfocl_la_CFLAGS = -I$(srcdir)/../../core \
 			 -I$(top_srcdir) @OVIS_LIB_INCDIR_FLAG@ \
 			 -I../../ -I$(srcdir)/../../ldmsd
 libappinfocl_la_LIBADD = ../../core/libldms.la \
-			 @LDFLAGS_GETTIME@ -lpthread -lovis_util -lcoll
+			 @LDFLAGS_GETTIME@  \
+			 $(top_builddir)/lib/src/ovis_util/libovis_util.la \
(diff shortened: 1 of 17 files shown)


tom95858:
We keep changing these back and forth. Let's get to the bottom of why this doesn't work on Ubunto because it should. binding with -l has other issues, in particular where the library actually comes from.

narategithub:
1) This patch removes the use of `-l`, and uses libtool `*.la` for the in-project libraries. Are you saying that we were against using libtool '*.la'?

2) I will dig into why it wasn't working on Ubuntu later. But, I was surprised it worked on CentOS (without this patch) because those `-l<our-lib>` should fail because the libraries are being built and installed. ... I'll dig into it when I have time 😄 .

baallan:
It appears configure.ac lost OPTION_LIB_FLAGS([OVIS]) and OPTION_INCLUDE_FLAGS in some refactorization.
Without that, or an equivalent, a lot of linking stuff is going to break.

narategithub:
@baallan Could you be more specific? This is master branch. You can see the single autoconf patch here https://github.com/ovis-hpc/ovis/commit/f4409ff1f8f64acac4e0abe2b6273b62a4150fd8. It is based on the combination of `configure.ac`, `lib/configure.ac` and `ldms/configure.ac`. I have also `make dist` and build from the generated dist tarball. Maybe I'm missing something?

baallan:
You may want to compare the ovis-4 top configure.ac to your configure.ac. Aside from any new plugins and directory renamings, I would expect them to be very similar. Lines 60 -110 of https://github.com/ovis-hpc/ovis/blob/OVIS-4/configure.ac#L10 cause lib and include flags to be well-defined for all the supporting libraries without using directory prefixes in includes.

PR #371, opened 2020-08-19
add better error messages for plugins usage query

### LINKED ISSUE #357: ovis-4 ldmsd -u meminfo produces excess warnings on multipart path
When successfully finding a requested plugin, no warnings about missing plugin path elements should be generated.

ldmsd_config.c

baallan:
fixes #357  . issue messages only after all checks are complete, except for invalid path type warnings.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -1021,6 +1021,27 @@ const char * blacklist[] = {
 
 static int ldmsd_plugins_usage_dir(const char *dir, const char *plugname);
 
+/* check path for existence, and if verbose != 0, whine if missing. */
+static int ldmd_plugins_check_dir(const char *path, int verbose) {
+	struct stat buf;
+	memset(&buf, 0, sizeof(buf));
+	int serr = stat(path, &buf);
+	int err = 0;
+	if (serr < 0) { 
+		err = errno;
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -1045,27 +1066,37 @@ int ldmsd_plugins_usage(const char *plugname)
...
 		if (trc)
 			rc = trc;
+		else
+			if (plugname)
+				plugfound = 1;
+	}
+	if (plugname && !plugfound) {
+		fprintf(stderr, "%s: no library in %s for %s\n", APP, path, plugname);
+		strncpy(library_path, path, sizeof(library_path) - 1);
+		while ((libpath = strtok_r(pathdir, ":", &saveptr)) != NULL) {
```
tom95858, line 1082: I'm not sure I understand how this can work. Line 1070 above has already modified pathdir (adding '\0' after each token), so after that logic runs, there will only be a single entry in pathdir, specifically, the 1st one.

baallan: the strncpy reinits library_path which is aliased by pathdir.
the aliasing of library_path exists because the code was cloned from new_plugin.

tom95858: Ah, OK, thanks for the explanation @baallan

PR #381, opened 2020-08-26
Refactor the test_sampler plugin

nichamon:
Without the patch, the test_sampler plugin requests multiple 'config'
lines to define a schema and to add a set. This design is incompatible
with the new JSON configuration protocol. Moreover, the test_sampler
plugin was extended many times, and its code has become complicated.

The change simplifies the plugin code and makes it compatible with the
JSON protocol.


nichamon:
@narategithub If you could skim through whether the change violated LDMSD plugin interface or not. I refactored it following the plugin interface document.

PR #382, opened 2020-08-26
Handle Updater push mode

nichamon:


nichamon:
@narategithub I forgot to extract the push attribute from the JSON request object of Updaters.

PR #383, opened 2020-09-01
Fix failover configuration buffer reset issue

nichamon:
The bug is that the buffer that reads the content from a configuration
file isn't reset if the previous request goes though the deferred
filter.


tom95858:
please add a summary commit message like:

Fix deferred filter buffer reset issue

before the remainder of your commit message and resubmit your request.

tom95858:
To be specific:

Fix failover configuration buffer reset issue

If the configuration file contains the commands 
failover_start, prdcr_start_regex, prdcr_start, 
updtr_start or strgp_start, the 
LDMSD will fail at the configuration stage

nichamon:
@tom95858 I revised the commit message as requested.

PR #390, opened 2020-09-03
Added test script that executes ldms-static-test.sh for all enabled plugin samplers

Snell1224:
Invoke ldms-static-test.sh for all enabled plugins during installation. Man Page included

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -813,6 +813,8 @@ AC_CONFIG_FILES([
 	ldms/scripts/ldms-wrapper
 	ldms/scripts/ldms-static-test.sh
 	ldms/scripts/pll-ldms-static-test.sh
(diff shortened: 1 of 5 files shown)


tom95858:
Hi Sara,

Please change the commit message to something like the following:

<Summary of the change (50 chars or so)>

<Any detail about the change>

The commit message should inform someone about what the change is and why it was made. It should not contain questions like: "Let me know if you have any questions." That message will go into the commit log and be there forever; long after any relevance the question may have had is gone.

tom95858:
Oops, I got mark-downed. The commit message in general should be something like:

```code
50 char summary message

Any necessary detail of the change
```

Snell1224:
Hi @tom95858 I revised the commit message to the one above. Thank you for letting me know.

PR #391, opened 2020-09-03
Added variable KILL_LDMSD_STRICT 

Snell1224:
Added variable KILL_LDMSD_STRICT to ldms-static-test.sh


tom95858:
Hi @Snell1224 ,

Same comment regarding the commit message for this change.

Snell1224:
@tom95858 Revised the commit message here as well.

PR #395, opened 2020-09-09
zap_fabric listen without specifying the address

narategithub:
REMARK: The network addresses discussed in this patch are IPv4 as
current zap applications use only IPv4 addressing. Generic network
address suport (also not limited to sockaddr) for zap shall be handled
separately in the future.

zap_fabric used `fi_info.src_addr` to specify the listening address
for the passive endpoint. The applications use address 0 to specify
*any* addresses to listen to (i.e. only the port is specified). Even
though this worked perfectly on other zap transports, it did not work
on libfabric. For example, on cygnus-08, a machine with a regular
on-board ethernet and a Converged Ethernet card, `ldmsd -x
fabric:10000` reported a successful listen on port 10000 with `verbs`
provider (the best provider on the machine), but `ldms_ls -x fabric
-p 10000 -h cygnus-08-iw` could not connect. When the passive side
listen with a specific address, e.g. `ldmsd -x
fabric:10000:cygnus-08-iw`, the active side could connect and read,
i.e. `ldms_ls -x fabric -p 10000 -h cygnus-08-iw` worked.

Instead of using `fi_info.src_addr`, this patch uses `node` and
`service` string parameters in `fi_getinfo()` call with `FI_SOURCE`
flags to let the provider resolve for the appropriate address. The "*"
string would be supplied to `node` if the address is 0, and a regular
IPv4 "AAA.BBB.CCC.DDD" address would be used if the address is not 0.

A quick test shows that the following scenarios work:

1) listen without specifying the address
   `ldmsd -x fabric:10000` <-->
   `ldms_ls -x fabric -p 10000 -h cygnus-08-iw`

2) listen with specific address (verbs provider)
   `ldmsd -x fabric:10000:cygnus-08-iw` <-->
   `ldms_ls -x fabric -p 10000 -h cygnus-08-iw`

3) listen with specific address (tcp provider)
   `ldmsd -x fabric:10000:cygnus-08` <-->
   `ldms_ls -x fabric -p 10000 -h cygnus-08`
   (Notice the missing `-iw`)

### DIFF
--- a/lib/src/zap/fabric/zap_fabric.c
+++ b/lib/src/zap/fabric/zap_fabric.c
@@ -185,7 +185,7 @@ static void dlog_(const char *func, int line, char *fmt, ...)
 #endif
 
 #define ZAP_FABRIC_INFO_LOG "ZAP_FABRIC_INFO_LOG"
-static int z_fi_info_log_on = 0;
+static int z_fi_info_log_on = 1;
 
 #define Z_FI_INFO_LOG(rep, fmt, ...) do { \
 	if (z_fi_info_log_on && rep && rep->ep.z && rep->ep.z->log_fn) \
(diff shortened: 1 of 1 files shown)


baallan:
So what is this supposed to do when there are two hsn interfaces? We commonly have both omnipath and mellanox interfaces active.

narategithub:
I think it will end up selecting only one domain & provider (per `fi_getinfo(3))`.

baallan:
can we be sure to log (info) the ip addr and protocol of each fabric listener?
That way admins can at least detect surprising results.

tom95858:
If you with to listen on both addresses, you should be able to use
```
ldmsd -x fabric:ib-addr:port -x fabric:port:fabric-addr ...
```
@baallan can you please confirm this works on the machine in question?

narategithub:
@baallan the address logging is now added be81885e47a9af7c1a34e7d2387ed43c74d6ac64 (a part of this pull request).

baallan:
I tested ldmsd -x fabric:ib-addr:port -x fabric:port:fabric-addr
using an ethernet and an omnipath address and both worked.
Testing the combination of omnipath and mellanox will take a bit longer -- I have to build on a different system in /tmp.

PR #396, opened 2020-09-10
fix V4 fabric xprt mem inits

baallan:
this fixes information leak to network by initializing raw buffers and fixes insufficiently initialized addr and event data in ldms_xprt.c..

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1917,7 +1917,8 @@ static void ldms_zap_handle_conn_req(zap_ep_t zep)
 {
 	static char rej_msg[64] = "Insufficient resources";
 	struct sockaddr lcl, rmt;
-	socklen_t xlen;
+	memset(&rmt, 0, sizeof(rmt));
+	socklen_t xlen = sizeof(lcl);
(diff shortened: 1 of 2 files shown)


ldms/src/core/ldms_xprt.c
```diff
@@ -2469,10 +2470,9 @@ static int __ldms_conn_msg_verify(struct ldms_xprt *x, const void *data,
  */
 static void ldms_zap_cb(zap_ep_t zep, zap_event_t ev)
 {
-	struct ldms_xprt_event event = {
-			.type = LDMS_XPRT_EVENT_LAST,
-			.data = NULL,
-			.data_len = 0};
+	struct ldms_xprt_event event;
```
tom95858, line 2473: How is this different than the static initialization?

baallan: holes in the structure { (int, ptr, int) doesn't align nicely } get init with memset but not the initializer. 
An alternate fix would be put .data last (int, int, ptr),  but I'm not sure if this goes over the wire or not.

tom95858: I see. OK. Thanks for the explanation.

PR #398, opened 2020-09-11
Create a worker/event library

nichamon:
This API provides a mechanism to schedule events to worker threads. This
will replace the ldmsd_task paradigm in order to simplify the ldmsd
state machine.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,6 +78,7 @@ OPTION_INCLUDE_FLAGS([OVIS],[ \
 	lib/src/ovis_util \
 	lib/src/ovis_auth \
 	lib/src/ovis_ctrl \
+	lib/src/ovis_ev \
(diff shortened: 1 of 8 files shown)


baallan:
Do we have any documentation (other than code) of the current or planned revised state machines?

tom95858:
@baallan: This change is just staging master code into the OVIS-4 branch as discussed in the LDMS UG meeting. This does not  change anything in the current state machines.

@nichamon, please:
- Remove the comment in the commit message about simplifying the state machines and replacing the ldmsd_task paradigm.
- Remove the configure option of the event library itself. There is no need/benefit to not build this library, and the configure is already a bit overwrought with options such as this. The EV_TEST conditional is fine.

Thanks, Tom

PR #400, opened 2020-09-17
Fix ovis_json CFLAGS and compilation warnings

narategithub:
`CFLAGS := $(filter-out -Werror, ${CFLAGS})` statement in
`ovis_json/Makefile.am` was actually a Makefile statement. Automake took
that literally into Makefile that was generated by configure script.
Since `CFLAGS` was usually not defined in the environment variable nor
given when `make` was called, `CFLAGS` became an empty string.

Without the filter-out statement, the value of the CFLAGS in the
corresponding Makefile generated by configure is the value of CFLAGS
given in configure command.

This patch removed the filter-out statement that made CFLAGS an empty
string and addressed the build errors caused by `-Wall -Werror`.


narategithub:
FYI @nichamon

tom95858:
@narategithub the original reason this was required was because bison/flex was generating code that caused compile to fail with -Wall -Werror. It was not due to the warnings fixed here which were, I think, added after the fact. What platforms did you test build this on?

narategithub:
@tom95858 I tested this on CentOS7. Do you have a list of distros I should cover?

tom95858:
Did you test build it with -Wall -Werror?

narategithub:
Yes I did.

baallan:
@narategithub sles12, sles15, rhel server >=7.6, rhel8 (rhel 4-6 are now off the list). python/cython from rhel 7 server software collections are good.

tom95858:
Thanks @baallan for the testing.

PR #401, opened 2020-09-18
fix installcheck error for all plugins enabled

Snell1224:
Generate the static-test-list file in the same directory as run-static-test.sh. All plugins are listed in static-test-list for "max" installcheck (via Jenkins).


morrone:
This should not be landed, because static-test-list should not be installed in bin (nor should run-static-test.sh). While you are in this code, could you please take a look at issue #402 where I reported some things that need fixing with run-static-test.sh and static-test-list?

morrone (changes requested):
static-test-list should not be install in bin.

tom95858:
@Snell1224 could you please change the install location as the issue suggests?

PR #405, opened 2020-09-21
stream_unsubscribe + prdcr_unsubscribe

narategithub:
[x] Manages clients of ldmsd stream republishing service in a
    red-black-tree with `<PEER_ADDR>:<STREAM_NAME>` keys. The clients
    are registered by `LDMSD_STREAM_SUBSCRIBE_REQ`.
    - RECALL: The ldmsd-to-ldmsd stream republishing service
      subscription is driven by `prdcr_subscribe` config command.
[x] `prdcr_unsubscribe` config command to instruct ldmsd to unsubscribe
    the republishing stream from peers.
    [x] ldmsctl
    [x] ldmsd_controller
    [x] Add `LDMSD_STREAM_UNSUBSCRIBE_REQ` request. This is an
        ldmsd-to-ldmsd request to stop the stream republishing.
[x] automatically remove clients of stream republishing service
    (registered by `LDMSD_STREAM_SUBSCRIBE_REQ`) on connection
    termination.
    [x] ldms_xprt_sockaddr() caching: cache the remote/local addresses
	on the success zap_get_name() so that the remote/local addresses
	are still available even after the endpoint became disconnected
	(but not yet destroyed), which is required for the republish
	client key construction. The caching is valid because the
	local/remote addresses are not changed after the endpoint
	becomes connected.
[x] Add `stream_client_dump` config command to dump the current list of
    all stream clients (not just the republishing ones) for debugging
    and test verification.
    [x] ldmsctl
    [x] ldmsd_controller
[x] test: ldmsd_stream_test in ldms-test (covers both manual unsubscribe
    and automatic client removal on disconnection).

### DIFF
--- a/ldms/python/ldmsd/ldmsd_config.py
+++ b/ldms/python/ldmsd/ldmsd_config.py
@@ -103,6 +103,7 @@ LDMSD_CTRL_CMD_MAP = {'usage': {'req_attr': [], 'opt_attr': ['name']},
                       'prdcr_set_status': {'opt_attr': ['producer', 'instance', 'schema']},
                       'prdcr_hint_tree': {'req_attr':['name'], 'opt_attr': []},
                       'prdcr_subscribe': {'req_attr':['regex', 'stream'], 'opt_attr': []},
+                      'prdcr_unsubscribe': {'req_attr':['regex', 'stream'], 'opt_attr': []},
                       ##### Updater Policy #####
                       'updtr_add': {'req_attr': ['name'],
                                     'opt_attr': ['offset', 'push', 'interval', 'auto_interval']},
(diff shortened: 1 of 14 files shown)


ldms/python/ldmsd/ldmsd_config.py
```diff
@@ -132,6 +133,7 @@
                       ##### Streams ###
                       'publish': {'req_attr': ['name'], 'opt_attr': []},
                       'subscribe': {'req_attr': ['name'], 'opt_attr': []},
+                      'stream_client_dump': {'req_attr': [], 'opt_attr': []},
```
tom95858, line 136: Would it be better to call this something like show_stream_clients?

PR #406, opened 2020-09-21
Cython version check and SWIG + lib/python removal

narategithub:
The primary objective of this pull request is to add Cython version checking in the configure script (commit https://github.com/narategithub/ovis/commit/980c02ad29539b1345a873437609ed4e398a7d37). The lib/python removal (commit https://github.com/narategithub/ovis/commit/84134bd5a20b668078754cdf9d3bc68d8b7f0651) and SWIG removal (https://github.com/narategithub/ovis/commit/885725260c95e11ccd506fa6aa009215ef479d8c) are also included as they were found unused. Splitting these commits into their own pull requests will create a merge conflict afterward as they modify the same section in configure.ac file.


configure.ac
```diff
@@ -548,12 +546,41 @@ if test -z "$ENABLE_SOS_TRUE"; then
...
 	pkgpythondir="${pythondir}/ovis_ldms"
 	pkgpyexecdir="${pkgpythondir}"
 	AX_PYTHON_MODULE_VERSION(["argparse"], 1.1, "python")
+
+	dnl cython check
+	if test "${CYTHON}" == "no"; then
+		dnl cython not found, try cython3 or cython3.6
+		AC_MSG_NOTICE(cython not found ... trying cython3 or cython3.6)
+		CYTHON=$(which cython3 2>/dev/null || \
+			 which cython3.6 2>/dev/null || \
```
morrone, line 573: EPEL package for RHEL4 is cython-3.8 (Yes, that one really has a dash in the name, sigh). We'll need that too.

configure.ac
```diff
@@ -548,12 +546,41 @@ if test -z "$ENABLE_SOS_TRUE"; then
 	fi
 fi
 
-dnl ldms-python API's
+dnl ldms-python API's and programs
+AX_WITH_PROG([CYTHON], [cython], [no])
+CYTHON_REQ_VER="0.28.5"
```
morrone, line 551: How about "CYTHON_MIN_VER"? Makes it clearer that it is a minimum rather than a single required version.

configure.ac
```diff
@@ -548,12 +546,41 @@ if test -z "$ENABLE_SOS_TRUE"; then
...
+		CYTHON=$(which cython3 2>/dev/null || \
+			 which cython3.6 2>/dev/null || \
+			 echo "no")
+		test "${CYTHON}" == "no" || AC_MSG_NOTICE(Using cython: ${CYTHON})
+	fi
+	test "${CYTHON}" != "no" || \
+		AC_MSG_ERROR(`cython` program not found. $CYTHON_MSG)
+	CYTHON_VER=$(${CYTHON} --version 2>&1 | sed -e 's/.*\s//')
+	AC_MSG_NOTICE(Cython version: ${CYTHON_VER})
+	LE_VER=$(echo -e "${CYTHON_VER}\n${CYTHON_REQ_VER}"| sort -V|head -n1)
```
morrone, line 581: Please use AX_COMPARE_VERSION for this.

configure.ac
```diff
@@ -548,12 +546,41 @@ if test -z "$ENABLE_SOS_TRUE"; then
...
+CYTHON_REQ_VER="0.28.5"
+CYTHON_MSG="
+
+    Please install Cython ver >= ${CYTHON_REQ_VER}. It is recommended to install
+    Cython via pip3 (\`pip3 install Cython\`) to get the latest version of
+    Cython.  Some installation package (e.g. python36-Cython from epel) may
+    install cython program with cython<VER> form (e.g. /bin/cython3.6). In such
+    case, --with-cython=FULL_PATH_TO_CYTHON_PROGRAM option is required (e.g.
+    --with-cython=/bin/cython3.6).
+
```
morrone, line 560: This message isn't accurate. The code below is checking for cython3.6. I'm not sure this comment is needed either. "--with" is pretty standard for specifying a custom path to something, and pip3 is not at all a universally recommended way to handle the installation. It probably isn't the way many HPC installations would do it.

configure.ac
```diff
@@ -548,12 +546,41 @@ if test -z "$ENABLE_SOS_TRUE"; then
 	fi
 fi
 
-dnl ldms-python API's
+dnl ldms-python API's and programs
+AX_WITH_PROG([CYTHON], [cython], [no])
```
morrone, line 550: How about using AC_CHECK_PROGS instead? I think we can use that to do something like:

AC_PATH_PROGS([CYTHON], [cython3.4 cython3.6 cython-3.8 cython], [no])

I believe that this macro will test each binary name in the list one by one, and if one is found in PATH, it stops and sets CYTHON to that.

The next thing to try is using AC_PATH_PROGS_FEATURE_CHECK instead of AC_CHECK_PROGS. It looks like that allows you to include a "feature-test", which could be the version tests (using AX_COMPARE_VERSION, for instance). This macro also allows the user to set our CYTHON variable using the CYTHON environment variable, which seems more consistent with how other compilers/linkers/etc are usually set (CC, LD, etc.). It also has an action-if-not found parameter where we can cat AC_MSG_ERROR(). It looks like this macro could wrap up everything into one section.

morrone (changes requested):

configure.ac
narategithub: Do we support RHEL4?

configure.ac
narategithub: Sounds good. Somehow I couldn't think of "MIN" last night :laughing:

configure.ac
narategithub: Why?

configure.ac
narategithub: The code below additionally looks for cython3 or cython3.6 so that on Unbuntu 18.04 (the one use for CI/CD script) and on RHEL7 we can `./configure` without having to specify `--with-cython=/usr/bin/cython3` or `--with-cython=/bin/cython3.6` respectively.

configure.ac
narategithub: `AC_PATH_PROGS([CYTHON], ...)` does not add `--with-cython` option to the configure script. The reason `AX_WITH_PROGS` is used is to so that users can specify the path to cython binary using the `--with-cython` option.

morrone: I don't think that --with-cython is necessarily the right way to do that. We don't do "--with-cc=gcc8932", we do "CC=gcc8932". Likewise, we should probably have "CYTHON=" instead of "--with-cython". The AC_PATH_PROGS_FEATURE_CHECK seems to support the "CYTHON=" method.

configure.ac
morrone: Because it is almost always better to use an existing macro than to invent our own sh scripting.

configure.ac
morrone: Sorry, brain fart. I meant RHEL8. (TOSS4 is RHEL8-based, which is where the "4" snuck in.)

.github/workflows/test-make-dist.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
 
     steps:
     - run: sudo apt install gettext
+    - name: cython
+      run: sudo pip3 install Cython
```
morrone, line 17: I would like to see this removed. Since ldms can build without Cython, we should not require Cython just to run "make dist". (And even if ldms some day requires Cython, it would be nice to be able to skip such a test to allow us to get to "make dist.) The cython check should be non-fatal and allow configure to complete so that make-dist works when cython is missing. In a build enviroment where we want to guarantee that Cython is included, we would add "--enable-cython" to thte configure command line, and that _would_ error out when Cython is missing.

tom95858:
@morrone I agree with you on all of this. BTW, I don't want to make the container a requirement, I'm advocating there creation as an "ease of use" thing.

configure.ac
narategithub: I would agree if AX_COMPARE_VERSION is a built-in. It is not though (on my CentOS7 machine). I need to download the macro from http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_compare_version.m4. Comparing to just the two-liner here ... I think I'll keep it as-is.

narategithub:
@tom95858 container? Maybe you're on the wrong thread here :)

narategithub:
This thread is already pretty long. So, here's a summary of the current status of the pull request:

- `--disable-python` is an alias of `--disable-ldms-python`. Giving either `--disable-python` or `--disable-ldms-python` will disable the build in ldms/python. Disabling python also drop Cython checking.
- If `CYTHON=<path_to_cython>` is specified (e.g. `configure CYTHON=/usr/bin/cython-3.8`), it will be used as the Cython program when building ldms/python. If `CYTHON` is specified to a non-existing path or non-executable path, configure will give an error. Cython version is still be checked.
- If `CYTHON` is not specified, search for it using `AC_PATH_PROGS` with `cython cython3 cython3.6 cython-3.8`. If no cython is found in that list, an error is reported. Cython version is also checked as well.
- Since ldms/python depends on Cython, and `python` is enabled by default, Cython check (existence/version) is also performed by default as well. Hence, a bare `./configure` will fail if Cython is not installed :(

morrone:
Thats pretty good! I apologize that I didn't think of everything all at once, but I have a few more suggestions.

I think the logic should probably go like this:

1. AC_ARG_ENABLE for python - this needs to be tri-state: enable/disable/did-not-specify. LDMS's own custom macros lose the normal tri-state capability of AC_ARG_ENABLE, and that is key here.
2. Test for python, extract the major/minor version for the Cython check later (e.g. PYTHON_VERSION="3.6")
3. If python is found, test for Cython. The searched for binary names should be: [cython${PYTHON_VERSION} cython-${PYTHON_VERSION} cython] (and it should be in that order with "cython" last)
4. Finally, resolve the findings of the checks with the AC_ARG_ENABLE results. If AC_ARG_ENABLE was "enable" and either python or cython not found, then AC_MSG_ERROR. If IF AC_ARG_ENABLE is just "did-not-specify", then either python of cython not found is only an AC_MSG_WARN.

Does that make sense? I can take a stab at it if I'm not being clear.

morrone:
I wonder too if the ARG should be named "python-api" instead of just "python". LDMS's python API requires both python and cython, but I can imagine in the future there might be some feature in LDMS that only requires python. Giving the API option a more specific name could avoid future conflict.

tom95858:
@morrone, @narategithub are you two happy enough with this to merge it, or do we need to tweak anything else?

morrone:
@tom95858 , there are still things that need additional work for me, but I wouldn't be opposed to landing this and I could follow up with additional changes.

morrone:
@baallan I do agree with that, but this PR is merged (so no one will remember to look at it), so it would be good to open an new issue to track that.

PR #410, opened 2020-09-22
Consolidate sampler default authentication values

morrone:
We modify the sampler code to eliminate the many different hard coded
sampler permission defaults, and instead inherit the default consistently
from ldms_set_new().

Also change the the default permissions from the too-permissive default
of 0777 to a more secure 0440.

This consolidation work is useful on its own as refactoring (to eliminate redundant
hard coded defaults), but it also lays the ground work for introducing 
global defaults for sampler uid, gid, and perm that can be set once at the ldmsd
level instead of having to repeat the configuration over and over again for each
smapler plugin in the configuration.

We do still preserve the capability to individually set uid/gid/perm values per
plugins if so desired.

### DIFF
--- a/ldms/man/ldmsctl.man
+++ b/ldms/man/ldmsctl.man
@@ -1,4 +1,4 @@
-.\" Manpage for ldmsctl
+\" Manpage for ldmsctl
 .\" Contact ovis-help@ca.sandia.gov to correct errors or typos.
 .TH man 8 "19 Nov 2019" "v4.3" "ldmsctl man page"
 
@@ -151,7 +151,7 @@ The group id of the set's owner. The default is the returned value of getegid().
 .TP
 .BI [perm " perm"]
 .br
-The sampler plugin instance access permission. The default is 0777.
+The sampler plugin instance access permission. The default is 0440.
 .RE
 .RE
 
--- a/ldms/man/ldmsd_controller.man
+++ b/ldms/man/ldmsd_controller.man
@@ -151,7 +151,7 @@ The group id of the set's owner. The default is the returned value of getegid().
 .TP
 .BI [perm " perm"]
 .br
-The sampler plugin instance access permission. The default is 0777.
+The sampler plugin instance access permission. The default is 0440.
 .RE
 .RE
 
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -1172,7 +1172,7 @@ ldms_set_t ldms_set_new_with_auth(const char *instance_name,
 
 ldms_set_t ldms_set_new(const char *instance_name, ldms_schema_t schema)
 {
-	return ldms_set_new_with_auth(instance_name, schema, geteuid(), getegid(), 0777);
+	return ldms_set_new_with_auth(instance_name, schema, geteuid(), getegid(), 0440);
 }
 
 int ldms_set_config_auth(ldms_set_t set, uid_t uid, gid_t gid, mode_t perm)
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -1205,14 +1205,32 @@ uint32_t ldms_set_uid_get(ldms_set_t s)
 	return __le32_to_cpu(s->set->meta->uid);
 }
 
+int ldms_set_uid_set(ldms_set_t s, uid_t uid)
+{
+	s->set->meta->uid = __cpu_to_le32(uid);
+        return 0;
+}
+
 uint32_t ldms_set_gid_get(ldms_set_t s)
 {
 	return __le32_to_cpu(s->set->meta->gid);
 }
 
+int ldms_set_gid_set(ldms_set_t s, gid_t gid)
+{
+	s->set->meta->gid = __cpu_to_le32(gid);
+        return 0;
+}
+
 uint32_t ldms_set_perm_get(ldms_set_t s)
 {
 	return __le32_to_cpu(s->set->meta->perm);
+}
+
+int ldms_set_perm_set(ldms_set_t s, mode_t perm)
+{
+	s->set->meta->perm = __cpu_to_le32(perm);
+        return 0;
 }
 
 extern uint32_t ldms_set_meta_sz_get(ldms_set_t s)
--- a/ldms/src/core/ldms.h
+++ b/ldms/src/core/ldms.h
@@ -1147,6 +1147,15 @@ extern uint32_t ldms_set_card_get(ldms_set_t s);
 uint32_t ldms_set_uid_get(ldms_set_t s);
 
 /**
+ * \brief Set the UID of the LDMS set.
+ * \param s The set handle.
+ * \param uid The UID to set
+ * \retval errno If failed.
+ * \retval 0     If succeeded.
+ */
+int ldms_set_uid_set(ldms_set_t s, uid_t uid);
+
+/**
  * \brief Retreive the GID of the LDMS set.
  * \param s The set handle.
  * \retval gid The GID of the set.
@@ -1154,11 +1163,29 @@ uint32_t ldms_set_uid_get(ldms_set_t s);
 uint32_t ldms_set_gid_get(ldms_set_t s);
 
 /**
+ * \brief Set the GID of the LDMS set.
+ * \param s The set handle.
+ * \param uid The GID to set
+ * \retval errno If failed.
+ * \retval 0     If succeeded.
+ */
+int ldms_set_gid_set(ldms_set_t s, gid_t gid);
+
+/**
  * \brief Retreive the permission of the LDMS set.
  * \param s The set handle.
  * \retval perm The permission of the set.
  */
 uint32_t ldms_set_perm_get(ldms_set_t s);
+
+/**
+ * \brief Set the permissions of the LDMS set.
+ * \param s The set handle.
+ * \param perm The UNIX mode_t bits (see chmod)
+ * \retval errno If failed.
+ * \retval 0     If succeeded.
+ */
+int ldms_set_perm_set(ldms_set_t s, mode_t perm);
 
 /**
  * \brief Get the size in bytes of the set's meta data
--- a/ldms/src/sampler/app_sampler/app_sampler.c
+++ b/ldms/src/sampler/app_sampler/app_sampler.c
@@ -1240,14 +1240,12 @@ int __handle_task_init(app_sampler_inst_t inst, json_entity_t data)
 	if (!app_set)
 		return ENOMEM;
 	app_set->task_pid = task_pid->value.int_;
-	set = ldms_set_new_with_auth(setname, inst->base_data->schema,
-				     inst->base_data->uid,
-				     inst->base_data->gid,
-				     inst->base_data->perm);
+        set = ldms_set_new(setname, inst->base_data->schema);
 	if (!set) {
 		free(app_set);
 		return errno;
 	}
+        base_auth_set(&inst->base_data->auth, set);
 	ldms_metric_set_u64(set, BASE_JOB_ID, job_id->value.int_);
 	ldms_metric_set_u64(set, BASE_COMPONENT_ID, inst->base_data->component_id);
 	ldms_metric_set_u64(set, inst->task_rank_idx, task_rank->value.int_);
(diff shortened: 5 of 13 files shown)


baallan:
@morrone re 777->440 change, will this change affect anyone but munge users? If so, this breaks the use of ldms_ls already present in production administrators scripts.
If the perms affect only munge (as I expect) then this is fine.

morrone:
> @morrone re 777->440 change, will this change affect anyone but munge users? If so, this breaks the use of ldms_ls already present in production administrators scripts.
> If the perms affect only munge (as I expect) then this is fine.

This should apply to any auth mechanism that uses uid/gid/perm for protecting access to metric sets. I am guessing that "munge" and "naive" are the ones that currently employs those.

With the auth "ovis" (and I assume auth "none") plugin, uid/gid/perm are not used. There is no sense of user or groups in auth "ovis", and I'm not aware of the perms doing anything either. At least, ldms_ls reports "-rwxrwxrwx" for metrics regardless of what the perms are set to in the code when auth_ovis is used. The only thing that appears to matter for auth_ovis is if you know the secret key, and then you get full access.

But if anyone knows otherwise, that is why we have code review. :)

tom95858:
@morrone, just for the purpose of clarification, the auth mechanisms don't protect access to metric sets. They allow or disallow a connection from a peer, and they set the uid, gid in the ldms transport structure. 

Regardless of the authentication method, the uid/gid from the ldms transport structure are used to allow or disallow requests received on that transport.  They are also used to filter content for requests such as LDMS_DIR by returning only the metric set names for which the uid/gid in the transport allow read access.

However, all auth methods except munge unconditionally set the uid/gid to zero. Munge sets this to the effective uid and gid of the connecting peer.

We really need to document what the mode bits mean for the different object types, e.g. metric-set, vs. producer, vs updater, etc...

morrone:
> @morrone, just for the purpose of clarification, the auth mechanisms don't protect access to metric sets. They allow or disallow a connection from a peer, and they set the uid, gid in the ldms transport structure.

So does that imply that the peer creates separate connections for every metric set that it wants to read?

> However, all auth methods except munge unconditionally set the uid/gid to zero. Munge sets this to the effective uid and gid of the connecting peer.

auth_naive claims to allow uid and gid to be set to anything they want. I would have assumed that it worked like auth_munge,
except without any actual security value (but maybe testing value, or some other use). Is that not how it works?

> We really need to document what the mode bits mean for the different object types, e.g. metric-set, vs. producer, vs updater, etc...

Yes, that would be great!

tom95858:
> > @morrone, just for the purpose of clarification, the auth mechanisms don't protect access to metric sets. They allow or disallow a connection from a peer, and they set the uid, gid in the ldms transport structure.
> 
> So does that imply that the peer creates separate connections for every metric set that it wants to read?

Only if you are using munge and the user wants to be both mary and bob. The process would have to seteuid to 'mary', then connect to the ldmsd, and then seteuid to 'bob' and connect to ldmsd.

Otherwise the peer would have to be root to see both bob and mary's sets at the same time. Unless, of course, bob and mary are in the same group, and the set perms were 0440 in which case it would work just fine with one connection.

> 
> > However, all auth methods except munge unconditionally set the uid/gid to zero. Munge sets this to the effective uid and gid of the connecting peer.
> 
> auth_naive claims to allow uid and gid to be set to anything they want. I would have assumed that it worked like auth_munge,
> except without any actual security value (but maybe testing value, or some other use). Is that not how it works?
> 

You are correct, that's how naive works and it should not be used in production in my opinion and is useful only for testing.


I 
> > We really need to document what the mode bits mean for the different object types, e.g. metric-set, vs. producer, vs updater, etc...
> 
> Yes, that would be great!

I'll add it to my list.

morrone:
Oh, I think I understand. When you said " the auth mechanisms don't protect access to metric sets. They allow or disallow a connection from a peer, and they set the uid, gid in the ldms transport structure.", you were just clarifying that the auth mechanisms only do authentication, not authorization. Authorization is handled at another layer. Got it.

Nevermind about my earlier comment ldms_ls auth_ovis reporting the wrong perms. I must have flubbed my testing and pointed at the wrong server earlier. I am seeing the perms as 0440 under ldms_ls with my patches and auth_ovis.

So yes, this changes the metric perms to 0440 in all cases by default. But that does mean that all metrics are still readable by default with auth_ovis. I don't think the executable bit is really used for metrics.

That just leaves the write bit. Do any metrics need to be writable to avoid breaking "production administrators scripts"?

morrone:
It occurs to me that the perm bits irrelevant for metric sets when the uid/gid is 0. uid/gid 0 is allowed full access in ldms. So it is unlikely that this patch would cause any issue for auth_ovis users.

baallan:
@morrone afaik all present production scenarios treat sets as write-once, read everywhere. 
I think all past development work on set transformations has proceeded on the basis of read-only and create-new.

morrone:
> @morrone afaik all present production scenarios treat sets as write-once, read everywhere.
> I think all past development work on set transformations has proceeded on the basis of read-only and create-new.

I don't understand. What type of sort of metric would use the "w" write bit associated with the metric set?

baallan:
In a present ldms-only context, I can't think of anything that would write into a metric set after the originating ldms_transaction_end. So I don't see an immediate use for the write bit. I can imagine down-stream (sosdb, future ldms, etc) a set might have a need to be written to (e.g. data or metadata items not known at the point of origin)  but known/derived elsewhere in the transmission chain.

We have periodically dreamed of (probably event driven) uses of sets to communicate among agent plugins within the ldms hierarchy; enabling write on the sets used by these agents might make sense where sets are part of some load-management scenario by an application.

I do wonder also what we do with the execute bit. Shouldn't it be 550 by default and you don't even know the set name exists, as with ls, if the execute bit is 0 instead of 1. I guess 440 is the 'paranoid' default.

tom95858:
> > @morrone afaik all present production scenarios treat sets as write-once, read everywhere.
> > I think all past development work on set transformations has proceeded on the basis of read-only and create-new.
> 

@baallan I don't know what you are trying to say here. The ldms_metric_set_xxx code does not look at a sets owner/access bits. Only remote changes use the owner/access rights.

> I don't understand. What type of sort of metric would use the "w" write bit associated with the metric set?

There is only one bit of logic that uses the 'w' bit in a metric set and that is if a remote user requests to change a local sets user-data. See ldmsd_set_udata_regex and ldmsd_set_udata are the functions that do this. They are called from ldmsd_request.c.

Configuration objects use the 'w' bit to allow a remote peer to change it. The remote peer in this case is almost always the ldmsd_controller.

morrone:
>I do wonder also what we do with the execute bit. Shouldn't it be 550 by default and you don't even know the set name exists, as with ls, if the execute bit is 0 instead of 1.

No, I don't think so. That doesn't map from filesystem concepts (and granted, ldms can do anything it wants), nor does it seem to currently behave that way.  Removing the x bit from a file doesn't make it so you don't see the file name any more. You remove the x bit from the directory _containing_ the file to make the entire contents of the directory inaccessible.

The x bit doesn't currently do anything that I am aware of (for metric sets), so it would be better to not have it set by default when we don't know what it might do in the future. If it does something useful in the future, we can set it then.

morrone:
How exactly does the jobid make it into the ldmsd...for instance with the slurm spank plugin? Is there some jobid metric that needs to be writable, or does that come in through ldms streams or something and use some other authorization path?

baallan:
I think of sets as leaf directories and metrics as the files, wrt x bits.

The job data for most sets is read from another set. Another sampler (which depends on config) creates the set containing the original job data. Implicit in this scheme is that (for correctness) the job data sampler must run before the other samplers in the same interval. There is a spank plugin and paired sampler as a method of getting the data and there are other schemes (eg. jobid sampler and prolog/epilog mods to populate a local data file on the compute nodes).

tom95858:
> How exactly does the jobid make it into the ldmsd...for instance with the slurm spank plugin? Is there some jobid metric that needs to be writable, or does that come in through ldms streams or something and use some other authorization path?

It depends on the job sampler; but the job_id is "just-like" any other metric and has the type uint64_t. This is somewhat simplified, but:

- In the jobinfo case, the sampler reads from a file and sets the job_id based on the contents of the file.
- For the spank plugin/sampler plugin case, the sampler subscribes to an ldmsd stream and receives job update information from which it sets the job_id field

The point is really that the job_id setting is accomplished by the owning ldmsd call ldms_metric_set_u64.

morrone:
> I think of sets as leaf directories and metrics as the files, wrt x bits.

@baallan, you may think of them that way, but ldms doesn't appear to. :)

> For the spank plugin/sampler plugin case, the sampler subscribes to an ldmsd stream and receives job update information from which it sets the job_id field

@tom95858 OK, thanks!  So the spank plugin doesn't need write access to the jobid metric set.

As an aside, why doesn't the spank plugin just connect as a peer and write directly to the metric set? The jobid plugin could just create an empty metric set with the correct uid/gid/perms, and then the spank plugin could just write into it directly, right? Actually, I probably know the answer...the jobid spank plugin was initially more about collecting information about the individual processes in the job than just identifying what job was running on the node. So that probably explains the added complexity.

Is it possible for peers to write into metric sets at this time?

Man, now I'm getting off topic. :)

baallan:
@morrone the fact that ldms does not yet allow per-file permissions but only directory permissions is something we will be driven to correct by down-stream security models if we wish to allow admins to follow unix mode semantics through an entire ovis pipeline (ldms-sosdb-web).

baallan:
@morrone re the actual code in this PR:

You've introduced a race condition on the permissions of a set by removing the mode bits from set_new_.... E.g. fopen(name, mode).

A better (more comprehensible) naming of the several mutators you've added would be in the style of ldms_set_chown, ldms_set_chmod, especially if you are recycling the UNIX mode_t in the argument list for the calls instead of introducing an ldms_mode_t. Granted we may have been abusing mode_t before you looked at this area...

morrone:
> @morrone re the actual code in this PR:
> 
> You've introduced a race condition on the permissions of a set by removing the mode bits from set_new_.... E.g. fopen(name, mode).

For that to be the case, metric sets would need to be accessible before ldms_set_publish() time. Are metric sets accessible before publishing? And if so, what the heck does "publish" actually mean?

And if that were correct then your next suggestion of separate chonw and chmod doesn't follow, because there would always be a race between those two calls.

> A better (more comprehensible) naming of the several mutators you've added would be in the style of ldms_set_chown, ldms_set_chmod, especially if you are recycling the UNIX mode_t in the argument list for the calls instead of introducing an ldms_mode_t. Granted we may have been abusing mode_t before you looked at this area...

No, I don't think that is at all consistent with the established naming in this API.

First of all, my ldms_set_{uid|gid|perm}_set() functions are just the direct inverse of the existing ldms_set_(uid|gid|perm)_get() calls. Using a different name would be strange an surprising. If we want to change the naming, we should do so everywhere to be consistent, and I think that change is out of scope for this PR.

Also, the mode_t is exactly consistent with everywhere else in ldms.h. "ldms_mode_t" does not appear in ldms.h, only "mode_t". If we want to change that, that is fine. But again I think that is out of scope for this PR.

Another thing to look at is that ldms_set_(uid|gid|perm)_get() all return int32_t instead of uid_t/gid_t/mode_t, whcih would make them consistent with the rest of ldms.h. But again, out of scope for this PR.

baallan:
@morrone wrote: Are metric sets accessible before publishing? And if so, what the heck does "publish" actually mean?

I second the question. We really don't have a v4 documentation of the life-cycle of set instances and how this life-cycle relates to api calls. At least not in headers or man pages.

morrone:
> @morrone wrote: Are metric sets accessible before publishing? And if so, what the heck does "publish" actually mean?
> 
> I second the question. We really don't have a v4 documentation of the life-cycle of set instances and how this life-cycle relates to api calls. At least not in headers or man pages.

My guess is that the metrics aren't avaible until publish time. But that was new in v4 (or somewhere in the v4 release stream). I don't think v3 either had, or honored the publish calls. I seem to remember there being publish calls that did nothing at one point, and then later actually worked and metrics only appeared after the metric was published.

So at the very least I would suspect that right now a metric set won't show up in ldms_ls until ldms_set_publish() is called. I don't know if there is some hidden gotcha where it is accessible through some other means before publishing. But if so, that is probably a design or implementation bug. Admittedly if such a loophole existed, it would have to be dealt with before this PR could land it in its current state. But my working assumption that it does not. I am assuming that publish works as one might expect.

baallan:
my understanding is that the set may be visible locally with or without publish when it is added to the local set tree at the end of set new. I could easily be wrong...

PR #414, opened 2020-09-23
add pre-push test to verify make dist does not depend on cython

baallan:
This action file (when connected) will verify that 'make dist' is not broken for rhel7/vanilla toss3.

### DIFF
--- a/.github/workflows/test-make-dist-nocython.yaml
+++ b/.github/workflows/test-make-dist-nocython.yaml
@@ -1,0 +1,22 @@
+name: Test "make dist r7"
+
+on:
+  push:
+    branches: [ OVIS-4 ]
+  pull_request:
+    branches: [ OVIS-4 ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-18.04
+
+    steps:
+    - run: sudo apt install gettext
+    - uses: actions/checkout@v2
+    - name: autogen
+      run: sh autogen.sh
+    - name: configure
+      run: ./configure
+    - name: make dist
+      run: make dist


baallan:
@morrone so does adding this push test make sense to you?
@narategithub I don't understand why you changed the existing 'make dist' test instead of fixing the problem it diagnosed. Cython needs to be not required unless explicitly asked for. It's fine to test both configuration cases, of course.

narategithub:
@baallan The reason is that `ldmsd_controller` requires Cython LDMS to communicate with ldmsd. Other python module in `ldms/python/ldmsd/*` are there to assist `ldmsd_controller`.

`ldms/python/cmds/kq` is the other python program that depends on SOS Python module ... and SOS Py also uses Cython.

So, I don't see anything in `ldms/python` that does not depend on Cython.

baallan:
@narategithub ldmsd_controller is optional. The c-program interface is all that is required.

narategithub:
@baallan please also see my discussion with @morrone here: https://github.com/ovis-hpc/ovis/pull/406#discussion_r492373938

If ldmsd_controller is optional, then why shouldn't we disable python by default?

baallan:
@narategithub you should disable the cython dependent bits (call it python-libs or python-api) by default. Anything that 'normally requires pip install X' on any of our supported platforms should be disabled by default.

baallan:
@narategithub perhaps we should just revert the whole python3 patch list if you can't make it pass './configure && make dist' on vanilla rhel7 by default.

morrone:
I don't think that we should necessarily add another test workflow for this. But I do think that the "make dist" workflows should be switched back to not downloading Cython. Cython is need for real builds (when desitred), not configure for make dist. See PR #415.

As to whether configure should default to not using python/cython if cython isn't found: I do basically agree. I mentioned in a PR review that I think python selection should really be tri-state. The test for the AC_ARG_ENABLE should record, "yes" (--enable provided on the command line), "no" (--disable provided on the command line), and "notspecified" (no opinion expressed on the command line...could also be named "test" or "check" or something). This is a very common idiom in autoconf.

The only reason I didn't implement that in my cython check update is that the AX_PYTHON_DEVEL macro appears to error out if python isn't found rather than allowing us to provide an action-if-not-found. So we can't run it speculatively in the "notspecified" case.

But now that consider it again, python is almost always available. So we could still implement "nospecified" for Cython. We could make it so it disables python use when cython is not found, even though missing python would always throw an error.

morrone:
Oh, and I minor note on terminology. This is not a "pre-push" test you are making, it is actually post-push. _After_ a push happens to the OVIS-4 branch it would trigger, and _after_ a pull-request is initiated that targets OVIS-4, the test would trigger.

The PR test is the important one to catch things before they are merged to the OVIS-4 branch. The OVIS-4 test is just a backup test to make sure nothing when wrong in the merge process.

baallan:
Since we are "supporting cython bindings where available" and it is available in the standard ubuntu 20 environment currently in beta(ish) on github actions, I would like to see enabling it and building ldms with python/cython enabled as one of the pre-push actions. Since we're paying for github, lets use it as much as possible. 

This is separate, of course, from having the default configure command work out of the box (no explicit disables or enables required).

Exactly how we name the actions is esthetics; if we accept PR #415 then I would rename the new file in this one to make-dist-cython and the perhaps figure out how to make it run build/install from the dist tar ball generated.

baallan:
pre-push/post-PR, tomAto, tomahto

morrone:
Yes, I think everyone agrees that full build tests would be great to add. We just need someone to decide to do it.

If/When we change the cython check to be optional, the two existing workflows should be updated to put them back to original state where the did not specify "--disable-cython". In that case this patch will still not be needed.

A build test instead would certainly be great to add. Yes, that build test could certainly build from a the make dist tarball.

morrone:
Alright, I think I've got it! See PR #417. (I dropped PR #415, 417 reverts the workflow files like I suggested here).

Give it a try, let me know if I missed anything.

baallan:
Dropping in favor of #417.

PR #417, opened 2020-09-24
Make the python & cython autoconf checks optional by default

morrone:
Make the python & cython autoconf checks optional by default.
Also allow python & cython to be required with --enable-python,
and disabled with --disable-python.

It turns out that the AX_PYTHON_DEVEL macro is dependent on the
AM_PATH_PYTHON macro. While the AX_PYTHON_DEVEL macro does not
have action-if-found/-not-found options, the AM_PATH_PYTHON macro
does. While it this still allows that AM_PATH_PYTHON can succeed
and AX_PYTHON_DEVEL can fail, in most instances this won't be
the case.

The CYTHON_CHECK macro is upgraded to have action-if-found and
action-if-not-found options. Also the macro is moved into its
own file in the m4 directory.

Since the cython check is now conditional, we remove the special
cython handling from the make dist and create release workflows.

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -12,8 +12,6 @@ jobs:
 
     steps:
     - run: sudo apt install gettext
-    - name: cython
-      run: sudo pip3 install Cython
     - uses: actions/checkout@v2
     # actions/checkout@v2 breaks annotated tags by converting them into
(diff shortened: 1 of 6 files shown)


baallan:
@morrone Thanks. builds out of the box for me on our toss3 machines.

Issue #423, opened 2020-09-25
ldmsd throws Floating point exception on startup


morrone:
I am attempting to set up an ldmsd as an aggregator, and my current configuration is reliable tripping a Floating point exception at start up. This is basically at commit 0746b85, except it also has my commits from PR #410. I don't suspect they are related yet, but full disclosure.

Here is a run under gdb with the tail of the -v DEBUG output, and the backtrace from the FPE.

```
[cut]
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 245 flags 3 rec_len 28 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : ldmsd_prdcr_new(name opal181, xprt rdma, host opal181, port 411, type 0, intv 2000000
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 247 flags 3 rec_len 28 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : ldmsd_prdcr_new(name opal182, xprt rdma, host opal182, port 411, type 0, intv 2000000
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 249 flags 3 rec_len 28 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : ldmsd_prdcr_new(name opal183, xprt rdma, host opal183, port 411, type 0, intv 2000000
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 251 flags 3 rec_len 28 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 253 flags 3 rec_len 43 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : Checking for store_csv in /usr/lib64/ovis-ldms
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 255 flags 3 rec_len 28 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : ldmsd_plugattr_create: called without conf file.
Fri Sep 25 01:18:09 2020: DEBUG     : store_csv: parsed (null)
[New Thread 0x2aaaaf4f2700 (LWP 59266)]
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 256 flags 3 rec_len 28 rsp_err 0
Fri Sep 25 01:18:09 2020: DEBUG     : msg_no 257 flags 3 rec_len 28 rsp_err 0
[...]

PR #424, opened 2020-09-25
Verify the given interval and offset values of an updater are valid

nichamon:
The checks are
 * the given interval value is not an empty string,
 * the given interval value is a number,
 * the given interval value is greater than 0,
 * the given offset value is not an empty string,
 * the given offset value is a number.

`ldmsd_updtr_new_with_auth()` receives the interval and offset values as
char *. The checks are made before the function is called to keep the
patch small, i.e., no changes to the function signature and its
references. If the checks are made in the fucntion, there are no trivial
ways to send informative error messages to remote clients, e.g.,
`ldmsd_controller`.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -2632,7 +2632,9 @@ static int updtr_add_handler(ldmsd_req_ctxt_t reqc)
 	gid_t gid;
 	int perm;
 	char *perm_s = NULL;
+	char *endptr;
 	int push_flags, is_auto_task;
+	long interval, offset;
 
 	reqc->errcode = 0;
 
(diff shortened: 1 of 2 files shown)


baallan:
I thought the interval == 0 was allowed for one-time immediate events.

If trapping errors later is hard, this location is also the place where we need a check for abs(offset ) <= interval/2 (or however we code the check to be consistent with the scheduling use of offset).

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -2660,9 +2662,56 @@ static int updtr_add_handler(ldmsd_req_ctxt_t reqc)
...
+			cnt = Snprintf(&reqc->line_buf, &reqc->line_len,
+				"The given update interval value (%s) is not a number.",
+				interval_str);
+			goto send_reply;
+		} else {
+			if (0 >= interval) {
+				reqc->errcode = EINVAL;
+				cnt = Snprintf(&reqc->line_buf, &reqc->line_len,
+						"The update interval value must "
+						"not be or less than 0.");
```
tom95858, line 2688: @nichamon, please restate this as:
```
The update interval must be greater than zero
```

tom95858:
@nichamon, also add argument checking in ldmsd_updtr_new_with_auth.

tom95858:
@nichamon sorry one more thing, add an 'invalid argument' test(s) to the ldmsd_updtr_new_with_auth test cases.

ldms/src/ldmsd/ldmsd_request.c
nichamon: Done

nichamon:
> @nichamon, also add argument checking in ldmsd_updtr_new_with_auth.

Done

nichamon:
> @nichamon sorry one more thing, add an 'invalid argument' test(s) to the ldmsd_updtr_new_with_auth test cases.

@tom95858 I'll add this to my list.

nichamon:
> I thought the interval == 0 was allowed for one-time immediate events.
> 
> If trapping errors later is hard, this location is also the place where we need a check for abs(offset ) <= interval/2 (or however we code the check to be consistent with the scheduling use of offset).

@baallan I added the verification of offset. The logic used in this patch is the same as the one used to verify the sample offset.

tom95858:
@nichamon, can you please add 
```
Fixes #423 
```
To the end of the commit message.

nichamon:
> @nichamon, can you please add
> 
> ```
> Fixes #423 
> ```
> 
> To the end of the commit message.

@tom95858 Done.

PR #426, opened 2020-09-27
`ldms_init()` returns errno for consistency

narategithub:
`ldms_init()` returned `-1` on error, which is inconsistent to other
ldms interfaces. It shall return `errno` for consistency.


baallan:
I'm inclined to think that init for the package would be better returning -1 and setting errno as many other unix tools do.

PR #428, opened 2020-09-27
LDMS Cython with proper exceptions

narategithub:
LDMS Cython was using `RuntimeError` to raise an error from within the
module to the Python application. Even with string describing the error,
it lacks useful attribute such as `errno` from the underlying C code.

In this patch, `RuntimeError` usages were replaced with
`ConnectionError` if the error concerned a transport. For other errors
not directly concerning a transport but stemmed from C function calls,
`OSError` (the parent of all exceptions that have `errno` attribute) is
used with the corresponding `errno`. For all other errors not concerning
C function calls, the more appropriate exception classes from the Python
"builtins" were selected.

### DIFF
--- a/ldms/python/ldms.pyx
+++ b/ldms/python/ldms.pyx
@@ -203,8 +203,8 @@ def init(int max_sz):
     """init(max_sz) - initialize LDMS with memory pool `max_sz` bytes"""
     cdef int rc = ldms_init(max_sz)
     if rc:
-        raise RuntimeError("ldms_init({}) error: {}"\
-                           .format(max_sz, ERRNO_SYM(rc)))
+        raise OSError(rc, "ldms_init({}) error: {}" \
+                          .format(max_sz, ERRNO_SYM(rc)))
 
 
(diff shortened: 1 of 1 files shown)


ldms/python/ldms.pyx
```diff
@@ -1947,7 +1947,8 @@ cdef class Xprt(object):
                 this function will just be blocked indefinitely.
         """
         if self._conn_cb:
-            raise RuntimeError("The callback has been supplied to `connect()`. "
+            raise TypeError("Bad `Xprt.recv()` call. "
```
tom95858, line 1950: @narategithub I think this should be a ValueError, from the Python3 docs:

```
Raised when an operation or function receives an argument that has the right type 
but an inappropriate value...
```

narategithub: Done. I was reluctant between `TypeError` and `ValueError`. Neither of them are exactly fit because this function does not receive an argument, but I think they are the closest to this error: using synchronous recv() on an asynchronous transport.

PR #429, opened 2020-09-27
Add LDMS Cython tests (also serves as examples)

narategithub:


narategithub:
NOTE: Since the test uses lookup by schema, it needs #427.

PR #430, opened 2020-09-28
Extend the ovis_json library functionalities

nichamon:
The following functionalities are added.
* Dump a JSON entity to a string (json_entity_dump)
* Copy a JSON entity (json_entity_copy)
* Create a JSON dictionary entity by giving the list of its
attribute-value pairs (json_dict_build)
* Return the number of dictionary attributes (json_attr_cnt)
* Remove an attribute from a dictionary (json_attr_rem)
* Pop an element of a list (json_item_pop)
* Remove an element of a list (json_item_rem)

For ease of use, the following changes are made.
* To create a JSON ATTR entity, json_entity_new() receives char *
instead of json_entity_t for the attribute key
* json_attr_add() receives char * as key and json_entity_t as attribute
value, instead of a single attribute entity itself

### DIFF
--- a/lib/src/ovis_json/ovis_json.c
+++ b/lib/src/ovis_json/ovis_json.c
@@ -94,6 +94,8 @@ json_entity_t json_attr_first(json_entity_t d)
 	hent_t ent;
 	assert(d->type == JSON_DICT_VALUE);
 	ent = htbl_first(d->value.dict_->attr_table);
+	if (!ent)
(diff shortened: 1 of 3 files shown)


nichamon:
@tom95858 The changes to json_entity_new() and  json_attr_add() don't affect any other source files because there are no callers. I tested the changes with the FVT that is to be merged into the TADA repo.

PR #433, opened 2020-09-29
Update the README

morrone:
This is a first pass at updating the REAME before the 4.3.4 release. It could probably use more work, but perhaps it will be a good starting point.


README.md
```diff
@@ -23,88 +22,101 @@ applications yield the processor and transported using RDMA-like operations, res
...
-to discover sequence of occurrence patterns of log messages and to perform causal analysis.
+  https://github.com/ovis-hpc/ovis/releases
 
-Baler has been moved out of the OVIS tree. Please check it out at <https://github.com/opengridcomputing/baler>
+The tarball is avialble in the "Assets" section of each release. Be sure to
+download the tarball that has a name of the form "ovis-ldms-X.X.X.tar.gz".
 
-# Obtaining OVIS source code
+The links that are named "Source code (zip)" and "Source code (tar.gz)" are
+automatic GitHub links that we are unable to remove. Yhey will be missing the
```
tom95858, line 42: *They will be missing

README.md
```diff
@@ -23,88 +22,101 @@ applications yield the processor and transported using RDMA-like operations, res
...
+## Build Dependencies
+
+* autoconf (>=2.63)
+* automake
+* libtool
+* make
+* bison
+* flex
+* gettext development package
+* libevent developmetn package
```
tom95858, line 62: *development

README.md
```diff
@@ -23,88 +22,101 @@ applications yield the processor and transported using RDMA-like operations, res
...
 For cray-related LDMS sampler plug-in dependencies, please see the man page of the
 plug-in in `ldms/man/`.
 
+### RHEL7/CentOS7 dependencies
+
+RHEL7/CentOS7 systems will require a the following packages at a minimum:
 
-# Building OVIS
+* autoconf
+* automake\
```
tom95858, line 79: trailing backslash?

README.md
```diff
@@ -23,88 +22,101 @@ applications yield the processor and transported using RDMA-like operations, res
...
 * Ubuntu and friends
 * CentOS and friends
 * Cray XE6, Cray XK, Cray XC
 
 # Unsupported features
-LDMS sampler plugins
+
+The following LDMS sampler plugins are considered unsupported. Use are your own risk:
 * perfevent sampler
 * papi sampler
```
tom95858, line 121: Do you mean the hweventpapi sampler? The papi sampler is supported.

README.md
narategithub: @morrone actually, we don't need libevent anymore. I think the comment in `configure.ac` stating that we do was a copy-paste error.

README.md
morrone: Yep.

README.md
morrone: OK.

README.md
morrone: Yes.

README.md
morrone: I don't know on that one. That was left over from the previous version. I'll assume hweventpapi is the right one and make that change.

morrone:
In addition to the fixes for the review comments, in this version I delete a few of the other README files that are now obsolete. I also delete the obsolete Changelog files in the second commit.

PR #442, opened 2020-10-01
Fix ldmsctl crashes when it receives a multiple-record message

nichamon:
Without the fix, ldmsctl reused a single buffer to receive message
records. It assumed that the buffer would be handled before the next
record arrives. The assumption is not always true if a response message
consists of multiple records. The symptom is as if ldmsctl does not
receive some records, resulting in ldmsctl crashes.

### DIFF
--- a/ldms/src/ldmsd/ldmsctl.c
+++ b/ldms/src/ldmsd/ldmsctl.c
@@ -108,8 +108,15 @@ extern int read_history ();
 
 static char *linebuf;
 static size_t linebuf_len;
-static char *buffer;
-static size_t buffer_len;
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 @narategithub found this bug by running the ldmsd_interface_msg_boundary.py test in the source tree. The patch passes the test.

nichamon:
@tom95858 @narategithub The last version passes the test script. After I created the pull request, I found some bugs and a leak. I also initially removed the union from the ldmsctl_ctrl because currently we only use ldms_xprt. However, the change to the ldmsctl_ctrl is unrelated to the patch and makes it hard to follow the diff.

PR #443, opened 2020-10-01
Remove SLURM-related configure options

morrone:
Instead of having individual module build options, just build
slurm plugins if their dependecies are met.

The slurm_sampler plugin appears to have no external dependencies,
so it is build unconditionally.

The jobinfo_slurm and slurm_notifier slurm spank plugins are built
it slurm is detected.  They can be required by using "--with-slurm"
and turned off by using "--without-slurm".

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -294,9 +294,6 @@ OPTION_DEFAULT_DISABLE([switchx], [ENABLE_SWITCHX])
 OPTION_WITH([switchx], [SWITCHX],[/usr/local])
 OPTION_DEFAULT_DISABLE([kokkos], [ENABLE_KOKKOS])
 OPTION_DEFAULT_ENABLE([jobinfo-sampler], [ENABLE_JOBINFO])
-OPTION_DEFAULT_DISABLE([jobinfo-slurm], [ENABLE_SLURM_JOBINFO])
-OPTION_DEFAULT_DISABLE([spank-plugin], [ENABLE_SPANK_PLUGIN])
-OPTION_DEFAULT_ENABLE([slurm-sampler], [ENABLE_SLURM_SAMPLER])
 OPTION_DEFAULT_DISABLE([papi-sampler], [ENABLE_PAPI_SAMPLER])
 OPTION_DEFAULT_DISABLE([ibm_occ], [ENABLE_IBM_OCC_SAMPLER])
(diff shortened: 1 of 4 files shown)


tom95858:
I'm worried that this will cause confusion for users that have a build machine that is missing the necessary slurm bits but nonetheless expect slurm support. With this change, they won't discover they're missing the slurm bits on their build server until they try to deploy their build. Am I missing something?

morrone:
> I'm worried that this will cause confusion for users that have a build machine that is missing the necessary slurm bits but nonetheless expect slurm support. With this change, they won't discover they're missing the slurm bits on their build server until they try to deploy their build. Am I missing something?

If a user requires slurm bits, then they would want to specify "--with-slurm". If "--with-slurm" is slurm is not found, an error will be thrown at configure time. "--with-slurm" means "build all the things that depend on slurm". "--without-slurm" means "don't build that things that depend on slurm". Specifying nothing means "build the slurm dependent components if you happen to find slurm".

In other words, they definitely will see an error on their build server if they specify "--with-slurm", but slurm is missing. (Barring any bugs, of course.)

The main change here is that it cleans up and simplifies things for users at configure time. This change makes it so there is just one option that is focused on the dependency (slurm) instead of three different configure options focused on plugins (and maybe a fourth one to set the path to slurm, but I'm not looking at the code at the moment).

This is a lot easier for people getting started with LDMS. Instead of having to learn the details of LDMS's plugins (Why are there at least three different slurm-related options? which ones might I need and which do I not? Why wouldn't I want to just have them all built?), the user can just go "Oh, I have SLURM, I'll set --with-slurm". Then the LDMS configuration issues can be dealt with later at ldmsd run/configuration (here I mean ldmsd -c, type configuration, not the configure script). At build time, they can just specify the simpler dependencies without already needing deep understanding of ldms.

This also helps if the person packaging ldms is not the same as the person the makes the run-time configuration of ldmsd. The latter might have a deeper understanding of ldms plugins than the former.

All in all, I'd like to see us move away from the current excess of configuration options (many that don't really provide a great services). I see this as one step in the cleanup process.

And FYI, I don't necessarily expect to see this land until after 4.3.4 is tagged. If there something else that absolutely needs to land and triggers a whole new testing cycle, then sure, this is probably low enough impact to be included as well. But I wouldn't want it to add any more time to 4.3.4.

PR #445, opened 2020-10-02
V434 mem leak repairs

baallan:
This fixes several memory leaks in configuration and ldms_ls handling paths and adds the test scripts that uncovered them.
Each bug fix is in its own commit.


lib/src/ovis_json/ovis_json_parser.y
```diff
@@ -109,9 +109,11 @@ attr_list: /* empty */ { $$ = new_dict_val(); }
 	 | string ':' value {
 	     json_entity_t e = new_dict_val();
 	     $$ = add_dict_attr(e, $1, $3);
+	     json_entity_free($1);
```
tom95858, line 112: @nichamon can you please confirm this leak?

lib/src/ovis_json/ovis_json_parser.y
```diff
@@ -109,9 +109,11 @@ attr_list: /* empty */ { $$ = new_dict_val(); }
...
 	 }
 	 | attr_list ',' string ':' value {
 	     $$ = add_dict_attr($1, $3, $5);
+	     json_entity_free($3);
```
nichamon, line 116: @tom95858 With the same reason above, this is also legit.

tom95858:
Thanks @baallan for finding and fixing this.

lib/src/ovis_json/ovis_json_parser.y
baallan: @nichamon is there an input file that exercises the entire grammar (lex and yacc) that we can test routinely as the config tools evolve?

PR #446, opened 2020-10-02
Maintenance of ldmsd tests that use python scripts

narategithub:
Maintenance of ldmsd tests that use python scripts due to the change of python2->python3 and xprt.connect() interface change. Also fix bugs found along the way.


narategithub:
@tom95858 @nichamon FYI, the change in `ldmsd_controller`:
- only affect the `--debug` mode to raise the exception and making it easier to debug, and
- a python3-ify in `greeting` (a test command) handler

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5050,7 +5050,7 @@ static int __greeting_path_req_handler(ldmsd_req_ctxt_t reqc)
 		attr.attr_len = strlen(myself);
 		ldmsd_hton_req_attr(&attr);
 		ldmsd_append_reply(reqc, (char *)&attr, sizeof(attr), LDMSD_REQ_SOM_F);
-		ldmsd_append_reply(reqc, myself, strlen(myself), 0);
+		ldmsd_append_reply(reqc, myself, strlen(myself)+1, 0);
```
nichamon, line 5053: Why do you +1 here?

narategithub: so that the content is terminated by `'\0'`. Otherwise, the next level ldmsd will have to add `'\0'` before adding its name to the reply.

tom95858: pls. make certain that the ldms_ctxt_alloc takes that into account.

narategithub: I'm confused. This is in `greeting` request message handler in ldmsd.

PR #448, opened 2020-10-05
fix V4 auth plugin location issue and log library access problem

baallan:
This request creates an ldms (rather than ldmsd) logging library from the logging API defined in ldmsd and then uses it in the auth plugins loader. Because ldms_xprt and ldms_auth are part of core/ and both need to log errors appropriately, logging  belongs in core and not in ldmsd/. The changes to the auth plugin loader are in a separate patch for easier review of the auth logic adjustment.

The refactored 'ldms_log.h" api provides back compatiblity with the former ldmsd declarations. This can be easily removed with sed over the entire code base, but to avoid conflicts we should do that sed at a point where everyone is pretty much checked in. 

This change allows all plugin and core library writers to assume that ldms_log() just exists instead of passing around logging pointers.
Refactoring plugin apis to stop passing log pointers is a little more complicated than sed and should also be done at a community negotiated point.

### DIFF
--- a/ldms/src/core/ldms_log.c
+++ b/ldms/src/core/ldms_log.c
@@ -1,0 +1,339 @@
+/* -*- c-basic-offset: 8 -*-
+ * Copyright (c) 2010-2019 National Technology & Engineering Solutions
+ * of Sandia, LLC (NTESS). Under the terms of Contract DE-NA0003525 with
+ * NTESS, the U.S. Government retains certain rights in this software.
+ * Copyright (c) 2010-2019 Open Grid Computing, Inc. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the BSD-type
+ * license below:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *      Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ *
+ *      Redistributions in binary form must reproduce the above
+ *      copyright notice, this list of conditions and the following
+ *      disclaimer in the documentation and/or other materials provided
+ *      with the distribution.
+ *
+ *      Neither the name of Sandia nor the names of any contributors may
+ *      be used to endorse or promote products derived from this software
+ *      without specific prior written permission.
+ *
+ *      Neither the name of Open Grid Computing nor the names of any
+ *      contributors may be used to endorse or promote products derived
+ *      from this software without specific prior written permission.
+ *
+ *      Modified source versions must be plainly marked as such, and
+ *      must not be misrepresented as being the original software.
+ *
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "ldms_log.h"
+#include "ovis_util/util.h"
+#include <sys/time.h>
+
+#include <unistd.h>
+#include <inttypes.h>
+#include <stdarg.h>
+#include <getopt.h>
+#include <stdlib.h>
+#include <sys/errno.h>
+#include <stdio.h>
+#include <syslog.h>
+#include <string.h>
+#include <sys/queue.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/fcntl.h>
+#include <sys/mman.h>
+#include <pthread.h>
+#include <netinet/in.h>
+#include <signal.h>
+
+#ifdef DEBUG
+#include <mcheck.h>
+#endif /* DEBUG */
+
+static char *logfile = NULL;
+static pthread_mutex_t log_lock = PTHREAD_MUTEX_INITIALIZER;
+static FILE *log_fp;
+static int log_level_thr = LDMSD_LERROR;  /* log level threshold */
+static int quiet = 0; /* Is verbosity quiet? 0 for no and 1 for yes */
+
+const char* ldms_loglevel_names[] = {
+	LOGLEVELS(LDMSD_STR_WRAP)
+	NULL
+};
+
+int ldms_loglevel_set(char *verbose_level)
+{
+	int level = -1;
+	if (0 == strcmp(verbose_level, "QUIET")) {
+		quiet = 1;
+		level = LDMSD_LLASTLEVEL;
+	} else {
+		level = ldms_str_to_loglevel(verbose_level);
+		quiet = 0;
+	}
+	if (level < 0)
+		return level;
+	log_level_thr = level;
+	return 0;
+}
+
+enum ldms_loglevel ldms_loglevel_get()
+{
+	return log_level_thr;
+}
+
+int ldms_loglevel_to_syslog(enum ldms_loglevel level)
+{
+	switch (level) {
+#define MAPLOG(X,Y) case LDMSD_L##X: return LOG_##Y
+	MAPLOG(DEBUG,DEBUG);
+	MAPLOG(INFO,INFO);
+	MAPLOG(WARNING,WARNING);
+	MAPLOG(ERROR,ERR);
+	MAPLOG(CRITICAL,CRIT);
+	MAPLOG(ALL,ALERT);
+	default:
+		return LOG_ERR;
+	}
+#undef MAPLOG
+}
+
+/* Impossible file pointer as syslog-use sentinel */
+#define LDMSD_LOG_SYSLOG ((FILE*)0x7)
+
+void __ldms_log(enum ldms_loglevel level, const char *fmt, va_list ap)
+{
+	if ((level != LDMSD_LALL) &&
+			(quiet || ((0 <= level) && (level < log_level_thr))))
+		return;
+	if (log_fp == LDMSD_LOG_SYSLOG) {
+		vsyslog(ldms_loglevel_to_syslog(level),fmt,ap);
+		return;
+	}
+	time_t t;
+	struct tm tm;
+	char dtsz[200];
+
+	pthread_mutex_lock(&log_lock);
+	if (!log_fp) {
+		pthread_mutex_unlock(&log_lock);
+		return;
+	}
+	t = time(NULL);
+	localtime_r(&t, &tm);
+	if (strftime(dtsz, sizeof(dtsz), "%a %b %d %H:%M:%S %Y", &tm))
+		fprintf(log_fp, "%s: ", dtsz);
+
+	if (level < LDMSD_LALL) {
+		fprintf(log_fp, "%-10s: ", ldms_loglevel_names[level]);
+	}
+
+	vfprintf(log_fp, fmt, ap);
+	fflush(log_fp);
+	pthread_mutex_unlock(&log_lock);
+}
+
+void ldms_log(enum ldms_loglevel level, const char *fmt, ...)
+{
+	va_list ap;
+	va_start(ap, fmt);
+	__ldms_log(level, fmt, ap);
+	va_end(ap);
+}
+
+/* All messages from the ldms library are of e level.*/
+#define LDMSD_LOG_AT(e,fsuf) \
+void ldms_l##fsuf(const char *fmt, ...) \
+{ \
+	va_list ap; \
+	va_start(ap, fmt); \
+	__ldms_log(e, fmt, ap); \
+	va_end(ap); \
+}
+
+LDMSD_LOG_AT(LDMSD_LDEBUG,debug);
+LDMSD_LOG_AT(LDMSD_LINFO,info);
+LDMSD_LOG_AT(LDMSD_LWARNING,warning);
+LDMSD_LOG_AT(LDMSD_LERROR,error);
+LDMSD_LOG_AT(LDMSD_LCRITICAL,critical);
+LDMSD_LOG_AT(LDMSD_LALL,all);
+
+static char msg_buf[4096];
+void ldms_msg_logger(enum ldms_loglevel level, const char *fmt, ...)
+{
+	va_list ap;
+	va_start(ap, fmt);
+	vsnprintf(msg_buf, sizeof(msg_buf), fmt, ap);
+	ldms_log(level, "%s", msg_buf);
+}
+
+enum ldms_loglevel ldms_str_to_loglevel(const char *level_s)
+{
+	int i;
+	for (i = 0; i < LDMSD_LLASTLEVEL; i++)
+		if (0 == strcasecmp(level_s, ldms_loglevel_names[i]))
+			return i;
+	if (strcasecmp(level_s,"QUIET") == 0) {
+		return LDMSD_LALL;
+				}
+	if (strcasecmp(level_s,"ALWAYS") == 0) {
+		return LDMSD_LALL;
+	}
+	if (strcasecmp(level_s,"CRIT") == 0) {
+		return LDMSD_LCRITICAL;
+	}
+	return LDMSD_LNONE;
+}
+
+const char *ldms_loglevel_to_str(enum ldms_loglevel level)
+{
+	if ((level >= LDMSD_LDEBUG) && (level < LDMSD_LLASTLEVEL))
+		return ldms_loglevel_names[level];
+	return "LDMSD_LNONE";
+}
+
+/** return a file pointer or a special syslog pointer */
+int ldms_log_open(const char *progname, const char* xlogfile, const char **errstr)
+{
+	if (!progname || ! xlogfile || !errstr) {
+		ldms_log(LDMSD_LERROR, "invalid arguments to ldms_open_log: %p %p %p\n",
+			progname, xlogfile, errstr);
+		return EINVAL;
+	}
+	FILE *f;
+	if (strcasecmp(xlogfile,"syslog")==0) {
+		ldms_log(LDMSD_LDEBUG, "Switching to syslog.\n");
+		f = LDMSD_LOG_SYSLOG;
+		openlog(progname, LOG_NDELAY|LOG_PID, LOG_DAEMON);
+		log_fp = f;
+		return 0;
+	}
+
+	logfile = strdup(xlogfile);
+	if (!logfile) {
+		return ENOMEM;
+	}
+	f = fopen_perm(logfile, "a", LDMS_DEFAULT_FILE_PERM);
+	if (!f) {
+		ldms_log(LDMSD_LERROR, "Could not open the log file named '%s'\n",
+							logfile);
+		*errstr = "log open failed";
+		return 9;
+	} else {
+		int fd = fileno(f);
+		if (dup2(fd, 1) < 0) {
+			ldms_log(LDMSD_LERROR, "Cannot redirect log to %s\n",
+							logfile);
+			*errstr =  "error redirecting stdout";
+			return 10;
+		}
+		if (dup2(fd, 2) < 0) {
+			ldms_log(LDMSD_LERROR, "Cannot redirect log to %s\n",
+							logfile);
+			*errstr =  "error redirecting stderr";
+			return 11;
+		}
+		stdout = f;
+		stderr = f;
+	}
+	log_fp = f;
+	return 0;
+}
+
+int ldms_logrotate() {
+	int rc;
+	if (!logfile) {
+		ldms_log(LDMSD_LERROR, "Received a logrotate command but "
+			"the log messages are printed to the standard out.\n");
+		return EINVAL;
+	}
+	if (log_fp == LDMSD_LOG_SYSLOG) {
+		/* nothing to do */
+		return 0;
+	}
+	struct timeval tv;
+	char ofile_name[PATH_MAX];
+	gettimeofday(&tv, NULL);
+	sprintf(ofile_name, "%s-%ld", logfile, tv.tv_sec);
+
+	pthread_mutex_lock(&log_lock);
+	if (!log_fp) {
+		pthread_mutex_unlock(&log_lock);
+		return EINVAL;
+	}
+	fflush(log_fp);
+	fclose(log_fp);
+	rename(logfile, ofile_name);
+	log_fp = fopen_perm(logfile, "a", LDMS_DEFAULT_FILE_PERM);
+	if (!log_fp) {
+		printf("%-10s: Failed to rotate the log file. Cannot open a new "
+			"log file\n", "ERROR");
+		fflush(stdout);
+		rc = errno;
+		goto err;
+	}
+	int fd = fileno(log_fp);
+	if (dup2(fd, 1) < 0) {
+		rc = errno;
+		goto err;
+	}
+	if (dup2(fd, 2) < 0) {
+		rc = errno;
+		goto err;
+	}
+	stdout = stderr = log_fp;
+	pthread_mutex_unlock(&log_lock);
+	return 0;
+err:
+	pthread_mutex_unlock(&log_lock);
+	return rc;
+}
+
+int ldms_log_parse_verbosity(const char *optarg) {
+	if (0 == strcmp(optarg, "QUIET")) {
+		quiet = 1;
+		log_level_thr = LDMSD_LLASTLEVEL;
+	} else {
+		log_level_thr = ldms_str_to_loglevel(optarg);
+	}
+	if (log_level_thr < 0) {
+		return -1;
+	}
+	return 0;
+}
+
+void ldms_log_close() {
+	if (logfile) {
+		free(logfile);
+		logfile = NULL;
+	}
+}
+
+static __attribute__((constructor))
+void __init()
+{
+	log_fp = stdout;
+}
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -172,7 +172,8 @@ struct ldmsd_plugin_cfg *new_plugin(char *plugin_name,
 			 "function.", plugin_name);
 		goto err;
 	}
-	lpi = pget(ldmsd_msg_logger);
+	// lpi = pget(ldmsd_msg_logger);
+	lpi = pget(ldmsd_log);
 	if (!lpi) {
 		snprintf(errstr, errlen, "The plugin '%s' could not be loaded.",
 								plugin_name);
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -89,7 +89,7 @@ LIST_HEAD(host_list_s, hostspec) host_list;
 LIST_HEAD(ldmsd_store_policy_list, ldmsd_store_policy) sp_list;
 pthread_mutex_t sp_list_lock = PTHREAD_MUTEX_INITIALIZER;
 
-#define LDMSD_PLUGIN_LIBPATH_MAX	1024
+#define LDMSD_PLUGIN_LIBPATH_MAX	LDMS_PLUGIN_LIBPATH_MAX
 struct plugin_list plugin_list;
 
 void ldmsd_cfg_unix_cleanup(ldmsd_cfg_xprt_t xprt)
--- a/packaging/make-all-top.sh
+++ b/packaging/make-all-top.sh
@@ -43,7 +43,7 @@ if test -f ldms/src/sampler/meminfo/meminfo.c; then
 	cd $build_subdir
 	expected_ovislib_prefix=$prefix
 	expected_sos_prefix=/badsos
-	allconfig="--prefix=$prefix --disable-rdma --enable-ssl --with-libevent=$expected_event2_prefix --disable-sos --disable-perfevent --disable-rpath --enable-zap --disable-zaptest --enable-authentication --disable-sysclassib --disable-ibnet"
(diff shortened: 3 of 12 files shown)


baallan:
@morrone Please review this intermediate step (at least treating logging as a built-in part of libldms) toward getting consistent handling of error messages, particularly missing in auth and transport layers as you have noted. I expect switching to an entirely new logging api is something we should do at the point of the "new apis for plugins" later.

I note that the auth plugins do not have a help (usage) string in their api as the other plugins do. They do need it, imo. After that is sorted in some other commit, I will need to update the ldmsd plugin usage dump function to include auth plugins.

baallan:
Note: all the now ifdef'd code in ldmsd.c I will be happy to clean up once folk have had a chance to review this.
The refactored code has been built, tested with valgrind, and check with nm to verify no lingering ldmsd_log symbols.

morrone:
I am not entirely clear on the approach here. I, as a plugin author, want to be able to insert messages into the daemon's normal logging stream, not to create my own independent log files. And probably that is what we want all plugins to do. So the only thing that we need to export to the plugins is an "ldms_plugin_log()" style function.

I don't think we need nor want to have each plugin opening and managing their own individual log files, together with the configuration that implies.

In other words, perhaps most of this really should stay in ldmsd rather than moving to ldms. If you intend that the code reuse here is intended for, say, SOS and other daemons, rather than for plugins, then maybe the API needs a refresh rather than a direct move to ldms.

If we are creating a new API, then I don't think we would necessarily want to perpetuate the _logrotate() function. Modern daemons should really not trouble themselves with rotating logs. The "logrotate" daemon has been around for a long time, and systemd/journald is growing common in recent times.

To be most compatible with the "logrotate" program (although logrotate does offer multiple ways of interacting with daemons), the logging API should just offer a "_reload()" function. It would basically just do:

```
  /* take logging mutex */
  fclose()
  fopen()
  /* release logging mutex */
```

Then the ldmsd (or other daemons) would call that _reload() function on, say, receipt of signal USR1/USR2/HUP. (There is a case to be made that USR1/USR2 is better, because HUP implies a full reconfiguration. USR1 or USR2 could just be the log reload.)

For even more modern Linux systems, ldmsd would just write its log to stdout, and let systemd/journald handle all of the log file arrangements.

If this is about creating logging function for plugins:

There was some desire (not by me, by the way), that the logging levels of individual plugins be adjusted separately. Certainly parties thought that it was verboten to do log(DEBUG, ) because someone might be trying to see the log(DEBUG, ) from other plugins or the main daemon, and that would be too much noise (I didn't see a problem with plugins just using log(DEBUG,)). So to satisfy that requirement, I think the plugin logging function might be something like this:

```ldms_plugin_log(instance, severity, msg);```

This would be dependent on the multi-instance code that is coming in the future. But this would allow the parent daemon (e.g. ldmsd) to decide on a per-instance basis whether or not the logging level is high enough to emit the message.

**Comment by baallan** [comment]
@morrone As noted at ldms_log.h:191, plugin writers just use ldms_log() and move on with their lives. This request lets us get rid of passing log pointers to plugins near dlopen. No need to mess with opening or closing logs. The log model here is a global, thread-safe stream of messages that can be composed anywhere in the code.

The refactoring of ldms_logrotate, while a worthy project now that RHEL6/sysv init can be dropped from support and systemd is ubiquitous, is orthogonal to this simple refactoring to get the basic logging operations visible everywhere in ldms without the excessive pointer passing currently in place. Actually, is systemd ubiquitous? Seems to be missing in many container models.

There is no intent here that the ldms_log functionality is a standalone library for reuse with, e.g., SOS. It's lumped in with libldms.so in the makefile. Perhaps it should be lumped in with lib/ovis_util/util.h instead. That would make it available to zap plugins.

Any refactoring to extend log filtering control to a per-plugin basis is beyond the scope of this request. This request is about just getting existing logging to be uniform everywhere in the auth, plugin, and zap foundations of ldms so we can then remove the convoluted or just plain missing) log pointer handling and get useful messages emitted where they should.

**Comment by morrone** [comment]
> @morrone As noted at ldms_log.h:191, plugin writers just use ldms_log() and move on with their lives. This request lets us get rid of passing log pointers to plugins near dlopen. No need to mess with opening or closing logs. The log model here is a global, thread-safe stream of messages that can be composed anywhere in the code.

Yes, but I don't think that makes sense to export a whole raft of functions to plugins (and the entire world) that should never use those functions. We should make them internal to ldmsd.

> The refactoring of ldms_logrotate, while a worthy project now that RHEL6/sysv init can be dropped from support and systemd is ubiquitous, is orthogonal to this simple refactoring to get the basic logging operations visible everywhere in ldms without the excessive pointer passing currently in place.

> Actually, is systemd ubiquitous? Seems to be missing in many container models.

It is common in base Linux distros, but I can't speak to its adoption in containers. Keeping logs in containers probably isn't a common thing either. I think the container way is to send all log traffic somewhere off node. I don't know what mechanisms are used to do that though. The logrotate daemon is always an option if one wants logs kept in a container without journald. But yes, this can all be future work, it doesn't have to happen in this PR.
 
> There is no intent here that the ldms_log functionality is a standalone library for reuse with, e.g., SOS. It's lumped in with libldms.so in the makefile. Perhaps it should be lumped in with lib/ovis_util/util.h instead. That would make it available to zap plugins.

This doesn't make sense to me either. Again, the only place this code is supposed to be used is inside of ldmsd. So it should only be inside ldmsd. The only function declaration that should appear in a header that gets installed and is used by various plugins is the "ldms_log" function (and associated log levels and such). Everything else should be in ldmsd-internals.

Plugins should only get the implementation of the function when linked into ldmsd at run time. The function implementations should not be exported in libraries that anyone can use if that is not what we intend.

**Comment by baallan** [comment]
@morrone I take your point that plugin-writers don't need to see the header of the open/close functions. However, we do need to have access to those functions for drivers other than ldmsd.c for testing purposes in a few nontrivial cases.

Do I just split the header into ldms_log.h and ldms_log_int.h or some such that only drivers need to include if they need to tinker?

**Comment by morrone** [comment]
Sure, that could be reasonable. If the functions declared in ldms_log_int.h need to be used by binaries other than ldmsd (e.g. tests, as you mention), then ldms_log_int.h should be in noinst_HEADERS, and the library that implements ldms_log_int.c (or its like) should be a noinst convenience library, not something that is installed.

But it is not at clear to me why tests would need to use the ldmsd's logging library to implement their own log files. I would think that writing messages to stdout/stderr error would be sufficient for tests.

**Comment by baallan** [comment]
Think about unit and coverage testing of plugins and libraries separate from the ldmsd.c driver. Filling those with printf is a problem.

**Comment by tom95858** [comment]
If we're going to change the API, then I would very much prefer if we moved away from the log-level model to something more descriptive, e.g.

```
log_msg(service-id, msg-type-id, ...)
```

Each service gets a number of bits that comprises a mask. The 'severity' level is a piece of that mask. For the sake of argument if you have 64 bits, and 4 levels, each level is 16bits and each service can have 16 msg-type-ids in each severity level.

This would provide very fine grained control over which messages the admin would like to see from the daemon.

**Comment by morrone** [comment]
@tom95858 I'm not fully following your example. I'm not clear on why a service would have a baked in severity level. You talk about "a mask", but there are two parameters in your prototype, so I'm a little fuzzy on where you are suggesting things go.

But in any event, it is not clear to me that combining who is doing the logging ("service") with the level of logging in one field would be a useful thing to do.

And I think we should be careful about inventing our own severity levels, because what admins are going to need to map ldms levels into the levels they are used to so they can fit ldms into the frameworks that they use with everything else on the system. This mapping almost certainly needs to be built into ldms, because no admin wants to write level conversion scripts for every product that they work with.

Could you clarify what might qualify as a "service" under this model?

This is an area where we probably shouldn't reinvent the wheel unless we have _really_ good reasons to do so. Maybe all the messages in your proposal pop out of the daemon in the end labelled with severity level that admins are already familiar with. I don't know. That isn't clear to me yet.

**Comment by morrone** [comment]
> Think about unit and coverage testing of plugins and libraries separate from the ldmsd.c driver. Filling those with printf is a problem.

I can't imagine why it would be a problem to have printf in unit or coverage tests.

**Comment by tom95858** [comment]
> @tom95858 I'm not fully following your example. I'm not clear on why a service would have a baked in severity level. You talk about "a mask", but there are two parameters in your prototype, so I'm a little fuzzy on where you are suggesting things go.

A service doesn't have a baked in severity level, the msg-type-id do. It's the same thing we do today, but instead of a 


> But in any event, it is not clear to me that combining who is doing the logging ("service") with the level of logging in one field would be a useful thing to do.
> 

It's a proposed implementation. The purpose of using a bit mask is to make the decision of whether or not to log fast.

> And I think we should be careful about inventing our own severity levels, because what admins are going to need to map ldms levels into the levels they are used to so they can fit ldms into the frameworks that they use with everything else on the system. This mapping almost certainly needs to be built into ldms, because no admin wants to write level conversion scripts for every product that they work with.
> 

We already have our own severity levels, CRITICAL through DEBUG. We could certainly normalize those if you feel that admins expect something different.

> Could you clarify what might qualify as a "service" under this model?
>

Auth
Transport
Updater
Producer
Store
...

> This is an area where we probably shouldn't reinvent the wheel unless we have _really_ good reasons to do so. Maybe all the messages in your proposal pop out of the daemon in the end labelled with severity level that admins are already familiar with. I don't know. That isn't clear to me yet.

The msg-id, etc... would not be in the log message. That's just a way to 'filter' what log messages are presented to the admin by allowing them to target what they are after when things get weird.

We would map the current 'levels' into a mask that gets us the equivalent course grained logging. 

The only value of this is to reduce the message volume when you're trying to determine what's going on and to make debugging easier by allowing you to get very fine grained messaging (currently DEBUG) for one 'service' (Transport, etc..) and ERROR for everything else.

If we're going to tear up the carpet, I see an opportunity to allow for more targeted logging.

I'm envisioning something like this:

ldmsd_controller --host <somewhere> --port ...
> log_options service=.* level=ERROR

Messages  of severity ERROR (or worse) for all services are logged

> log_option service=Transport level=DEBUG

All services except Transport still have ERROR level, but Transport messages are logged at the DEBUG level

> log_option service=uGNI msg-id=GNI_ERR_RESOURCE level=OFF

The uGNI plugin would log only resource errors, all other messages from that plugin are turned off

Anyway, that's the idea.

The bitmask talk was just about making the capability above very fast.

**Comment by morrone** [comment]
OK, I see what you are going for. We probably need to hash out some details.

Are you suggesting that there can only be one log_option rule mask, or that they can somehow stack?

I guess my only concern is that this all seems like overkill, when it is pretty darn easy to just grep (or whatever your favorite tool is) for what you care about in the log. A course syslog-style log level seems more than adequate to me. (And I think plugins should be allowed to log at those normal coarse levels...the "no DEBUG from plugins" never seemed right to me.)

**Comment by baallan** [comment]
@tom95858 With _this_ request I am not proposing a new API, but only moving the one we have already so it can be accessed uniformly everywhere in our code, eliminating the passing of log pointers. I can split the header as @morrone suggested to keep the open/close functions invisible to plugins.

@morrone I would much prefer the logging api we already have over the undisciplined use of printf. I could maybe be persuaded that we should _in a separate commit_ eliminate the support of syslog and alternate file locations for the log, but I'm not sure I know the unix shell for redirecting stdout of a daemon process to a log file for managing test results-- that seems to be a systemd-specific trick. Do you know that foo? Or the foo to have a unit test run as non-root under systemd?

Re the filter-controlled logging, this seems to be mainly driven in practice by the desire to log way too much debug info all the time and expecting the user to turn it off. This (formatting and then suppressing) makes performance impacts on sampler->sample() and store->store() and probably some transport calls that we must not have.; otherwise I wouldn't mind much since we're already pretty verbose in logging. 

I could see an extension on the existing logging to always automatically provide a context string: 
#define ldms_log(level, fmt, ...) ldms_log_(level, __FILE__ ":"__FUNCTION__":" fmt, __VA_ARGS__) 
or some such. This would get us out of working in undisciplined location information such as the 'SAMP' macro in many samplers. It (as written) is a gcc-ism.

**Comment by baallan** [comment]
To be clear: I'm trying to get v4.3 out the door with the logging style we have already extended to all the places that have been identified as needing it to make diagnosing TOSS ldmsd deployments feasible (auth, transport, plugins of either, etc). I really don't want to be introducing __more__ passing of log pointers such as we currently see in sampler plugins.  

The api redesign discussion is well and good, but not the scope here.  It feels very much like better being the enemy of good on the path an admin-friendlier 434 release. 

Is there any other feedback on the content of this request?

**Comment by tom95858** [comment]
> To be clear: I'm trying to get v4.3 out the door with the logging style we have already extended to all the places that have been identified as needing it to make diagnosing TOSS ldmsd deployments feasible (auth, transport, plugins of either, etc). I really don't want to be introducing **more** passing of log pointers such as we currently see in sampler plugins.
> 
> The api redesign discussion is well and good, but not the scope here. It feels very much like better being the enemy of good on the path to an admin-friendlier 434 release.
> 
> Is there any other feedback on the content of this request?


Hi @baallan, It's not clear to me how this patch achieves any of these goals. You want to extend logging to "all the places that have been identified as needing it..." and yet all the places you mention already have access to a logging service.

The auth plugins notably do not currently log messages but they could. The transport provided in the auth_xprt_bind function has a pointer to an ldms_log_fn_t function. 

Perhaps for the 4.3.4 time frame you could use the existing code and add the messages you believe are required to the auth plugins.

**Comment by morrone** [comment]
> I would much prefer the logging api we already have over the undisciplined use of printf.

First, note that the following is a diversion about future work. Having the _int.[ch] files will satisfy me for this PR.

printf would not be any more or less disciplined than using the logging API. The logging API implies no discipline when it comes to test messaging. The tests can all use any output file they wish with the current logging API, and the logging levels are probably of very little use in testing context. The logging levels may even conflict with the output expected by automake's 

Discipline comes from following a testing protocol, whether it is something invented from scratch, or something already in wide use such as TAP: http://testanything.org/tap-specification.html.  Harnesses to aid in following the testing protocol can be downloaded or written. For instance, for TAP there is "sharness" for shell tests, libtap for C, etc.

TAP is even mentioned in the automake manual's section about support for test suites.

**Comment by baallan** [comment]
@tom95858 the ldms_log_fn_t is the wrong type (it does not take a level argument, which is only visible via ldmsd.h currently). This is not sufficient for solving the current logging issues. zap_log_fn_t has the same problem in the transport arena. 

By far the easiest fix to get the same logging api available everywhere would be to put the proposed ldms_log in a library peer of zap and just use it whereever needed. Existing code passing around log pointers would continue to work unchanged (clean up to follow in future change), and code following the level-tagged api can be added quickly.  

So my putting the new library extracted from ldmsd.c in ldms/src/core/libldms needs to be revised to putting it in lib/src/[ovis_util | zap ].

Alternatively, we could just move the log level enum over to lib/src/ovis_util/ovis_log_levels.h and update the auth and zap log pointer types to take a log level. This would force an immediate rewrite of all the existing log calls to add the appropriate level (which needs doing eventually anyway). This is very minimal risk, as the changes all cause compiler errors if missed.

**Comment by baallan** [comment]
@morrone We also noticed the tap bits in the automake docs, but lacking a credible, complete, and apparently maintainable example, we haven't yet managed to apply it. If you happen to know of a real tap deployment that is compatible with automake 1.13.4 (toss3/rhel7), we'd love to see it.

File: ldms/src/core/ldms_log.c [hunk 1]
```diff
@@ -0,0 +1,339 @@
...
+}
+
+const char *ldms_loglevel_to_str(enum ldms_loglevel level)
+{
+	if ((level >= LDMSD_LDEBUG) && (level < LDMSD_LLASTLEVEL))
+		return ldms_loglevel_names[level];
+	return "LDMSD_LNONE";
+}
+
+/** return a file pointer or a special syslog pointer */
```
tom95858, line 218: Nit: this actually sets a global variable and returns an errno.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -89,7 +89,7 @@ LIST_HEAD(host_list_s, hostspec) host_list;
 LIST_HEAD(ldmsd_store_policy_list, ldmsd_store_policy) sp_list;
 pthread_mutex_t sp_list_lock = PTHREAD_MUTEX_INITIALIZER;
 
-#define LDMSD_PLUGIN_LIBPATH_MAX	1024
+#define LDMSD_PLUGIN_LIBPATH_MAX	LDMS_PLUGIN_LIBPATH_MAX
```
tom95858, line 92: Maybe just PATH_MAX from limits.h?

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -172,7 +172,8 @@ struct ldmsd_plugin_cfg *new_plugin(char *plugin_name,
 			 "function.", plugin_name);
 		goto err;
 	}
-	lpi = pget(ldmsd_msg_logger);
+	// lpi = pget(ldmsd_msg_logger);
```
tom95858, line 175: nit: should be removed?

packaging/make-all-top.sh
```diff
@@ -43,7 +43,7 @@ if test -f ldms/src/sampler/meminfo/meminfo.c; then
 	cd $build_subdir
 	expected_ovislib_prefix=$prefix
 	expected_sos_prefix=/badsos
-	allconfig="--prefix=$prefix --disable-rdma --enable-ssl --with-libevent=$expected_event2_prefix --disable-sos --disable-perfevent --disable-rpath --enable-zap --disable-zaptest --enable-authentication --disable-sysclassib --disable-ibnet"
+	allconfig="--prefix=$prefix --disable-rdma --enable-ssl --with-libevent=$expected_event2_prefix --disable-sos --disable-perfevent --disable-rpath --enable-zap --disable-zaptest --enable-authentication --disable-sysclassib --disable-ibnet --disable-python"
```
tom95858, line 46: I don't think disable-python belongs in this patch.

tom95858:
@baallan, I'm guessing you will be following up with log related changes to auth plugins, etc...?

packaging/make-all-top.sh
baallan: yes. will need to back out changes to make-all-top. Esp. if python bits have moved to default python disabled as needed.

ldms/src/ldmsd/ldmsd_config.c
baallan: no, path_max is unportably small sometimes (256 has been seen in the wild).
In any case, this is just an alias of existing code, not a change in behavior.

ldms/src/core/ldms_log.c
baallan: I tried to keep this close to the original code as possible without being broken.  Will take another look.

ldms/src/ldmsd/ldmsd_config.c
baallan: yes.

baallan:
@tom95858 The second commit of this pr adds some auth changes, and there would also be some followup logging improvements to zap and its plugins. The first commit needs to be rearranged to put the log header into lib someplace zap can see it.

baallan:
@tom95858 do we go with this after it's cleaned up, or would you prefer the alternative of adjusting the location of the ldmsd_log_level enum to someplace in lib/ and the existing log_fn_t typedefs equivalent to printf?

packaging/make-all-top.sh
morrone: > @morrone We also noticed the tap bits in the automake docs, but lacking a credible, complete, and apparently maintainable example, we haven't yet managed to apply it. If you happen to know of a real tap deployment that is compatible with automake 1.13.4 (toss3/rhel7), we'd love to see it.

You are asserting some strong generalities that have not been true in my experience. If you want to discuss particular problems you are having in another issue, we could do that. I suspect they can all be worked through.

Flux makes extensive use of TAP. https://github.com/flux-framework/flux-core

morrone:
> @morrone We also noticed the tap bits in the automake docs, but lacking a credible, complete, and apparently maintainable example, we haven't yet managed to apply it. If you happen to know of a real tap deployment that is compatible with automake 1.13.4 (toss3/rhel7), we'd love to see it.

You are asserting some strong generalities that have not been true in my experience. If you want to discuss particular problems you are having in another issue, we could do that. I suspect they can all be worked through.

Flux makes extensive use of TAP. https://github.com/flux-framework/flux-core

nichamon:
@baallan Could you please state the objectives of the commit that creates the log library? The commit message states that the patch is to get rid of passing the log handle. It is not clear to me why passing the log handle is a problem though. After following the thread, it sounds to me that another objective is that all OVIS libraries could have the same logging capabilities, especially the log level feature.

What are the requirements this log library satisfies? 

Lastly could you please explain why this patch needs to be in the GA? We want to have the GA release ASAP. This patch and the needed changes in other files will postpone the release.

Thanks

baallan:
@nichamon the log library (as to be revised following the discussion) provides access to a uniform api like that currently available in src/ldmsd.h but not available currently in lib/src/zap or ldms/src/auth. Adding and/or adjusting more function pointer passing interfaces to get uniform logging everywhere is more work than the currently proposed back-compatible (at compile time) approach.

This will allow numerous missing log messages to be added for a level of administrative friendliness (explaining auth and zap/xprt failures) we need for GA.  If there are critical bugs that we need to get a labeled release in place to address, we can just add a beta tag instead of rushing GA unnecessarily. 

If this approach is acceptable, I can get the patch rearranged, tested, and extended with most of the needed log messages this week. Independent of that, I still have scale testing in progress for rdma and fabric transports, which are part of the GA requirements-- likely that will take longer to complete than these user-oriented logging changes.

tom95858:
> @nichamon the log library (as to be revised following the discussion) provides access to a uniform api like that currently available in src/ldmsd.h but not available currently in lib/src/zap or ldms/src/auth. Adding and/or adjusting more function pointer passing interfaces to get uniform logging everywhere is more work than the currently proposed back-compatible (at compile time) approach.

I don't think that is true, however, as you pointed out, the signature for the log function does include the log level.

> 
> This will allow numerous missing log messages to be added for a level of administrative friendliness (explaining auth and zap/xprt failures) we need for GA.

Let's get these messages specified so we can reasonably assess whether or not this large change is important enough to hold up G/A.

> If there are critical bugs that we need to get a labeled release in place to address, we can just add a beta tag instead of rushing GA unnecessarily.
> 

It takes quite a long time to run the scalability testing. Whenever we add new code, it requires us to re-run these tests. Presently we are continually adding new code that requires restarting large-scale testing. If we continue along this path, we will never G/A.

> If this approach is acceptable, I can get the patch rearranged, tested, and extended with most of the needed log messages this week. Independent of that, I still have scale testing in progress for rdma and fabric transports, which are part of the GA requirements-- likely that will take longer to complete than these user-oriented logging changes.

baallan:
@tom95858 I recall the users meeting discussion ended with agreement that fixing the auth plugin location (and attendant need to fix logging) be a post 434-ga feature. While I would much prefer plugin location and logging to be fixed in 4.3.4 ga, it doesn't have to hold up things.  We just won't use 4.3.4-ga for a toss release.

It (plugin location, logging) should be one of the first things we fix in 4.3.5; we also have a bunch of sampler extensions to fix "wide set" issues (lustre, procnetdev, sysclassib) that can be fixed in short order as well to form an admin-friendly 4.3.5 without changes to transport that would need extensive testing.

baallan:
@nichamon, @tom95858  so an approximate list of approximate requirements for fixing logging would be as follows. A multistep update process may be desirable. A more detailed requirements list would take less time to just implement in code than to compose here:
- Somehow or other (this pr is but a single way) expose the ldmsd log level enum to the entire lib&ldms code base, most conveniently in some widely used existing header so that no new #include lines are needed. Renaming the enum (ovis_, ldms_, etc log_level is a matter of taste rather than a requirement.
- Update the auth, xprt, zap code and any other existing passed log function pointers to accept a level argument, or alternatively eliminate all passing of log pointers and use a common library everywhere.
- Add messages with appropriate log levels for all failures in the auth and zap plugins and zap.c/ldms_auth.c. Particularly notice of auth key/user failures to the log of the rejecting daemon is needed, but any other useful messages about undefined addresses or failed/missing hardware or missing plugin libraries.
- Where ever feasible, add fast local logic to suppress redundant logging of the same messages coming from iterative failed attempts to satisfy a single administrative request. E.g. issue the 'connection failed' once per connection target at the beginning of a retry logic and then a 'retry succeeded' when the connection is finally made. The success event clears whatever flag was suppressing redundant messages so that we get a failed message once again if the connection is closed/failed and (unsuccessfully) reopened.

Re some of the code where we have a bucket of debugging messages, but they are all ifdef'd out under normal build conditions. This is fine (for obvious performance reasons) but some of the messages need to be reviewed for promotion to admin-visible diagnostics.

Re the whole per-message filtering discussion, if we get a common log api in place throughout the code base then it becomes much easier to do sed-like things to the code base to introduce log per-message/per-plugin filtering. As it is now, we have too many printf-like pointers with funny aliases to automate such evolution without heavy duty code refactoring tools.

morrone:
> @tom95858 I recall the users meeting discussion ended with agreement that fixing the auth plugin location (and attendant need to fix logging) be a post 434-ga feature. While I would much prefer plugin location and logging to be fixed in 4.3.4 ga, it doesn't have to hold up things. We just won't use 4.3.4-ga for a toss release.

While I completely agree that I would like to see the location of the plugins and other install paths improved, this is a problem in the current toss packages. So it isn't like we would be taking a step backwards. So this wouldn't necessarily prevent its use in TOSS. I don't have an issue with 4.3.4 having the current paths. 4.3.4 is already behind schedule, so I would like to see any PRs that would delay the schedule be put on hold.

baallan:
Re deferring PRs: 449 and 450 are bug fixes that should go through. 443, 446, 448 can all wait.

@morrone the current toss version of ldms is v3 which has no auth plugin problems. They were introduced in v4 and do represent a step backward. However, the reluctance to push 434 officially into toss is due to the difficulty in diagnosing transport and auth configuration issues, not where the libraries get installed. Our first local toss update cycle which could possibly consume a toss-provided ovis release is Jan/Feb upcoming, so I don't share the urgency others are feeling about 4.3.4 becoming 'ga'. Obviously we want nice things for the cray systems that sometimes update faster than toss, but I still hear lots of hammering and sawing in the cray related bits.

morrone:
> Re deferring PRs: 449 and 450 are bug fixes that should go through. 443, 446, 448 can all wait.
> 
> @morrone the current toss version of ldms is v3 which has no auth plugin problems.

The latest TOSS 3 (and pre-4) release has ldms v4.3.3_9toss. Our tag is actually quite a bit beyond 4.3.3 on the OVIS-4 branch, but that was the most recent official tag at the time. We are using that on most of our systems now where we have LDMS installed. I'll be keen to get 4.3.4 when it is out because of at least one important fix. If there are auth problems in 4.3.X, I'd like to see issues opened on them so I can follow along. I haven't encountered any show stoppers with auth munge yet (at least not with the head of OVIS-4).

baallan:
@tom95858 @morrone based on Tom's comments about filtering logging, I would like to propose the following header/code (actually it runs as a proof of signature validity if you compile it -Wall).

It retains Tom's O(1) cost to filter, with improvements of this proposal:
The number (and corresponding name) of msg-type-id (in my case called 'facility') is dynamically determined so that
- plugins unknown at daemon build time can define their own names for the configuring of their logs.
- not just code, but object instances, can have their own names (imagine the debugging power). No bit-fiddling.
- I included the proof of back-compatibility macros with the existing ldmsd log, so we don't have to fix all code in one giant commit when the library is added.

Please take a look before Monday's LDMS users' meeting.

```
#include <stdarg.h>
#include <stdio.h>
#include <stdint.h>
#include <inttypes.h>

enum ldms_log_level {
	LDMS_LNONE = -1,
	/* usual syslog levels */
	LDMS_LDEBUG,
	LDMS_LINFO,
	LDMS_LWARNING,
	LDMS_LERROR,
	LDMS_LCRITICAL,
	/* developer demands logging under all circumstances; the nerve of that developer... */
	LDMS_LALL,
	/* deprecated names for back compatibility */
	LDMSD_LNONE = LDMS_LNONE,
	LDMSD_LDEBUG = LDMS_LDEBUG,
	LDMSD_LINFO = LDMS_LINFO,
	LDMSD_LWARNING = LDMS_LWARNING,
	LDMSD_LERROR = LDMS_LERROR,
	LDMSD_LCRITICAL = LDMS_LCRITICAL,
	LDMSD_LALL = LDMS_LALL,
};

/* back compatibility macros */
/* map enum name */
#define ldmsd_log_level ldms_log_level
/* map ldmsd_log to ldms_log with level 0, that will be taken by the driving main.
 * Migration to use of log facilities can be gradual and separate from adding the library
 * header in ldms.h and linking the library.
 */
#define ldmsd_log_helper(level, fmt, ...) ldms_log(0, level, fmt "%s", __VA_ARGS__)
#define ldmsd_log(...) ldmsd_log_helper(__VA_ARGS__, "")

/* the utility */

/* get threshold by string name. Cost per order of lookup in underlying hash.  */
enum ldms_log_level ldms_log_threshold_get_byname(const char *facility_name);

/* get threshold by facility number. Cost  O(1). 
 * Underlying implementation probably a growable 2d array
 * indexed[facility/BARSIZE][facility % BARSIZE] for O(1) lookup.
 */
enum ldms_log_level ldms_log_threshold_get(uint64_t facility_number);

/* Get the facility map formatted for display; lines of name,number,threshold.
 * Alternately, expose an iterator or some such.
 * Caller frees result.
 */
char * ldms_log_facilities_to_string( );


/* Set the threshold for the facility by name.
 * If this is never called, the facility will have the default threshold of 
 * CRITICAL.
 */
int ldms_log_threshold_set(const char *facility_name, enum ldms_log_level lvl);

void __ldms_log(enum ldms_log_level level, const char *fmt, va_list ap);
int ldms_log(uint64_t facility, enum ldms_log_level lvl, const char *fmt, ...)
{
	// func signature test, not full thread-safe library. lock for threshold and output channel are one.
	if (lvl < ldms_log_threshold_get(facility))
		return -1;
        va_list ap;
        va_start(ap, fmt);
        __ldms_log(lvl, fmt, ap);
        va_end(ap);
	return 0;
}

/** \brief Register the name to get an integer tag.
 *
 * The name is used in parsing configuration information that sets the
 * desired log threshold for the facility.
 *
 * Good facility names are . structured to allow pattern matching of related
 * facilities. E.g. 
 * sampler.meminfo, zap.rdma, xprt.*, updtr.*, prdcr.*, auth.munge, store.*.
 * Facility names may even be scoped to configuration object instance names,
 * hence facilities are uint64_t.
 *
 * The facility number is used to support O(1) lookup of log_level threshold
 * when filtering log output.
 * Expected use model is that the caller caches the result and passes it to
 * all logging calls, then calls ldms_log_facility_get when done operating.
 *
 * It is not possible to know all facility names at build time of the logging
 * library, due to our plugin architecture.
 */
uint64_t ldms_log_facility_get(const char *facility_name);

/** \brief Unregister the facility and name given.
 * Name and facility must match so that others cannot unregister
 * a facility accidentally, e.g due to a memory corruption of the cached value.
 */
void ldms_log_facility_put(const char *facility_name, uint64_t facility);

/* functions to parse level specification are an exercise left to the reader.
 * Following a hierarchical string naming convention for various code areas
 * allows convenient specifications.
 */
int ldms_log_threshold_set(const char *facility_name, enum ldms_log_level lvl);

/* reclaim log-associated memory and ignore all further log callse */
void ldms_log_facility_close();

/* demo */
int main(int argc, char ** argv)
{
	uint64_t logfac = ldms_log_facility_get("ldmsd"); /* first call also handles init of internals */

	/* legacy code still works */
	ldmsd_log(LDMSD_LINFO, "old signature work via macro magic\n");

	/* new code */
	ldms_log(logfac, LDMS_LINFO, "new signature work with no args\n");
	ldms_log(logfac, LDMS_LINFO, "new signature work with %d args\n", 1);

	ldms_log_facility_put("ldmsd", logfac);

	ldms_log_facility_close();
	return 0;
}

/* dummy implementation for compiler test of signatures. */
void ldms_log_facility_put(const char *name, uint64_t fac)
{
}

uint64_t ldms_log_facility_get(const char *name)
{
	return 0;
}

void ldms_log_facility_close() {}
enum ldms_log_level ldms_log_threshold_get(uint64_t facility_number)
{
	return LDMS_LDEBUG;
}
void __ldms_log(enum ldms_log_level level, const char *fmt, va_list ap)
{
	vfprintf(stdout, fmt, ap);
}
```

baallan:
Following the 10/26/20 user group telecon I created https://github.com/ovis-hpc/ovis/wiki/Proposal-5.

I am updating the header/code proposed above following the meeting feedback.

PR #451, opened 2020-10-08
Fix listening endpoint leak and cleanup LDMSD's listening path

nichamon:


ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -1005,29 +1001,20 @@ int listen_on_ldms_xprt(ldmsd_listen_t listen)
...
-			  "failed, error: %s(%d). args='%s'. Please check transport "
-			  "configuration, authentication configuration, "
-			  "ZAP_LIBPATH (env var), and LD_LIBRARY_PATH.\n",
-			  listen->xprt,
-			  auth_name,
-			  ovis_errno_abbvr(errno),
-			  errno, args ? args : "(empty conf=)");
-		free(args);
-		cleanup(6, "error creating transport");
+		ldmsd_log(LDMSD_LCRITICAL, "The listening endpoint '%s:%d' "
```
tom95858, line 1005: I think this error message is misleading. What does 'before LDMSD listens on it mean?' Doesn't the change in the error message remove any clues to the administrator as to why the transport creation might have failed.

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_config.c
nichamon: With the change, `listen->x`  must be created before `listen_on_ldms_xprt` is called. The condition `if (!listen->x)` verifies that `listen->x` has really been created beforehand instead of verifying whether the call to `ldms_xprt_new_with_auth()` succeeds or not.

The change removes the creation of `listen->x` in the `listen_on_ldms_xprt()` function because without this change `listen->x` is already created in `main()` when it processes the -x option. Since the function name is `listen_on_ldms_xprt()`, I made the function only listens on the LDMS transport.

tom95858:
@nichamon, I understand what the code change is doing. The caller (ldmsd.c) has already created the transport when listen_on_ldms_xprt is called. Please just remove the duplicated logic and assert if listen->x is NULL. If listen->x is NULL, it is a logic error and there is nothing the user can do about that and therefore nothing they could do with this error message.

nichamon:
> @nichamon, I understand what the code change is doing. The caller (ldmsd.c) has already created the transport when listen_on_ldms_xprt is called. Please just remove the duplicated logic and assert if listen->x is NULL. If listen->x is NULL, it is a logic error and there is nothing the user can do about that and therefore nothing they could do with this error message.

@tom95858  The change has been made.

PR #453, opened 2020-10-12
V4 transport bugs

baallan:
This fixes two information remote disclosure nits and documents (instead of excessively initializing) valid suppression of epoll_ctl event argument initialization warnings.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -340,7 +340,7 @@ static void send_dir_update(struct ldms_xprt *x,
 	hdr_len = sizeof(struct ldms_reply_hdr)
 		+ sizeof(struct ldms_dir_reply);
 	buf_len = ldms_xprt_msg_max(x);
-	reply = malloc(buf_len);
+	reply = calloc(1, buf_len);
 
 	cnt = snprintf(reply->dir.json_data, buf_len - hdr_len,
 		       "{ \"directory\" : [ %s ]}", json);
--- a/lib/src/zap/rdma/zap_rdma.c
+++ b/lib/src/zap/rdma/zap_rdma.c
@@ -198,6 +198,7 @@ static int __enable_cq_events(struct z_rdma_ep *rep)
 	struct epoll_event cq_event;
 	cq_event.data.ptr = rep;
 	cq_event.events = EPOLLIN | EPOLLOUT;
+	/* safe to suppress "Syscall param epoll_ctl(event)" valgrind msg here. */
 
 	/* Release when deleting the cq_channel fd from the epoll */
 	__zap_get_ep(&rep->ep);
@@ -714,6 +715,7 @@ static zap_err_t z_rdma_connect(zap_ep_t ep,
 
 	cm_event.events = EPOLLIN | EPOLLOUT;
 	cm_event.data.ptr = rep;
+	/* safe to suppress "Syscall param epoll_ctl(event)" valgrind msg here. */
 	__zap_get_ep(&rep->ep); /* Release when disconnected */
 	rc = epoll_ctl(cm_fd, EPOLL_CTL_ADD, rep->cm_channel->fd, &cm_event);
 	if (rc)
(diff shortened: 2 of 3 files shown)


lib/src/zap/rdma/zap_rdma.c
```diff
@@ -198,6 +198,7 @@ static int __enable_cq_events(struct z_rdma_ep *rep)
 	struct epoll_event cq_event;
 	cq_event.data.ptr = rep;
 	cq_event.events = EPOLLIN | EPOLLOUT;
+	/* safe to suppress "Syscall param epoll_ctl(event)" valgrind msg here. */
```
tom95858, line 201: Let's not put these kinds of comments in the code please.

tom95858 (changes requested):

lib/src/zap/rdma/zap_rdma.c
tom95858: Hi @baallan, I think it would be preferable to add information to the README.md under a valgrind section. What do you think?

baallan:
I think readme format for suppressions is both unmaintainable and not directly usable as input to valgrind and other tools that consume stack-list format suppression information.
Comments adjacent to the C code lines evaluated is most maintainable but aesthetically unappealing and not directly usable as input to tools.
.supp files adjacent to the C code files are adequately maintainable and easily assembled into a global suppression list for testing.
Hand-maintained global suppression files are unmaintainable, due to both human factors and platform variations in what should be suppressed.

baallan:
suppression comments change has been removed.

tom95858:
@baallan, ok great. Then it would seem helpful to refer to the valgrind suppression files in the README.md Valgrind section with information on how to use them.

baallan:
@tom95858 I will look at adding suppression files and a readme pointer to them as a separate PR later.

ldms/src/core/ldms_xprt.c
```diff
@@ -340,7 +340,7 @@ static void send_dir_update(struct ldms_xprt *x,
 	hdr_len = sizeof(struct ldms_reply_hdr)
 		+ sizeof(struct ldms_dir_reply);
 	buf_len = ldms_xprt_msg_max(x);
-	reply = malloc(buf_len);
+	reply = calloc(1, buf_len);
```
tom95858, line 343: @ballan, This is a bzero of a megabyte on the socket and uGNI transports. The only bits that won't be initialized are the padding bytes in the header. Can you please just bzero the header (24B) and rerun valgrind to confirm this?

tom95858 (changes requested):

ldms/src/core/ldms_xprt.c
baallan: @tom95858 I can try that, but the error message says the issue is '96 bytes inside a block of 420', which appears to be the middle of the json space. I will rerun the tests; it may be that subsequent changes to json payload have caused the issue to disappear. I need to verify my test is still running the same sampler set as well, if something is payload dependent.

tom95858: Did you confirm that this change fixed the valgrind complaint? Regardless we can't bzero 1MB of data on every message, you need to figure out what the actual problem is.

tom95858: If the problem is in the JSON payload, then this change will not fix it because the json payload is copied into the 1MB buffer calloc bzero'd. Also, the error would be '96B inside a block of 1MB, not 420B. What transport was this problem exhibited on?

baallan: I am working on getting my hands on the data where the hole is found. I'm seeing this kind of message on both sock and rdma on stria. Started adding more logging yesterday, and will also see about getting valgrind+gdb to stop on detection and let me look around (though that's always messier with multithreaded code).

baallan:
I don't see why this is being closed, when the issue is still being shown as late as yesterday in the bundle of output I sent to Narate.

tom95858:
Hi @baallan, so far, the fix is to bzero a 1MB buffer for every message sent. Since this pull request has been idle for 10 days and now can't be merged without conflict, I closed it.

If there is still a problem here, please fix it; know why the fix works, does not break the build, does not cause a regression error, actually fixes a problem, does not negatively affect performance, and then we'll move on from there. Thank you!

PR #456, opened 2020-10-13
Add the zap_send_mapped() API to the zap library

nichamon:
Currently the zap_send() API receives application buffer and copies it
to another buffer that has been registered with the hardware using the
zap_map() interface. The zap_send() API is simple to use but is
inefficient when the same buffer is to be sent to many clients because
the same buffer is mapped and copied many times.

The zap_send_mapped() API has the following signature.

   int zap_send_mapped(zap_ep_t ep, zap_map_t map, void *buf,
                       size_t len, void *context)

The zap_send_mapped() API posts the RDMA_SEND operation directly from
the previously mapped buffer (map) without requiring registration or
copy by the transport. This significantly improves performance for
sending operation. Unlike the zap_send() API, the zap_send_mapped() API
delivers a completion event via the application callback function unless
the function return an error.

### DIFF
--- a/lib/src/zap/fabric/zap_fabric.c
+++ b/lib/src/zap/fabric/zap_fabric.c
@@ -1,8 +1,8 @@
 /* -*- c-basic-offset: 8 -*-
- * Copyright (c) 2013-2019 National Technology & Engineering Solutions
+ * Copyright (c) 2013-2020 National Technology & Engineering Solutions
  * of Sandia, LLC (NTESS). Under the terms of Contract DE-NA0003525 with
  * NTESS, the U.S. Government retains certain rights in this software.
- * Copyright (c) 2013-2019 Open Grid Computing, Inc. All rights reserved.
+ * Copyright (c) 2013-2020 Open Grid Computing, Inc. All rights reserved.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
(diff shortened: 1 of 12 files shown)


lib/src/zap/sock/zap_sock.c
```diff
@@ -113,7 +113,8 @@ static zap_err_t __sock_send_msg(struct z_sock_ep *sep, struct sock_msg_hdr *m,
 static zap_err_t __sock_send_msg_nolock(struct z_sock_ep *sep,
 					struct sock_msg_hdr *m,
 					size_t msg_size,
-					const char *data, size_t data_len);
+					const char *data, size_t data_len,
+					struct z_sock_io *io);
```
nichamon, line 117: @narategithub `struct z_sock_io *io` isn't used in the function. I want to remove it from the signature if I didn't miss anything.

nichamon:
The patch has been tested by the zap_test program for all zap transports. I created this as a draft to prevent the patch get merged before the release.

lib/src/zap/fabric/zap_fabric.c
```diff
@@ -2275,6 +2375,7 @@ static zap_err_t z_fi_read(zap_ep_t ep,
 
 static pthread_t cm_thread, cq_thread;
 
+__attribute__((unused))
```
nichamon, line 2378: @narategithub  Shall we remove this function?

lib/src/zap/fabric/zap_fabric.c
```diff
@@ -2321,7 +2422,7 @@ static int init_once()
 		z_fi_info_log_on = atoi(env);
 
 	init_complete = 1;
-	atexit(z_fi_cleanup);
+//	atexit(z_fi_cleanup);
```
nichamon, line 2425: @narategithub Shall we remove this line?

PR #459, opened 2020-10-16
Implement TAP protocol for test script

Snell1224:
Adding the Test Anything Protocol (TAP) to generate tests during "make check" for the ldms-run-static-test script. During make check (with test-scripts enabled), a test suite will be generated containing the xfails, fails and successes for each plugin listed in "ldms-static-test-list". 

The main goal is to provide the output status of nested test scripts so we can see which plugin test failed/xfailed/succeeded within the test script during the build process. 

Before adding this, the test suite would only provide a pass or fail of the entire "ldms-run-static-test" script so the tests running within this script were not being outputted and tracked in the testsuite.
 
Documentation for TAP Protocol - https://www.gnu.org/software/automake/manual/html_node/Use-TAP-with-the-Automake-test-harness.html#Use-TAP-with-the-Automake-test-harness


ldms/src/test/Makefile.am
```diff
@@ -41,10 +41,14 @@ test_metric_LDFLAGS = $(AM_LDFLAGS) -pthread -lm
...
+mypkglib_SCRIPTS = ldms-run-static-tests.test
 mypkglib_DATA = ldms-static-test-list.sh ldms-static-test-bypass
 
+# invoke tap-driver
+TEST_LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) \
+                  $(top_srcdir)/config/tap-driver.sh
+
 installcheck-local: 
-	ldms-run-static-tests.sh
+	ldms-run-static-tests.test
```
baallan, line 52: make check works.

A minor improvement on 'make check' with tap added would be to give the count of tests skipped (and possibly show them as SKIP in the output); but this would be additional work, not a reason to hold up this PR.

make installcheck on a fresh checkout fails with:
```
make[3]: Leaving directory `/home/baallan/ovis/repos/o4sara/LDMS_objdir/ldms/src/sampler'
Making installcheck in test
make[3]: Entering directory `/home/baallan/ovis/repos/o4sara/LDMS_objdir/ldms/src/test'
ldms-run-static-tests.test
make[3]: ldms-run-static-tests.test: Command not found
make[3]: *** [installcheck-local] Error 127
make[3]: Leaving directory `/home/baallan/ovis/repos/o4sara/LDMS_objdir/ldms/src/test'
make[2]: *** [installcheck-recursive] Error 1
make[2]: Leaving directory `/home/baallan/ovis/repos/o4sara/LDMS_objdir/ldms/src'
make[1]: *** [installcheck-recursive] Error 1
make[1]: Leaving directory `/home/baallan/ovis/repos/o4sara/LDMS_objdir/ldms'
make: *** [installcheck-recursive] Error 1
```

Is install-check-local a holdover from previous development, or is the file it's trying to call missing?

Snell1224: I believe the file naming was wrong. The installcheck hook was looking explicitly for a file named "ldms-run-static-tests.test" in that directory, however the file was actually "ldms-run-static-tests.test**.in**. I made this change and make installcheck worked for me.

ldms/src/test/Makefile.am
```diff
@@ -41,10 +41,14 @@ test_metric_LDFLAGS = $(AM_LDFLAGS) -pthread -lm
...
+mypkglib_SCRIPTS = ldms-run-static-tests.test
 mypkglib_DATA = ldms-static-test-list.sh ldms-static-test-bypass
 
+# invoke tap-driver
+TEST_LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) \
+                  $(top_srcdir)/config/tap-driver.sh
+
 installcheck-local: 
-	ldms-run-static-tests.sh
+	ldms-run-static-tests.test.in
```
baallan, line 52: @Snell1224 installcheck-local here is still failing when I run make installcheck. 
It may be you need to include ./ in front of the command, since . is not in the PATH by default.

Do we need both check and installcheck here (do they do something different?). If not, we can just remove installcheck-local.

Snell1224: @baallan Yes, you're right, make installcheck did not work for me either. I realized that I had forgot to "enable-ldms-test" when running make installcheck earlier. 

I included the ./, enabled the ldms-test option, and installcheck worked. I've already made the changes to this branch so you can look it over again. 

About the check and installcheck, yes I believe they both will be needed. Some of Jenkins test scripts test with make check and others with make installcheck (install-test.sh and unit-test.sh). Those that use make installcheck will fail whenever the 'ldms-test' option is enabled. Yes, they are slightly different in that they test different parts of the program. Make check tests the files in the source tree while the make installcheck tests the their installed copies. So, I think it might be easier/better to keep both.

baallan:
@Snell1224 now this is merged, please clone  ovis/.github/workflows/build-ddebug-centos7.yaml  to a new name and modify it to build (including the test options to configure) and run make check. And then do a second clone to run make installcheck. For both, enable all current plugins that don't require special libraries (so for example dstat should be enabled but sysclassib should not).

PR #466, opened 2020-10-20
fix redundant connection err and duplicate set messages.

baallan:
This adds (at no additional memory cost due to word alignment) the warned flag(s) to producer and producer-set objects so that warnings about 'failed connections that will be retried' and about 'duplicate set instance names' are issued once per event.

In addition:
- related messages that are informational instead of unexpected errors have been demoted to INFO.
- prdcr names have been added to debug messages about set update and update scheduling.
- the aggdup test demonstrates correct handling when 2 samplers produce the same set instance, then one sampler disappears and reappears.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_prdcr.c
+++ b/ldms/src/ldmsd/ldmsd_prdcr.c
@@ -338,7 +338,8 @@ static void _add_cb(ldms_t xprt, ldmsd_prdcr_t prdcr, ldms_dir_set_t dset)
 {
 	ldmsd_prdcr_set_t set;
 
-	ldmsd_log(LDMSD_LINFO, "Adding the metric set '%s'\n", dset->inst_name);
+	ldmsd_log(LDMSD_LINFO, "Adding the metric set '%s' from producer %s\n",
+		dset->inst_name, prdcr->obj.name);
 
 	/* Check to see if it's already there */
 	set = _find_set(prdcr, dset->inst_name);
@@ -523,22 +524,22 @@ static void __prdcr_remote_set_delete(ldmsd_prdcr_t prdcr, ldms_set_t set)
 	assert(prdcr_set->ref_count);
 	switch (prdcr_set->state) {
 	case LDMSD_PRDCR_SET_STATE_START:
-		ldmsd_log(LDMSD_LERROR,
+		ldmsd_log(LDMSD_LINFO,
 			  "Deleting %s in the START state\n",
 			  prdcr_set->inst_name);
 		break;
 	case LDMSD_PRDCR_SET_STATE_LOOKUP:
-		ldmsd_log(LDMSD_LERROR,
+		ldmsd_log(LDMSD_LINFO,
 			  "Deleting %s in the LOOKUP state\n",
 			  prdcr_set->inst_name);
 		break;
 	case LDMSD_PRDCR_SET_STATE_READY:
-		ldmsd_log(LDMSD_LERROR,
+		ldmsd_log(LDMSD_LINFO,
 			  "Deleting %s in the READY state\n",
 			  prdcr_set->inst_name);
 		break;
 	case LDMSD_PRDCR_SET_STATE_UPDATING:
-		ldmsd_log(LDMSD_LERROR,
+		ldmsd_log(LDMSD_LINFO,
 			  "Deleting %s in the UPDATING state\n",
 			  prdcr_set->inst_name);
 		break;
@@ -561,6 +562,12 @@ static void prdcr_connect_cb(ldms_t x, ldms_xprt_event_t e, void *cb_arg)
 	}
 	switch (e->type) {
 	case LDMS_XPRT_EVENT_CONNECTED:
+		if (prdcr->warned ) {
+			prdcr->warned = 0;
+			ldmsd_log(LDMSD_LINFO, "Producer %s: connection error cleared for %s %s:%d\n",
+					prdcr->obj.name, prdcr->xprt_name,
+					prdcr->host_name, (int)prdcr->port_no);
+		}
 		ldmsd_log(LDMSD_LINFO, "Producer %s is connected (%s %s:%d)\n",
 				prdcr->obj.name, prdcr->xprt_name,
 				prdcr->host_name, (int)prdcr->port_no);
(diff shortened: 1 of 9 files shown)


ldms/src/ldmsd/ldmsd_prdcr.c
```diff
@@ -561,6 +562,12 @@ static void prdcr_connect_cb(ldms_t x, ldms_xprt_event_t e, void *cb_arg)
 	}
 	switch (e->type) {
 	case LDMS_XPRT_EVENT_CONNECTED:
+		if (prdcr->warned ) {
```
tom95858, line 565: just set this to 0. There's no need for a 'connection error cleared' message.

ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -537,28 +548,35 @@ void __ldmsd_prdset_lookup_cb(ldms_t xprt, enum ldms_lookup_status status,
...
 		status = (status < 0 ? -status : status);
 		if (status == ENOMEM) {
-			ldmsd_log(LDMSD_LERROR,
-				"Error %d in lookup callback for set '%s' "
-				"Consider changing the -m parameter on the "
-				"command line to a larger value. "
-				"The current value is %s\n",
-				status, prd_set->inst_name,
-				ldmsd_get_max_mem_sz_str());
+			if (! (prd_set->warned & LDMSD_PRDCR_SET_WARNED_NOMEM)) {
```
tom95858, line 551: You're daemon is basically hosed at this point. Hiding these messages will make it seem as if everything is ok, when in fact, these sets are being ignored.

ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -537,28 +548,35 @@ void __ldmsd_prdset_lookup_cb(ldms_t xprt, enum ldms_lookup_status status,
...
-					"It is likely that there are more "
-					"than one producers pointing to "
-					"the set.\n",
-					prd_set->prdcr->obj.name, prd_set->inst_name);
+			warn_once_prdcr_set_exists(prd_set);
 		} else {
-			ldmsd_log(LDMSD_LERROR,
-				  "Error %d in lookup callback for set '%s'\n",
-					  status, prd_set->inst_name);
+			if (! (prd_set->warned & LDMSD_PRDCR_SET_WARNED_LOOKUP)) {
```
tom95858, line 564: This is a configuration error, again it will make it seem like everything is ok when in fact, there is a serious configuration problem.

ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -537,28 +548,35 @@ void __ldmsd_prdset_lookup_cb(ldms_t xprt, enum ldms_lookup_status status,
...
+				prd_set->warned |= LDMSD_PRDCR_SET_WARNED_LOOKUP;
+				ldmsd_log(LDMSD_LERROR,
+					  "Error %d in lookup callback for set '%s'\n",
+						  status, prd_set->inst_name);
+			}
 		}
 		prd_set->state = LDMSD_PRDCR_SET_STATE_START;
 		goto out;
 	}
+	if (prd_set->warned != 0) {
```
tom95858, line 574: No need to tell anyone we're clearing the error, just set it to zero.

ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -575,7 +593,8 @@ void __ldmsd_prdset_lookup_cb(ldms_t xprt, enum ldms_lookup_status status,
 			goto out;
 	}
 	prd_set->state = LDMSD_PRDCR_SET_STATE_READY;
-	ldmsd_log(LDMSD_LINFO, "Set %s is ready\n", prd_set->inst_name);
+	ldmsd_log(LDMSD_LINFO, "Set %s is ready from %s\n", prd_set->inst_name,
```
tom95858, line 596: It would be nice to normalize these messages. maybe "prdcr <name>: blah" for all of them instead of switching back and forth between whether the prdcr name comes before or after the set name.

ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -636,20 +655,24 @@ static void schedule_prdcr_updates(ldmsd_updtr_task_t task,
...
 				if (rc == EEXIST) {
-					ldmsd_log(LDMSD_LERROR, "Prdcr '%s': "
-						"lookup failed synchronously. "
-						"The set '%s' already exists. "
-						"It is likely that there are more "
-						"than one producers pointing to "
-						"the set.\n",
-						prd_set->prdcr->obj.name,
-						prd_set->inst_name);
+					warn_once_prdcr_set_exists(prd_set);
```
tom95858, line 658: Again this is a serious error.

tom95858 (changes requested):
In general this change will make dodgy producers look "ok" after the initial report. I don't think this is a good idea. Let's socialize this concept a bit more.

ldms/src/ldmsd/ldmsd_prdcr.c
baallan: We found with v3 users that having a 'cleared' message was useful. They aren't left wondering if something is perpetually broken. If the debug level is DEBUG, then it's obvious, but nobody runs production ldmsd at DEBUG level.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: We are not hiding the messages, we're simply emitting each message once per bad set.
The goal here is to prevent redundantly flooded logs, not hiding the messages.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: Again, not hiding anything-- just logging it exactly once instead of every time the updater or reconnect loop re-executes.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: Again, clear messages were found admin-friendly in v3.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: I appended the detail so that anyone who already has an expression match against the prior string in some log filter will continue to match.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: The fact that it's serious doesn't mean we should flood the log with it. Logging it once per set on which it occurs.

ldms/src/ldmsd/ldmsd_prdcr.c
baallan: @tom95858 @valleydlr There is also the question, good to discuss here, if "connection error" (which is expected in normal cluster operation of aggregators) should be an LERROR or LWARNING level message routinely. 

One could also easily add a producer parameter "warn_after=$seconds" that emits an LERROR level message only after $seconds of connection retry failures. 

This would make the typical case (admin/agg nodes come up before compute nodes) quiet for the frivolous errors caused by the compute nodes still being in their boot phase. Normal operation at 1000 nodes scale shouldn't cause a log storm of frivolous error/warning messages due to variations in daemon boot order.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: Jim has requested a presentation and discussion at the monday ldms user meeting.
I will put some ppt together.

tom95858:
@valleydlr, @baallan: I think we agreed to close this.

baallan:
@tom95858 I think we agreed that the "all encompassing log api refactor" would help some of this. But the core of this (don't emit redundant messages, even after the api refactor) is still needed.

An item within it which you and I didn't converge on was "issue a connection ok after a connection has failed". I suppose I can live without that, if it bothers you too much. I would prefer the major improvement over doing nothing at all.

It would be good if we can agree on a path forward (timing, resources) for extending the logging api to include the facility tag and log level per facility user control. Personally, I'd like to see that api change before or with the plugin api change.

PR #458, opened 2020-10-16
fix most cluster specific and sock assumptions in pll-ldms-static-test

baallan:
this is a test scripts only request and is part of testing 4.3.4/can be merged before GA.

PR #467, opened 2020-10-21
adding munge to ldms-static-test list

Snell1224:
Adding munge to plugin list for ldms-run-static-test.

### DIFF
--- a/ldms/src/test/ldms-static-test-bypass.in
+++ b/ldms/src/test/ldms-static-test-bypass.in
@@ -42,4 +42,4 @@ procdiskstats
 kgnilnd
 hello_stream
 llnl
-
+munge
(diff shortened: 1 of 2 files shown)


baallan:
Isn't there a .in file involved here? We should be using $pkglibdir and not having your home directory appear in the repo.

(admittedly, I need to fix such things in my tests occasionally)-- in fact I have one in #458 still pending.

Snell1224:
Ah, yes you're right. That's not supposed to be there. I got confused on which one I was one and accidently pushed to the wrong branch. Thanks for letting me know!

baallan:
@Snell1224 Is there a revision to this or a replacement PR? or is it ready to go?

Snell1224:
@baallan yes it is ready to go. Already revised it and fixed the issue.

PR #468, opened 2020-10-22
Fix zap_fabric:get_name() so that it can receive sa_len of 0.

nichamon:
The sa_len parameters of the zap_get_name() API is an output parameter.
The given sa_len value could be 0. Without the change, zap_fabric() would
return an error if the value is 0. This change makes the behavior of
zap_fabric's get_name() compatible with those in the other zap
transports.

### DIFF
--- a/lib/src/zap/fabric/zap_fabric.c
+++ b/lib/src/zap/fabric/zap_fabric.c
@@ -2052,14 +2052,29 @@ static zap_err_t z_get_name(zap_ep_t ep, struct sockaddr *local_sa,
 			    struct sockaddr *remote_sa, socklen_t *sa_len)
 {
 	int ret = 0;
-	size_t sz = *sa_len;
+	size_t sz;
 	struct z_fi_ep *rep = (struct z_fi_ep *)ep;
 
 	if (!rep->fi_ep) {
 		return ZAP_ERR_NOT_CONNECTED;
 	}
+	if (0 == *sa_len) {
+		/*
+		 * fi_getname() returns -FI_ETOOSHORT if *sa_len is 0.
+		 * Set it now.
+		 */
+		*sa_len = sizeof(struct sockaddr);
+	}
+	sz = *sa_len;
 	ret = fi_getname(&rep->fi_ep->fid, local_sa, &sz);
-	ret = ret || fi_getpeer(rep->fi_ep, remote_sa, &sz);
+	if (ret)
+		return ZAP_ERR_NO_SPACE;
+
+	sz = *sa_len;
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 Without this change, zap_fabric cannot be used with munge. ldms_xprt cannot begin the munge authentication process because it fails to get the transport name, i.e., zap_get_name() returns an error.

tom95858:
Hi @nichamon, I think on input the sa_len specifies the size of the output structure. On output it specifies the bytes actually written. The reason for this is because different transports (IPv4/IPv6) have different sizes for the sockaddr. So the point is that if the caller gives you a sockaddr_in, and the transport is sockaddr_in6, the interface would overwrite the output buffer.

nichamon:
HI @tom95858, I understand the idea. 

When I tested zap_fabric, I found that I could not use zap_fabric with Munge. The causes were 
* fi_getname() will return an error (-FI_ETOOSMALL) if the input size is 0. In turn zap_fabric:get_name() returns an error.
* ldms_xprt_sockaddr() always passes sa_len with the value 0 (instead of using caller's value).

I created this pull request to make zap_fabric:get_name() compatible with the get_name() interface of the other zap transports, i.e., it won't return an error if the input size is 0.

Regarding ldms_xprt_sockaddr(), I forgot to create a pull request for it. The pull request is #474. There are 2 issues in ldms_xprt_sockaddr(). One is what I mentioned above. The other issue is that ldms_xprt_sockaddr() may cause a buffer overflow.

zap_fabric can be used with Munge by applying either this pull request or pull request #474. However, the pull requests were created for different reasons.

lib/src/zap/fabric/zap_fabric.c
```diff
@@ -2052,14 +2052,29 @@ static zap_err_t z_get_name(zap_ep_t ep, struct sockaddr *local_sa,
...
 {
 	int ret = 0;
-	size_t sz = *sa_len;
+	size_t sz;
 	struct z_fi_ep *rep = (struct z_fi_ep *)ep;
 
 	if (!rep->fi_ep) {
 		return ZAP_ERR_NOT_CONNECTED;
 	}
+	if (0 == *sa_len) {
```
tom95858, line 2061: Is this still a pull request that you want pulled in? Shouldn't this check if the sa_len is < the required address length?

nichamon: No, this doesn't need to be pulled in. The fix to `ldms_xprt_sockaddr` fixes the problem that zap_fabric can't be used with Munge. I'll close this pull request. Thanks.

nichamon:
It is an unnecessary change.

PR #469, opened 2020-10-22
Remove the atexit() call in zap_fabric

nichamon:


tom95858:
Hi @nichamon what symptom does this fix?

nichamon:
@tom95858 When I tested zap_fabric, sometimes ldms_ls hang when it connected to an aggregator using the verbs provider. This change is one of my efforts to fix it. I just tested it again. It seems it doesn't fix ldms_ls hang problem. I might have tested with the wrong provider before I created the pull request. I'll close this pull request.

PR #473, opened 2020-10-23
Add Zap's shared map feature

nichamon:


nichamon:
**Tests that have been performed**

- _zap_sock_ and _zap_rdma_ passed the zap_test program.
- _zap_ugni_ has not been tested yet.
- _zap_fabric_ needs to be tested more. There are underlying problems that prevent me from properly testing the change.
- _ldmsd_ has been tested: 1 sampler daemon and 1 aggregator. ldms_ls -l to both. They work as expected.

PR #474, opened 2020-10-26
Fix ldms_xprt_sockaddr()

nichamon:
This change fixes 2 issues in ldms_xprt_sockaddr().
* ldms_xprt_sockaddr() always passes 0 as sa_len to zap_get_name()
  instead of the given value by the caller
* ldms_xprt_sockaddr() may cause memory overflow if the given buffers
  are smaller than the actual addresses.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -3764,20 +3764,28 @@ int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
 {
 	struct ldms_xprt *x = _x;
 	zap_err_t zerr;
- again:
-	if (x->sa_len) {
-		/* use cached addresses */
+
+	if (!x->sa_len) {
+		/*
+		 * The addresses have never been gotten.
+		 * Get the addresses first.
+		 */
+		x->sa_len = sizeof(x->local_sa);
+		zerr = zap_get_name(x->zap_ep, (struct sockaddr *)&x->local_sa,
+				    (struct sockaddr *)&x->remote_sa, &x->sa_len);
+		if (zerr)
+			return zap_zerr2errno(zerr);
+	}
+
+	if (x->sa_len > *sa_len) {
 		*sa_len = x->sa_len;
-		memcpy(local_sa, &x->local_sa, x->sa_len);
-		memcpy(remote_sa, &x->remote_sa, x->sa_len);
-		return 0;
+		return E2BIG;
 	}
-	zerr = zap_get_name(x->zap_ep, &x->local_sa, &x->remote_sa, &x->sa_len);
-	if (zerr) {
-		x->sa_len = 0;
-		return -1;
-	}
-	goto again;
+
+	memcpy(local_sa, &x->local_sa, x->sa_len);
+	memcpy(remote_sa, &x->remote_sa, x->sa_len);
+	*sa_len = x->sa_len;
+	return 0;
 }
 
 static void __attribute__ ((constructor)) cs_init(void)
--- a/ldms/src/core/ldms_xprt.h
+++ b/ldms/src/core/ldms_xprt.h
@@ -429,8 +429,8 @@ struct ldms_xprt {
 
 	/* for addr caching */
 	socklen_t sa_len;
-	struct sockaddr local_sa;
-	struct sockaddr remote_sa;
+	struct sockaddr_storage local_sa;
+	struct sockaddr_storage remote_sa;
 };
 
 void __ldms_xprt_term(struct ldms_xprt *x);


tom95858:
@nichamon, the bug is not in ldms_xprt_sockaddr, it is in the callers of this interface. All of the callers should be specifying the size of the local_sa buffer they are providing. It is an application error not to do so and this function should assert or return EINVAL if *sa_len == 0 on input. 

On output it should be the size of the address actually written.

ldms/src/core/ldms_xprt.c
```diff
@@ -3764,20 +3764,24 @@ int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
...
 	zap_err_t zerr;
- again:
-	if (x->sa_len) {
-		/* use cached addresses */
-		*sa_len = x->sa_len;
-		memcpy(local_sa, &x->local_sa, x->sa_len);
-		memcpy(remote_sa, &x->remote_sa, x->sa_len);
-		return 0;
-	}
-	zerr = zap_get_name(x->zap_ep, &x->local_sa, &x->remote_sa, &x->sa_len);
```
nichamon, line 3775: @tom95858 `ldms_xprt_sockaddr` pass `&x->sa_len` instead of `sa_len`. `x->sa_len` is always zero at this line. This is a bug.

ldms/src/core/ldms_xprt.c
```diff
@@ -3764,20 +3764,24 @@ int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
...
-	if (zerr) {
-		x->sa_len = 0;
-		return -1;
+	socklen_t __sa_len = *sa_len;
+
+	if (!x->sa_len) {
+		zerr = zap_get_name(x->zap_ep, &x->local_sa,
+					&x->remote_sa, &__sa_len);
```
nichamon, line 3771: @tom95858 Here `ldms_xprt_sockaddr` passes the `*sa_len` value to `zap_get_name`.

nichamon:
@tom95858 I also updated the patch so that `ldms_xprt_sockaddr` returns EINVAL if the given `*sa_len` is 0.

ldms/src/core/ldms_xprt.c
tom95858: @nichamon, on input to these address functions, sa_len is the size of the buffer. On output sa_len is the size of the address written to the buffer. It should be something like this:
```
socklen_t sa_len = sizeof(x->local_sa);
zap_get_name(..., &x->local_sa, &sa_len);
x->sa_len = sa_len;
```

ldms/src/core/ldms_xprt.c
```diff
@@ -3764,20 +3764,27 @@ int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
...
-	zerr = zap_get_name(x->zap_ep, &x->local_sa, &x->remote_sa, &x->sa_len);
-	if (zerr) {
-		x->sa_len = 0;
-		return -1;
+	socklen_t __sa_len = *sa_len;
+
+	if (*sa_len == 0)
+		return EINVAL;
+
+	if (!x->sa_len) {
```
tom95858, line 3772: Why are you checking this? I'm actually not sure why sa_len is in the transport structure at all. The input value is always going to be sizeof(local/remote_sa). 

The local_sa and remote_sa addresses should be sockaddr_storage (128B), not sockaddr(14B). The current definitions are too small to accept an IPv6 address. Right now it's not symptomatic because we're not using IPv6.
That said, it's not the correct way to handle addresses like this.

ldms/src/core/ldms_xprt.c
```diff
@@ -3764,20 +3764,27 @@ int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
...
-		*sa_len = x->sa_len;
-		memcpy(local_sa, &x->local_sa, x->sa_len);
-		memcpy(remote_sa, &x->remote_sa, x->sa_len);
-		return 0;
-	}
-	zerr = zap_get_name(x->zap_ep, &x->local_sa, &x->remote_sa, &x->sa_len);
-	if (zerr) {
-		x->sa_len = 0;
-		return -1;
+	socklen_t __sa_len = *sa_len;
```
tom95858, line 3767: local variables like this don't need the __ prefix. The symbols are not visible outside the function. I'm not sure why it's here at all actually. IF we are going to keep x->sa_len (and I'm not sure it's needed), then it the call would look something like this:
```
x->sa_len = sizeof(x->local_sa);
zerr = zap_get_name..., &x->local_sa, &x->remote_sa, &x->sa_len);
```

ldms/src/core/ldms_xprt.c
```diff
@@ -3764,20 +3764,27 @@ int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
...
+		zerr = zap_get_name(x->zap_ep, &x->local_sa,
+					&x->remote_sa, &__sa_len);
+		if (zerr)
+			return zap_zerr2errno(zerr);
+		x->sa_len = __sa_len;
 	}
-	goto again;
+
+	/*
+	 * Truncate the address if the given buffers are smaller
```
tom95858, line 3781: This is an EINVAL. If this is the case, then the address written is garbage, truncating it avoids the buffer overflow, but the address written is invalid.

tom95858 (changes requested):

ldms/src/core/ldms_xprt.c
nichamon: @tom95858 I'm aware that truncated address is garbage. I'm mimic what `getsockname()` does. `getsockname()` doesn't return an error if the input buffer is smaller than the actual address size. I'll return EINVAL instead then.

nichamon:
@tom95858 I have updated the code.

What I did,

`socklen_t __sa_len` was removed.

Given that `x->local_sa` and `x->remote_sa` already contain the actual addresses:
If `x->sa_len > *sa_len`, `ldms_xprt_sockaddr()` returns EINVAL, set `*sa_len = x->sa_len` doesn't set the given `local_sa` and `remote_sa`.

nichamon:
@tom95858 I have changed `x->local_sa` and `x->remote_sa` to be `struct sockaddr_storage` and modified `ldms_xprt_sockaddr()` to work similarly to `getsockname()`. The differences are

- `ldms_xprt_sockaddr()` will return E2BIG if the input `sa_len` is smaller than the actual address size. `getsockname()` will return 0 in this case.

- `ldms_xprt_sockaddr` caches the addresses in the LDMS transport (`x->local_sa` and `x->remote_sa`) by calling `zap_get_name`. If they have been cached already, it copies the address in the transport to the input buffers.

tom95858 (approved):

PR #475, opened 2020-10-27
Fix endpoint leak when zap_connect failed synchronously

narategithub:
When `zap_connect()` failed synchronously, zap endpoint were left alone
and the ldms xprt reference that was taken when setting zap ucontext was
not put back. Hence, the LDMS xprt reference would never reached 0 and
the associated resources would not be freed. This patch releases the
zap endpoint and put down the associated LDMS xprt reference.


tom95858:
@narategithub, can you please confirm that this issue is resolved in OVIS-4.3.4-GA. I believe that it is, but I would like you to verify. If it is, in fact, fixed, please close this pull request.

narategithub:
@tom95858 without this patch, the `OVIS-4` branch (456bbcd), still have the endpoint leak problem. I have updated the patch to call `__ldms_xprt_resource_free()` -- making it consistent with connection error and connection reject asynchronous cases.

PR #478, opened 2020-10-28
Fix extra args appearing in ldmsd command when munge used.

baallan:
also set default key file name for ovis auth only if ovis type is specified.


baallan:
@tom95858 If going GA, please include this. We need it for munge to work as expected for TOSS folk using genders config.

PR #480, opened 2020-10-29
fix masking of perm bits definition error on sampler_base

baallan:
The sampler_base parsing of the octal permissions bits allowed parse errors to go through. This could lead to an untrapped configuration error leaving set access security at a level different than the user intended.

### DIFF
--- a/ldms/src/sampler/sampler_base.c
+++ b/ldms/src/sampler/sampler_base.c
@@ -414,13 +414,17 @@ int base_auth_parse(struct attr_value_list *avl, struct base_auth *auth,
 	}
 	value = av_value(avl, "perm");
 	if (value) {
-		if (value[0] != '0') {
-			log(LDMSD_LINFO,
-			    "%s: Warning, the permission bits '%s' are not specified "
-			    "as an Octal number.\n",
-			    value);
+		long lval, pval;
+		errno = 0;
+		lval = strtol(value, NULL, 8);
+		pval = lval & 0777;
+		if (errno || pval != lval || pval == 0) {
+			log(LDMSD_LERROR,
+			    "The permission bits must specified "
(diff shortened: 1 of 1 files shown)


baallan:
@tom95858 please include this security fix in 434 ga if it has not already been released.

morrone:
I think a better approach would be to improve the parsing.

I can't imagine that anyone wants to use base 10 to express permission bits, and your proposed change would make that an error anyway. So why don't we just always interpret numbers for perms fields as base 8? Just change the "base" in the strtol() to 8 instead of 0, and then the test for leading zero can just be eliminated. And then folks can specify the numbers like they would with the chmod command line tool. No leading zero necessary.

baallan:
@tom95858 Do you have a preference or a consistency-with-sos-configs reason for keeping the 0-prefix check on permissions-config arguments? If not, I'm inclined to take Chris' suggestion.

tom95858:
I agree with @morrone too.

baallan:
@tom95858 I updated to assume octal and range check. Is it correct that the valid permission bits are at most 777?

ldms/src/sampler/sampler_base.c
```diff
@@ -414,13 +414,17 @@ int base_auth_parse(struct attr_value_list *avl, struct base_auth *auth,
...
-			    "as an Octal number.\n",
-			    value);
+		long lval, pval;
+		errno = 0;
+		lval = strtol(value, NULL, 8);
+		pval = lval & 0777;
+		if (errno || pval != lval || pval == 0) {
+			log(LDMSD_LERROR,
+			    "The permission bits must specified "
+			    "as a non-zero octal number < 1000; got %s.\n", value);
```
tom95858, line 424: This error message is a bit dodgy. Might want to say <= 777 or something like that.

PR #486, opened 2020-11-05
add file descriptor count option to dstat

baallan:
The dstat sampler can now collect a simple or detailed summary
of /proc/self/fd open file descriptors. option fd=1 collects
just the count and fdtypes=1 also collects file descriptor counts
by type (anon, pipe, dev, /path, socket). fd info collection is
off by default.

this also adds the option auto-schema=1 (off by default)
which changes the default schema name from dstat to dstat_$hash
where hash is 1 or 2 hex digits derived from the option flags
which control the content of the schema. A utility ldms_dstat_schema_name
provides the schema name for storage policy setup.

When running an aggregator, using dstat with fdtypes=1 enabled will help identify leaks of file descriptors.

### DIFF
--- a/ldms/scripts/examples/dstat.job
+++ b/ldms/scripts/examples/dstat.job
@@ -2,19 +2,23 @@
 export plugname=dstat
 portbase=62086
 JOBDATA $TESTDIR/job.data 1 2 3
+export dsname=$(ldms_dstat_schema_name mmalloc=1 io=1 fd=1 auto-schema=1)
+echo DSNAME=$dsname
 LDMSD -p prolog.jobid  1 2
 LDMSD -p prolog.jobid -p prolog.jobid.store3 3
 MESSAGE ldms_ls on host 1:
 JOBDATA $TESTDIR/job.data 1 2 3
-LDMS_LS 1 -l
+SLEEP 2
+LDMS_LS 1 -lv
 MESSAGE ldms_ls on host 2:
-LDMS_LS 2 -l
+LDMS_LS 2 -lv
 MESSAGE ldms_ls on host 3:
-LDMS_LS 3
+LDMS_LS 3 -lv
 JOBDATA $TESTDIR/job.data 1 2
 SLEEP 3
 JOBDATA $TESTDIR/job.data 1 3
 SLEEP 2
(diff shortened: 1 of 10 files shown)


baallan:
@narategithub I've added file counters to the dstat sampler, so that fd leaks are more easily tracked.
It would be good to get this merged into 434 before GA to simplify testing and post-GA debugging at deployment scale.

tom95858:
@baallan, I don't think we want this done this way since it walks the directory entry to count the file descriptors. That will take a very long time on systems with large numbers of connections and therefore file descriptors ... exactly the ones you are interested in.

baallan:
@tom95858 Neither Narate nor google came up with a better method of counting and optionally classifying fds when I asked, so I added it and documented it as being a potentially expensive option and made it disabled by default.
If you have a better method that doesn't require kernel access, I can certainly improve the sampler implementation with it. 

The sampler gives admins & researchers/testers a valuable ability to build confidence about resource consumption during early deployments; I expect once admins get confident with an updated ldmsd configuration, they'll probably turn off dstat.

tom95858:
Hi @baallan, well this is water under the bridge now since you pushed this change without completing the review. This change will, I believe, give misleading information about the CPU utilization of ldmsd. In addition, the same information is available with ps, lsof, and any number of other utilities that will not affect the CPU, memory and network utilization attributed to ldmsd.

A review of configure.ac indicates that dstat is enabled by default and unfortunately, this change breaks the Centos-7 build with -Wall -Werror.

tom95858:
Hi @baallan, I misspoke, it is Ubuntu 18.04 that the failure occurred on.

baallan:
Do we not have a test-make-dist.yaml equivalent that runs automatically with ubuntu18/Wall/Werr on the github actions yet? Should I try to make one, or do you/Mon already have one pending someplace?

baallan:
@tom95858 Near the top of my testing development list is jenkins-driven performance quantification/regression testing per plugin and overall. We'll see how dstat and the other heavyweights (lustre, procnetdev, edac, vmstat, procstat) do in various configurations. This is one of the internal requirements for getting to deploy on our ARM platforms.

PR #487, opened 2020-11-05
Add 1 sec sampling, fabric-sock test options, file counting

baallan:
incorporated file leak checks in cluster test, and other improvements


baallan:
this is a man/test scripts only change

PR #490, opened 2020-11-06
Fix ldms_xprt build error with `-DDEBUG`

narategithub:


ldms/src/core/ldms_xprt.c
```diff
@@ -549,7 +549,7 @@ void __ldms_xprt_resource_free(struct ldms_xprt *x)
 		case LDMS_CONTEXT_DIR_CANCEL:
 			x->active_dir_cancel--;
 			break;
-		case LDMS_CONTEXT_LOOKUP:
+		case LDMS_CONTEXT_LOOKUP_REQ:
```
tom95858, line 552: Is all this debug code still used/relevant/useful? The code would be a lot easier to read without it.

PR #493, opened 2020-11-10
Split apart libpapi and libfm configuration

morrone:
We remove the libpfm linkage from the libhwevent and rapl samplers.
It appears that neither of those had a direct dependency upon libpfm.

With that change made, it is clear that our --enable-papi option should
not also depend upon pfm. We therefore remove that and instead introduce
an --enable-pfm option to require or disable libpfm inclusion. The
--with-libpfm-prefix option remains to specify a path to check for
libpfm.

For instance, the following should now work:

  --enable-papi --with-libpapi-prefix=/usr/papi600 --disable-pfm

This would result in only the samplers that depend on papi being built.
Those samplers (only one at time of writing) that also require libpfm
would not be build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -319,7 +319,7 @@ if test -z "$ENABLE_INFLUX_TRUE"; then
 fi
 
 AC_ARG_ENABLE([papi],
-	[AS_HELP_STRING([--enable-papi], [require components that depend upon libpapi (and libpfm) @<:@default=check@:>@])],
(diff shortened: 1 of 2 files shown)


morrone:
@tom95858 , I checked that the samplers built as expected with this patch, but I have never actually used the papi/pfm related samplers. It would be good to verify that one or both of libhweventpapi and/or rapl still work with the explicit -lpfm linkage removed.

morrone:
Abandoning.

PR #497, opened 2020-11-10
Add the zap_send_mapped() API to the zap library

nichamon:
Currently the zap_send() API receives application buffer and copies it
to another buffer that has been registered with the hardware using the
zap_map() interface. The zap_send() API is simple to use but is
inefficient when the same buffer is to be sent to many clients because
the same buffer is mapped and copied many times.

The zap_send_mapped() API has the following signature.

   int zap_send_mapped(zap_ep_t ep, zap_map_t map, void *buf,
                       size_t len, void *context)

The zap_send_mapped() API posts the RDMA_SEND operation directly from
the previously mapped buffer (map) without requiring registration or
copy by the transport. This significantly improves performance for
sending operation. Unlike the zap_send() API, the zap_send_mapped() API
delivers a completion event via the application callback function unless
the function return an error.


narategithub:
Part of #783

PR #498, opened 2020-11-11
Allow root to access anything and remove duplicated logic

nichamon:


nichamon:
@tom95858 The problem I found is that updtr_stop given in ldmsd_controller returns Permission Denied when ldmsd uses -a none or -a ovis. The cfgobj's UID and GID are the process UID and GID when the object is created according to configuration files. Both auth_none and auth_ovis set LDMS transport's remote UID and GID to 0. Since I started ldmsd as a non-root user, the transport's remote UID and the object's UID don't match. In turn, updtr_stop returns Permission Denied.

I modified `ovis_access_check` so that it allows the access if root is the one who requests the access. Then, I found that `ldms_access_check` has the same logic so I made `ldms_access_check` to call `ovis_access_check`. I didn't remove one of the two functions to make the change small.

PR #499, opened 2020-11-11
Remove set_info from ldms_lookup message

nichamon:


tom95858:
Hi @nichamon, have you tested if this works with OVIS-4.3.3?

nichamon:
> Hi @nichamon, have you tested if this works with OVIS-4.3.3?

Hi @tom95858, it passes the Github action that Narate's added to test the compatibility with V4.3.3. The path isn't used since  V4.3.3. However, it is not a bug fix. At this release stage, IMO, let's not merge it. It is a ported code from the master branch.

PR #504, opened 2020-11-11
Nit fix zap_rdma build error with `-DCTXT_DEBUG`

narategithub:


lib/src/zap/rdma/zap_rdma.c
```diff
@@ -105,7 +105,7 @@
 #ifdef CTXT_DEBUG
```
tom95858, line 105: @narategithub we should probably either make this ifdef DEBUG, or add CTXT_DEBUG to the github DEBUG build test. I really don't want to discover these things one at a time.

narategithub: sure. I should also make github `-DDEBUG` build job to build zap_rdma and zap_fabric (the library dependencies are available on CentOS7 repo). By default they aren't built.

narategithub: @tom95858 I put in various DEBUG flags in debug build github workflow in #509.

PR #507, opened 2020-11-11
remove configure macro for unused constant

baallan:


baallan:
this was a holdover from v3 afaict

PR #508, opened 2020-11-11
trap/fix use of uninit pointer/ins in ldmsctl:__handle_cmd

baallan:
bug picked up by gcc 4.8.5 -Wall

### DIFF
--- a/ldms/src/ldmsd/ldmsctl.c
+++ b/ldms/src/ldmsd/ldmsctl.c
@@ -1948,7 +1948,7 @@ static int __handle_cmd(struct ldmsctl_ctrl *ctrl, char *cmd_str)
 	/*
 	 * Send all the records and handle the response now.
 	 */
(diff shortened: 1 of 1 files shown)


tom95858 (commented):
@baallan, this code is broken, but although your change may silence the static analysis tool, it didn't fix the error. If you'll notice resp comes from recv_buf which is pulled off of the recv_buf_q, copied into lbuf, and then freed on line 1995, so resp->err_msg is pointing to freed memory. @nichamon, please look at this for an actual fix. I think that in the event that there is an error,  we'll want to do something different, probably inside the while(1).

nichamon:
@tom95858 I's working on this patch shortly because I swapped my attention to the dir_update path last week. I'll be back on this tonight.

baallan:
@nichamon please close this PR after you create a better PR.

nichamon:
The fix is #517 .

PR #509, opened 2020-11-12
Add various DEBUG flags to github debug build job

narategithub:


narategithub:
This patch needs #504 and #505 to succeed.

PR #511, opened 2020-11-12
Fix a possible list corruption in ldmsd_set_deregister()

nichamon:


ldms/src/ldmsd/ldmsd.c
```diff
@@ -843,22 +843,25 @@ int ldmsd_set_register(ldms_set_t set, const char *pluing_name)
...
-	ldmsd_plugin_set_t set;
+	ldmsd_plugin_set_t set, nxt_set;
 	ldmsd_plugin_set_list_t list;
 	struct rbn *rbn;
 	ldmsd_set_tree_lock();
 	rbn = rbt_find(&set_tree, plugin_name);
 	if (!rbn)
 		goto out;
 	list = container_of(rbn, struct ldmsd_plugin_set_list, rbn);
-	LIST_FOREACH(set, &list->list, entry) {
```
tom95858, line 854: I think it would be better if you kept the LIST_FOREACH and  moved the delete outside the loop. It doesn't make sense to continue after you find the item anyway; there shouldn't be two sets with the same instance name  in the list correct?

```
LIST_FOREACH(set, &list->list, entry) {
    if (0 == strcmp(set->inst_name, isnt_name))
         break;
}

if (set) {
   LIST_REMOVE(set, entry);
   ...delete everything...
}
```

nichamon: @tom95858 Got it! I'll let you know ASAP I updated the patch.

nichamon: @tom95858 I made the change per your request. The updated code has been tested.

PR #512, opened 2020-11-12
Initialize the buffer length before passing to ldms_xprt_sockaddr

nichamon:
…addr

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -6080,6 +6080,7 @@ static int stream_subscribe_handler(ldmsd_req_ctxt_t reqc)
 			       len, 256);
 		goto send_reply;
 	}
+	key->addr_len = sizeof(key->addr);
 	rc = ldms_xprt_sockaddr(reqc->xprt->ldms.ldms, &local_sa, &key->addr,
 				&key->addr_len);
 	if (rc) {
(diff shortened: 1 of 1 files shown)


tom95858 (commented):
What is the symptom here? Are streams broken after the libfabric change for salen == 0?

nichamon:
> What is the symptom here? Are streams broken after the libfabric change for salen == 0?

There are no symptoms if Munge is in used. The reason is that ldms_auth_munge calls `ldms_xprt_sockaddr` so the remote and local addresses are cached, given that the address buffers are large enough. (If`ldms_xprt_sockaddr` returns an error at this state, the connection will be terminated).

For the other authentication methods, stream_subscribe and strean_subscribe requests may fail. I said 'may' here because the input buffer length isn't initialized so the value is unknown. Thus, `prdcr_subscribe` command may not work properly as the stream_subscribe sent by the aggregator to the sampler daemon could fail. I've seen this symptom when I don't use any authentication method.

nichamon:
The bug doesn't affect slurm sampler plugin and papi sampler plugin because both of them subscribe to a stream by calling `ldmsd_stream_subscribe`.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6226,6 +6228,7 @@ void stream_xprt_term(ldms_t x)
 	struct __RSE_key_s *key = (void*)_buff;
 	int rc;
 
+	key->addr_len = sizeof(key->addr);
```
nichamon, line 6231: @tom95858 Without this line, `key->addr_len` is 0 and, thus, `ldms_xprt_sockaddr` would return an error.  The function `stream_xprt_term` removes the client from the client list when the connection to the client is disconnected. Since it cannot get the address, it won't remove the client from the list upon disconnection.

PR #513, opened 2020-11-12
Fix handling slurm include in alternate install path

morrone:
Needs testing, but I think that will fix the SLURM_CFLAGS in an alternate location problem.


tom95858:
That's what I'm testing now...

morrone:
No longer required.

PR #514, opened 2020-11-12
Improve the dir_update path

nichamon:


ldms/src/core/ldms_xprt.c
```diff
@@ -795,7 +802,6 @@ static void process_dir_request(struct ldms_xprt *x, struct ldms_request *req)
 				goto out;
 			}
 			reply->hdr.len = htonl(last_cnt + hdrlen);
-			reply->dir.json_data_len = htonl(last_cnt);
```
tom95858, line 798: We should set this value to -1 or something and add doc to indicate that it is deprecated...

ldms/src/core/ldms_xprt.c
```diff
@@ -347,34 +347,42 @@ static void send_dir_update(struct ldms_xprt *x,
...
+		if (x->log)
+			x->log("%s: Out of memory\n", __FUNCTION__);
+		return;
+	}
+
 	reply->hdr.xid = x->remote_dir_xid;
 	reply->hdr.cmd = htonl(LDMS_CMD_DIR_UPDATE_REPLY);
 	reply->hdr.rc = 0;
 	reply->dir.type = htonl(t);
-	reply->dir.json_data_len = htonl(cnt);
```
tom95858, line 368: Let's set it to -1 to indicate that it is deprecated...

ldms/src/core/ldms_xprt.c
```diff
@@ -347,34 +347,42 @@ static void send_dir_update(struct ldms_xprt *x,
...
 	struct ldms_reply *reply;
 
 	assert(t != LDMS_DIR_LIST);
 
 	hdr_len = sizeof(struct ldms_reply_hdr)
 		+ sizeof(struct ldms_dir_reply);
-	buf_len = ldms_xprt_msg_max(x);
-	reply = malloc(buf_len);
+	/* Calculate the size of the string */
+	json_data_len = snprintf(NULL, 0, "{ \"directory\" : [ %s ]}", json);
```
tom95858, line 358: Not the most efficient way to do this, but let's start with this.

tom95858 (changes requested):
Please test this against a 4.3.3 client to ensure that this fixes the issue

nichamon:
> Please test this against a 4.3.3 client to ensure that this fixes the issue

I have tested this the dir_update path doesn't crash anymore, but the aggregator asserts when it received a SET_DELETE_REQUEST as it does not recognize the request.

ldms/src/core/ldms_xprt.c
nichamon: I added the line back so that the path in 4.3.4 is compatible with 4.3.3. __process_dir_reply() in v4.3.3 still uses reply->dir.json_data_len.

ldms/src/core/ldms_xprt.c
nichamon: I added the line back so that the path in 4.3.4 is compatible with 4.3.3. __process_dir_reply() in v4.3.3 still uses reply->dir.json_data_len.

ldms/src/core/ldms_xprt.c
nichamon: I have changed this to
`
json_data_len  = strlen("{ ....}") + json_sz;
`

nichamon:
@tom95858 I have tested the send_dir_update path. The setup is

- sampler daemon runs 4.3.3. It used the test_sampler plugin and created 3 sets.
- 1st-level agg (AGG1) runs 4.3.3
- 2nd-level agg (AGG2) runs 4.3.4
- ldms_ls -vl to all daemons reported the sets correctly
- The sampler daemon was killed.
- AGG1 unpublished the sets and in turn send_dir_update is called to send a dir_del to AGG2.
- AGG2 received the dir_del message and reseted the set.
- ldms_ls -vl to AGG1 and AGG2 reported no sets as expected.
- The sampler daemon restarted.
- AGG1 added the sets and published the sets. It sent dir_add to AGG2.
- AGG2 received the dir_add and added sets.
- ldms_ls -vl to all daemons reported the sets correctly.

All daemons survived. The fix is good.

ldms/src/core/ldms_xprt.c
tom95858: Does 4.3.3 use it correctly as ntohl(??) or is it expecting the unswapped value?

tom95858: Also, I don't see where this line has been added back.

tom95858:
@nichamon, can you please test that ldms_ls (from the sampler node running ovis-4.3.3) to the aggregator node (running ovis4.3.4) does not crash?

nichamon:
> @nichamon, can you please test that ldms_ls (from the sampler node running ovis-4.3.3) to the aggregator node (running ovis4.3.4) does not crash?

I have tested this. ldms_ls running ovis-4.3.3 with and without -vl works with an aggregator running v4.3.4 + this patch.

PR #496, opened 2020-11-10
Fix the hwevetnpapi sampler plugin

nichamon:

PR #515, opened 2020-11-13
github action build test update

narategithub:
- add "-Wall -Werror" to the build test.
- use ovishpc container image for the build. This is the same one used
  in ldms-test repository.


narategithub:
@tom95858  FYI, The github "Build Test - CentOS7 / build" job failed due to hweventpapi. Once #496 is merged and this PR has rebased, the github check should succeed.

PR #517, opened 2020-11-17
Fix a use-after-free in ldmsctl.c

nichamon:
A symptom of the bug is that the status commands will not report any
status because the error code value is a non-zero, garbage value. No
error messages are printed because the response is not really an error
response.


nichamon:
@tom95858 

> @nichamon, please look at this for an actual fix. I think that in the event that there is an error, we'll want to do something different, probably inside the while(1).

I'm not sure what errors are you referring to. In case of an error that LDMSD fails to handle the command, `cmd->resp()` is responsible to handle it. I added the code to assert if the `sem_wait()` wakes up but the recv list is empty.

PR #519, opened 2020-11-18
Add subcribe, xprt_stats and thread_stats to the ldmsctl

nichamon:
The commands have been added to ldmsd_controller already.


nichamon:
The patch has been tested that the subscribe command really made the LDMS daemon to subscribe the stream. The xprt_stats and thread_stats commands report the same information as in ldmsd_controller.

PR #520, opened 2020-11-18
Fix bugs in the stream path

nichamon:


nichamon:
The PR contains 2 commits. The first one is the one that takes the LDMS transport's reference when subscribes a stream and put it back when removes the client from the list. While I tested the patch, I found the bug that is fixed by the second commit. It was a typo that passes `&key` instead of `key` which is `struct __RSE_key_s *` already.

tom95858:
We need to be a bit more specific about the naming of the pull request/commit. "Fix bugs in XXX" is not particularly informative when reviewing a git log.

PR #521, opened 2020-11-19
Add a build test on Ubuntu-20.04

nichamon:
The test includes build job and distcheck job. The build uses the
default enable/disable configure options.

### DIFF
--- a/.github/workflows/build-test-ubuntu.yaml
+++ b/.github/workflows/build-test-ubuntu.yaml
@@ -1,0 +1,28 @@
+name: Build Test - Ubuntu-20.04
+
+on:
+  push:
+    branches: [ OVIS-4 ]
+  pull_request:
+    branches: [ OVIS-4 ]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    container:
+        image: ovishpc/ovis-ubuntu-build
+    steps:
+    - uses: actions/checkout@v2
+    - run: sh autogen.sh
+    - run: ./configure CFLAGS="-Wall -Werror"
+    - run: make
+
+  distcheck:
+    runs-on: ubuntu-20.04
+    container:
+        image: ovishpc/ovis-ubuntu-build
+    steps:
+    - uses: actions/checkout@v2
+    - run: sh autogen.sh
+    - run: ./configure
+    - run: make distcheck


baallan:
@nichamon which gcc version does this end up compiling with, or is there some way to force the output of gcc --version from the yaml script?

tom95858:
@nichamon, this seems to be failing because bison and flex are missing from the build container. Am I missing something?

nichamon:
> @nichamon, this seems to be failing because bison and flex are missing from the build container. Am I missing something?

@tom95858 You didn't miss anything. I just pushed a new docker image. Bison and Flex were missing from the previous docker image. Now it will fail because of the errors (-Wall and -Werror) in `ldms/src/sampler/dstat/parse_stat.c`. I'll address those and probably rebase this on top of the fix.

nichamon:
> @nichamon which gcc version does this end up compiling with, or is there some way to force the output of gcc --version from the yaml script?

The gcc version that the build test uses is gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0. The problem in #486 can be seen when we compile with both gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0 and the one in the build test.

baallan:
@nichamon the snprintf usage which is generating nuisance warnings is expected to truncate in copying from the longer string; it would be appropriate to pragma suppress that warning in the source.
The readdir_r issue is discussed at:  https://lwn.net/Articles/696474/
It's not obvious to me which way we want to go; will ldmsd be forever compiled with glibc? If not, we may want to pragma suppression that bunch of warnings too.

nichamon:
@baallan According to the man page, `readdir` is **MT-Unsafe race:dirstream**. According to the gnu documentation, this means `readdir` is not thread-safe in the case of concurrent calls to `readdir()` that specify the **same** directory stream. In the dstat plugin case, the dir stream object is local to the `parse_proc_pid_fd` function, so it is impossible that there will be more than one thread that pass the object to `readdir()` concurrently.

baallan:
@nichamon so if i readdir on /proc/$pid/fd simultaneously from two different threads (different plugins with overlapping functionality, which we may easily have happen), are the stream objects definitely different? Or is the same stream returned to both threads by opendir()?

nichamon:
@tom95858 I have rebased the branch on top of the latest OVIS-4. Now OVIS-4 passes the Ubuntu20.04 build test.

PR #522, opened 2020-11-19
fix #481 by applying sampler_base auth parse/set (security)

### LINKED ISSUE #481: jobid plugin does not follow perm bits option convention, leaving set data exposed
The ldms_jobid plugin should apply the sampler_base auth configuration function to parse uid/gid/parm in the conventional way.

baallan:
replace perm/uid/gid option code with more featureful sampler_base function call.
Fixes #481


baallan:
this security fix has been rebased to current OVIS-4. please merge before next release.

PR #525, opened 2020-11-20
Fix zap_fabric build error (w/o DEBUG)

narategithub:
When building zap_fabric without `-DDEBUG`, the static global variable
`z_fi_op_str` is not used. With `-Wall -Werror`, this resulted in
build error. This patch marks the variable unused with
`__attribute__((unused))`.

### DIFF
--- a/lib/src/zap/fabric/zap_fabric.c
+++ b/lib/src/zap/fabric/zap_fabric.c
@@ -90,6 +90,7 @@ static struct {
 	pthread_mutex_t		lock;
 } g;
 
+__attribute__((unused))
 static const char *z_fi_op_str[] = {
 	[ZAP_WC_SEND]        =  "ZAP_WC_SEND",
 	[ZAP_WC_RECV]        =  "ZAP_WC_RECV",


narategithub:
@tom95858 github actions did not catch this b/c they did not build zap_fabric. I think I should put another build test that enable as many plugins as possible (with and without DEBUG) to catch these kinds of build bugs.

baallan:
@narategithub Why wouldn't the more appropriate change be to move the #if DEBUG from line 110 up to line 92? Seems like unused would be wrong in the case where debug is defined.

baallan:
> 
> 
> @tom95858 github actions did not catch this b/c they did not build zap_fabric. I think I should put another build test that enable as many plugins as possible (with and without DEBUG) to catch these kinds of build bugs.

@narategithub 
The tricky bit about this is to label the vm/container that has all the extra headers and libraries which support exotic build as "build only" in some way. Otherwise one is tempted to proceed on to running tests in the same container which has no corresponding hardware.

tom95858:
Hi @narategithub, why didn't our pull build test catch this?

narategithub:
@tom95858 b/c zap_fabric is disabled by default. That's why I mentioned that we need another build test that enables everything.

PR #526, opened 2020-11-20
Fix the dstat sampler plugin

nichamon:


nichamon:
Please note that the pull request addresses 2 problems: 

- the -Werror and -Wall build error on Ubuntu
- the bug that the sampler reports the wrong number of file descriptors. The counter didn't ignore the current and parent directories returned by `readdir_r`.

nichamon:
@tom95858 The patches have been tested on both Ubuntu18.04 and Centos7.
The configuration of the dstat plugin below
`
config name=dstat fd=true fdtypes=true ....
`
is used to configure the dstat plugin so that the changes would be tested.

There are no aggregators. The sampler daemon ran only the dstat plugin.

I compared the result from `ldms_ls -l` with the number of file descriptors in `/proc/self/fd`. The results of the code with and without the patch were compared as well. All metric values of the two code versions were the same except `fd_count`. The `fd_count` value before applying the patch counts the current and parent directories of `/proc/self/fd`.

baallan:
@nichamon When implementing, I considered the correct 'count' to be that which matches "ls -a /proc/self/fd/|wc -l" (which would include . and ..).
But if you adjust the man page for dstat to note that . and .. are excluded from the count, then this patch is fine (assuming that opendir result objects are unique per call, which I have no documentation to prove).

nichamon:
@baallan I answered your question in #521 here.

> @nichamon so if i readdir on /proc/$pid/fd simultaneously from two different threads (different plugins with overlapping functionality, which we may easily have happen), are the stream objects definitely different? Or is the same stream returned to both threads by opendir()?

`opendir` is thread-safe. It returns different stream objects. You can see the opendir manpage or the source code.

nichamon:
> @nichamon When implementing, I considered the correct 'count' to be that which matches "ls -a /proc/self/fd/|wc -l" (which would include . and ..).

The dstat man page defines `fd_count` as

https://github.com/ovis-hpc/ovis/blob/f8523cbffa5c8cf07378b4344d783433d873afe6/ldms/src/sampler/dstat/Plugin_dstat.man#L98

I think there is nothing to change on the man page. The patch I made fixes the code so that it reports the `fd_count` value according to the dstat man page which provides a sensible definition of the `fd_count` metric. In your opinion, are '.' and '..' open file descriptors?

baallan:
> 
> `opendir` is thread-safe. It returns different stream objects. You can see the opendir manpage or the source code.

The man page on rhel7 says nothing about the returned object being unique. But if you've verified that in the opendir source code, then we're all good.

nichamon:
> > `opendir` is thread-safe. It returns different stream objects. You can see the opendir manpage or the source code.
> 
> The man page on rhel7 says nothing about the returned object being unique. But if you've verified that in the opendir source code, then we're all good.

Yes, I have verified the source code.

PR #528, opened 2020-11-20
Fix slurm configure

narategithub:
This pull request addresses the following issues.

When Slurm is not installed in a system default directory (`/usr`), and `--with-slurm=<SLURM_PREFIX>` is given to the configure script, the build system will silently ***not*** build plugins that depend on slurm because:

1. `SLURM_CFLAGS` includes the installation prefix (`withval`) instead of prefix/include (`withval/include`). `have_slurm` will end up being `no`.
2. There is a typo in the condition checking before printing an error that makes it not printing the error.

In addition, SLURM_CFLAGS is not exported to automake -- making the build failed later even though 1 and 2 is fixed.

So far, I have not experienced this because `ovishpc/ovis-centos-build` dockerhub image has slurm installed in the default system directories. Building on a system that has slurm installed in /opt/slurm exposes the issues.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -430,7 +430,7 @@ AC_ARG_WITH([slurm],
 	[AS_HELP_STRING([--with-slurm],
 			[support SLURM jobid information and more @<:@default=check@:>@])],
 			[AS_IF([test "x$withval" != xyes -a "x$withval" != xno],
-				[SLURM_CFLAGS="-I $withval"])],
+				[SLURM_CFLAGS="-I $withval/include"])],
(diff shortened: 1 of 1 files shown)


tom95858:
Hi @narategithub, This fix is already in post-ga-fixes branch. I'll post this soon for review.

narategithub:
@tom95858 Gotcha. I'll close this when patches from post-ga-fixes arr posted. Thanks!

tom95858:
I believe this is fixed in OVIS-4 top of tree

PR #530, opened 2020-11-24
zap_rdma omnipath workaround

narategithub:
omnipath evidently does not deliver all completions. This series of commits contain zap_rdma workaround for omnipath and zap_rdma debug log improvements introduced when debugging zap_rdma over omnipath.

### DIFF
--- a/lib/src/zap/rdma/zap_rdma.c
+++ b/lib/src/zap/rdma/zap_rdma.c
@@ -253,17 +253,74 @@ static int __enable_cq_events(struct z_rdma_ep *rep)
 	/* handle CQ events */
 	struct epoll_event cq_event;
 	cq_event.data.ptr = rep;
-	cq_event.events = EPOLLIN | EPOLLOUT;
+	cq_event.events = EPOLLIN;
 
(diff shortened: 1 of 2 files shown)


tom95858:
I would like this to be tested at "scale" on real Infiniband (i.e. not Pathscale), and iWarp. This change is fairly large and I'm concerned that we might regress OFA compliant providers.

narategithub:
Sure.

narategithub:
@tom95858 I experienced a dead lock inside `librdmacm.so:rdma_destroy_id()` on both iWarp and IB hardware. I'm closing this pull request + debugging it. I'll re-open this again when it passed the scale test.

narategithub:
- This has passed a sampler+L1 flap test at scale on Mellanox IB and Chelsio iWarp hardware.
- Still waiting for omnipath test results.

narategithub:
@tom95858 From the latest logs that @baallan sent to me, I believe this pull request is good to go. The passive-side (samplers) memory leak that happened in both zap_rdma and zap_sock cases are a separate issue that is addressed in #548.

PR #532, opened 2020-11-25
Fix zap.c so that LDMS processes can run as daemons on Ubuntu20.04

nichamon:
When an LDMSD starts on Ubuntu20.04 as a daemon, the process stuck after
the demonization process starts. It turns out that it is a deadlock when
glibc is handling the child hook registered in zap.c by calling
pthead_atfork(). In zap.c, cs_init() calls pthread_atfork() and
registers cs_init() itself as the child hook handler.

### DIFF
--- a/lib/src/zap/zap.c
+++ b/lib/src/zap/zap.c
@@ -994,18 +994,14 @@ void zap_thrstat_free_result(struct zap_thrstat_result *res)
 	free(res);
 }
 
-#ifdef NDEBUG
-#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
-#endif
-static void __attribute__ ((constructor)) cs_init(void)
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 Some background ... I upgraded my desktop to Ubuntu-20.04. When I run LDMSD from the OVIS-4 branch, ldmsd didn't start properly. I installed the glibc debugging symbols so I can see the steps in the gdb session. I also compared the source code of glibc in Ubuntu-20.04 and Ubuntu-18.04 (The version I had before I upgraded my machine).

The problem isn't just on my desktop. I saw the same problem on docker containers based on Ubuntu-20.04.

nichamon:
There are more problems with Ubuntu 20.04. I'm closing this pull request and will open a new one when I sort everything out.

PR #534, opened 2020-11-30
Various fixes to the OVIS-4 GA

tom95858:
These fixes resolve build, transport, sampler and ldmsd issues with the OVIS-4 GA release

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -123,6 +123,11 @@ ldms_t ldms_xprt_get(ldms_t x)
 	return x;
 }
 
+static int ldms_xprt_connected(struct ldms_xprt *x)
+{
+	assert(x && x->ref_count);
+	return (x->disconnected == 0 && x->zap_ep && zap_ep_connected(x->zap_ep));
+}
 LIST_HEAD(xprt_list, ldms_xprt) xprt_list;
 ldms_t ldms_xprt_first()
 {
@@ -350,6 +355,8 @@ static void send_dir_update(struct ldms_xprt *x,
 	size_t json_data_len;
 	struct ldms_reply *reply = NULL;
 
+	if (!ldms_xprt_connected(x))
+		return;
 	assert(t != LDMS_DIR_LIST);
 
 	hdr_len = sizeof(struct ldms_reply_hdr)
@@ -407,6 +414,9 @@ static void send_req_notify_reply(struct ldms_xprt *x,
 	size_t len;
 	int rc = 0;
 	struct ldms_reply *reply;
+
+	if (!ldms_xprt_connected(x))
+		return;
 
 	len = sizeof(struct ldms_reply_hdr) + e->len;
 	reply = malloc(len);
@@ -1439,9 +1449,8 @@ out:
 int __ldms_remote_update(ldms_t x, ldms_set_t s, ldms_update_cb_t cb, void *arg)
 {
 	assert(x == s->xprt);
-	if (x->disconnected || !zap_ep_connected(x->zap_ep)) {
+	if (!ldms_xprt_connected(x))
 		return ENOTCONN;
-	}
 
 	if (!s->lmap || !s->rmap)
 		return EINVAL;
@@ -1488,6 +1497,9 @@ int ldms_xprt_recv_request(struct ldms_xprt *x, struct ldms_request *req)
 {
 	int cmd = ntohl(req->hdr.cmd);
 	int rc;
+
+	if (!ldms_xprt_connected(x))
+		return ENOTCONN;
 
 	switch (cmd) {
 	case LDMS_CMD_LOOKUP:
@@ -2947,6 +2959,9 @@ int ldms_xprt_send(ldms_t _x, char *msg_buf, size_t msg_len)
 	struct ldms_context *ctxt;
 	int rc;
 
+	if (!ldms_xprt_connected(x))
+		return ENOTCONN;
+
 	assert(msg_len > 4);
 	if (!msg_buf)
 		return EINVAL;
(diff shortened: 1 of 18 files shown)


tom95858:
@morrone, @baallan: could you guys please take a look at this. The change is large and overlaps with areas that you are involved with.

If you get the time to pull, build and test that would be greatly appreciated.

baallan:
@tom95858 will try to do this for cts1x and stria by wednesday.

m4/lib-ld.m4
```diff
@@ -0,0 +1,119 @@
+# lib-ld.m4 serial 6
```
baallan, line 1: sucking in get-text-devel bits means we have to track developments in libtool ourselves; this has bitten my butt in previous projects more often than tracking gettext has. Has this change been tested (with --disable-rpath active) on rhel8.x?

baallan:
@tom95858 do I need to also apply any of the following when testing this:
#497 #499 #530 #532, particularly 530?

ldms/src/core/ldms_xprt.c
```diff
@@ -350,6 +355,8 @@ static void send_dir_update(struct ldms_xprt *x,
 	size_t json_data_len;
 	struct ldms_reply *reply = NULL;
 
+	if (!ldms_xprt_connected(x))
```
nichamon, line 358: @tom95858 Why do we test whether the state is connected here? Is it to prevent passing NULL zap endpoint to zap_send() at line 397?

nichamon (commented):
@tom95858 All changes in ldmsd_controller, ldmsd_request.py and ldmsd_request.c look good. I have tried daemon_status, thread_status and xprt_status in ldmsd_controller and they all look good. I have a question regarding the check whether the LDMS transport is connected in various places in ldms_xprt.c. Please see my in-line comment.

ldms/src/core/ldms_xprt.c
tom95858: The point is to avoid submitting a bunch of I/O on an endpoint that is closing. This is really a uGNI 'work-around'. Because we defer disconnect on the uGNI transport, the daemon will continue to submit requests to an endpoint that is effectively closed. This causes many PostRdma requests to be submitted on an endpoint (and associated CQ) that leak descriptors due to a bug in uGNI. So the idea is to detect this as early as possible and minimize the # of descriptors that get leaked.

It may well be that this logic is no longer required given that we now track (or try to) leaked descriptors and opportunistically re-create the CQ.

nichamon: Got it! Thanks for the explanation.

m4/lib-ld.m4
tom95858: Please test as you see fit. Absent this, we need to add gettext-devel  which is a bit odd IMO since that is a locale message translation service.

tom95858:
> @tom95858 do I need to also apply any of the following when testing this:
> 
* #497 - no
* #499 - no
* #530 - yes, but I believe that @narategithub  is refactoring this due to some issues he found on IB and iWARP
* #532 - no

baallan:
@tom95858 @narategithub I'm testing the following single-node, multidaemon scenario for sock transport (centos7):
Create N-L0s. Create P-L1s, each aggregating all L0s. Create 1 L2 aggregating from all L1s. (yes, expecting redundant sets handling).
Flap the L0s. No evident leak of FDs as determined with ls proc/pid/fd from L1 or L2.
Flap the L1s. Evident leak of socket FDs as determined with ls proc/pid/fd from L2, no apparent fd leak from L0s.

Attaching logs of ls (l2.txt.tar.gz) and log of L2 (1.txt.gz).
[L2.txt.tar.gz](https://github.com/ovis-hpc/ovis/files/5632588/L2.txt.tar.gz)
[1.txt.gz](https://github.com/ovis-hpc/ovis/files/5632600/1.txt.gz)

tom95858:
Hi @baallan,

> @tom95858 @narategithub I'm testing the following single-node, multidaemon scenario for sock transport (centos7):
> Create N-L0s. Create P-L1s, each aggregating all L0s. Create 1 L2 aggregating from all L1s. (yes, expecting redundant sets handling).
> Flap the L0s. No evident leak of FDs as determined with ls proc/pid/fd from L1 or L2.
> Flap the L1s. Evident leak of socket FDs as determined with ls proc/pid/fd from L2, no apparent fd leak from L0s.
>

Thank you for performing this test. This is not a supported configuration and the daemon is leaking where it would once just assert. We'll take a look at this, but I don't consider it a stop-ship issue.

tom95858:
@baallan, could you please repeat the test, but with a supported configuration? I am running this test on Voltrino and don't see the issue.

tom95858:
Hi @baallan,  I think what you are seeing are sockets that are in the CLOSE_WAIT state (according to my testing of this). This occurs when the peer fails to respond to the final TCP FIN that is occurring because the peer was killed. The TCP protocol will wait 2MSL (two trips around the planet) which is a few minutes before cleaning up that fd. You can confirm this by using lsof -p <pid> instead of ls /proc/<pid>/fd | wc -l.

baallan:
@tom95858 are the attached from netstat and lsof consistent with your theory?
I'm not seeing any CLOSE_WAIT states, but if the expected close sequence is order minutes maybe it hasn't gotten to close_wait yet?
[ldmsd.pid.1.ns.1607016759.gz](https://github.com/ovis-hpc/ovis/files/5638218/ldmsd.pid.1.ns.1607016759.gz)
[ldmsd.pid.1.lsof.1607016759.gz](https://github.com/ovis-hpc/ovis/files/5638219/ldmsd.pid.1.lsof.1607016759.gz)

Also, can you define what you mean here by 'supported configuration'? A cts1/toss machine? else?

morrone (commented):
@tom95858 , you might consider replacing this commit (the "Add gettext autoconf macros" one) with PR #544. I believe that gnulib is the actual upstream source of these files, not gettext. gettext is probably just one of the projects that includes m4 files from gnulib. I took the approach in that PR of using the gnulib-tool command and including the resulting files. I think that might be the more "correct" approach to take longer term, and will make updating these files in the future from the known upstream source a tiny bit easier.

That also removes the dependency on gettext-devel.

PR #536, opened 2020-11-30
fix information leak in sock plugin

baallan:
struct sock_msg_read_resp has an alignment hole after 'status'.
This change fixes the information leak to network of that hole (memset)
and the uninitialized dst_ptr space.


baallan:
@tom95858 apparently I can't request you review this PR with the web ui.
Don't know if something about github is currently broken or I just fell off some privilege list.

In any case, found a minimalist way to stop one of the information leaks in sock and made this pr.

PR #537, opened 2020-11-30
add define for sampler base attributes.

baallan:
This lets plugins apply attribute config filters without duplicating
a string list that can get out of date as the base evolves.
It also documents the list in code form instead of comments.
The list is the union of the attributes actually processed
in sampler_base.c and the ones reserved in the documentation
of sampler_base.h.

### DIFF
--- a/ldms/src/sampler/dstat/dstat.c
+++ b/ldms/src/sampler/dstat/dstat.c
@@ -418,16 +418,8 @@ static const char *usage(struct ldmsd_plugin *self)
 }
 
 static const char *dstat_opts[] = {
+	BASE_ATTRIBUTES,
 	"debug",
-	"schema",
-	"instance",
-	"producer",
-	"component_id",
-	"uid",
-	"gid",
-	"perm",
-	"job_set",
-	"set_array_card",
 	"io",
 	"stat",
 	"statm",
--- a/ldms/src/sampler/sampler_base.h
+++ b/ldms/src/sampler/sampler_base.h
@@ -119,6 +119,25 @@ typedef struct base_data_s {
 
 #define BASE_CONFIG_USAGE BASE_CONFIG_SYNOPSIS BASE_CONFIG_DESC
 
+/** Define the list of reserved attributes plugins derived from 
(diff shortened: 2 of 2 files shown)


baallan:
@tom95858 @narategithub I need this merged to sustainably remove cloned code in dstat.c  and avoid creating similar code in other samplers.

tom95858:
@baallan, The way this is coded, it duplicates the string list (if not the string constant itself, which is compiler optimization dependent).

Aren't these options processed the sampler_base code? Why is your dstat sampler doing with these values that isn't done within sampler base?

That said, it I can see how it would be useful to know what attributes are handled in sampler_base; if for no other reason than to avoid collisions. But to do that you would export a global array of strings, not a #define of string constants that are replicated everywhere they are used.

baallan:
@tom95858 The question that needs to be answered (for admin-friendliness) by each sampler config() function is:
"Is there an option that we don't recognize?" (e.g. spelling errors, retired formerly deprecated options).

There are multiple ways to crack that nut. It would be a mere matter of programming to add use of the plugattr checker routines into sampler base and have the samplers pass their list of sampler-specific attributes into the base, have the base merge the lists and do the check.  But if we're counting every string constant, then we also need to find a new approach to log statements that doesn't blow up the global string list.

Since my understanding is that there's a major refactor of sampler_base coming anyway, I went with the least intrusive solution for now. But if we can do something that is more directionally correct, that's acceptable. I would like to see nearly all of sampler_base become more "mix in"  (as is the case with the perm/uid/gid auth config function)  than strict "OO", if your aim is to strictly minimize things like the size of the global string constants table.

baallan:
@tom95858 so how would you propose the base check its options and the derived check its options if neither knows the other's option list and the definition of check is: "is this an attribute I don't recognize as a valid exact spelling match"?
Somewhere, someone has to see the entire valid attribute list, else each party thinks the other parties attributes are invalid.

baallan:
@tom95858 so If I change this to a global array as you suggested on 12/14 are you good to merge it?

PR #542, opened 2020-12-07
Prevent SIGPIPE in zap_sock

narategithub:
zap_sock used `write()` to write messages to the socket. It is possible
to write to a disconnected socket, and when it happened SIGPIPE was
sent to the calling process and the process terminated (default SIGPIPE
action).

Instead of `write()`, this patch uses `send()` with `MSG_NOSIGNAL` to
disable the SIGPIPE on disconnected socket.

### DIFF
--- a/lib/src/zap/sock/zap_sock.c
+++ b/lib/src/zap/sock/zap_sock.c
@@ -130,17 +130,6 @@ static pthread_mutex_t z_key_tree_mutex;
 static LIST_HEAD(, z_sock_ep) z_sock_list = LIST_HEAD_INITIALIZER(0);
 static pthread_mutex_t z_sock_list_mutex = PTHREAD_MUTEX_INITIALIZER;
 
-static int __set_sockbuf_sz(int sockfd)
-{
-	int rc;
-	size_t optval = SOCKBUF_SZ;
-	rc = setsockopt(sockfd, SOL_SOCKET, SO_SNDBUF, &optval, sizeof(optval));
-	if (rc)
-		return rc;
-	rc = setsockopt(sockfd, SOL_SOCKET, SO_RCVBUF, &optval, sizeof(optval));
-	return rc;
-}
-
 static int z_rbn_cmp(void *a, const void *b)
 {
 	uint32_t x = (uint32_t)(uint64_t)a;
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 this shall fix the SIGPIPE issue I asked you in the e-mail.

baallan:
I have been seeing the sigpipe in my testing. is this something that is on (or will be added) to the branch I'm testing?

narategithub:
@baallan this is not in the branch you're testing (narategithub/OVIS-4-fix). But, I thought you're testing zap_rdma over omnipath with that branch.

baallan:
@narate it's easy to run both at the same time and compare.

narategithub:
@baallan I'd just added this in narategithub/OVIS-4-fix. Please fetch & merge. It was not there originally b/c I just have this fix yesterday.

lib/src/zap/sock/zap_sock.c
```diff
@@ -1161,14 +1161,16 @@ static void sock_write(ovis_event_t ev)
...
 	/* msg */
 	while (wr->msg_len) {
-		wsz = write(sep->sock, wr->msg + wr->off, wr->msg_len);
+		wsz = send(sep->sock, wr->msg + wr->off, wr->msg_len, MSG_NOSIGNAL);
 		if (wsz < 0) {
 			if (errno == EAGAIN || errno == EWOULDBLOCK) {
 				__enable_epoll_out(sep);
 				goto out;
 			}
+			if (errno == EPIPE)
```
tom95858, line 1170: EPIPE is telling you it's EOF, so why are you skipping the shutdown call?

narategithub: Because I think that `shutdown()` will eventually be called when we handle EPOLLHUP. Calling `shutdown()` on an disconnected socket will just return ENOTCONN. Shall I change the code to call `shutdown()` in this case too?

baallan:
@narategithub if you could review and include 536 545 546 547 fixes found in yesterday and previous testing rounds, that would make seeing newer problems easier.

tom95858:
__sock_send_msg_nolock used to always return ZAP_ERR_OK; so this adds quite a few new error paths. I think a far simpler fix is to add this:

```
int set = 1;
setsockopt(sd, SOL_SOCKET, SO_NOSIGPIPE, (void *)&set, sizeof(int));
```

Looking at this bit of code I notice that we have three functions for setting socket options:

```
     rc = __set_sockbuf_sz(sep->sock);
     rc = __sock_nonblock(sep->sock);
     rc = __set_keep_alive(sep);
```
The reason this is bad is because someone could forget to add on of these calls to a path and we get inconsistent options on the active vs. passive side. In fact we _did_ do that. _z_sock_conn_request is missing the call to __set_keep_alive.

So could we please refactor this change to have all the socket options set in one function, add the new SO_NOSIGPIPE option and call this one function from __z_sock_conn_request and z_sock_connect?

Also leave the call to _write_ as is; as well as the return code from sock_write so that we don't add a bunch of new error paths?

narategithub:
@tom95858 OK, I'll do that. Where did you find SO_NOSIGPIPE? I searched PIPE in `socket(7)`, `tcp(7)` and `setsockopt(2)` and could not find it.

narategithub:
@tom95858 I don't think there is a `SO_NOSIGPIPE` in Linux. Here's the uapi/asm-generic/socket.h in Linus's linux master branch on github: https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/socket.h

How shall I proceed?

narategithub:
@tom95858 I think this is ready to merge. I also scale-flap test on cygus, but let it run just about 10 min, and all look fine. Could you please review it? Thanks!

narategithub:
@tom95858 just a reminder. Please don't forget to include this in the release.

PR #544, opened 2020-12-08
Import m4 files from gnulib instead of gettext

morrone:
Import m4 files from gnulib instead of gettext

gettext is not the true source of config.rpath,, lib-ld.m4,
lib-link.m4, and lib-prefix.m4. gnulib is the true source of those
files. This commit updates ovis to get the files from the true origin
rather than getting them indirectly from another project.

This approach both gets us more up-to-date files right now and also
give us a cleaner documented approach for updating an maintaining
those file in the future.

The gnulib-tool command from the gnulib project is used to import
the current latest m4 files in support of AC_LIB_HAVE_LINKFLAGS.
This also has the effect of including the AC_LIB_RPATH macro.

This was run when the gnulib project was at commit
e649a3bbd4be8a6af0db7275f6600bb5bed1ffb7 in its repository.

The following command was employed:

      gnulib-tool --no-vc-files --source-base=gnulib --import havelib


baallan:
@tom95858 @morrone can we figure out how to make the current set of github  checks run on this and the decide if we're taking it?

@morrone is there a little shell script involving "gnulib-tool --..." you could add to ovis/util/ to let a developer easily update what you have done when the need arises? Or is that somehow already there and I'm just not seeing it?

morrone:
m4/gnulib-cache,m4 contains the details of the details of how gnulib was configured, and a command line version to reproduce what was done in the future. I am reluctant to try to make this easier, because gnulib-tool is not a normal system-installed package. It is always something that needs to be checked out and used from its source tree. It is probably better that the person updating the gnulib stuff understand what they are doing. If copy-and-paste of the command from that file is a significant barrier, then they might not know what they are doing with gnulib in general.

I rebased this PR so all of the checks are now run against the head of OVIS-4 at this time.

PR #545, opened 2020-12-08
fix ldms_xprt remote memory leak in send_dir_update

baallan:
fix addresses one of the problems in #453

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -375,6 +375,7 @@ static void send_dir_update(struct ldms_xprt *x,
 	reply->hdr.xid = x->remote_dir_xid;
 	reply->hdr.cmd = htonl(LDMS_CMD_DIR_UPDATE_REPLY);
 	reply->hdr.rc = 0;
+	reply->dir.more = 0;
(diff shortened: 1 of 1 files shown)


ldms/src/core/ldms_xprt.c
```diff
@@ -375,6 +375,7 @@ static void send_dir_update(struct ldms_xprt *x,
 	reply->hdr.xid = x->remote_dir_xid;
 	reply->hdr.cmd = htonl(LDMS_CMD_DIR_UPDATE_REPLY);
 	reply->hdr.rc = 0;
+	reply->dir.more = 0;
```
narategithub, line 378: It doesn't hurt to set the flag to 0, but upon code inspection the `more` flag is ignored in the `DIR_UPD` message.

narategithub:
@tom95858 I think this is a good patch.

ldms/src/core/ldms_xprt.c
baallan: yes. this is an information leak patch, not a "functional" change. Thank you for confirming that more is unused at the other end; I thought so but it's good to have a second opinion on network code.

PR #546, opened 2020-12-08
fix local use of uninit data in rdma logging

baallan:


lib/src/zap/rdma/zap_rdma.c
```diff
@@ -1099,6 +1099,7 @@ static int cq_event_handler(struct ibv_cq *cq, int count)
 {
 	struct ibv_wc wc;
 	int rc, poll_count = 0;
+	memset(&wc, 0, sizeof(wc));
```
tom95858, line 1102: @baallan this doesn't do anything. A work completion does not go on the wire, it is a local data structure.

baallan: ok mislabeling. apologies for the confusion. it does go into log formatting which causes use of uninit data to control output write logic.

PR #548, opened 2020-12-09
Put back the RBD share_lookup reference when disconnected

nichamon:
LDMS takes the share_lookup references on RBDs and sets when it responds to a
lookup request. Without this patch, LDMS puts the references back only when
applications call ldms_set_delete(). Thus, LDMS leaks the RBDs and sets that
have been looked-up. The leak also affects LDMSDs when it respond to an ldms_ls
-l request.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -690,11 +690,15 @@ static void __destroy_set(void *v)
 	pthread_mutex_unlock(&__del_tree_lock);
 }
 
+void __put_share_lookup_ref(ldms_set_t rbd)
+{
+	ref_put(&rbd->set->ref, "share_lookup");
+	ref_put(&rbd->ref, "share_lookup");
+}
+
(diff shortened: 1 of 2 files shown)


nichamon:
@baallan found a leak in the Valgrind logs of sampler daemons. @narategithub pointed me to the leak upon his inspection.
After I tested OVIS-4 + post-ga-fixes, I found that the references taken on RBDs and sets (share_lookup) in send_lookup_reply are not put back in the case of disconnection. They are put back only when the sampler daemon receives the SET_DELETE reply.

I have tested the patch and confirmed that the references are put back when the connection between the sampler daemon and the aggregator is disconnected. After the disconnection, I also terminated the plugins to delete the sets. All RBDs associated with the sets were freed and the sets were freed. However, I haven't tested the patch at scale.

baallan:
@nichamon thanks for the quick fix. @narategithub do I assume correctly that a revised OVIS-4-fix with this weeks related patches all included will be out soonish for me to retest?

baallan:
in today's testing of ovis-4-fix the symptom this patch addresses is no longer present.

PR #549, opened 2020-12-10
remove buggy ldmsd_msg_logger function

baallan:
the ldmsd_msg_logger function:
is not thread safe
is passed to all the plugins that take a log pointer
contains an easily exploitable buffer overrun
is redundant with ldmsd_log.

This patch removes the function and redirects the only caller to ldmsd_log.
It leaves a #define for any client code that might have used it.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -277,15 +277,6 @@ LDMSD_LOG_AT(LDMSD_LERROR,error);
 LDMSD_LOG_AT(LDMSD_LCRITICAL,critical);
 LDMSD_LOG_AT(LDMSD_LALL,all);
 
-static char msg_buf[4096];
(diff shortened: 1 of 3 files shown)


baallan:
@nichamon please tell us if there is some use case for ldmsd_msg_logger that I overlooked.

nichamon:
> @nichamon please tell us if there is some use case for ldmsd_msg_logger that I overlooked.

The result of grep from the top of the OVIS-4 branch:

```
$ grep -rn 'ldmsd_msg_logger' ldms/*
ldms/src/ldmsd/ldmsd.c:281:void ldmsd_msg_logger(enum ldmsd_loglevel level, const char *fmt, ...)
ldms/src/ldmsd/ldmsd_config.c:175:      lpi = pget(ldmsd_msg_logger);
ldms/src/ldmsd/ldmsd.h:814:void ldmsd_msg_logger(enum ldmsd_loglevel level, const char *fmt, ...);
```
I agree that there is only one caller as you mentioned.

PR #550, opened 2020-12-10
Fix a possible use-after-freed in the cleanup path

nichamon:
The cleanup logic iterates through the storage policy tree to close all
stores. It also puts back a storage policy reference, which is
erroneous. It is unnecessary to put the reference back as we only want
to close the stores and not free the storage policies. Moreover, this
could cause a use-after-free if all producer sets associated with a
storage policy have been deleted. The storage policy will be freed when
the cleanup logic puts back the reference, and then the cleanup logic
reaccesses the storage policy to get the next storage policy.


nichamon:
@baallan found the use-after-free issue at cleanup while LDMSD was closing the stores after LDMSD was disconnected from its producers.

I have confirmed the use-after-free issue.

baallan:
I tested Narates fix branch that included this patch and I no longer get the use-after errors.

PR #552, opened 2020-12-11
fix buffer overrun in zap event queue definition.

baallan:
The thread name buffer is too small for the potential names, if thread limit
from user is high. This could potentially corrupt event queue setup in init_atfork, depending on stack layout determined by compiler.

### DIFF
--- a/lib/src/zap/zap.c
+++ b/lib/src/zap/zap.c
@@ -1010,8 +1010,8 @@ static void init_atfork(void)
 	assert(zev_queue);
 
 	for (i = 0; i < zap_event_workers; i++) {
-		char thread_name[16];
-		(void)snprintf(thread_name, 16, "zap:worker:%d", i);
+		char thread_name[24];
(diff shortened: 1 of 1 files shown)


tom95858:
There is no possibility of corruption because snprintf is being used. Bumping the buffer size does not accomplish anything because the maximum pthread name is 16

baallan:
@tom95858 in the light of day (and your comment), yes, snprintf prevents the overrun. But it doesn't prevent the incorrect name truncation of the stats. Based on the name format string, the maximum number  that fits in the name is '9999'. On the large systems coming down the pipe, this is considerably smaller than what might be read from the user via ZAP_EVENT_WORKERS env. Being limited to 16 characters in the current format leads to incorrect report formatting for thread worker stats names.

Do I increase the allowed name length and retitle/describe the patch, or put in guard code that limits worker count to 9999, or shorten the name format prefix to z:wkr:?

PR #554, opened 2020-12-11
Fix zap_sock close() not shutdown

narategithub:
The `shutdown()` calls in zap_sock close implementation were
accidentally checked in, making socket connection stayed connected in
the case of bad authentication test (e.g. mismatched keyes or wrong
munge domain) in `ldmsd_auth_ovis_test` and `ldmsd_auth_test` tests in
`ldms-test` repository. Re-introducing the `shutdown()` calls solved the
issue.

### DIFF
--- a/lib/src/zap/sock/zap_sock.c
+++ b/lib/src/zap/sock/zap_sock.c
@@ -256,12 +256,12 @@ static zap_err_t z_sock_close(zap_ep_t ep)
 	case ZAP_EP_CONNECTED:
 	case ZAP_EP_LISTENING:
 		sep->ep.state = ZAP_EP_CLOSE;
-		// shutdown(sep->sock, SHUT_RDWR);
+		shutdown(sep->sock, SHUT_RDWR);
 		break;
 	case ZAP_EP_ERROR:
 	case ZAP_EP_ACCEPTING:
 	case ZAP_EP_CONNECTING:
-		// shutdown(sep->sock, SHUT_RDWR);
+		shutdown(sep->sock, SHUT_RDWR);
 		break;
 	case ZAP_EP_CLOSE:
 		break;
@@ -857,8 +857,6 @@ static int __recv_msg(struct z_sock_ep *sep)
 		rqsz = sizeof(struct sock_msg_hdr) - buff->len;
 		rsz = read(sep->sock, buff->data + buff->len, rqsz);
 		if (rsz == 0) {
-			if (errno == EAGAIN)
-				return errno;
 			/* peer close */
 			rc = ENOTCONN;
 			from_line = __LINE__;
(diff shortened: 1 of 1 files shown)


tom95858 (changes requested):
I think these shutdown calls are superfluous which is why I removed them. Can you just remove the check for EAGAIN in the rsz == 0 path in sock_read. I think that's the actual problem that causes the %100 CPU utilization when auth fails.

narategithub:
@tom95858 I did try having only the EAGAIN removal last night, leaving the `shutdown()` calls commented out. The result was that EPOLLHUP is never delivered, and hence zap disconnect event is never delivered. `ldms_ls` with bad auth just stuck there forever, but CPU utilization was not at 100% anymore. I'd just forced-pushed the patch to include EAGAIN removal in the case of read() returning 0.

lib/src/zap/sock/zap_sock.c
```diff
@@ -256,12 +256,12 @@ static zap_err_t z_sock_close(zap_ep_t ep)
...
 	case ZAP_EP_LISTENING:
 		sep->ep.state = ZAP_EP_CLOSE;
-		// shutdown(sep->sock, SHUT_RDWR);
+		shutdown(sep->sock, SHUT_RDWR);
 		break;
 	case ZAP_EP_ERROR:
 	case ZAP_EP_ACCEPTING:
 	case ZAP_EP_CONNECTING:
-		// shutdown(sep->sock, SHUT_RDWR);
+		shutdown(sep->sock, SHUT_RDWR);
```
tom95858, line 264: shouldn't the endpoint state be set to ZAP_EP_CLOSE here? Seems like there is a race with sock_event where the endpoint could be left in ZAP_EP_ACCEPTING/CONNECTING, etc...

narategithub: Let me look into it .. I don't wanna break the state machine in the last minute :worried:

narategithub: @tom95858 We do not move the state to ZAP_EP_CLOSE because at this point, the underlying socket already became `connected` and we expect `EPOLLHUP` from epoll. The EPOLLHUP will be processed in `sock_event()` function. If we move the endpoint state to `ZAP_EP_CLOSE`, `sock_event()` will incorrectly deliver `ZAP_EP_DISCONNECTED` event, and that breaks zap event design (DISCONNECTED must only be delivered if CONNECTED has been delivered). By not moving the state in these cases, `sock_event()` knows that the underlying socket terminated while the zap endpoint is in connecting/accepting states and then delivers corresponding CONNECT_ERROR events.

tom95858:
Hi @narategithub, this breaks streams.

narategithub:
@tom95858 Could you please point me to your stream test? So that I can test/debug with `ldms-test` in one go?

tom95858:
publisher:
```
ldmsd_stream_publish -h nid00062 -x sock -a munge -p 12345 -f ffile -s test_stream
```
subscriber:
```
ldmsd_stream_subscribe -p 12345 -s test_stream -a munge
```
ffile contains the data to publish. Mine contains this:
```
1
2
3
4
5
```

narategithub:
Thanks! I'll debug this tonight.

tom95858:
You might experiment with SHUT_WR instead of SHUT_RDWR

PR #556, opened 2020-12-15
Fix rdma zap endpoint use-after-free

narategithub:
zap_rdma ignore TIMEWAIT_EXIT event as the endpoints are destroyed
since DISCONNECTED or ERROR events are handled, and we don't want to
wait around to free resources. Hence, if the event is TIMEWAIT_EXIT,
rep could already be gone and we shall not access it.


narategithub:
@tom95858 please don't merge this yet. It need to test over night.

PR #557, opened 2020-12-15
Fix rbd use after free

narategithub:
The releasing of rbd in __ldms_xprt_resource_free() could race with
ldms_xprt_set_delete(). In such case, __ldms_xprt_resource_free()
released the xprt lock, and then ldms_xprt_set_delete() took the lock
and call __ldms_rbd_xprt_release() which removed the rbd from the tree
and drop a reference. That reference could be the last reference to
the rbd, and hence once ldms_xprt_set_delete() released the lock and
__ldms_xprt_resource_free() took the lock again, the rbd would already
be freed -- causing a use-after-free issue.

This patch solved the issue by taking the rbd reference while the rbd
is still in the tree and xprt lock is still hold, and drop the
reference after we're done processing it.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -600,6 +600,7 @@ void __ldms_xprt_resource_free(struct ldms_xprt *x)
 	int drop_ep_ref = 0;
 	while ((rbn = rbt_min(&x->rbd_rbt))) {
 		rbd = RBN_RBD(rbn);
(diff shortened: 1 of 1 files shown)


tom95858:
@narategithub Did you see a crash from this, or is this by inspection.

narategithub:
@tom95858 I saw a crash on this on cygnus. I am also running scale-flap test on cygnus as well.

PR #558, opened 2020-12-15
Fix zap_rdma submit_pending() deadlock

narategithub:
submit_pendin() took endpoint lock to free the context while holding
credit lock, while the read/write/send took endpoint lock then credit
lock. This created a dead lock. submit_pending() should re-acquire the
credit lock after it is done freeing context and release the endpoint
lock, which is what this patch does.


tom95858:
@narategithub should I accept this?

narategithub:
@tom95858 Yes, please.

FYI, the slow leak with RDMA in scale-flap test is still there. I'm still hunting it.

PR #559, opened 2020-12-16
add config check for new ibv_wc_t* enums

baallan:
This patch checks at configure for enums that exist on TOSS3 infiniband/verbs.h but not some mellanox ib library versions and ifdefs them out if they are not all defined. Affects at least ARM platforms.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -137,8 +137,13 @@ if test "$disable_zap" != "yes"; then
 dnl we need libibverbs-devel and librdmacm-devel to support rdma
   if test "$enable_rdma" = "yes"; then
     AC_CHECK_HEADER([infiniband/verbs.h],
-    [AC_DEFINE([HAVE_VERBS_H], [1],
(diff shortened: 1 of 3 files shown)


baallan:
@narategithub let us know if you have something better coming for this problem. this is the quickie I made to let build finish on stria.

tom95858:
Does the build test include --enable-rdma? If not, we should add the necessary OFA RPM to the container and do so.

baallan:
Looks like "Build Test with `-DDEBUG` - CentOS7"  includes --enable-rdma in .github/workflows.
The patch has been manually tested on both 'with' and 'without' systems. (cts1x, stria).

baallan:
Oddly, when I look at the log for the build from "Build Test with -DDEBUG - CentOS7" I don't see rdma being enabled, so something fishy with the yaml or github launcher.

baallan:
rdma is enabled in the "Build Test with -DDEBUG - CentOS7" build, based on a closer reading of the logs.

PR #560, opened 2020-12-17
Add build-with-rdma github action step

narategithub:
Add another github action step to test build with rdma but without DEBUG
in both CentOS and Ubuntu build test actions.

### DIFF
--- a/.github/workflows/build-test-centos7.yaml
+++ b/.github/workflows/build-test-centos7.yaml
@@ -26,3 +26,13 @@ jobs:
     - run: sh autogen.sh
     - run: ./configure
     - run: make distcheck
+
+  build-with-rdma:
+    runs-on: ubuntu-20.04
+    container:
+        image: ovishpc/ovis-centos-build
+    steps:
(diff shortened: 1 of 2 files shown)


tom95858:
I think we can get rid of the build tests that don't include --enable-rdma correct? They are redundant?

narategithub:
I think the one **without** `--enable-rdma` is to test the bare minimum test. The case of `--enable-rdma` is currently a superset of bare minimum as `--enable-rdma` just build more stuff. However, I remembered in the past that we all build with `--enable-SOMETHING` and we never get a build error, but when @valleydlr build without that option, his build failed. So, I think the bare build test is still needed.

baallan:
yes, please, we should have a bare './configure && make install && make installcheck' test or equivalent on each flavor of rhel/suse/centos/ubuntu that we can manage on github actions. Obviously some flavors may not be possible due to unavailability of containers due to OS licensing issues.   I think the bare ./configure test is part of doing a 'make distcheck' test, so maybe this is covered if we pass "make distcheck".

PR #563, opened 2020-12-18
Fix a share_lookup RBD reference leak in the ldms_set_delete path

nichamon:
The leak occurs when

* there is a race between the ldms_set_delete() path and the disconnected
path (Note that the disconnected path releases the transport lock, takes
the set lock, releases the set lock, and retakes the transport lock),
*the ldms_set_delete() path removes the leak RBD from the transport’s RBD
tree while the disconnected path handles another RBD, and
* LDMS fails to send the SET_DELETE request to the disconnected client,
i.e., zap_send() synchronously fails due to the disconnected connection.
The SET_DELETE context is freed upon the synchronous error.

LDMS releases the share_lookup RBD references either in the disconnected
path or in the SET_DELETE reply handling path. Since LDMS fails to send
the SET_DELETE request to the client, it will not receive the SET_DELETE
reply. Besides, the ldms_set_delete() path has removed all RBDs from all
associated LDMS transport’s RBD trees. Thus, the disconnected path has
no access to the RBDs. Therefore, LDMS did not release the share_lookup
RBD references in this scenario.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -547,7 +547,6 @@ static void __ldms_drop_rbd_set_refs(struct ldms_rbuf_desc *rbd)
 
 void __ldms_xprt_resource_free(struct ldms_xprt *x)
 {
-	extern void __put_share_lookup_ref(ldms_set_t rbd);
 	pthread_mutex_lock(&x->lock);
 	x->remote_dir_xid = x->local_dir_xid = 0;
 
@@ -609,6 +608,11 @@ void __ldms_xprt_resource_free(struct ldms_xprt *x)
 			pthread_mutex_unlock(&set->lock);
 			pthread_mutex_lock(&x->lock);
 		}
+		/* Make sure that we didn't lose a set delete race */
+		if (!rbd->xprt) {
+			ref_put(&rbd->ref, __func__);
+			continue;
+		}
 		if (rbd->type == LDMS_RBD_LOCAL) {
 			/*
 			 * The only LOCAL RBDs that are associated with a transport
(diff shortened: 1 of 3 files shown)


nichamon:
@tom95858 The patch has been tested with 2 test setups. Both setups use RDMA. The 1st-level aggregator is the daemon that Narate and I focus on for the share_lookup RBD reference leak.

- Narate's setup has been running for about 5 hours. It is still running.
sampler daemons are flapped
One 1st-level aggregator
One 2nd-level aggregator is not flapped.

- Nichamon's setup was stopped and checked the Valgrind result.
200 sampler daemons with 11 sets are flapped every 30 seconds.
One 1st-level aggregator was started with Valgrind.
One 2nd-level aggregator is flapped every 30 seconds.
The Valgrind result indicates that we are not leaking any RBDs anymore.

I believe that I didn't catch this leak when I tested my previous patch #548 because I built the binaries differently (I didn't turn on -D_REF_TRACK_ this time) and different transport (Last time I tested the patch in #548 with zap_sock).

ldms/src/core/ldms_xprt.c
```diff
@@ -223,6 +223,8 @@ ldms_t ldms_xprt_by_remote_sin(struct sockaddr_in *sin)
 	return 0;
 }
 
+extern void __put_share_lookup_ref(ldms_set_t rbd);
```
tom95858, line 226: By putting this code:
```
	ref_put(&rbd->set->ref, "share_lookup");
	ref_put(&rbd->ref, "share_lookup");
```
inside the __put_share_lookup_ref function, you lose the function name and line number where the reference was dropped.

tom95858 (changes requested):

ldms/src/core/ldms_xprt.c
nichamon: I will change the __put_shate_lookup_ref() function to be a #define. I wrapped the ref_put() calls into a function because both the share_lookup references of RBD and set must be put together.

nichamon:
@tom95858 I made the change from a function to a #define.

PR #573, opened 2020-12-28
Bind job "init"/"exit" events to Job Prolog/Epilog

tom95858:
This is an initial refactoring of event binding in the
Spank plugin. This will have to be augmented because some
values that are available in the previous slurmstepd
callbacks have a zero value at prolog time.


tom95858:
Converting to draft...

PR #577, opened 2020-12-29
Revert slurm_notifier commit

tom95858:


morrone:
Please keep the standard git revert message, which would have been:

```
Revert "Fix Slurm step handling in Spank plugin"

This reverts commit 4a28e2c73b1409ef8378785729780c88d62d50b0.
```

For those of us that are fluent in git, that message format makes it possible to instantly identify which commit in the history is being reverted.

You can add additional commentary after the standard message if so desired.

morrone:
PR #579 has the correct commit message.

morrone (changes requested):
Use git revert's commit message.

PR #580, opened 2020-12-29
Fix build-test-ubuntu.yaml workflow

morrone:
The workflow had two jobs with the same name, which hid the fact
that the rdma build on the ubuntu docker container was not working.

Fix the name of the rdma job.


morrone:
@narategithub, does the ovishpc/ovis-ubuntu-build container have infiniband packages installed? The build is not passing when I fix the name of the job that uses "configure --enable-rdma".

Perhaps we should just remove this job from the ubuntu workflox rather than fixing the job name?

narategithub (approved):
:+1:

PR #583, opened 2021-01-02
Refactor Slurm notifier and sampler event handling

tom95858:
slurm_notifier Spank plugin changes

* Bind "init"/"exit" events to prolog/epilog spank callbacks
  * PrologFlags=Alloc will cause prolog events to be delivered
    immediately as opposed to at the 1st job step. This is
    particularly useful for ```salloc -N x bash``` jobs. 
  * At prolog time NTASKS and NNODES is not available and
    therefore before the first job step is launched, the
    nnodes and ntasks values are zero

* Add "step_init"/"step_exit" events
  * These events contain accurate values for ntasks and nnodes

slurm_sampler changes

* Sampler binds job start/end times to the "init"/"exit"
  events
* Job step start/end events do not modify the job start
  and end times
* Caveats:
  * Job task PID are updated at each job step and overwrite
    any previous value
  * Job size begins at zero at salloc time and is updated
    at each job-step

papi_sampler changes

* Remove chatty messages that result from papi_sampler data
  missing from subscriber_data

### DIFF
--- a/ldms/src/sampler/slurm/slurm_sampler.c
+++ b/ldms/src/sampler/slurm/slurm_sampler.c
@@ -720,6 +720,56 @@ static void handle_job_init(job_data_t job, json_entity_t e)
 	ldms_transaction_end(job_set);
 }
 
+static void handle_step_init(job_data_t job, json_entity_t e)
+{
+	int int_v;
+	json_entity_t attr, data, dict;
+
+	data = json_attr_find(e, "data");
+	if (!data) {
+		msglog(LDMSD_LERROR, "slurm_sampler: Missing 'data' attribute "
+		       "in 'init' event.\n");
+		return;
+	}
+	dict = json_attr_value(data);
+
+	ldms_transaction_begin(job_set);
+	attr = json_attr_find(dict, "nnodes");
+	if (attr) {
+		int_v = json_value_int(json_attr_value(attr));
+		ldms_metric_array_set_u32(job_set, node_count_idx, job->job_slot, int_v);
+	}
+
+	attr = json_attr_find(dict, "local_tasks");
+	if (attr) {
+		int_v = json_value_int(json_attr_value(attr));
+		ldms_metric_array_set_u32(job_set, task_count_idx, job->job_slot, int_v);
+	}
+
+	attr = json_attr_find(dict, "uid");
+	if (attr) {
+		int_v = json_value_int(json_attr_value(attr));
+		ldms_metric_array_set_u32(job_set, job_uid_idx, job->job_slot, int_v);
+	}
+
+	attr = json_attr_find(dict, "gid");
+	if (attr) {
+		int_v = json_value_int(json_attr_value(attr));
+		ldms_metric_array_set_u32(job_set, job_gid_idx, job->job_slot, int_v);
+	}
+
+	attr = json_attr_find(dict, "total_tasks");
+	if (!attr) {
+		msglog(LDMSD_LERROR, "slurm_sampler: Missing 'total_tasks' attribute "
+		       "in 'init' event.\n");
+		goto out;
+	}
+	int_v = json_value_int(json_attr_value(attr));
+	ldms_metric_array_set_u32(job_set, job_size_idx, job->job_slot, int_v);
+ out:
+	ldms_transaction_end(job_set);
+}
+
 static void handle_task_init(job_data_t job, json_entity_t e)
 {
 	json_entity_t attr;
@@ -763,8 +813,7 @@ static void handle_task_init(job_data_t job, json_entity_t e)
 	ldms_metric_array_set_u32(job_set, task_rank_idx + job->job_slot, task_id, int_v);
 
 	job->task_init_count += 1;
-	if (job->task_init_count == job->local_task_count)
-		ldms_metric_array_set_u8(job_set, job_state_idx, job->job_slot, JOB_RUNNING);
+	ldms_metric_array_set_u8(job_set, job_state_idx, job->job_slot, JOB_RUNNING);
  out:
 	ldms_transaction_end(job_set);
 }
@@ -851,7 +900,7 @@ static int slurm_recv_cb(ldmsd_stream_client_t c, void *ctxt,
 	pthread_mutex_lock(&job_lock);
 	rc = ENOENT;
 	if (0 == strncmp(event_name->str, "init", 4)) {
-		job = get_job_data(tstamp, job_id); /* protect against duplicate entries */
+		job = get_job_data(tstamp, job_id);
 		if (!job) {
 			uint64_t local_task_count;
 			attr = json_attr_find(dict, "local_tasks");
@@ -871,6 +920,15 @@ static int slurm_recv_cb(ldmsd_stream_client_t c, void *ctxt,
 			}
 			handle_job_init(job, entity);
 		}
+	} else if (0 == strncmp(event_name->str, "step_init", 9)) {
+		job = get_job_data(tstamp, job_id);
+		if (!job) {
+			msglog(LDMSD_LERROR, "slurm_sampler: '%s' event "
+			       "was received for job %d with no job_data\n",
+			       event_name->str, job_id);
+			goto out_1;
+		}
+		handle_step_init(job, entity);
 	} else if (0 == strncmp(event_name->str, "task_init_priv", 14)) {
 		job = get_job_data(tstamp, job_id);
 		if (!job) {
--- a/ldms/src/sampler/spank/slurm_notifier.c
+++ b/ldms/src/sampler/spank/slurm_notifier.c
@@ -140,6 +140,7 @@ static void msglog(const char *format, ...)
  * Events have the following syntax:
  *
  *   {
+ * 		"schema"    : <schema-name>,
  *      "event"     : <event-type>,
  *      "timestamp" : <timestamp>
  *      "context"   : <spank-context-name>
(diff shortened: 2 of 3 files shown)


ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -663,35 +775,45 @@ jbuf_t make_task_exit_data(spank_t sh, const char *event, const char *context)
...
-	jbuf_t jb;
-
-	if (0 == nnodes(sh))
-		/* Ignore events before node assignment */
-		return ESPANK_SUCCESS;
+	jbuf_t jb = NULL;
 
 	switch (context) {
+	case S_CTX_ALLOCATOR:
+		/* Called from sbatch/salloc which is likely running on the head node. */
```
morrone, line 785: I do not think that we should do anything at all in S_CTX_ALLOCATOR. I think the whole case statement can be removed and just be replace with a simple conditional on S_CTX_REMOTE.

ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -663,35 +775,45 @@ jbuf_t make_task_exit_data(spank_t sh, const char *event, const char *context)
...
+	if (jb) {
+		DEBUG2("%s", jb->buf);
+		send_event(argc, argv, jb);
+		jbuf_free(jb);
+	}
+	return ESPANK_SUCCESS;
+}
 
-	jb = make_init_data(sh, "init", context_str);
+int slurm_spank_slurmd_init(spank_t sh, int argc, char *argv[])
```
morrone, line 803: This hook, slurm_spank_slurmd_init, needs to be removed entirely.

ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -800,25 +892,31 @@ slurm_spank_task_exit(spank_t sh, int argc, char *argv[])
...
-	const char *context_str;
-	jbuf_t jb;
-
-	if (0 == nnodes(sh))
-		/* Ignore events before node assignment */
-		return ESPANK_SUCCESS;
+	jbuf_t jb = NULL;
 
 	switch (context) {
+	case S_CTX_ALLOCATOR:
```
morrone, line 899: Again, S_CTX_ALLOCATOR should not be used at all, and this case statement can be replaced with a conditional using S_CTX_REMOTE.

ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -612,7 +682,49 @@ jbuf_t make_exit_data(spank_t sh, const char *event, const char *context)
...
+	jb = _append_item_u16(sh, jb, "ncpus", S_JOB_NCPUS, ','); if (!jb) goto out_1;
+	jb = _append_item_u32(sh, jb, "nnodes", S_JOB_NNODES, ','); if (!jb) goto out_1;
+	jb = _append_item_u32(sh, jb, "local_tasks", S_JOB_LOCAL_TASK_COUNT, ','); if (!jb) goto out_1;
+	jb = _append_item_u32(sh, jb, "total_tasks", S_JOB_TOTAL_TASK_COUNT, ' '); if (!jb) goto out_1;
+	jb = jbuf_append_str(jb, "}}");
+ out_1:
+	return jb;
+}
+
+jbuf_t make_step_exit_data(spank_t sh, const char *event)
```
morrone, line 709: It is not clear why this function should have an "event" parameter when the only place this can possibly be called is at "step_exit".

ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -612,7 +682,49 @@ jbuf_t make_exit_data(spank_t sh, const char *event, const char *context)
 	return jb;
 }
 
-jbuf_t make_task_init_data(spank_t sh, const char *event, const char *context)
+jbuf_t make_step_init_data(spank_t sh, const char *event)
```
morrone, line 685: It is not clear why this function should have an "event" parameter when the only place this is called is at "step_init".

ldms/src/sampler/slurm/slurm_sampler.c
```diff
@@ -871,6 +920,15 @@ static int slurm_recv_cb(ldmsd_stream_client_t c, void *ctxt,
 			}
 			handle_job_init(job, entity);
 		}
+	} else if (0 == strncmp(event_name->str, "step_init", 9)) {
```
morrone, line 923: This looks good. Don't we also need a new handler for "step_exit"?

morrone (changes requested):

ldms/src/sampler/spank/slurm_notifier.c
tom95858: agreed, and my testing confirms that this event is not useful for our purposes

ldms/src/sampler/spank/slurm_notifier.c
tom95858: agreed, I had it in there to confirm it wasn't being called

ldms/src/sampler/slurm/slurm_sampler.c
tom95858: I don't think so because the only purpose of "step_init" is to update nnodes, and ntasks; however, what it won't do is update these back to zero between steps. If that is an issue, we can add that event.

morrone: Well, this is an area where the mismatch between slurm's model of a "job" and ldms's model of a "job" causes confusion.

I should think that since the ldms slurm sampler is just calling tasks started by slurm part of a "job", that each time there is a step_init event, the number of tasks should be incremented accordingly, and during step_exit events it should decrement accordingly.

So using an interleaved example of job steps we might have:

T0 step_init, jobid A, stepid X, 3 tasks
T100 step_init, jobid A, stepid Y, 2 tasks
T200 step_fini, jobid A, stepid X
T300 step_fini, jobid A, stepid Y

So I would expect ldms to show these tasks counts at the following times:

T50 3 tasks
T150 5 tasks
T250 2 tasks
T300 0 tasks

The same _cannot_ be done for nnodes though. Job steps do not necessarily run on the same exact subset of nodes as other job steps. For instance, one job step might run on node opal[1-50], and second job step in the same allocation might run on opal[25-75]. And the encompassing job allocation for those job sets might be opal[1-1000].

morrone:
@tom95858 , according to the slurm documentation, there is an environment variable named "SLURM_JOB_NODELIST" set when the job prolog script is called. It should be reasonable to parse that node list (in hostlist string format, I presume) to derive the job allocation's node count at prolog time.

ldms/src/sampler/slurm/slurm_sampler.c
morrone: I should note that I'm assuming task count on a particular node, not task count in the entire job step.

tom95858:
Yeah, that's a good idea. Let me take a look at that. I had originally discounted that imagining full-machine jobs on Cory with 20k nodes and the performance impact of parsing that, but it's only a job startup.

tom95858:
SLURM_JOB_NODELIST is not set at slurm_spank_job_prolog time. I checked using both spank_getenv and getenv.

morrone:
> SLURM_JOB_NODELIST is not set at slurm_spank_job_prolog time. I checked using both spank_getenv and getenv.

Yes, the documentation appears to be wrong. It is named "SLURM_NODELIST".

tom95858:
Ok, ill try that

tom95858:
Here are my findings:

```bash
salloc -N 8 bash
env | grep SLURM
SLURM_NODELIST=orion-iw-[01-08]
SLURM_JOB_NAME=bash
SLURM_NODE_ALIASES=(null)
SLURM_JOB_QOS=normal
SLURM_NNODES=8
SLURM_JOBID=698
SLURM_TASKS_PER_NODE=16(x8)
SLURM_JOB_ID=698
SLURM_SUBMIT_DIR=/home/tom/influx-test
SLURM_JOB_NODELIST=orion-iw-[01-08]
SLURM_CLUSTER_NAME=orion
SLURM_JOB_CPUS_PER_NODE=16(x8)
SLURM_SUBMIT_HOST=ovs-5416.ogc.int
SLURM_JOB_PARTITION=ldms
SLURM_JOB_NUM_NODES=8
```
This contains all the data we need to populate a reasonable "job_size" for the set. 

However on the compute nodes where the spank plugins are running, these values are node specific, e.g.

```
srun -N 1 env | grep SLURM
SLURM_NODELIST=orion-iw-[01-08]
SLURM_JOB_NAME=bash
SLURM_JOB_QOS=normal
SLURM_NNODES=1
SLURM_JOBID=698
SLURM_TASKS_PER_NODE=1
SLURM_JOB_ID=698
SLURM_SUBMIT_DIR=/home/tom/influx-test
SLURM_JOB_NODELIST=orion-iw-[01-08]
SLURM_CLUSTER_NAME=orion
SLURM_JOB_CPUS_PER_NODE=16(x8)
SLURM_SUBMIT_HOST=ovs-5416.ogc.int
SLURM_JOB_PARTITION=ldms
SLURM_JOB_NUM_NODES=8
SLURM_PRIO_PROCESS=0
SLURM_UMASK=0002
SLURM_NTASKS=1
SLURM_NPROCS=1
SLURM_DISTRIBUTION=cyclic
SLURM_STEP_ID=0
SLURM_STEPID=0
SLURM_SRUN_COMM_PORT=32859
SLURM_STEP_NODELIST=orion-iw-01
SLURM_STEP_NUM_NODES=1
SLURM_STEP_NUM_TASKS=1
SLURM_STEP_TASKS_PER_NODE=1
SLURM_STEP_LAUNCHER_PORT=32859
SLURM_SRUN_COMM_HOST=192.168.2.1
SLURM_TOPOLOGY_ADDR=orion-iw-01
SLURM_TOPOLOGY_ADDR_PATTERN=node
SLURM_CPUS_ON_NODE=16
SLURM_TASK_PID=37165
SLURM_NODEID=0
SLURM_PROCID=0
SLURM_LOCALID=0
SLURM_LAUNCH_NODE_IPADDR=192.168.2.1
SLURM_GTIDS=0
SLURM_CHECKPOINT_IMAGE_DIR=/var/slurm/checkpoint
SLURM_JOB_UID=1002
SLURM_JOB_USER=tom
SLURM_JOB_GID=1002
SLURMD_NODENAME=orion-iw-01
```
So although we can determine NNODES at the allocated nodes, we can't compute the TASKS_PER_NODE because that is step specific. 

So it's not clear to me that this is any better than what we have already.

morrone:
Right, that matches what I have been saying. "tasks per node" is _inherently_ a step specific concept. The allocation can't know ahead of time how many tasks a step will put on each node in the allocation. slurm allows one to set defaults for task count that a step launch can choose to employ or ignore. But those default numbers in the allocation state would not have been useful for us even if they were available in environment variables. You shouldn't try to fill in the task count at prolog/epilog time.

BUT you probably need to take care here too!  Here is an example of the sort of thing you would need to understand and parse if you attempt to use those fields:

```
SLURM_STEP_NUM_NODES=4
SLURM_STEP_NUM_TASKS=21
SLURM_STEP_TASKS_PER_NODE=6,5(x3)
```

What this is telling you is that 6 tasks were assigned to one node, and three nodes each got 5 tasks. I don't know if this string is ordered to match SLURM_NODELIST order or not.

baallan:
I cannot help wondering if, at the level of detail being discussed here, we haven't gone down the rabbit hole.

And by that I mean all this fine-grained job detail stuff would be far more reliably translated from the slurm (or other resource manager) system database into some sos/ovis/other database with a periodic/db-event-driven query engine that maps details from one database to another instead of trying to trap event streams. In fact we have application teams doing this (sans ldms) already.

The original jobid sampler (V2/V3) tried to capture at the compute node only the job id present on the node at sample time. This is the minimum needed to get most useful analyses done in the presence of clock discrepancies between compute node and slurm database. Yes clock issues should be rare, but we saw them happen. In particular, there (historically) have been skew bugs and bugs where the end time of a job does not get logged properly in the job history database, but we could easily pick the job end time (to the nearest sample interval) off the ldms compute node jobid data. All the other analysis types (step, task, nodelist etc) can be done better from directly db>db copied data using the node-recorded jobid list as a sanity check/fast query pre-filter. 
This approach also gets us out of the business of coding to the evolving event hooks of every hpc resource manager/scheduler.

tom95858:
On Wed, Jan 6, 2021 at 5:11 PM Christopher J. Morrone <
notifications@github.com> wrote:

> Right, that matches what I have been saying. "tasks per node" is
> *inherently* a step specific concept. The allocation can't know ahead of
> time how many tasks a step will put on each node in the allocation. slurm
> allows one to set defaults for task count that a step launch can choose to
> employ or ignore. But those default numbers in the allocation state would
> not have been useful for us even if they were available in environment
> variables. You shouldn't try to fill in the task count at prolog/epilog
> time.
>
Right, the point of the note was to prove that out in light of your
suggestion regarding SLURM_NODELIST, etc...

> BUT you probably need to take care here too! Here is an example of the
> sort of thing you would need to understand and parse if you attempt to use
> those fields:
>
> SLURM_STEP_NUM_NODES=4
> SLURM_STEP_NUM_TASKS=21
> SLURM_STEP_TASKS_PER_NODE=6,5(x3)
>
> What this is telling you is that 6 tasks were assigned to one node, and
> three nodes each got 5 tasks. I don't know if this string is ordered to
> match SLURM_NODELIST order or not.
>
According to the docs, it is, but there are spank variables that can be
queried directly with spank_get_item that doesn't require parsing strings.

One other thing before I post another version, should we set the app_id to
the current step_id? Right now, app_id is always zero. My sense is that it
could be useful.

PR #584, opened 2021-01-02
Fix endpoint ref count issue in CONNECT_ERROR path

tom95858:
The passive side CONNECT_ERROR path drops an extra reference
that causes the zap_interpose_cb to crash when it de-references
the endpoint.

Also took the connect reference before the call to zap_accept to ensure
that the reference is taken before the callback provided in zap_accept
is called.

The bug is triggered when the remote peer disconnects before the
zap_accept ACCEPT/ACK sequence is completed resulting in a
CONNECT_ERROR event being delivered to ldms_zap_auto_cb

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -46,6 +46,7 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+#define _GNU_SOURCE
 #include <sys/errno.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


ldms/src/core/ldms_xprt.c
```diff
@@ -2081,17 +2082,18 @@ static void ldms_zap_handle_conn_req(zap_ep_t zep)
 
 	__ldms_xprt_conn_msg_init(x, &msg);
 
+	/* Take a 'connect' reference. Dropped in ldms_xprt_close() */
+	ldms_xprt_get(_x);
+
 	zerr = zap_accept(zep, ldms_zap_auto_cb, (void*)&msg, sizeof(msg));
 	if (zerr) {
 		x->log("ERROR: %d accepting connection from %s.\n", zerr, rmt_name);
 		goto err1;
```
nichamon, line 2091: This should be
`goto err2;`

ldms/src/core/ldms_xprt.c
```diff
@@ -2081,17 +2082,18 @@ static void ldms_zap_handle_conn_req(zap_ep_t zep)
 
 	__ldms_xprt_conn_msg_init(x, &msg);
 
+	/* Take a 'connect' reference. Dropped in ldms_xprt_close() */
+	ldms_xprt_get(_x);
+
 	zerr = zap_accept(zep, ldms_zap_auto_cb, (void*)&msg, sizeof(msg));
 	if (zerr) {
 		x->log("ERROR: %d accepting connection from %s.\n", zerr, rmt_name);
```
nichamon, line 2090: If `zap_accept()` returns a synchronous error, `zap_reject()` should not be called at line 2101.

nichamon (changes requested):

PR #588, opened 2021-01-12
Fix spank_notifier init event

narategithub:
spank_notifier init event was created in job prolog, which is a
job_script context. User-specified environment variables (such as
SUBSCRIBER_DATA) are not available in job_script context. In fact,
SLURM_JOB_NAME, SLURM_JOB_USER are also not available in job_script
context. This made papi_sampler ignored all jobs.

This patch moves the creation/sending of the init event into
slurm_spank_init() and sends the init event only if it is the first step
(JOB_STEPID=0) on each participated node.

NOTE1: This has passed `papi_sampler_test` in `ldms-test` repository
       (commit id: 4d2f29a). Without this patch, the test failed.

NOTE2: This patch has not yet addressed `papi_sampler` to handle
       multi-step jobs. It shall be addressed separately.

### DIFF
--- a/ldms/src/sampler/spank/slurm_notifier.c
+++ b/ldms/src/sampler/spank/slurm_notifier.c
@@ -588,6 +588,28 @@ static int send_event(int argc, char *argv[], jbuf_t jb)
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


tom95858:
Narate, please read the commentary on the slurm_notifier pull request.

PR #589, opened 2021-01-15
Support for multi-step jobs in PAPI sampler.

tom95858:
These changes move the SUBSCRIBER_DATA field to the step_init event so that each job step can have it's own SUBSCRIBER_DATA value.

The PAPI sampler is updated to support multi-step jobs. Each step in the job has it's own metric set; the set name is constructed with a suffix of <job_id> "." <step_id>.

Both serial and overlapping job steps are supported.


narategithub (approved):
The patch passed both `papi_sampler_test` and `papi_store_test` in `ldms-test` repo.

narategithub:
@tom95858 I'll also update `papi_sampler_test` and `papi_store_test` to include multi-step jobs. But this patch passed the existing tests (that was previously failed).

PR #591, opened 2021-01-17
Add prdcr_stats and set_stats to ldmsctl

nichamon:


tom95858:
I know there is a lot of mixed messaging about this, but free() checks for NULL. So the additional check in our code is just a waste of ergs.

From the free man page
...
The  free() function shall cause the space pointed to by ptr to be deallocated; that is, made available for further allocation. If ptr is a __null__ pointer, no action shall occur. 
...

nichamon:
> I know there is a lot of mixed messaging about this, but free() checks for NULL. So the additional check in our code is just a waste of ergs.
> 
> From the free man page
> ...
> The free() function shall cause the space pointed to by ptr to be deallocated; that is, made available for further allocation. If ptr is a **null** pointer, no action shall occur.
> ...

Got it. I'll remove the commit that check for NULL points before calling free() and rename the pull request.

tom95858 (approved):

PR #592, opened 2021-01-18
Fix the push path and the test_ldms_push program

nichamon:


nichamon:
The bug was found because LDMS fails the test case of canceling push in the test_ldms_push program. The symptom is that push could not be canceled. The cause is that the push path called `ldms_lookup_rbd()`, which could return any of the RBDs that correspond to the transport and the set. A symptom is that push could not be canceled. Another symptom is that new RBDs could be allocated every time LDMS receives a push register request instead of updating an existing push RBD.

After I fixed the call of `ldms_lookup_rbd()`, I found the reference problems. An RBD reference is taken every time LDMS receives a push register request. This is unnecessary. Also, `process_cancel_push_request()` calls `__ldms_free_rbd()` which will put back the "rendezvous_push" reference twice. It puts back one reference extra.

nichamon:
Without the patch to `ldms_xprt`, clients may not be able to cancel push updates. However, it will receive push updates. If the updater with the push mode has been restarted, the client may receive multiple push updates when the set got changed on the server.

ldms/src/core/ldms_xprt.c
```diff
@@ -2510,30 +2569,32 @@ static void handle_rendezvous_push(zap_ep_t zep, zap_event_t ev,
...
+	for (; push_rbd; push_rbd = __rbd_by_set_next(push_rbd)) {
 		if (push_rbd->push_flags & LDMS_RBD_F_PUSH) {
-			ref_get(&push_rbd->set->ref, "rendezvous_push");
 			/* Update the push flags, but otherwise, do nothing */
 			goto out;
 		}
 	}
 
+create_push_rbd:
 	/* We will be the target of RDMA_WRITE */
```
tom95858, line 2585: This comment isn't true... We should remove it.

nichamon: Done.

tom95858:
@nichamon what is the status of this pull request?

nichamon:
@tom95858 I have tested the patch on Voltrino. No crashes and aggregators received the data as expected in the push mode. The cancel push path works properly.

- sampler daemons: 200 daemons x 5 nodes & 11 sets per daemon. All sampler daemons killed 2 minutes after they started and restarted again in 30 seconds.
- agg11 and agg12: each connected to 500 sampler daemons.
- agg21 connected to agg11 and agg12. agg21 had 2 updaters: one updater for each agg11 and agg12. Both Updaters used the onchange push mode. The updater for agg11's sets was stopped and started periodically to test the cancel push path.

tom95858:
It's showing as WIP, is this ready for review yet?

tom95858:
@nichamon, I don't think we need the push-mode RBD ... correct? I don't believe that it's used. Am I missing something here?

nichamon:
Currently, `ldms_xprt_push` iterates through `set->remote_rbd_list` and sends the set content when it finds a push rbd. The bottomline is that LDMS uses push rbd to tracks if a client has register for push or not. However, we don't really need it. We could add a push_flag to the rbd structure similar to the notify mechanism. Also, the push registration logic doesn't need `zap_share` anymore.

I didn't remove the push_rbd in this patch because they will be gone when the logic (ported from master) that completely removes the rbd structure from LDMS is merged.

PR #594, opened 2021-01-18
Move job_user and job_name to step_init

tom95858:
The job_user and job_name environment variables are not set in the
job_prolog context.


narategithub (approved):
It works for the job with srun. For a job without srun, job_name and job_user are <empty>. I guess the job script writer that does not use srun has to call ldmsd_stream_publish to notify ldmsd about the information.

PR #600, opened 2021-01-22
update gender config example to v4

baallan:
updating the plugins-conf subdir of the genders systemd example to v4.


morrone (changes requested):
The volume of new files and repetition in the contents of these files make this look a lot less like an update to "examples" than it does your local configuration settings stored in the upstream source. I do not think that is a particularly good direction to go with examples.

The Makefile.am does not appear to have been updated in this commit to include the files in the dist either.

baallan:
@morrone  thankyou for catching the makefile omission.

morrone:
@tom95858 the Makefile.am omission does not appear to have been addressed before landing.

tom95858:
@morrone, sorry I was told this had been addressed and was ready for merge. @baallan  please resolve this ASAP?

baallan:
I will add the makefile update to the same branch and open another request today. Apologies; it slipped my mind.

PR #602, opened 2021-01-28
Fix big ldmsd_stream

narategithub:
This pull request contains 2 commits:
- "Fix big ldmsd_stream data." This commit contains a (quite big) patch addressing ldmsd_stream with data too big to fit in a message. It also addresses the record splitting mechanisms of ldmsd_stream request that is mismatch that of the ldmsd_request.
- "Add sq status check." This commit addresses the issue of when to disconnect the xprt in ldmsd_stream_publish_file. Even though a new zap API is added, I believe it does not make it incompatible with ldmsd 4.3.6.

NOTE1: This pull request has passed all test cases in `ldms-test` repository.
NOTE2: A manual test between `ldmsd_stream_subscribe` and `ldmsd_stream_publish` is also performed, with `SOCKBUF_SZ 256` (see `zap_sock.h`) to encourage a lot of record splitting:
```sh
$ python3 <<EOF
> import json
> f = open('data.json', 'w')
> print(json.dumps([i for i in range(0, 65536)]), file=f)
> f.close()
> EOF

$ ldmsd_stream_subscribe -x sock -p 10000 -s test -f out.txt

$ ldmsd_stream_publish -h localhost -x sock -p 10000 -s test -t ${FORMAT} -f ${FILE}

# NOTE: The FORMAT being tested are both json and string.
```

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request_util.c
+++ b/ldms/src/ldmsd/ldmsd_request_util.c
@@ -873,7 +873,7 @@ ldmsd_req_attr_t ldmsd_req_attr_get_by_id(char *request, uint32_t attr_id)
 {
 	ldmsd_req_hdr_t req = (ldmsd_req_hdr_t)request;
 	ldmsd_req_attr_t attr = ldmsd_first_attr(req);
-	while (attr->discrim) {
+	while (attr && attr->discrim == 1) {
 		if (attr->attr_id == attr_id) {
 			return attr;
 		}
@@ -951,7 +951,7 @@ void ldmsd_ntoh_req_msg(ldmsd_req_hdr_t req)
 		return;
 
 	attr = ldmsd_first_attr(req);
-	while (attr && attr->discrim) {
+	while (attr && attr->discrim == ntohl(1)) {
 		ldmsd_ntoh_req_attr(attr);
 		attr = ldmsd_next_attr(attr);
 	}
@@ -983,7 +983,7 @@ void ldmsd_hton_req_msg(ldmsd_req_hdr_t resp)
 		return;
 
 	attr = ldmsd_first_attr(resp);
-	while (attr && attr->discrim) {
+	while (attr && attr->discrim == 1) {
 		next_attr = ldmsd_next_attr(attr);
 		ldmsd_hton_req_attr(attr);
 		attr = next_attr;
--- a/ldms/src/ldmsd/ldmsd_stream.c
+++ b/ldms/src/ldmsd/ldmsd_stream.c
@@ -135,6 +135,84 @@ void ldmsd_stream_close(ldmsd_stream_client_t c)
 	pthread_mutex_unlock(&c->c_s->s_lock);
 }
 
+struct req_buf {
+	ldms_t x;
+	int sz;
+	int off;
+
+	union {
+		struct ldmsd_req_hdr_s req[0];
+		char buff[0];
+	};
+};
+
+static void req_buf_init(struct req_buf *b, ldms_t x, int sz)
+{
+	b->x = x;
+	b->sz = sz;
+	b->off = sizeof(*b->req);
+	b->req->flags = LDMSD_REQ_SOM_F;
+	b->req->marker = LDMSD_RECORD_MARKER;
+}
+
+static struct req_buf *req_buf_new(ldms_t x, int sz)
+{
+	struct req_buf *b = calloc(1, sizeof(*b) + sz);
+	if (b)
+		req_buf_init(b, x, sz);
+	return b;
+}
+
+int req_buf_send(struct req_buf *b)
+{
+	int rc;
+	b->req->rec_len = b->off;
+	ldmsd_hton_req_hdr(b->req);
+	rc = ldms_xprt_send(b->x, b->buff, b->off);
+	ldmsd_ntoh_req_hdr(b->req);
+	if (rc)
+		return rc;
+	b->off = sizeof(*b->req);
+	b->req->flags &= ~LDMSD_REQ_SOM_F;
+	return 0;
+}
+
+/* Fill the available space of the req with data. Send when the space
+ * ran out, and repeat the filling/send. `off` is use to track the offset of the
+ * available space. */
+static int req_buf_append(struct req_buf *b, const void *data, int data_len, int flags)
+{
+	const char *d = data;
+	int dlen = data_len;
+	int cpsz, rc;
+	while (dlen) {
+		cpsz = b->sz - b->off;
+		cpsz = dlen < cpsz ? dlen : cpsz;
+		memcpy(&b->buff[b->off], d, cpsz);
+		dlen -= cpsz;
+		d += cpsz;
+		b->off += cpsz;
+		if (b->off == b->sz) {
+			/* space full, send */
+			if ((flags & LDMSD_REQ_EOM_F) && dlen == 0) {
+				b->req->flags |= LDMSD_REQ_EOM_F;
+				flags &= ~LDMSD_REQ_EOM_F;
+				return req_buf_send(b);
+			}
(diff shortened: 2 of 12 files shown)


ldms/src/ldmsd/ldmsd_request_util.c
```diff
@@ -873,7 +873,7 @@ ldmsd_req_attr_t ldmsd_req_attr_get_by_id(char *request, uint32_t attr_id)
 {
 	ldmsd_req_hdr_t req = (ldmsd_req_hdr_t)request;
 	ldmsd_req_attr_t attr = ldmsd_first_attr(req);
-	while (attr->discrim) {
+	while (attr && attr->discrim == 1) {
```
nichamon, line 876: Why do we need to compare  `attr->discrim == 1`? Can we leave it as `attr && attr->discrim`? I didn't go through whether all places have initialized attr->discrim to either 1 or 0. Are you certain?

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -317,68 +431,26 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
...
+			rc = ldmsd_stream_publish(x, stream, LDMSD_STREAM_STRING,
+					buffer, data_len);
 			if (rc)
 				goto err;
-			/* Turn off start of message flag, for subsequent sends */
-			flags = 0;
 		}
 	}
 
+	while (ldms_xprt_sq_status(x) == EBUSY) {
```
tom95858, line 450: This can't be correct right? If this is required, then the transport is broken.

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
```diff
@@ -134,29 +183,68 @@ static int stream_publish_handler(ldmsd_req_hdr_t req)
...
+	/* we got all request data */
+	ldmsd_req_hdr_t req = (ldmsd_req_hdr_t)ctxt->buff;
+	req->rec_len = htonl(ctxt->off); /* set aggregateg rec len */
+	ldmsd_ntoh_req_msg(req);
+	int rc = stream_publish_handler(req);
 
+	/* request processed, reset data buffer */
+	ctxt->off = 0;
+
+	/* send the reply */
```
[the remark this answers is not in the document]
narategithub: @tom95858 FYI, I copied the send-reply from the existing code. Shall I remove them?

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
```diff
@@ -57,6 +60,69 @@ static const char *short_opts = "p:f:s:x:a:A:D";
 
 #define AUTH_OPT_MAX 128
 
+struct xprt_ctxt {
```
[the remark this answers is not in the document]
narategithub: I don't think so. It get created when the xprt connected, and destroyed when disconnected or error.

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
```diff
@@ -186,12 +279,22 @@ static void recv_msg(ldms_t x, char *data, size_t data_len)
...
+			msglog("xprt_ctxt_new() failed, errno: %d\n", errno);
+			ldms_xprt_close(ldms);
+		}
 		break;
 	case LDMS_XPRT_EVENT_DISCONNECTED:
 	case LDMS_XPRT_EVENT_REJECTED:
 	case LDMS_XPRT_EVENT_ERROR:
+		ctxt = xprt_ctxt_find(x);
+		if (ctxt) {
+			xprt_ctxt_free(ctxt);
```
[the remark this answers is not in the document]
narategithub: It is not being used when xprt is disconnected or error.

ldms/src/ldmsd/ldmsd_request_util.c
narategithub: I think I am. I changed it to 1 because all the setter set it to 1.

ldms/src/ldmsd/ldmsd_stream.c
narategithub: It is "broken" because we don't have send completion. For a big data case, multiple records may still be in the io_q (the pending io) or sq (in sock, not submitted to the system yet). Since it is asynchronous, we may close the xprt before all records are submitted to the underlying transport. And I found that from my test last night.

narategithub: @tom95858 shall we pull in `send_mapped()` from master instead? I believe `send_mapped()` has send completion and we can use it for the case like this.

ldms/src/ldmsd/ldmsd_request_util.c
nichamon: I just checked. Either 0 or 1 is assigned to `attr->discrim`. This is safe for now. The attribute data structure would be gone when we merge the new JSON configuration protocol.

ldms/src/ldmsd/ldmsd_stream.c
nichamon: I remember last you guys address the send without completion problem for stream, we have an assumption that users will use stream to send small data. That's why we didn't put in the zap send_mapped() at the time or made a bigger patch.

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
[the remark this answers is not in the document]
nichamon: Got it!. I was confused it with LDMSD's request contexts.

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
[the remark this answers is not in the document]
nichamon: @narategithub handled the `LDMS_XPRT_EVENT_RECV` event in `ldmsd_stream.c:event_cb()` to do nothing. Given that the program that sends stream to `ldmsd_stream_subscribe` is `ldmsd_stream_publish`,  the responses sent by `ldmsd_stream_subscribe` won't cause any problems. However, if streams are published by ldmsd, ldmsd will receive an unexpected message. The symptom is the same as what Eric saw last year.

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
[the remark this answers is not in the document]
nichamon: Got it!

ldms/src/ldmsd/ldmsd_stream.c
tom95858: The problem is that the close is synchronous, but the send is asynchronous. IOW, we queue sends to a send_q, but we don't queue close. We had this same issue on receive a while back, i.e. the inbound disconnect would get delivered right away which could then be followed by i/o completions. 

The other issue is that the chunking of configuration messages should be done in one place not be copied here and there so that it has to be maintained in multiple places.

It's to clear to me why configuration messages should be handled differently than any other request, e.g. dir_reply, that may be very large.

narategithub: I agree with you. The config was done outside of ldms xprt (i.e. ldmsd) because historically, according to @nichamon, the config was only used by ldmsd and at the time no other LDMS was doing splitting/joining messages like `push` and `dir` do today.

I think queuing `close` command is a fine, but we still need a send completion for `ldmsd_stream_publish` program, because the program may exited before all messages got submitted (the routine of the program was connect, repeatedly send, close, exit).

This patch was a hurried/last-minute "make it work" kind of effort. I was actually thinking something similar to you but the change will be much bigger and more chance of disruption. Shall I or @nichamon started a proposal and we go from there?

I'm closing this pull request in favor of the better solution in a near future.

PR #603, opened 2021-01-31
Fix wrong error in listen config command when auth not found.

narategithub:


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6498,6 +6498,16 @@ static int listen_handler(ldmsd_req_ctxt_t reqc)
 	if (!listen) {
 		if (errno == EEXIST)
 			goto eexist;
+		else if (errno == ENOENT) {
+			reqc->errcode = ENOENT;
+			(void) snprintf(reqc->line_buf, reqc->line_len,
+					"Listen error: authentication domain "
```
tom95858, line 6504: I think 'domain' should be 'plugin' right?

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6498,6 +6498,16 @@ static int listen_handler(ldmsd_req_ctxt_t reqc)
...
+					"'%s' not found. Please make sure "
+					"that it is created with `auth_add` "
```
tom95858, line 6506: If -a <auth-plugin> is used on the command line is 'auth_add' still required?

ldms/src/ldmsd/ldmsd_request.c
narategithub: I think it used to be (in master branch). It used to be that we don't have `auth` config object. And we introduced the `auth` config object as an entity holding authentication domain information. Since then, `auth` refers to auth config object, not auth config plugin. If I remember correctly, the motivation to auth config object is to reduce the repetitive authentication parameter in the use case of multiple authentication domain like the following:

```
prdcr_add name=p1 host=host1 ... auth=munge socket=/var/run/munge-dom1.sock
prdcr_add name=p2 host=host2 ... auth=munge socket=/var/run/munge-dom1.sock
...
prdcr_add name=p100 host=host100 ... auth=munge socket=/var/run/munge-dom1.sock

prdcr_add name=p101 host=host101 ... auth=munge socket=/var/run/munge-dom2.sock
...
prdcr_add name=p200 host=host200 ... auth=munge socket=/var/run/munge-dom2.sock

prdcr_add name=p201 host=host201 ... auth=ovis conf=/var/run/ovis-auth1.conf
...
prdcr_add name=p300 host=host300 ... auth=ovis conf=/var/run/ovis-auth1.conf
``` 

The above config example is reduced to
```
auth_add name=dom1 plugin=munge socket=/var/run/munge-dom1.sock
auth_add name=dom2 plugin=munge socket=/var/run/munge-dom2.sock
auth_add name=dom3 plugin=ovis conf=/var/run/ovis-auth1.conf

prdcr_add name=p1 host=host1 ... auth=dom1
prdcr_add name=p2 host=host2 ... auth=dom1
...
prdcr_add name=p100 host=host100 ... auth=dom1

prdcr_add name=p101 host=host101 ... auth=dom2
...
prdcr_add name=p200 host=host200 ... auth=dom2

prdcr_add name=p201 host=host201 ... auth=dom3
...
prdcr_add name=p300 host=host300 ... auth=dom3
```

The `auth` in `listen` command then also refers to auth object that represents an authentication domain.

ldms/src/ldmsd/ldmsd_request.c
narategithub: If `-a munge` was supplied at the CLI, then `listen` and `prdcr_add` will use it by default if `auth` attribute is not given. However, if `auth=munge` is given, it will look for auth config object named "munge". In this case, `auth_add name=munge plugin=munge` is still needed.

PR #606, opened 2021-02-01
Change the deafult auth reference to "auth_default"

narategithub:
Change the default authentication reference to `auth_default` so that
users can explicitly specify it in `listen` and `prdcr_add`
configuration commands.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.h
+++ b/ldms/src/ldmsd/ldmsd.h
@@ -1224,7 +1224,7 @@ typedef struct ldmsd_auth {
 
 /* Key (name) of the default auth -- intentionally including SPACE as it is not
  * allowed in user-defined names */
-#define DEFAULT_AUTH " _DEFAULT_AUTH_ "
+#define DEFAULT_AUTH "auth_default"
 
 ldmsd_auth_t
 ldmsd_auth_new_with_auth(const char *name, const char *plugin,


narategithub:
@tom95858 I tested it with the following config
```conf
auth_add name=none plugin=none
listen  xprt=sock port=10002 auth=auth_default
listen  xprt=sock port=10003
listen  xprt=sock port=10004 auth=none
load name=meminfo
config name=meminfo component_id=1 instance=node-1/meminfo producer=node-1 perm=0777
start name=meminfo interval=1000000 offset=0
load name=vmstat
config name=vmstat component_id=1 instance=node-1/vmstat producer=node-1 perm=0700
start name=vmstat interval=1000000 offset=0
```

and with the following ldmsd CLI:
```sh
$ ldmsd -a munge -x sock 10001 -c /etc/ldmsd.conf -l /var/log/ldmsd.log -v INFO
```

The following is the `ldms_ls` results:
```sh
$ ldms_ls -x sock -p 10001 
Connection failed/rejected.

$ ldms_ls -x sock -p 10001 -a munge
node-1/vmstat                      
node-1/meminfo                     
                                   
$ ldms_ls -x sock -p 10002 -a munge
node-1/vmstat                      
node-1/meminfo       

$ ldms_ls -x sock -p 10003 -a munge
node-1/vmstat
node-1/meminfo

$ ldms_ls -x sock -p 10004 -a munge
Connection failed/rejected.

$ ldms_ls -x sock -p 10004
node-1/vmstat
node-1/meminfo

```

ldms/src/ldmsd/ldmsd.h
```diff
@@ -1224,7 +1224,7 @@ typedef struct ldmsd_auth {
 
 /* Key (name) of the default auth -- intentionally including SPACE as it is not
  * allowed in user-defined names */
-#define DEFAULT_AUTH " _DEFAULT_AUTH_ "
+#define DEFAULT_AUTH "auth_default"
```
tom95858, line 1227: Is a configuration object created for "auth_default?" IOW, if a user does this:

```
prdcr_add name=X host=node-1 ... auth="auth_default"
```

It will use the -a, -A provided on the command line?

narategithub: Yes.

The default auth config object was already there, with a name that cannot be specified by users because it contains spaces. The reason was probably to avoid name collision.

tom95858: Thanks for the explanation @narategithub

PR #608, opened 2021-02-02
Allow adding listen endpoints after init

nichamon:


narategithub (approved):
Looks OK to me.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6513,6 +6503,16 @@ static int listen_handler(ldmsd_req_ctxt_t reqc)
 		else
 			goto enomem;
 	}
+
+	if (ldmsd_is_initialized()) {
```
tom95858, line 6507: What is this check really doing? What happens if ldmsd is not initialized, the request results in an error and is ignored? It looks like the configuration object remains, is it retried as part of the initialization? Also, the error message isn't really correct. the listen did not fail, it was never even attempted.

nichamon: LDMSD only creates configuration objects during the initialized time. The initialized time ends when LDMSD processes all configuration files. It starts the existing cfgobjs afterward. There is no 'listen_start' command, so the 'listen' command combines creating a listen cfgobj and starting the listen cfgobj. Starting a listen cfgobj means creating an LDMS transport endpoint and listening on the endpoint.

The check `ldmsd_is_initialized()` is to separate the cfgobj creation and the action that starts a listen cfgobj. If LDMSD receives a 'listen' command after it has been initialized, it will create the listen cfgobj and start the listen cfgobj. On the other hand, if LDMSD is not initialized yet, it will only create a listen cfgobj. LDMSD will start the listen cfgobjs after it processes all configuration files.

PR #610, opened 2021-02-04
Fix papi_sampler segmentation fault on job exit event

narategithub:
`handle_job_exit()` in `papi_sampler` incorrectly assumed that
`job->set` was always successfully created before the job exit event and
accessed the set. This resulted in segmentation fault. This patch
modifies `papi_sampler` to check the set before accessing it.


narategithub:
@tom95858 I found this when I tested the new `zap_sock` against tests in `ldms-test` repo.

tom95858:
Thanks @narategithub

PR #611, opened 2021-02-04
[PLS-REVIEW] New zap thread model

narategithub:
NOTE: Please see `lib/src/zap/ZAP_THREAD_NOTE.md` for an overview of
          the zap thread mechanism.
    
    [x] busy-ness calculation (using Tom's  thrstat)
    [x] sock
        [x] zap_test
        [x] ldmsd: samp->agg1->agg2 (w/sos)
        [x] all tests in ldms-test (sock)
    [x] rdma
        [x] samp - agg1 - agg2 - store test
        [x] flap samp (sniff)
        [x] flap agg1 (sniff)
            [x] flap agg2 (sniff)
        [x] fd check
            [x] ldms_ls - ldmsd test
    [x] ugni
       [x] ldmsd - ldms_ls
           - with debug print, the endpoint references are correct, the
             endpoints (both active and passive) are destroyed at the end.
       [x] samp - agg1 - agg2 - ldms_ls, with flapping agg1
       [x] samp (many sets) - agg1 - ldms_ls
       [x] scale test: many samp -> agg11+agg12 -> agg2
       [*] NOTE1: Need to protect GNI calls with the global `ugni_lock`.
           Not protecting them at all gave us SIGSEGV and SIGABRT down in
           the libugni. Protecting only the CQ of each thread using
           thread's own mutex also did not work. We got GNI_DlaDrain()
           hanged when we have multiple io threads.
       [*] NOTE2: `GNI_MemRegister()` call with recv CQ to register SMSG
           mbox memory took 0.6-0.7 seconds to complete. It is the source
           of the connection slowness seen by `ldms_ls`. Fortunately, this
           happens only once for each io thread initialization.
    [x] fabric over tcp: 64 samps -> 2 agg1 -> 1 agg2 - ldms_ls
    [x] fabric over verbs: 64 samps -> 2 agg1 -> 1 agg2 - ldms_ls

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -3900,7 +3900,16 @@ void __ldms_xprt_term(struct ldms_xprt *x)
 
 int ldms_xprt_term(int sec)
 {
-	return zap_term(sec);
+	int i, rc, tmp;
+	zap_t z;
+	rc = 0;
+	for (i = 0; i < ldms_zap_tbl_n; i++) {
+		z = ldms_zap_tbl[i].zap;
+		tmp = zap_term(z, sec);
+		if (tmp)
+			rc = tmp;
+	}
+	return rc;
 }
 
 int ldms_xprt_sockaddr(ldms_t _x, struct sockaddr *local_sa,
--- a/ldms/src/ldmsd/ldms_ls.c
+++ b/ldms/src/ldmsd/ldms_ls.c
@@ -733,7 +733,7 @@ int main(int argc, char *argv[])
 			free(xprt);
 			xprt = strdup(optarg);
 			if (!xprt) {
-				printf("ERROR: out of memory\n"); 
+				printf("ERROR: out of memory\n");
 				exit(1);
 			}
 			break;
@@ -1036,6 +1036,12 @@ done:
 		free(dir);
 	}
 
+	/* gracefully close, and wait at most 2 seconds before exit */
 	ldms_xprt_close(ldms);
+	struct timespec _t;
+	clock_gettime(CLOCK_REALTIME, &_t);
+	_t.tv_sec += 2;
+	sem_timedwait(&conn_sem, &_t);
+
 	exit(0);
 }
--- a/ldms/src/ldmsd/ldmsd_prdcr.c
+++ b/ldms/src/ldmsd/ldmsd_prdcr.c
@@ -548,10 +548,31 @@ static void __prdcr_remote_set_delete(ldmsd_prdcr_t prdcr, ldms_set_t set)
 	prdcr_reset_set(prdcr, prdcr_set);
 }
 
+const char *_conn_state_str[] = {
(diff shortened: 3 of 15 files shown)


baallan:
re point 4.1 of the zap thread note. There should be, even in the case of 'too busy' be some administratively defined hard upper bound on the number of threads created for zap io. First because it's a good idea to let admins control their daemons and second because elsewhere in the code there is a message format which (as @narategithub commented in another PR) expects the number of threads to be never > 1000. The limit and its control should be added to the code and the documentation and the design doc/discussion notes.

morrone:
It looks like there is some good documentation about the new model in the comment. But could you please summarize what the general motivation is for this change? What goal(s) is/are trying to be accomplished through this change?

tom95858:
The current transport event flow is implemented privately in each transport provider and is single threaded, i.e. all transports of the same type share the same thread. This logic monitors activity (via epoll) on a set of file descriptors and generates Zap events that are ultimately delivered to the application. The single thread that monitors the I/O events limits scalability for large endpoint counts such as at the 1st level aggregators. The figure below illustrates this implementation.

image

This change implements a shared, multi-thread event subsystem in libzap.so to monitor the endpoint file descriptors as shown in the figure below.
 
image

A secondary benefit is that the system administrator does not have to guess how many threads are required to support a given workload. The implementation will scale the thread counts up and down as latency and load dictate. As mentioned by others, the system administrator will likely want to set an upper limit on the thread count, however a core-count reasonable upper limit will be used as a default

morrone:
Thanks! That is great context.

If the diagrams are accurate, then of course the new model has the tradeoff that the "Application callback function" is now called under the context of the transport thread, and blocks processing of new messages (for the set of endpoints that it monitors, of course) until it is complete. That was not an issue under the previous model.

Perhaps if we can guarantee that the application callback function is short and bounded in processing time, this would be acceptable, but otherwise it looks like it introduces a new problem.

tom95858:
> If the diagrams are accurate, then of course the new model has the tradeoff that the "Application callback function" is now called under the context of the transport thread, and blocks processing of new messages (for the set of endpoints that it monitors, of course) until it is complete. That was not an issue under the previous model.

Right, but only if we assume that the the queue into which I/O events are posted would grow unbounded. The depth of this queue is 4096 by default and can be tweaked with ZAP_EVENT_QDEPTH; but effectively it has the same issue, if **ldmsd** (i.e. the application) cant keep up, I/O will back up. The intent of this new model is to detect this and grow the thread pool as required.

tom95858:
@narategithub, it looks like the build fails with -D DEBUG.

narategithub:
@tom95858 The debug-build test failed because of `zap_fabric`. This work-in-progress has not touched fabric transport yet. I tried building with `-DDEBUG -Wall -Werror` with only sock and rdma and it succeeded.

narategithub:
@tom95858 this is ready for you to review.

narategithub:
Part of #783

PR #612, opened 2021-02-04
Unify the code that chunks & gathers records to a message and refactor ldmsd_stream_subcribe

nichamon:


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -253,6 +258,52 @@ static void event_cb(ldms_t x, ldms_xprt_event_t e, void *cb_arg)
 	sem_post(&conn_sem);
 }
 
+#define ROUND(len) ((((len)-1)|0XFFF)+1)
```
tom95858, line 261: If you scavenge code from ldmsd_stream_publish, you can send the file's contents a record at a time. This will avoid reading the whole file into memory.

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -317,72 +366,27 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
...
-			req->req_id = LDMSD_STREAM_PUBLISH_REQ;
-			req->rec_len = this_rec;
-			if (!data_len)
-				/* No more data, turn on last message flag */
-				flags |= LDMSD_REQ_EOM_F;
-			req->flags = flags;
-			ldmsd_hton_req_hdr(req);
-
-			rc = ldms_xprt_send(x, (char *)req, this_rec);
+	if (stream_type == LDMSD_STREAM_JSON) {
```
tom95858, line 370: If we don't try to reuse ldmsd_stream_subscribe, then we can combine these two paths. 
```
while count=fread max_msg_len:
    if (count < max_msg_len)
        flags |= LDMSD_REQ_EOM_F
    stream_send(..., flags, ...)
```

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -317,72 +366,27 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
...
+			rc = ldmsd_stream_publish(x, stream, LDMSD_STREAM_STRING,
+					buffer, data_len);
 			if (rc)
 				goto err;
-			/* Turn off start of message flag, for subsequent sends */
-			flags = 0;
 		}
 	}
 
+	sleep(1);
```
tom95858, line 385: This should not be here...

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6136,6 +6079,7 @@ static int stream_republish_cb(ldmsd_stream_client_t c, void *ctxt,
 	if (rc)
 		goto out;
 	rc = ldmsd_req_cmd_attr_term(rcmd);
+	ldmsd_log(LDMSD_LDEBUG, "%s\n", data);
```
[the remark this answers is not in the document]
nichamon: I removed in the patch I just pushed.

ldms/src/ldmsd/ldmsd_stream.c
nichamon: I refactored to code so that it reads the file in multiple chunks if the file is larger than max message length.

ldms/src/ldmsd/ldmsd_stream.c
nichamon: I refactored the code so that `ldmsd_stream_publish_file()` does not call `ldmsd_stream_publish()`, so the path was combined in the patch I just pushed.

ldms/src/ldmsd/ldmsd_stream.c
nichamon: Removed.

PR #613, opened 2021-02-05
Address build warnings (errors) when '-O2' is given

narategithub:
This patch addresses warnings, which become errors, found when building
with `CFLAGS='-O2 -Wall -Werror'` on CLE 7.0.UP02 (SuSE 15.1) system.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -3278,7 +3278,7 @@ int ldms_register_notify_cb(ldms_t x, ldms_set_t s, int flags,
 void ldms_xprt_set_delete(ldms_set_t s, ldms_set_delete_cb_t cb_fn, void *cb_arg)
 {
 	struct ldms_request *req;
-	struct ldms_rbuf_desc *rbd, *next_rbd;
+	struct ldms_rbuf_desc *rbd, *next_rbd = NULL;
 	ldms_t xprt;
 	struct ldms_context *ctxt;
 	struct ldms_set *set = s->set;
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -5585,6 +5585,8 @@ static char *__xprt_stats_as_json(size_t *json_sz)
 	struct ldms_xprt_rate_data rate_data;
 	int reset = 0;
 
+	(void)clock_gettime(CLOCK_REALTIME, &start);
+
 	ldms_xprt_rate_data(&rate_data, reset);
 
 	buff = malloc(sz);
--- a/ldms/src/sampler/dstat/dstat.c
+++ b/ldms/src/sampler/dstat/dstat.c
@@ -166,14 +166,15 @@ static char * compute_pidopts_schema(struct attr_value_list *kwl, struct attr_va
 		return strdup(schema_name);
 	}
 	schema_name = "dstat";
-	bsz = strlen(schema_name) + 16;
+	bsz = strlen(schema_name) + 18; /* len(schema_name) + len("_<HEX>") + 1 */
 	buf = malloc(bsz);
-	char hbuf[bsz];
-	memset(hbuf, 0, bsz);
+	if (!buf)
+		return NULL;
 	if (dosuffix) {
-		snprintf(hbuf, bsz, "_%x", pidopts);
+		snprintf(buf, bsz, "%s_%x", schema_name, pidopts);
+	} else {
+		snprintf(buf, bsz, "%s", schema_name);
 	}
-	snprintf(buf, bsz, "%s%s", schema_name, hbuf);
 	return buf;
 }
 
--- a/lib/src/zap/zap.c
+++ b/lib/src/zap/zap.c
@@ -1011,7 +1011,7 @@ static void init_atfork(void)
 
 	for (i = 0; i < zap_event_workers; i++) {
 		char thread_name[16];
-		(void)snprintf(thread_name, 16, "zap:worker:%d", i);
+		(void)snprintf(thread_name, 16, "zap:wkr:%hd", (short)i);
 		zap_event_queue_init(&zev_queue[i], zap_event_qdepth,
 					thread_name, stats_w);
 		rc = pthread_create(&zev_queue[i].thread, NULL,
--- a/ldms/src/ldmsd/ldmsctl.c
+++ b/ldms/src/ldmsd/ldmsctl.c
@@ -175,18 +175,23 @@ static void ldmsctl_buffer_free(ldmsctl_buffer_t buf)
 	free(buf);
 }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpragmas"
+#pragma GCC diagnostic ignored "-Walloc-size-larger-than="
 static void ldmsctl_buffer_mem_append(ldmsctl_buffer_t buf, void *src, size_t n)
 {
 	if (buf->len - buf->off < n) {
-		buf->buf = realloc(buf->buf, buf->len + n);
-		if (!buf->buf) {
+		char *tmp = realloc(buf->buf, buf->len + n);
+		if (!tmp) {
 			fprintf(stderr, "Out of memory\n");
 			exit(ENOMEM);
 		}
+		buf->buf = tmp;
 		buf->len += n;
 	}
 	memcpy(&(buf->buf[buf->off]), src, n);
 }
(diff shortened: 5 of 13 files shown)


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5679,8 +5681,8 @@ static char *__xprt_stats_as_json(size_t *json_sz)
 					    (struct sockaddr *)&ss_local,
 					    (struct sockaddr *)&ss_remote,
 					    &socklen);
-			strncpy(xprt_type, ldms_xprt_type_name(op->op_min_xprt),
```
tom95858, line 5682: What 's the warning that you got here?

ldms/src/sampler/dstat/parse_stat.c
```diff
@@ -127,7 +127,7 @@ int parse_proc_pid_stat(struct proc_pid_stat *s, const char *pid) {
 	commstart++;
 	if ((commend - commstart) >= COMM_SZ)
 		return ENAMETOOLONG;
-	strncpy(s->comm, commstart , COMM_SZ);
+	snprintf(s->comm, COMM_SZ, "%s", commstart);
```
tom95858, line 130: Are these the warnings related to the size being on the source buffer instead of the destination buffer?

ldms/src/ldmsd/ldmsd_request.c
narategithub: It warned about the combination of size and string not terminating with '\0'.

ldms/src/sampler/dstat/parse_stat.c
narategithub: I think at this spot it warned about not terminating with '\0'.

ldms/src/ldmsd/ldmsd_request.c
tom95858: I don't quite get that. The man page says that the result will always be terminated with a '\0' What are we missing here.

narategithub: From `strncpy(3)`

>The  strncpy()  function is similar, except that at most n bytes of src
>are copied.  Warning: If there is no null byte among the first n  bytes
>of src, the string placed in dest will **not** be null-terminated.

tom95858: I feel like we're going round in circles here. This used to be strcpy, and then someone changed a bunch of these to strncpy because some tool was warning about strcpy.  So now we're going to snprintf because yet another tool is warning about the potentially missing '\0'.  strncpy is less efficient, writing "" to a 1MB buffer will write 1MB of '\0' to the destination, and snprintf is slower still. These changes are all non-functional, progressively less efficient, and succeed only in quieting a warning about a non-issue. What platform is this on? I'm leaning towards just quieting that warning globally on that platform and calling it a day.

tom95858:
Is this the Intel toolchain or the GNU toolchain? IOW, did you load PrgEnv_Gnu before configuring/building?

ldms/src/ldmsd/ldmsd_request.c
narategithub: I agree with you. Shall we use `memcpy()` and suppress the warning if it came up instead? We already know the lengths of both source and destination.

narategithub:
This is in PrgEnv_Gnu environment. @nichamon and I also tried `CFLAGS='-Wall -Werror -O2` on CentOS, some of the warnings (that turned into errors) in this patch also came up. When we build with `-O0` or not giving any optimization option, these warnings didn't come up though.

ldms/src/ldmsd/ldmsd_request.c
tom95858: If we know the length of the source and the destination, then why isn't strcpy sufficient? Does gcc complain about strcpy? Also, I see a bunch of places where someone is using strncpy, but the length is dependent on the *source* buffer, not the destination. Those changes were no doubt made due to a warning and created a bug in the process. Does every use of strncpy create this warning, or only some. And if it's only some, why?

tom95858:
Let's start with the real issues, i.e. the potential use of an uninitialized variable. Put those in their own pull request, and then let's try and figure out what's up with strcpy/strncpy, etc... Also, there are a bunch of trailing spaces that you cleaned up. I wonder if we can add a pull check that will ensure a pull request is not adding white-space pollution.

narategithub:
OK. I'll separate the strcpy/strncpy out. I'll also add a github action check for trailing spaces.

tom95858:
I think the strncpy warning comes up if the size specified is equal the destination buffer size. To make it go away, I think you can just make the size argument sz - 1.

narategithub:
I tried, it didn't :laughing:

narategithub:
@tom95858 the `strcpy/strncpy/snprintf` have been removed from the patch. It is ready for you to review.

ldms/src/sampler/dstat/dstat.c
```diff
@@ -173,7 +173,11 @@ static char * compute_pidopts_schema(struct attr_value_list *kwl, struct attr_va
 	if (dosuffix) {
 		snprintf(hbuf, bsz, "_%x", pidopts);
 	}
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpragmas"
+#pragma GCC diagnostic ignored "-Wformat-truncation"
```
tom95858, line 178: I think this is a real bug? bsz is 21B. hbuf and buf are both 21B. If you concatenate those two you get 42B which results in the truncation warning.

narategithub: Thanks for catching this. Let me put in a proper fix.

narategithub:
@tom95858 the commit has been fixed up.

lib/src/zap/zap.c
```diff
@@ -1010,8 +1010,8 @@ static void init_atfork(void)
 	assert(zev_queue);
 
 	for (i = 0; i < zap_event_workers; i++) {
-		char thread_name[16];
-		(void)snprintf(thread_name, 16, "zap:worker:%d", i);
+		char thread_name[64];
+		(void)snprintf(thread_name, 64, "zap:worker:%d", i);
```
tom95858, line 1014: A pthread name can't be bigger than 16B (see pthread_setname_np). Probably just clamp the %d to a width of 4 and I think it will go away.

narategithub: Oh. Thanks for pointing that out. I'll fix it now.

lib/src/zap/zap.c
```diff
@@ -1011,7 +1011,7 @@ static void init_atfork(void)
 
 	for (i = 0; i < zap_event_workers; i++) {
 		char thread_name[16];
-		(void)snprintf(thread_name, 16, "zap:worker:%d", i);
+		(void)snprintf(thread_name, 16, "zap:wkr:%hd", (short)i);
```
narategithub, line 1014: @tom95858 I couldn't limit the space of int in the format (unlike float or string). So, I shorten the name "worker" -> "wkr" and specifying "%hd" (short) made the warning went away. I think the short int should be OK .. as the number of zap worker should be below 1,000.

ldms/src/core/ldms_xprt.c
```diff
@@ -3278,7 +3278,7 @@ int ldms_register_notify_cb(ldms_t x, ldms_set_t s, int flags,
 void ldms_xprt_set_delete(ldms_set_t s, ldms_set_delete_cb_t cb_fn, void *cb_arg)
 {
 	struct ldms_request *req;
-	struct ldms_rbuf_desc *rbd, *next_rbd;
+	struct ldms_rbuf_desc *rbd, *next_rbd = NULL;
```
baallan, line 3281: @narategithub It appears next_rbd might also be left pointing at something else inappropriate (but not null) from the previous while(rbd) loop when the goto next_1 occurs.
Does this init need to be moved to inside the top of the complaining loop?

narategithub: Nah. If the statements in the first loop got executed, `next_rbd` must be NULL, otherwise, `rbd` won't be NULL and we won't exit the loop.

Thanks for asking though. I'd just saw that the 2nd loop is incorrect. If the first `rbd->xprt` happened to be NULL, then `rbd = next_rbd` will be NULL at `next_1`. I'll add this in another commit on top of this.

narategithub: @baallan please see #619

PR #614, opened 2021-02-05
Improve code comments about ldmsd messages

morrone:
Update the code comments about the ldmsd message
Standard Record Format, and remove the information about
the no longer used String Record Format.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.h
+++ b/ldms/src/ldmsd/ldmsd_request.h
@@ -228,28 +228,24 @@ typedef struct ldmsd_req_hdr_s {
 #pragma pack(pop)
 
 /**
- * The format of requests and replies is as follows:
+ * The format of requests and replies is as follows.
+ *
+ * Note that "attr_id" is usually one of the values in
+ * enum ldmsd_request_attr. However, the values in ldmsd_request_attr
+ * do not have a prescribed meaning or data type. The meaning and data
+ * type are left to the implementation of each individual message.
  *
  *                 Standard Record Format
  * +---------------------------------------------------------+
  * | ldmsd_req_hdr_s                                         |
  * +---------------------------------------------------------+
- * | attr_0, attr_discrim == 1                               |
+ * | ldmsd_req_attr_s: discrim=1, attr_id=?                  |
  * +---------------------------------------------------------+
- * S                                                         |
- * +---------------------------------------------------------+
- * | attr_N, attr_discrim == 1                               |
- * +---------------------------------------------------------+
- * | attr_discrim == 0, remaining fields are not present     |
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_request.h
```diff
@@ -228,28 +228,24 @@ typedef struct ldmsd_req_hdr_s {
...
  * | ldmsd_req_hdr_s                                         |
  * +---------------------------------------------------------+
- * | attr_0, attr_discrim == 1                               |
- * +---------------------------------------------------------+
- * S                                                         |
- * +---------------------------------------------------------+
- * | attr_N, attr_discrim == 1                               |
- * +---------------------------------------------------------+
- * | attr_discrim == 0, remaining fields are not present     |
+ * | ldmsd_req_attr_s: discrim=1, attr_id=?                  |
```
tom95858, line 242: @morrone, to be clear attr_id specific data follows after the id.

morrone: I'm not sure what you are trying to clarify there. Did you want me to change something in my comment?

After the attr_id in ldmsd_req_attr_s are only "attr_len" and a union of multiple possible data types and array lengths. There is no specific information in the attr structure itself about what the data type might be. That can only be determined by reading the code of the associated message constructors and handlers. Or am I missing something?

Or maybe you just wanted to add the union { attr_value[], attr_u16[], attt_u32[], attr_u64[] } to each ldmsd_req_attr_s line in the comment for some reason? I though by making it clear that the structure was an ldmsd_req_attr_s (where no such information was present in the comment previously, we just had to read a bunch of code), it is already clear the the elements of ldmsd_req_attr_s are part of ldmsd_req_attr_s.

But I'm open to further clarification.

tom95858: I thought it might be informative to say after "attr_discrim=1,attr_id=?, attr_id specific data"

BTW, this is all modeled after sunrpc.

morrone: I'm not sure I can see a lot of sunrpc influence. For instance, sunrpc uses XDR to encode and decode the data, whereas the ldms rpc has it own ad hoc internal encoding rather than using an external data description language. Where would I look to find analogies in sunrpc?

The field is names "disrim", not "attr_discrim". That is one of the things I'm correcting in the documentation.

As to the data, whether or not it is actually present depends on whether attr_len is > sizeof(struct ldmsd_req_attr_s), does it not? Is there anything in the protocol that enforces that? If not, then the inclusion of attr_id specific data is optional. It may be easier to describe that in words than adding attr_len and the union to the diagram .

PR #618, opened 2021-02-09
Introduce metric units to LDMS

nichamon:


baallan:
Why are we shortening the space available for the metric name to accommodate the unit string?  Widen the fields, and if width is a problem, make display of units optional. (-U)

morrone (changes requested):
I think that struct ldms_value_desc in ldms_core.h needs to be changed to make its new contents clear. For instance, "vd_name" should probably be renamed to "vd_strings" or something more generic like that. (vd_name_len should be renamed to match). Then the code comment should clearly define what is found under that name. It should clearly explain that the vd_strings _may_ contain multiple strings, and explain what the different strings might be, and how one might parts the array along with vd_name_len to figure out how many strings are in the field.

In other words, there is no way that anyone looking at the definition of ldms_value_desc would guess that vd_name is potentially more than one string. We need to make that explicit rather than hiding that detail in the function implementations..

ldms/src/core/ldms.c
```diff
@@ -1032,11 +1032,13 @@ size_t __ldms_value_size_get(enum ldms_value_type t, uint32_t count)
 	return roundup(value_sz, 8);
 }
 
-void __ldms_metric_size_get(const char *name, enum ldms_value_type t,
-		uint32_t count, size_t *meta_sz, size_t *data_sz)
+void __ldms_metric_size_get(const char *name, const char *unit,
+			    enum ldms_value_type t,
+			    uint32_t count, size_t *meta_sz, size_t *data_sz)
 {
 	/* Descriptors are aligned on eight byte boundaries */
```
morrone, line 1039: ldms_value_desc is preceded by "pragma pack(4)". 4 means 4 bytes, right? Why are we assuming 8 byte/ 64-bit boundaries? Isn't that a bug?

tom95858: We're not assuming, we're requiring 4B alignment. These structures are RDMA'd around, so the consumer and producer have to have the exact same layout.

morrone: Right, 4B. But the code says 8B. Why is that?

tom95858:
> Why are we shortening the space available for the metric name to accommodate the unit string? Widen the fields, and if width is a problem, make display of units optional. (-U)

@baallan, @morrone, the reason for the "clever code" is because we require binary compatibility with earlier releases. So we can't change the size of these things or it will break.

@morrone, we can change the name of the field, but vd_strings is less informative in my opinion than vd_name.

@nichamon, did you look at using vd_user_data for the units string? I don't think this field is used. That would give us 7B (1 for the terminating '\0') for the unit string, which I think is sufficient.

morrone:
> > Why are we shortening the space available for the metric name to accommodate the unit string? Widen the fields, and if width is a problem, make display of units optional. (-U)
> 
> @baallan, @morrone, the reason for the "clever code" is because we require binary compatibility with earlier releases. So we can't change the size of these things or it will break.
> 
> @morrone, we can change the name of the field, but vd_strings is less informative in my opinion than vd_name.
> 
> @nichamon, did you look at using vd_user_data for the units string? I don't think this field is used. That would give us 7B (1 for the terminating '\0') for the unit string, which I think is sufficient.

I completely understood that this was for backwards compatibility reasons.

But I'm not sure that you got my point about renaming vd_name, because my point was more about change than advocating for a particular new name. My point is that it needs to be renamed so that people will not think it represents a single field when it in fact represents two fields. It also needs to be well documented how a single field is being overloaded to pack in two fields. Being too-clever-for-our-own good to avoid changing the protocol will bite us in the long term in terms of code complexity and technical debt, unless we take careful measures to avoid confusion.

Are we really sure that 7 bytes would be enough? Have we done a survey of the units that we currently use? I'm a little concerned that we'll wind up with some strange abbreviations that deprive us of the clarity that we would hope for by adding the units field.

ldms/src/core/ldms.c
narategithub: @morrone The structure itself if 4B pack. However, the structure has a variable length data. The start/end of the entire structure+data has to be 8B aligned due to RDMA requirement.

baallan:
@tom95858 I'm talking about (and just about) the io format to screen, not wire protocol. The currently proposed change is actual back-incompatible. Adding a -U to enable printing units only when present would preserve back-compatibility.

tom95858:
@morrone, I think I understand about changing the name and I think changing it is fine, maybe vd_name_units or somesuch. I think we need a comment too.

The support for metric units (in master) had a fixed size string of 8B that was part of the mdesc structure; since that fits in the unused field I mentioned it as a possibility. If we feel that's not long enough, @nichamon 's current implementation allows for that, but has this somewhat in-obvious embedded '\0'.

tom95858:
@baallan, please explain how exactly this is backward in-compatible.

baallan:
> 
> 
> @baallan, please explain how exactly this is backward in-compatible.

If I'm reading the screen and suddenly a new column appears and my old column is shorter and is truncating its contents unlike it did before, I'd call that back incompatible. That's what the string format change does.

nichamon:
> @nichamon, did you look at using vd_user_data for the units string? I don't think this field is used. That would give us 7B (1 for the terminating '\0') for the unit string, which I think is sufficient.

Hi @tom95858 ,

There are configuration commands to set user data. I don't have information on whether users are using the commands to set the user data. I prefer not to have a length-restriction on the unit lengths. Historically, the length limitation has bitten us a few times, and we had to change the data structure/code.  Do you have any concerns over long unit strings?

ldms/src/ldmsd/ldms_ls.c
```diff
@@ -321,18 +321,22 @@ void metric_printer(ldms_set_t s, int i)
...
+	const char *metname, *metunit;
 	if (type != LDMS_V_NONE) {
 		metname = ldms_metric_name_get(s, i);
+		metunit = ldms_metric_unit_get(s, i);
 	} else {
 		metname = "SET_ERROR";
+		metunit = "";
 	}
-	if (user_data)
-		sprintf(name_str, "%-42s 0x%" PRIx64,
```
nichamon, line 331: > @tom95858 I'm talking about (and just about) the io format to screen, not wire protocol. The currently proposed change is actual back-incompatible. Adding a -U to enable printing units only when present would preserve back-compatibility.

Hi @baallan ,

`ldms_ls` is backward-compatibility. The change passes the compatibility test on Github, which tests the result format printed by the `ldms_ls`. In more detail, `ldms_ls` in this patch printed the result the same way as the current version, including spaces between the metric name string and the metric value. You could use the printing formats on line 331 (current version) and line 333 (in the patch) as a reference.

nichamon:
> @morrone, I think I understand about changing the name and I think changing it is fine, maybe vd_name_units or somesuch. I think we need a comment too.
> 
> The support for metric units (in master) had a fixed size string of 8B that was part of the mdesc structure; since that fits in the unused field I mentioned it as a possibility. If we feel that's not long enough, @nichamon 's current implementation allows for that, but has this somewhat in-obvious embedded '\0'.

@tom95858 @morrone 

I think we are on the same page that the field `vd_name` and `vd_name_len` should be renamed. I will change them to `vd_name_unit` and `vd_name_util_len`. I will add comments to clarify the field definitions.

ldms/src/ldmsd/ldms_ls.c
baallan: @nichamon 
How does replacing: 
sprintf(name_str, "%-42s 0x%" PRIx64, metname,...
with 
sprintf(name_str, "%-27s %-14s -0x%" PRIx64, metname, metunit
maintain compatibility? 
You've shortened the maximum length that will be output on a metric name (a problem we've had past problems with in things like the lustre samplers).
You've introduced a new column, which (if I parse based on whitespace (a very frequent thing) and not character-column number) will be pushing unit data into a parse routine that expects udata value or something else.
If this passed a 'back compatibility test', the test was not well formulated.

Back compatibility for _ls would be "do i get exactly the same output before and after the change".
If I want additional (new) columns, I should have to pass a new option or perhaps a multiple of an old one. -U? "-l -l"?

nichamon: @baallan, I thought your only concern is the appearance of the `ldms_ls` result on the screen when the metric unit doesn't present.

Now let me make sure that I have got all of your concerns.

- To your knowledge, there are metrics with names longer than 26 characters. (I guess this because you mentioned your past Lustre problem)
- You already have use cases that parse the `ldms_ls` results. Is this correct? Are the use cases in your test suite?

Did I miss anything?

baallan: There are metrics with long names (>26) ; there will be metrics with longer names coming, as we expect them from C++ function names in user-defined performance sets. Truncating the names must be avoided, or the name length limit must be enforced as a failure at set definition time (adding a limit I don't recommend having).  We already implicitly (unfortunately not explicitly, which should be fixed in a separate patch) disallow using whitespace and comma in metric names to keep down-stream tools happy.

I have production scripts that parse ldms_ls output that one would expect to break under the change if samplers which naturally have units available (e.g. meminfo, lustre) displayed in the revised format automatically.  These cases are not in the ldms-static-test.sh suite. 

Most automated scripting will already have assumed the units that go with the metric they are examining via ldms, so I'm not sure I see the value in changing the default behavior of ldms_ls. Adding units as optional behavior is fine. Are the units intended to be mutable strings like udata, or to be constants like metric_name?

nichamon: Thanks for the information.

> There are metrics with long names (>26)

I will keep the space for the metric name at 42 characters then.

> Most automated scripting will already have assumed the units that go with the metric they are examining via ldms, so I'm not sure I see the value in changing the default behavior of ldms_ls.

I don't follow your logic at all. What are the automated scriptings you referred to here? How do they relate to `ldms_ls` at all?

> I have production scripts that parse ldms_ls output that one would expect to break under the change if samplers which naturally have units available (e.g. meminfo, lustre) displayed in the revised format automatically. 

There are a couple of things to address here. 

- Will `ldms_ls` report metric units when they present?
- Should we add/change the cmd-line options to configure how `ldms_ls` print metric units?
- What format will `ldms_ls` print the metric information?

My answer to the first question is yes, `ldms_ls` should report metric units if they are defined. I answered based on the `ldms_ls` program's objective, which gives set information to users, and the significance of metric units. Metric units are important information to understand a metric, which we have been missing for a long time. We design `ldms_ls` to mainly display set data on the screen.

My answer to the second issue is not to add/change options to disable/enable `ldms_ls` printing metric units. Yes, it could be a pain in the neck to update the existing scripts, primarily since this could involve changing >100 lines of code. However, this should not be a reason to complicate `ldms_ls` usage by adding/changing the cmd-line options. Also, it should not be a reason that prevents `ldms_ls` from reporting informative results, i.e., not reporting metric units.

To address the metric information format, we could decide on the format that is easy to read on the screen and avoid changing 100 lines of code in the existing tools. How large would the code change in your scripts be if the metric unit is added with a space as the delimiter?

nichamon: I came up with another format that doesn't add a new column when metric units present. The metric unit follows the metric name in the parentheses. If the metric unit doesn't exist, the result will look the same as the current version, see M2.

```
$ ldms_ls -x sock -p 10001 -l
set_0: consistent, last update: Tue Feb 16 04:51:50 2021 +0000 [1284us]
D u64        M1(Bytes)                                  3
D u64        M2                                             3

```

nichamon:
@tom95858 @morrone I have addressed the field names and added comments as well as the API documents.

ldms/src/ldmsd/ldms_ls.c
baallan: This is not a satisfactory solution as we have metrics with names that already end with (xxxxxxx). please make units a separate column and make that column controllable by a format control flag.
Unix 'ls' has numerous format control flags, and there's no reason ldms_ls shouldn't.

If we're on a spree to make ldms_ls compatible and useful and extensible, please consider the example of slurm 'sacct' which supports a "machine parseable" mode. In machine parseable mode, columns are separated by a documented character (| in the case of sacct; we could use \t instead) and a corresponding column header is optionally produced using the same separator. If you aren't using "parseable" mode, you get no expectation of back compatibility.  If you add a documented parseable mode simultaneously, I withdraw any objections  about  whimsical human-oriented formatting. Notice in the sacct example that you get a "format" option to specify which columns you want (for either parsable or whimsical format).

baallan:
> I do think it is reasonable to have a flag that allows us to export data as CSV or JSON for example. This would make it much easier to write and maintain scripts that process command output.

So to be specific about something that has proved very useful:
sacct -o FORMAT -p 
produces something very import friendly to spreadsheets and python/perl tabular parsing in spite of a lot of gunk that may appear inside the report columns. FORMAT defines the columns selected, not tabular vs json. In particular, commas and whitespace do not interfere because the pipe separator is used. See 'man sacct'. Arguably, we could support "-o format -f<json|tabular[:separator]>" since -p is already taken and we should probably reserve -d for debugging output not yet provided.

tom95858:
@baallan, @nichamon , then let's please take a look at sacct and do this work in a different patch? I think it will be a bit substantial.

baallan:
A separate patch to add sacct-style output make sense to me. I would prefer to see it merged/released at the same time as anything that modifies ldms_ls output, so that the downstream repair to filters broken by changes to ldms_ls human format are all addressable by going directly to the machine parsable formats instead of needing another iteration of ldms_ls human output parsing.

nichamon:
@tom95858 IMO, any changes to `ldms_ls` that do not relate to metric units should go to another patch.

baallan:
@valleydlr Given that very few of the data sources actually supply units, why are we adding units to the data collection infrastructure? What requirement is this addressing? It seems that the efficient solution would be to supply a separately curated database that can be independently queried with a key of plugin_name:metric_name for the units and many other useful attributes.

narategithub:
Part of #783

PR #619, opened 2021-02-09
Fix rbd leak/infinite loop/crash in `ldms_set_delete()`

narategithub:
`next_rbd` was not set before jumping to `next_1` label. If the first
`rbd` did not have `xprt`, then the loop stopped prematurely (as
`rbd = next_rbd` being NULL). If the first `rbd` had `xprt`, but one of
the other `rbd` did not have `xprt`, `rbd` value would stay the same
since `next_rbd` value won't change and `rbd = next_rbd`. This resulted
either in an inifinite loop (rbd won't be NULL) or a crash (too many
`ref_put()` on the `rbd`).


narategithub:
This has passed all test cases in `ldms-test`.

tom95858 (approved):
:+

PR #620, opened 2021-02-09
Update the test_ldms_push test program

nichamon:
The patch addresses two issues.
- The range of the update status has been changed, but the test hasn’t
been updated accordingly. Without the changes, some test conditions are
always true, and some are always false.
- As clients, the test program did not use the given host value to connect
to servers. Clients could not connect to a server on a remote machine.


tom95858 (approved):
👍

PR #621, opened 2021-02-10
Resolve `-O2` warnings on `strncpy`

narategithub:
On Cray Linux Environment (CLE) 7.0.UP02, gcc-9 seems to be the
preferred GNU C compiler. When '-O2' is supplied to the CFLAGS,
`-Werror=stringop-truncation` (which is not present in GCC 4.8.5 on
CentOS 7) is also enabled and gives the following warnings in many
places in our source tree (which results in error by `-Werror`):
```
    error: ‘strncpy’ specified bound <NUMBER> equals destination size
```

This patch addresses places where the warnings occur, and provides
length checking if necessary.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -2816,7 +2816,8 @@ static int __rbd_rbt_key_comp(void *tree_key, const void *key)
 void __ldms_xprt_init(struct ldms_xprt *x, const char *name,
 					ldms_log_fn_t log_fn)
 {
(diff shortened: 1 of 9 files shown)


narategithub:
This pull request has been build-test with `CFLAGS='-O2 -Wall -Werror'` on: CentOS7, openSUSE15 (gcc-9) and CLE7. The binaries build on CentOS7 are also tested against all test cases in `ldms-test` repository.

tom95858 (approved):
👍

PR #622, opened 2021-02-10
Implement tx2mon sampler plugin

Snell1224:
Adding a sampler called tx2mon. This sampler is designed to run on Stria because this machine contains the kernel module "tx2mon_kmod" that produces the decoded sysfs files that are used to map to operating region metrics. 

As with the tx2mon utility, the tx2mon sampler reports cpu and system-on-chip information from the node files located in the /sys/bus/platform/devices/tx2mon/[socinfo, node<i>_raw], decodes and provides the metric values of the SoC management controller (MC) operating region, current temperature from sensors and the frequencies and power measurements that is logged by the MC to host memory. 

The tx2mon sampler reports the same units and variables as the tx2mon command-line utility. 
The sampler contains two options where:
1.) User can enable/disable set system metrics and additional information of the internal block frequencies. (extra=true/false)
2.) User can view the frequency and temperature readings of each core or only the maximum and minimum frequency and temperature of all cores. (array=true/false)

Test and auto-schema naming option scripts have been created for ldms-static-test.sh.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -352,6 +352,20 @@ AS_IF([test "x$enable_papi" != xno],[
 AM_CONDITIONAL([HAVE_LIBPAPI], [test "x$HAVE_LIBPAPI" = xyes])
 AM_CONDITIONAL([HAVE_LIBPFM], [test "x$HAVE_LIBPFM" = xyes])
 
+# enable tx2mon by default only when sensible.
+AC_ARG_ENABLE([tx2mon],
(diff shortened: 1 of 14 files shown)


baallan:
@tom95858 please review or delegate as appropriate so we can include this in the upcoming builds for thunder x2 platforms (stria/astra). I have tested it, including configure script changes and valgrind runs, on stria nodes with and without tx2mon present in the OS.

PR #623, opened 2021-02-11
Fix secondary group authentication

morrone:
Secondary group authentication is broken in 4.3.5+. A fairly innocuous looking deduplication of code revealed a bug in the implementation to which we switch. This fixes the bug and cleans up another function that was made redundant in the original patch.


tom95858:
Thanks @morrone, good catch.

PR #626, opened 2021-02-11
zap close to wait for empty sq

narategithub:
Modify zap close to wait for empty send queue. This is to ensure
that the send operations requested by the application got submitted
to the underlying socket before the socket get destroyed by close().

This patch has been tested with test cases in `ldms-test` repository.

### DIFF
--- a/lib/src/zap/fabric/zap_fabric.c
+++ b/lib/src/zap/fabric/zap_fabric.c
@@ -734,6 +734,7 @@ static void flush_io_q(struct z_fi_ep *rep)
 		rep->ep.cb(&rep->ep, &ev);
 		__context_free(ctxt);
 	}
+	pthread_cond_signal(&rep->io_q_cond);
 }
 
 /*
@@ -869,6 +870,7 @@ static void submit_pending(struct z_fi_ep *rep)
 		if (post_wr(rep, ctxt))
 			__context_free(ctxt);
 	}
+	pthread_cond_signal(&rep->io_q_cond);
  out:
 	pthread_mutex_unlock(&rep->credit_lock);
 }
@@ -921,6 +923,15 @@ static zap_err_t __post_send(struct z_fi_ep *rep, struct z_fi_buffer *rbuf)
 static zap_err_t z_fi_close(zap_ep_t ep)
 {
 	struct z_fi_ep *rep = (struct z_fi_ep *)ep;
+	pthread_t self = pthread_self();
+
+	if (self != ep->event_queue->thread) {
+		pthread_mutex_lock(&rep->credit_lock);
+		while (!TAILQ_EMPTY(&rep->io_q)) {
+			pthread_cond_wait(&rep->io_q_cond, &rep->credit_lock);
+		}
+		pthread_mutex_unlock(&rep->credit_lock);
+	}
 
 	pthread_mutex_lock(&rep->ep.lock);
 	if (!rep->cm_fd)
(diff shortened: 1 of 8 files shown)


narategithub:
@tom95858 I think this change has a high risk of deadlock. `ldms_xprt_close()` which calls `zap_close()` actually get called from various places in the zap callback path. Since `zap_close()` is blocked in the thread working on the callback which also works on processing sq entries, the sq may never be emptied. Maybe we have to wait for the send completion patch instead?

tom95858:
Perhaps the upcall path in zap_sock.c should not call zap_close directly; it should call an internal function that doesn't wait. The case where we have an issue is when the application calls exit() immediately after calling zap_close(). To test out that theory, could add a zap_close_wait() function and see that it does what you expect.

narategithub:
@tom95858 I made the change so that zap sock close would not block if it is called from the endpoint callback chain (zap worker thread of the endpoint). I think this should suffice .. (but I may be wrong :sweat_smile:).

Adding `zap_close_wait()` implies `zap.close_wait()` for each transport, and `ldms_xprt_close_wait()` too, right? Seems like a bigger change.

I'll remove the draft status so that you can merge if it looks OK.

narategithub:
@tom95858 I've updated other transports + sniff-tested them.

tom95858:
@narategithub, how much testing has this change had? Is it ready to be merged? Have we done any scalability testing on it?

tom95858:
It looks like this needs to be rebased.

narategithub:
@tom95858 I've not scale-test it yet. I'll do the sock & rdma on Cygnus. @nichamon  will help test this out on uGNI.

narategithub:
@tom95858 Updates:

I use the following setup on cygnus:
- 100 samplers on cygnus-01..04 (400 samplers total), each sampler has 1 test set
- 4 agg1, each connect to 100 samplers
- 2 agg2, each connect to 2 agg1
- 1 agg3, connect to the 2 agg2

sampler daemons flap (restart) every 1 minute. agg2 flap every 5 cycle of sampler flaps.

sock seems OK (ran for about 20 minutes).
rdma has been running for about the same time now .. still OK (data on agg3 is up-to-date). I'll let it run overnight and check again tomorrow (rdma).

@nichamon is still working on testing this on uGNI xprt.

narategithub:
@tom95858 zap_rdma is good. @nichamon is working on uGNI test.

nichamon:
The patch passes the same tests as I ran for the fix to `ldmsd_stream_publish_file()` https://github.com/ovis-hpc/ovis/pull/644#issuecomment-787014214. It passes both for `zap_sock` and `zap_ugni`.

nichamon:
I ran into a problem when testing `zap_rdma`. `ldmsd_stream_publish` hang at exit because `zap_rdma` blocked at `pthread_join()` in `zap_rdma_fini`. The problem is irrelevant from the patch, but it prevents me from completing the test. I'll modify my test script to handle the problem.

nichamon:
The patch also passed the test when using `zap_rdma`.

nichamon:
For zap_fabric, `ldmsd_stream_subscribe` didn't receive all the stream data sent by `ldmsd_stream_publish` regardless of the stream size, smaller or bigger than the max msg length. @narategithub and I have been investigating this.

tom95858:
@nichamon, is this ready to go? Should we fix the pthread_timejoin_np in this patch and retest with RDMA?

narategithub:
@tom95858 `pthread_timedjoin_np()` has been added (commit 8144b208785e0eeaa165986b612166f96af7b284) into this pull request.

nichamon:
I think this is ready to be merged.

I retested `zap_rdma` with the `pthread_timejoin_np()`. I ran `ldmsd_stream_publish` 3000 times, and I could not reproduce the hang. I would say that the `zap_rdma` close and `pthread_timedjoin_np` patches are good.

Narate and I also tested `zap_fabric` in this patch, and `ldmsd_stream_subscribe` receives all streams sent by `ldmsd_stream_publish`. I ran `ldmsd_stream_publish` 3000 times, and `ldmsd_stream_subscribe` received all the streams data.

> For zap_fabric, ldmsd_stream_subscribe didn't receive all the stream data sent by ldmsd_stream_publish regardless of the stream size, smaller or bigger than the max msg length. @narategithub and I have been investigating this.

My test script was the cause of the problem above.

PR #629, opened 2021-02-12
Set the effective uid/gid to the peer's uid/gid

tom95858:
This security enhancement ensures that a connected peer cannot cause the
ldmsd to undertake any action with elevated privilege.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1971,21 +1971,22 @@ static int ldms_xprt_recv_reply(struct ldms_xprt *x, struct ldms_reply *reply)
 
 static int recv_cb(struct ldms_xprt *x, void *r)
 {
+	int rc, euid_set = 0;
 	struct ldms_request_hdr *h = r;
 	int cmd = ntohl(h->cmd);
+	uid_t euid = geteuid();
+	if ((int)x->ruid > 0 && euid == 0) {
+		euid_set = 1;
+		seteuid(x->ruid);
+	}
 	if (cmd > LDMS_CMD_REPLY)
 		return ldms_xprt_recv_reply(x, r);
 
-	return ldms_xprt_recv_request(x, r);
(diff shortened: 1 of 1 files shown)


baallan:
Does this change not update the uid for all threads in the daemon? Doesn't that stop any sampler which requires euid 0 to access its /proc entries to fail from the moment of the id change?

tom95858:
The short answer is that /proc does not require root access although it is possible to mount /proc with options that restrict access to /proc/pid. I did, however, mislabel this commit, only the uid is set, not the gid.

baallan:
Let me revise my question to include "euid 0 to access its /proc entries or other sampled devices"

I'm thinking in particular of sysclassib/ibnet, opa2, tx2mon, lustre (/sys/kernel/debug/lustre/*)

baallan:
@tom95858 this generates (would fail with -Wall -Werror) on rhel7 (below):
You might want to check the return code and log something at INFO level, since it's sort of an expected case to run unprivileged.

../../../../ldms/src/core/ldms_xprt.c: In function 'recv_cb':
../../../../ldms/src/core/ldms_xprt.c:1980:10: warning: ignoring return value of 'seteuid', declared with attribute warn_unused_result [-Wunused-result]
   seteuid(x->ruid);
          ^
../../../../ldms/src/core/ldms_xprt.c:1987:10: warning: ignoring return value of 'seteuid', declared with attribute warn_unused_result [-Wunused-result]
   seteuid(euid);

baallan:
@tom95858 two potential problems with this patch.
1) assuming a typical attack on a complex code (arbitrary stack corruption achieved somewhere during the callback which is wrapped in the seteuid/revert-euid), this patch doesn't prevent the attack's subsequent privileged action after the euid-managing stack frame returns.
2) the approach is racy: all the threads in the daemon get their privileges changed, while some of them are sampling or storing using privilege to open devices.

PR #630, opened 2021-02-12
Stream performance improvements

tom95858:
These changes significantly reduce the overhead of publishing
JSON data if that data is being forwarded and not otherwise consumed
at the subscriber.

The publisher is also acknowledged before delivering the data to the
subscribers. This reduces the latency at the producer waiting
for the ACK.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -102,6 +102,7 @@ void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap);
 
 static char * __thread_stats_as_json(size_t *json_sz);
 static char * __xprt_stats_as_json(size_t *json_sz);
+extern const char *prdcr_state_str(enum ldmsd_prdcr_state state);
 
 __attribute__((format(printf, 1, 2)))
 static inline
@@ -1697,7 +1698,6 @@ send_reply:
 
 int __prdcr_status_json_obj(ldmsd_req_ctxt_t reqc, ldmsd_prdcr_t prdcr, int prdcr_cnt)
 {
-	extern const char *prdcr_state_str(enum ldmsd_prdcr_state state);
 	ldmsd_prdcr_set_t prv_set;
 	int set_count = 0;
 	int rc = 0;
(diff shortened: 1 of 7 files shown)


nichamon:
@tom95858 Could you please rebase this on top of the latest OVIS-4? There are many changes I believe are unrelated to the patch.

tom95858:
Sure, testing now...

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6012,50 +6009,38 @@ static int stream_publish_handler(ldmsd_req_ctxt_t reqc)
 		goto err_reply;
 	}
 
+	reqc->errcode = 0;
+	ldmsd_send_req_response(reqc, "ACK");
```
nichamon, line 6013: LDMSD will send two responses back to the requester if no data is provided. Should we validate the stream before sending the ACK response (we won't validate if the JSON-formated string can be parsed or not)?

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -44,57 +46,90 @@ struct ldmsd_stream_s {
...
 			goto err_2;
-		}
 		pthread_mutex_init(&s->s_lock, NULL);
 		LIST_INIT(&s->s_c_list);
 		rbn_init(&s->s_ent, (char *)s->s_name);
+		pthread_mutex_lock(&s_tree_lock);
 		rbt_ins(&s_tree, &s->s_ent);
-	} else {
-		s = container_of(rbn, struct ldmsd_stream_s, s_ent);
+		pthread_mutex_unlock(&s_tree_lock);
```
nichamon, line 131: `pthread_mutex_lock(&s->s_lock)` is missing after the call to `pthread_mutex_unlock(&s_tree_lock)`.

nichamon (changes requested):

ldms/src/ldmsd/ldmsd_stream.c
tom95858: This is the path where the stream is newly allocated and therefore not locked. Am I missing something?

ldms/src/ldmsd/ldmsd_request.c
tom95858: The purpose of the change is to reduce the overhead of streams data if it is only being forwarded and not delivered. If we validate the data, ldmsd will incur the CPU hit even it is not delivering the data to any client. Good catch on the dup response. I'll fix it.

ldms/src/ldmsd/ldmsd_stream.c
tom95858: Ah...ok, I misread your comment.

PR #633, opened 2021-02-16
Fix a bug in zap_ugni that prevents active close transports

nichamon:
zap_ugni moves the endpoint state to CLOSE before going to the switch
cases that determine whether to shut down the socket or not. The bug
causes zap_ugni not to shut down the socket in any case. This results in
applications cannot actively close Zap ugni transports.

### DIFF
--- a/lib/src/zap/ugni/zap_ugni.c
+++ b/lib/src/zap/ugni/zap_ugni.c
@@ -513,13 +513,10 @@ static zap_err_t z_ugni_close(zap_ep_t ep)
 	DLOG_(uep, "Closing xprt: %p, state: %s\n", uep,
 			__zap_ep_state_str(uep->ep.state));
 	pthread_mutex_lock(&uep->ep.lock);
-	uep->ep.state = ZAP_EP_CLOSE;
 	switch (uep->ep.state) {
 	case ZAP_EP_LISTENING:
 	case ZAP_EP_CONNECTED:
 	case ZAP_EP_PEER_CLOSE:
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 With the bug, `z_ugni_close` won't shut down the socket, so `zap_ugni` won't deliver the disconnect event to applications. In the LDMSD layer, `prdcr_stop` will cause producers to be stuck in the STOPPING state.

tom95858 (commented):
I guess there must be a race here right? So what was happening is that setting EP_CLOSE before shutdown meant that the sock_event  callback would wake up and find the state already closed. Moving the state change to after the shutdown allowed for sock_event to find the event in one of the other states; but that is a race, right? sock_event is called in another thread; if sock_event is delayed for some reason, the original problem still exists. But maybe I'm missing something?

nichamon:
I missed the race. However, the bug that causes zap_ugni to not shutdown the socket at all is still there. I'll push an update to address the race and the not shutting down problem shortly.

nichamon:
At another look, we don't have the race between`z_ugni_close` and `ugni_sock_event`. Both paths held the endpoint lock. In my patch, `z_ugni_close` moves the endpoint state to CLOSE while it holds the endpoint lock. `ugni_sock_event` takes the lock before checking the endpoint state. There is no way that `ugni_sock_event` will find the state in the other states besides CLOSE if `z_ugni_close` is called before `ugni_sock_event` awake. If there is a race regarding the endpoint state in another path, it must be that that path does not take the endpoint lock before checking the endpoint state, right?

PR #635, opened 2021-02-17
Include auto_interval test scripts

Snell1224:
Include test scripts for the "auto_interval" option when updating the aggregator daemons. 

There are 3 sampler daemons with interval hints of 1sec (fast), 5secs (medium), and 10 secs (slow) and two aggregator daemons that aggregate the data at 1 sec and 10 secs, respectively. 

Test scripts have been checked will Valgrind.

### DIFF
--- a/ldms/scripts/examples/.canned
+++ b/ldms/scripts/examples/.canned
@@ -1,4 +1,5 @@
 # working
+auto-interval
 aggl2 # ldms_ls inconsistent, set duplicates at L2
 aggl2simple
 all_example
--- a/ldms/scripts/examples/auto-interval
+++ b/ldms/scripts/examples/auto-interval
@@ -1,0 +1,28 @@
+export plugname=meminfo
+portbase=61200
+VGARGS="--leak-check=full --track-origins=yes"
+DAEMONS $(seq 1 5)
+vgoff
+LDMSD 1 2 3 4 5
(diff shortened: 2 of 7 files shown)


baallan:
[patch.txt](https://github.com/ovis-hpc/ovis/files/6006539/patch.txt)
After running the new test, attached are suggested changes with the aims of:

- extending test length to make sure the csv files output slow/medium data.
- including test case in the list output by ldms-static-test.sh -l.
- making the initial updater intervals 2,15 so that if they unexpectedly stick, it is obvious in the csv timestamp deltas.

Does the script that runs most cases as part of make check pick up this new one automatically, or need a modification?

Snell1224:
Okay I can make these changes. 
And if you are talking about ldms-run-static-tests.sh then no, I'll need to add auto_interval to the test case list in order to be tested during make check....Come to think of it, I'll also need to add the tx2mon sampler as well so thanks for the reminder!.

tom95858:
Why aren't we adding these tests to the ldms-test repository instead of here in the examples directory?

baallan:
These tests and the rest of the "make check/example" scripts are executable as part of a build on any deployment platform. It may be appropriate do something similar in the separate ldms-test repository or another containerized testing environment, but that does not eliminate the need for them as part of 'make check'. You are welcome to clone the logic to another project where appropriate.

PR #636, opened 2021-02-18
preserve command line format in ldmsd at runtime.

baallan:
This applies strdup to keep -x and -A options from being edited in place, making ps output more useful. 
It also makes the -A error messages more useful by echoing the offending arg string.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -1657,6 +1657,7 @@ int main(int argc, char *argv[])
 	}
 
 	opterr = 0;
+	char * dup_auth = NULL;
 
 	while ((op = getopt(argc, argv, FMT)) != -1) {
 		switch (op) {
@@ -1814,23 +1815,39 @@ int main(int argc, char *argv[])
 			break;
 		case 'A':
 			/* (multiple) auth options */
-			lval = strtok(optarg, "=");
+			dup_auth = strdup(optarg);
+			if (!dup_auth) {
+				printf("Not enough memory parsing options!\n");
+				exit(1);
+			}
+			lval = strtok(dup_auth, "=");
 			if (!lval) {
-				printf("ERROR: Expecting -A name=value\n");
+				printf("ERROR: Expecting -A name=value. Got %s\n",
+					optarg);
+				free(dup_auth);
 				exit(1);
 			}
 			rval = strtok(NULL, "");
 			if (!rval) {
-				printf("ERROR: Expecting -A name=value\n");
+				printf("ERROR: Expecting -A name=value. Got %s\n",
+					optarg);
+				free(dup_auth);
 				exit(1);
 			}
 			if (auth_opt->count == auth_opt->size) {
-				printf("ERROR: Too many auth options\n");
+				printf("ERROR: Too many (> %d) auth options %s\n",
+					auth_opt->size, optarg);
+				free(dup_auth);
 				exit(1);
 			}
-			auth_opt->list[auth_opt->count].name = lval;
-			auth_opt->list[auth_opt->count].value = rval;
+			auth_opt->list[auth_opt->count].name = strdup(lval);
+			auth_opt->list[auth_opt->count].value = strdup(rval);
+			if (!auth_opt->list[auth_opt->count].name || !auth_opt->list[auth_opt->count].value) {
+			    printf("Out of memory parsing auth options: %s\n", optarg);
+			    exit(1);
+			}
 			auth_opt->count++;
+			free(dup_auth);
 			break;
 		case 'n':
 			snprintf(myname, sizeof(myname), "%s", optarg);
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd.c
```diff
@@ -1814,23 +1815,37 @@ int main(int argc, char *argv[])
...
+				printf("ERROR: Expecting -A name=value. Got %s\n",
+					dup_auth);
+				free(dup_auth);
 				exit(1);
 			}
 			rval = strtok(NULL, "");
 			if (!rval) {
-				printf("ERROR: Expecting -A name=value\n");
+				printf("ERROR: Expecting -A name=value. Got %s\n",
+					dup_auth);
```
tom95858, line 1835: I don't think this is doing what you expect. Is there anyway that this can fail unless dup_auth is not '\0' terminated?

ldms/src/ldmsd/ldmsd.c
```diff
@@ -1814,23 +1815,37 @@ int main(int argc, char *argv[])
...
+				free(dup_auth);
 				exit(1);
 			}
 			if (auth_opt->count == auth_opt->size) {
-				printf("ERROR: Too many auth options\n");
+				printf("ERROR: Too many auth options %s\n",
+					dup_auth);
```
tom95858, line 1841: Did you test these error paths? If everything works above, dup_auth will be the 1st word in the string that precedes the '=' sign. You should be dumping optarg in all these cases I think.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -1814,23 +1815,37 @@ int main(int argc, char *argv[])
 			break;
 		case 'A':
 			/* (multiple) auth options */
-			lval = strtok(optarg, "=");
+			
+			dup_auth = strdup(optarg);
+			if (!dup_auth) {
+				printf("Not enough memory parsing options!\n");
+				free(dup_auth);
```
tom95858, line 1822: The very next line is exit(1). This just makes the arg parsing harder to read and bloats the code. I know you're trying to quiet valgrind, but at what cost?

ldms/src/ldmsd/ldmsd.c
```diff
@@ -2094,12 +2109,18 @@ int main(int argc, char *argv[])
 		case 'x':
 			if (check_arg("x", optarg, LO_NAME))
 				return EINVAL;
+			char *dup_xtuple = strdup(optarg);
+			if (!dup_xtuple) {
+				printf("out of memory parsing arguments\n");
+				return ENOMEM;
```
tom95858, line 2115: It seems like this should be cleanup(some-error, error-string)

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd.c
baallan: I'm just following the existing return/cleanup convention of the function. refactoring error handling overall in main seems to me out of scope for just fixing nits.

ldms/src/ldmsd/ldmsd.c
baallan: nothing to do with valgrind. just a mistake I will fix in freeing null unnecessarily. Certainly all calls to strdup should be checked for success, though.

ldms/src/ldmsd/ldmsd.c
baallan: dumping optarg would be better here. Adding the count that has been exceeded would be better also.

ldms/src/ldmsd/ldmsd.c
baallan: My understanding of Narates check here is that we are parsing " -A arg" where arg is expected to be of the form $x=$y. lval becomes $x (which if = is missing becomes the entire arg). If arg is of the form "$x=" or just "$x", then the 2nd call to strtok returns NULL, according to the man page and we will have rval=NULL.

baallan:
@tom95858 @narategithub this has been revised for most of Tom's comments and the whitespace bug.

ldms/src/ldmsd/ldmsd.c
tom95858: The dup_auth should be optarg I think.

baallan: @tom95858 it is optarg. is your browser caching something? dup_auth is used in no print statements now.

ldms/src/ldmsd/ldmsd.c
tom95858: This code would be a lot simpler if we just had two stack variables to hold these strings.

```
char xtuple_arg[80];
char auth_arg[128];
```
or some such. All those out of memory checks would just go away.

baallan: That would be trading straightforward malloc/free/oom for a bunch of array overrun/input truncation conditions checking. Not a win.

PR #641, opened 2021-02-24
V436 genders example update (makefile.am)

baallan:
fix unaddressed review issue from #600.


baallan:
@tom95858 I added the file list update @morrone pointed out as missing.

PR #644, opened 2021-02-26
Fix ldmsd_stream_publish not send all data

nichamon:


nichamon:
@tom95858  I haven't had a chance to test this yet. I'll continue tonight and over the weekend. 

Without the patch, I cannot reproduce the issue that the ldmsd_stream_publish program doesn't send all the data when it is >64k. In the test, the ldmsd_stream_publish program sent a stream of 85M to ldmsd_stream_subscribe. I tested this on Voltrino with zap_sock, so I'll see how I could make sure that the patch really addresses the issue.

 Also, I need to do some magic before I can use uGNI on Voltrino.

nichamon:
@tom95858 I can reproduce the problem that ldmsd_stream_publish may not send all the data.

@narategithub  and I tested the patch and it fixes the problem.

We tested the following scenarios.

- `ldmsd_stream_publish` program publishes a JSON stream to the `ldmsd_stream_subscribe` program.
- `ldmsd_stream_publish` program publishes a string stream to the `ldmsd_stream_subscribe` program.
- `ldmsd_stream_publish` program published a JSON stream to LDMSD using the `test_stream_sampler` plugin in the TADA repo.
- `ldmsd_stream_publish` program published a string stream to LDMSD using the `test_stream_sampler` plugin in the TADA repo.

For each scenario, the data sizes and the number of repetitions are as follows.

- 5kB repeated 1000
- 7M repeated 100

Narate implemented test scripts that compare the output with the input data. ldmsd_stream_subscribe program and the LDMSD sampler plugin received all data in all test runs.

PR #645, opened 2021-02-26
remove csv store file event notification support

baallan:
This removes the csv file event notification functionality (and attendant ovis_util/notification.*). This functionality is redundant with the newer "rename-on-close" functionality of store_csv. Anything done with notification is done more reliably (and with less ldmsd complexity) using the rename with a template argument that moves the closed files into another directory (spool). Any independent utility can watch the spool directory for new files and know that they are complete when they appear as the result of an mv. This change eliminates race conditions inherent in combining file renames with file event notifications.

This change is back incompatible with v3 storage configurations which used the notify functionality. No known v4 installations use the these file event notification hooks. The hooks were originally added because rollover did not support rename/mv and NFS3 does not support inotify.


tom95858:
@baallan, please fix the pull request checks. @oceandlr, is this change ok with you?

baallan:
@tom95858 looks like all are passing. whitespace fixed.

oceandlr:
I haven't looked since I was not an assignee. I can do so tonight....

oceandlr:
I did not get to this yesterday. @baallan proposed removing this functionality. That seemed reasonable and it appears to be encapsulated, so the removal is probably clean. If anyone can verify that the store still works as usual, then, that would be an ok check. We never used this functionality in the Cray system installs which were the ones I was familiar with.

It is possible that I can get to this tomorrow, but you shouldn't have to wait on me.

baallan:
I ran the store with all the configurations in the ldms-static-test suite, including the non-noitification store-specific tests and some valgrind runs.
The only distant concerns would be if someone somewhere using v3 has quietly deployed the notification feature in their data management pipeline. If they have, we have workarounds available when they upgrade to v4.

oceandlr:
I agree the distant concern should not be a blocker. If @tom95858 is ok with me not testing this, then I am fine with it. Otherwise I will do my best to try tomorrow.

PR #647, opened 2021-03-01
Add rdc plugin to collect AMD GPU metrics

miketxli:
Signed-off-by: Mike Li <Tianxinmike.Li@amd.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -239,6 +239,12 @@ dnl Will need to have the NVIDIA library in the LD_LIBRARY_PATH
 OPTION_DEFAULT_DISABLE([cray-hss-devel], [ENABLE_CRAY_HSS_DEVEL])
 OPTION_WITH([cray-hss-devel], [CRAY_HSS_DEVEL],[/usr])
 
+OPTION_DEFAULT_DISABLE([rdc_sampler], [ENABLE_RDC])
+OPTION_DEFAULT_DISABLE([rocm_rdc_inc], [ENABLE_ROCM_RDC_INC])
+OPTION_WITH([rocm_rdc_inc], [ROCM_RDC_INC], [/opt/rocm/rdc]) dnl for the path to rdc include and lib
+OPTION_DEFAULT_DISABLE([rocm_smi_inc], [ENABLE_ROCM_SMI_INC])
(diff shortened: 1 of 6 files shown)


baallan:
@miketxli  I've backported the amd gpu sampler to https://github.com/baallan/ovis.git branch backport_rdc derived from the OVIS-4 branch.

If you or a colleague have time and an environment to test in, please try to build (and ideally run) it and let me know what problems you find. ovis/ldms/src/sampler/rdc_sampler/Makefile.am will need some minor tweaking to find the right headers/libs.

miketxli:
@baallan Great, let me build and test it. I'll let you know the result.

baallan:
@miketxli Thanks for testing and letting me know the revised rdc_sampler worked with your setup.

I'm trying now to build with rocm-4.1.0 and run with mi50 and v320 based cards; the library says the RDC_FI_ECC_CORRECT_TOTAL, RDC_FI_ECC_UNCORRECT_TOTAL, and XGMI metrics are unsupported. Is this the expected behavior?

baallan:
Is there a convenient way to tell if they are xgmi connected, perhaps in the output of rocm-smi -a?
I suspect they are not connected. The more concerning thing would be the ECC metrics; are they no longer using ecc memory?

miketxli:
> Is there a convenient way to tell if they are xgmi connected, perhaps in the output of rocm-smi -a?
> I suspect they are not connected. The more concerning thing would be the ECC metrics; are they no longer using ecc memory?

1) Yes, rocm-smi --showtopo
2) They use different way to get those metrics. There is patch for this.  See https://github.com/RadeonOpenCompute/rdc/tree/master/ras_prebuild. You need to copy all those files in the same folder as the librdc.so. I haven't try it myself. But this will be fixed in next release.

baallan:
Thanks for the clues. So it would show link type other than PCIE (xgmi instead?) if we had the direct connections, yes? And by next release you mean a rocm 4.3 or some other?

It seems in the meantime, I should drop the ECC metrics from the default counter set. Would you agree?

miketxli:
> Thanks for the clues. So it would show link type other than PCIE (xgmi instead?) if we had the direct connections, yes? And by next release you mean a rocm 4.3 or some other?
> 
> It seems in the meantime, I should drop the ECC metrics from the default counter set. Would you agree?

yes to xgmi.
I think you drop ECC from default set until it is stabilized in later release. I will find out target release for you.

PR #649, opened 2021-03-03
remove  dead store api

baallan:
Would prefer to convert to deleting these, but folk need to look first.
These two functions stubs date from 2013/2015 and are nowhere implemented.


tom95858:
@baallan, I am OK with just removing these.

baallan:
@tom95858 I converted patch to delete them. I assume you can still squash this when pulling/merging.

PR #650, opened 2021-03-04
add cpu_num metric and more error checking to tx2mon

baallan:
This records the socket number (cpu_num) in the data set so it doesn't have to be extracted from the set instance name.
Additional nits fixed:
 * It adds return checking so that config fails if a metric add fails.
 * It fixes an array overrun that happens and is trapped only when compiled in hardened mode
per ARM/rhel7 RPM build defaults (-Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong).
 * It converts a lingering printf into a log call.


Snell1224:
@baallan Tested the build on stria with make install/installcheck/check and it was successful. Also tested with ldms-static-test and the tx2mon sampler ran without errors.

baallan:
@tom95858 @valleydlr  as Sara has checked it, please merge this at your earliest convenience.

PR #654, opened 2021-03-07
L2 streams store 

oceandlr:
current l2 stream store. This is in development, as indicated. Directory location (hello_streams) reflects that since that is where current streams capabilities which are for demonstration are located. 

This store will evolve for performance reasons. It will also be used to assess commonality to go into store base capabilities.

### DIFF
--- a/ldms/src/sampler/hello_stream/hello_cat_publisher.c
+++ b/ldms/src/sampler/hello_stream/hello_cat_publisher.c
@@ -62,7 +62,7 @@
 #include "ldms.h"
 #include "ldmsd_stream.h"
 
-#define STRLEN 1024
+#define STRLEN 8096
 static ldms_t ldms;
 static sem_t conn_sem;
 static int conn_status;
--- a/ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
+++ b/ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
@@ -1,0 +1,151 @@
+.\" Manpage for Plugin_l2_stream_csv_store
+.\" Contact ovis-help@sandia.gov to correct errors or typos.
+.TH man 7 "6 Mar 2021" "v4" "LDMS Plugin l2_stream_csv_store man page"
+
+.SH NAME
+Plugin_l2_stream_csv_store - man page for the LDMS l2_stream_csv_store plugin
+
+.SH SYNOPSIS
+Within ldmsd_controller or a configuration file:
+.br
+config name=l2_stream_csv_store [ <attr>=<value> ]
+
+.SH DESCRIPTION
+With LDMS (Lightweight Distributed Metric Service), plugins for the ldmsd (ldms daemon) are configured via ldmsd_controller or a configuration file. The l2_stream_csv_store plugin is a DEVELOPMENTAL store that writes out either a single stream's data to csv format if the input type is a well-known json format or writes out the raw messages if the input type is str. Input type will be determined by the hello_cat_publisher or similar.
+
+
+.SH CONFIGURATION ATTRIBUTE SYNTAX
+.TP
+.BR config
+name=l2_stream_csv_store path=<path> container=<container> stream=<stream> [buffer=<0/1>] [rollover=<N>]
+.br
+configuration line
+.RS
+.TP
+name=<plugin_name>
+.br
+This MUST be l2_stream_csv_store.
+.TP
+path=<path>
+.br
+path to the directory of the csv output file
+.TP
+container=<container>
+.br
+directory of the csv output file
+.TP
+stream=<stream_a,stream_b>
+.br
+csv list of streams to which to subscribe.
+.TP
+buffer=<0/1>
+.br
+Optional buffering of the output. 0 to disable buffering, 1 to enable it with autosize (default)
+.TP
+rollover=<N>
+.br
+Optional rollover of the output. Only rollover type available is by number of records in the CSV (one per dict).
+.RE
+
+.SH JSON FORMAT AND OUTPUT HEADER AND FORMAT
+.PP
+The json is expected to be something like:
+.nf
+{"foo":1, "bar":2, "zed-data":[{"count":1, "name":"xyz"},{"count":2, "name":"abc"}]}
+.fi
+Note the brackets. There will be at most one list. It is expected that each dictionary in the list will have the same item names. Everything else must be singleton data items.
+.PP
+The header is only generated once, off the first received json. If the first json is missing the list, or if the list has no entries, then list data will not appear in the header and will not be parsed in subsequent data lines. The header values will be the singleton names (e.g., foo, bar) and a list will be broken up into and item per dictionary item with names listname:dictname (e.g., zed_data:count, zed_data:name).
+.PP
+There can be any number of dictionaries in a list. Data lines with multiple dictionaries will be written out in the csv as separate lines, with the singleton items repeated in each line like:
+.nf
+#foo,bar,zed-data:count,zed-data:name
+1,2,1,xyz
+1,2,2,abc
+.fi
+.PP
+There will be a header in every output file (can be more than 1 because of rollover).
+
+.SH STORE OUTPUT FILENAME
+.PP
+The filename will be '<streamname>.<timestamp>' (e.g., foo-123456789). The timestamp is determined when the store is started or rolledover and the file is created. That may be considerably earlier than when data is streamed to the store.
+
+
+.SH STORE COLUMN ORDERING
+.PP
+There is only column ordering for 'json' format. There is no column ordering for 'str' format. 'str' format will always be written out, no matter what the 'json' header keys may be. The json order is arbitrary.
+
+.SH TIMING INFORMATION
+There are a few options for timing information. These are driven by #defines in the code source right now.
+.TP
+NDATA_TIMING
+.br
+Set by #define or #undef NDATA_TIMING and NDATA. This will write out an absolute timestamp in the log at level LINFO  every NDATA streaminput (not dictionary) lines. The format looks like "10000 lines timestamp 123456789". The timestamp is only gotten once, if the timing information is to be written out.
+.TP
+TIMESTAMP_STORE
+.br
+Set by #define or #undef TIMESTAMP_STORE. THis will write out an absolute timestamp in the file as the last item in the csv and is called 'store_recv_time' in the header. The timestamp is only gotten once, when the function is entered (e.g., if a data line has multiple dicts, this will result in multiple output lines each of which will have the same additional timestamp value.
+.PP
+
+
+.SH BUGS
+No known bugs.
+
+.SH NOTES
+.PP
+This store is in development and may be changed at any time.
+.PP
+Supports more than 1 stream. There is currently no performance guidence about number of streams and amount of data.
+.PP
+There is no way to know if a stream will actually be used or if a final value is received. Therefore, this store will need to be restarted if you want to use it with a new stream or if you want use the same stream name, but with different fields in the json.
+.PP
+It is possible that with buffering, if a stream's sends are ended, there still may be unflushed data to a file.
+.PP
+There is no way to remove a stream from the index nor to unsubscribe. That is, there is nothing that is akin to open_store and close_store pair as in an actual store plugin.
+.PP
+Note the restrictions on the data input above. Also how that affects the header.
+.PP
+This is not full featured, like the csv store. There is only one rollover, yet, for example.
+
+
+
+.SH EXAMPLES
+.PP
+Within ldmsd_controller or a configuration file:
+.nf
+load name=l2_stream_csv_store
+config name=l2_stream_csv_store path=XYZ/store container=csv stream=foo buffer=1
+# dont call anything else on the store. the store action is called by a callback triggered by the stream.
+
+prdcr_add name=localhost1 host=localhost type=active xprt=sock port=52001 interval=20000000
+prdcr_subscribe stream=foo regex=localhost*
+prdcr_start name=localhost1
+.fi
+
+.PP
+Testdata:
+.nf
+cat XXX/testdata.txt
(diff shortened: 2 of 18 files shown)


morrone:
What does "L2" mean?

tom95858:
@oceandlr, maybe we should rename this app_event_store or something else that indicates what it actually does.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+
+
+/**
+ * \brief Configuration
+ */
+static int config(struct ldmsd_plugin *self,
+                  struct attr_value_list *kwl, struct attr_value_list *avl)
+{
+	char* s;
+        char* streamlist;
```
baallan, line 976: init streamlist to null to avoid wild free in early exit cases.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+        goto out;
+
+err:
+        //delete any created streamstores
+        idx_traverse(stream_idx, close_streamstore, NULL);
+        idx_destroy(stream_idx);
+        stream_idx = idx_create();
+
+out:
+        if (templist) free(templist);
```
baallan, line 1088: remove if (templist) and if (streamlist) tests before free. free accepts NULL correctly.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+		if (i == 0){
+			dataline->dictkey =
+                                (char**) calloc(idict, sizeof(char*));
+			if (!(dataline->dictkey)) return ENOMEM;
+			dataline->ndict = idict;
+		}
+	}
+	return 0;
+
+};
```
baallan, line 281: remove ; after function body close.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+
+	// process again to fill. will parse dicts here
+	i = 0;
+	isingleton = 0;
+	for (a = json_attr_first(e); a; a = json_attr_next(a)){
+		json_attr_t attr = a->value.attr_;
+		switch (attr->value->type){
+		case JSON_LIST_VALUE:
+			dataline->listkey = strdup(attr->name->value.str_->str);
+			rc = _parse_list_for_header(dataline,attr->value);
```
baallan, line 371: check strdup return for OOM.

baallan (changes requested):
I may find more issues under valgrind, I but haven't the needed test config/input/stream injection method yet.

oceandlr:
Thanks, everyone, for the comments. I will check them out.

Re name: it is a good point that it should be named meaningfully. This one assumes an input format that the L2 team has decided on. If they change to something else, then I will change the innards. I think it is useful for a variety of reasons to have in the release and I am treating hello_streams as a demo location as mentioned above.

I could:
- rename for now. if we need something different for the L2, then maybe I should just spin that into some other named thing
- make a decision now about putting this into contrib somewhere.

@valleydlr  @tom95858

ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
```diff
@@ -0,0 +1,151 @@
...
+
+prdcr_add name=localhost1 host=localhost type=active xprt=sock port=52001 interval=20000000
+prdcr_subscribe stream=foo regex=localhost*
+prdcr_start name=localhost1
+.fi
+
+.PP
+Testdata:
+.nf
+cat XXX/testdata.txt
```
baallan, line 128: add hello_cat_publisher or other injector example here.

```
 echo '{"job-id" : 10364, "rank" : 1, "kokkos-perf-data" : [ {"name" : "SPARTAFOO0", "count": 0, "time": 0.0000},{"name" : "SPARTAFOO1", "count": 1, "time": 0.0001},{"name" : "SPARTAFOO2", "count": 2, "time": 0.0002},{"name" : "SPARTAFOO3", "count": 3, "time": 0.0003},{"name" : "SPARTAFOO4", "count": 4, "time": 0.0004},{"name" : "SPARTAFOO5", "count": 5, "time": 0.0005},{"name" : "SPARTAFOO6", "count": 6, "time": 0.0006},{"name" : "SPARTAFOO7", "count": 7, "time": 0.0007},{"name" : "SPARTAFOO8", "count": 8, "time": 0.0008},{"name" : "SPARTAFOO9", "count": 9, "time": 0.0009}] }' | hello_cat_publisher -p $port1 -a none -s foo -t json -x sock -h localhost
```

or cat the example to a file and add -f $file to invocation without pipe.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+        sprintf(dpath, "%s/%s", root_path, container);
+	sprintf(tmp_basename, "%s/%s/%s", root_path, container, stream);
+        sprintf(tmp_filename, "%s/%s/%s.%lu",
+                root_path, container, stream, tspath);
+
+	msglog(LDMSD_LDEBUG, PNAME ": stream '%s' will have file '%s'\n",
+               stream, tmp_filename);
+
+	/* create path if not already there. */
+	rc = mkdir(dpath, 0777);
```
baallan, line 811: use instead here create_outdir from store_csv_common.h or at least  f_mkdir_p from ovis_util/util.h

ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
```diff
@@ -0,0 +1,151 @@
...
+.TP
+stream=<stream_a,stream_b>
+.br
+csv list of streams to which to subscribe.
+.TP
+buffer=<0/1>
+.br
+Optional buffering of the output. 0 to disable buffering, 1 to enable it with autosize (default)
+.TP
+rollover=<N>
```
baallan, line 45: for reasons which as yet escape me, rollover never happens when configured with:

```
config name=l2_stream_csv_store path=${STOREDIR} container=csv stream=foo buffer=1 rollover=3
```

baallan:
Consider adding the following files to ldms/scripts/examples/, so that anyone can run a simple demo/test case on their machine with: "ldms-static-test.sh stream_store"

```
::::::::::::::
stream_store
::::::::::::::
export plugname=stream_store
portbase=61020
echo '{"job-id" : 10364, "rank" : 1, "kokkos-perf-data" : [ {"name" : "SPARTAFOO0", "count": 0, "time": 0.0000},{"name" : "SPARTAFOO
1", "count": 1, "time": 0.0001},{"name" : "SPARTAFOO2", "count": 2, "time": 0.0002},{"name" : "SPARTAFOO3", "count": 3, "time": 0.00
03},{"name" : "SPARTAFOO4", "count": 4, "time": 0.0004},{"name" : "SPARTAFOO5", "count": 5, "time": 0.0005},{"name" : "SPARTAFOO6", 
"count": 6, "time": 0.0006},{"name" : "SPARTAFOO7", "count": 7, "time": 0.0007},{"name" : "SPARTAFOO8", "count": 8, "time": 0.0008},
{"name" : "SPARTAFOO9", "count": 9, "time": 0.0009}] }' > $LDMSD_RUN/x.json
LDMSD 1
for i in $(seq 10); do
	SLEEP 1
	hello_cat_publisher -p $port1 -a none -s foo -t json -x sock -h localhost -f $LDMSD_RUN/x.json
done

SLEEP 1
KILL_LDMSD 1
file_created $STOREDIR/csv/foo.*
::::::::::::::
stream_store.1
::::::::::::::
load name=l2_stream_csv_store
config name=l2_stream_csv_store path=${STOREDIR} container=csv stream=foo buffer=1 rollover=3

prdcr_add name=localhost1 host=localhost type=active xprt=sock port=${port1} interval=20000000
prdcr_subscribe stream=foo regex=localhost*
prdcr_start name=localhost1
```

ldms/src/sampler/hello_stream/hello_cat_publisher.c
```diff
@@ -62,7 +62,7 @@
 #include "ldms.h"
```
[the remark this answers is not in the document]
oceandlr: I don't know what this comment means. This has been working for testing for months.

oceandlr:
@tom95858 @valleydlr - please give me your advice....

I spoke with @valleydlr about where to put this. Since this is testing and possibly specific to one case, it probably shouldn't be in the release. for the l2, I could keep this in my own repo, but then it won't be in the RPMs that are being built on the machines. Is it reasonable for me to rebase on top of the release, build locally and then give you a binary of library to drop in by hand when we need it?

ldms/src/sampler/hello_stream/hello_cat_publisher.c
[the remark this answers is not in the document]
baallan: When compiled/run on centos7 (presumably on rhel also) with gcc4, the cat program loops forever instead of sending a message once. Instead of a single burst of the lines documented in the man page example appearing in the store file, I get 1000 events/sec being logged. 

Is this intended behavior?

ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
baallan: The example and how to run it is provided in the next comment re ldms/scripts/examples.

morrone:
> I spoke with @valleydlr about where to put this. Since this is testing and possibly specific to one case, it probably shouldn't be in the release. for the l2, I could keep this in my own repo, but then it won't be in the RPMs that are being built on the machines. Is it reasonable for me to rebase on top of the release, build locally and then give you a binary of library to drop in by hand when we need it?

With samplers, I build and rpm package samper plugins separately from (after) the ovis-ldms builds. That may be more to learn than you are looking for, but I thought I would mention that as a possibility. It would allow you to make an rpm of your development store for install on your machines. If that is of interest,

Granted, I don't know for sure that ovis-ldms is exporting everything needed to build store plugins the way that it exports everything needed for sampler plugins.

baallan:
It is certainly the intent that stores can be built externally the same way as samplers. If there is header information missing, there's a bug.
But that said, for test packaging/test system install purposes, it's trivial to cherry pick this store addition into the packaging process so long as the development branch is rebased to OVIS-4 before packaging. I don't see a rush to merge before things are fully baked. There's few enough of us involved that coordinating shouldn't be a problem.

ldms/src/sampler/hello_stream/hello_cat_publisher.c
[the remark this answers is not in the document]
baallan: I ran the cat program under gdb and it appears that indeed, it is sending the message only once.
But somehow that message gets recycled into the store file at ~1kHz until the store is closed.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+static void l2_stream_csv_store_init()
+{
+        stream_idx = idx_create();
+	pthread_mutex_init(&cfg_lock, NULL);
+}
+
+static void __attribute__ ((destructor)) l2_stream_csv_store_fini(void);
+static void l2_stream_csv_store_fini()
+{
+	pthread_mutex_destroy(&cfg_lock);
```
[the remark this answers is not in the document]
baallan: To avoid further use after free errors, reorder _fini like so:
{
        if (rothread_used){
                void * dontcare = NULL;
                pthread_cancel(rothread);
                pthread_join(rothread, &dontcare);
        }
        pthread_mutex_destroy(&cfg_lock);
        idx_t stridx = stream_idx;
        stream_idx = NULL;
        // may need additional code from term() here.
        idx_destroy(stridx);

}
otherwise rollover thread and stream_cb may try to use stream_idx after free.

ldms/src/store/hello_stream/l2_stream_csv_store.c
```diff
@@ -0,0 +1,1165 @@
...
+	}
+
+out:
+
+	pthread_mutex_unlock(&stream_handle->lock);
+	return rc;
+}
+
+
+static int open_streamstore(char* stream){
```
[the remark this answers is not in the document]
oceandlr: ???

ldms/src/store/hello_stream/l2_stream_csv_store.c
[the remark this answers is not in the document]
oceandlr: Not understanding. If this gets called, why would that happen?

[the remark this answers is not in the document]
baallan: the destructor tag function of a plugin library gets called by the loader (not directly us) when the library is closed at any kind of program exit or a plugin unload. This happens asynchronously to the threads such as rollover or the driver's calls to config. If we do not explicitly handle the plugin exit process as suggested here, then the race conditions can/do lead to unexplained crashes and (worse) hangs which make us seem flakey to admins. We need the memory and thread cleanups to be correct to avoid those problems and also so that unexpected leaks can be distinguished with dynamic analysis tools like valgrind.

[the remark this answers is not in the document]
morrone: @ballan, I think you are looking at the store_csv.c file, which is not part of this PR. :)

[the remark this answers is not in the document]
morrone: Also, the real reason to use _fini() to clean up isn't so much to avoid crashes or hangs at exit time. That is often bad programming. _fini() is mainly used to clean up things at fork() time when there is no other way to reinitialize things that need to be in a known intialized state after the fork happens.

ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
oceandlr: This has been updated.

ldms/src/store/hello_stream/l2_stream_csv_store.c
oceandlr: done. going with 0777 since that is what is in store_flatfile

ldms/src/store/hello_stream/Plugin_l2_stream_csv_store.man
oceandlr: 60 sec default to check for roll. extend your sequence

oceandlr:
After discussion with various stakeholders, it was decided that it is desirable to get this store into the mainline. It's the most reasonable, flexible format for json that results in an unambiguous CSV format. I will be closing this and putting in a new request with the store under a different name. This will probably additionally replace the hello_streams_store; I think having that one soley for tutorial purposes is unnecessary.

oceandlr:
closing this. a new one will go in on OVIS-4.

PR #655, opened 2021-03-10
store_csv: Repair constant file rename errors

morrone:
When altheader=0, the default setting for the store_csv store:

The store_csv code has a bug that results in chown/chmod error
messages every time a file is renamed. A sanitized version of
the error looks like:

  DEBUG: store_csv: rename_output: rename(A, B)
  ERROR: store_csv: rename_output: unable to chmod(A,660): No such file or directory.
  ERROR: store_csv: rename_output: unable to chown(A, <UID>, <GID>: No such file or directory.
  DEBUG: store_csv: rename_output: rename(A, B)

It _appears_ that the second redundant rename of "A" to "B" succeeds, even
though the chmod and chown of "A" fail immediately before. However, this is
not the case. In fact only the first rename succeeds as we would expect, and
the error from the second rename is simply ignored. The code comment
explaining the ignored error says:

 "enoent happens if altheader = 0 or typeheader = 0"

So it was recognized that the code was incorrect, but the error was ignored
rather than fixing the problem.

The issue seems to be that when altheader=0, the header is not a separate
file, but added as a comment in the main CSV file. However the code redundantly
fills in the same file name for the headerfile rather than recognizing this
fact.

This code has become a bit spaghettified over the years, and probably
needs an overhaul to make a fully clean fix.

In the mean time we use a small fix to avoid the incorrect file rename 
attempts when altheader is zero and when typeheader is zero.

### DIFF
--- a/ldms/src/store/store_csv.c
+++ b/ldms/src/store/store_csv.c
@@ -739,7 +739,8 @@ static void *get_ucontext(ldmsd_store_handle_t _s_handle)
 
 /*
  * note: this should be residual from v2 where we may not have had the header info until a store was called
- * which then meant we had the mvec. ideally the print_header will always happen from the store_open.
+ * which then meant we had the mvec. ideally the print_header will always happen from the open_store,
+ * but in point of fact is currently _never_ called in open_store.
  * Currently we still have to keep this to invert the metric order
  */
 static int print_header_from_store(struct csv_store_handle *s_handle, ldms_set_t set,
@@ -768,13 +769,13 @@ static int print_header_from_store(struct csv_store_handle *s_handle, ldms_set_t
 		else
 			ec = snprintf(tmp_path, PATH_MAX, "%s.HEADER",
 				s_handle->path);
-			} else {
+	} else {
 		if (rolltype >= MINROLLTYPE)
 			ec = snprintf(tmp_path, PATH_MAX, "%s.%d",
 				s_handle->path, (int)s_handle->otime);
 		else
 			ec = snprintf(tmp_path, PATH_MAX, "%s", s_handle->path);
-				}
+	}
 	(void)ec;
 	csv_format_header_common(fp, tmp_path, CCSHC(s_handle), s_handle->udata,
 		&PG, set, metric_array, metric_count);
--- a/ldms/src/store/store_csv.c
+++ b/ldms/src/store/store_csv.c
@@ -277,11 +277,11 @@ static void roll_cb(void *obj, void *cb_arg)
 	if (s_handle->headerfile) {
 		fclose(s_handle->headerfile);
 	}
-	if (s_handle->headerfilename) {
+	if (s_handle->altheader != 0 && s_handle->headerfilename) {
 		rename_output(s_handle->headerfilename, FTYPE_HDR,
-			CSHC(s_handle), cps);
+			      CSHC(s_handle), cps);
 	}
-	if (s_handle->typefilename) {
+	if (s_handle->typeheader != 0 && s_handle->typefilename) {
 		rename_output(s_handle->typefilename, FTYPE_KIND,
 			CSHC(s_handle), cps);
 		snprintf(tmp_typepath, PATH_MAX, "%s.KIND.%d",
--- a/ldms/src/store/store_csv_common.c
+++ b/ldms/src/store/store_csv_common.c
@@ -421,12 +421,9 @@ void rename_output(const char *name,
 	err = rename(name, newname);
 	if (err) {
 		int ec = errno;
-		if (ec != ENOENT) {
-			strerror_r(ec, errbuf, EBSIZE);
-			cps->msglog(LDMSD_LERROR,"%s: rename_output: failed rename(%s, %s): %s\n",
-				cps->pname, name, newname, errbuf);
-		}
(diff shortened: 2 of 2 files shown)


baallan:
@morrone please rebase to top of tree. 

Do you have an input store config example you can share that produces incorrect files, or is this strictly a style issue?

The order of calls in rename_output is chmod/chown/rename, so your explanation here cannot be correct, unless you intended to start it with "For the case when altheader=0,...". 

Less the commingled unrelated nit changes, the proposed change looks ok, though I haven't run it though the csv tests.

morrone:
> @morrone please rebase to top of tree.

It is already at the top of the OVIS-4 tree.
 
> Do you have an input store config example you can share that produces incorrect files, or is this strictly a style issue?

While not a fatal issue (the files are still copied), the code is attempting operations in the filesystem that cannot possibly work. That makes it a bug. While it does that it is throwing a constant stream of error messages in the logs, and makes it a bug with an annoyance. So this is not an issue of (code?) style at all.
 
> The order of calls in rename_output is chmod/chown/rename, so your explanation here cannot be correct, unless you intended to start it with "For the case when altheader=0,...".

The order of the calls in my example is consistent with the chmod/chown/rename in the code. And yes, this is for the default mode of operation, altheader=0.

In case it isn't entirely clear, the code is calling "rename_output()" twice on the same file. That is a bug. It should only be calling rename_output() once for a particular file. rename_output() does the following:

  chmod()
  chown()
  rename()

so two calls to rename_output() for the same file would do:

  chmod()
  chown()
  rename()
  chmod()
  chown()
  rename()

If one were to set "-v DEBUG", there would be no mention of the first two chmod() and chown() in the logs because those two calls would _succeed_. Then the first rename() is mentioned in a DEBUG level message. Then the second set of chmod() and chown() _fail_, so we see ERROR messages in the logs about those. Finally, the second rename() is mentioned at DEBUG level in the logs (and fails, but the code ignores that). In other words:

  chmod() no message in the logs
  chown() no message in the logs
  rename() message in logs at DEBUG level
  chmod() ERROR in logs
  chwon() ERROR in logs
  rename() message in logs at DEBUG level

Hopefully that explains why the log example I gave is correct in its order.

> Less the commingled unrelated nit changes, the proposed change looks ok, though I haven't run it though the csv tests.

Testing and thorough review would be good on this one. I'm not entirely sure that I accounted for every side condition.

baallan:
I will run this through the related ldms-static-test.sh cases and review the logs. Will take a little while.

tom95858:
Hi @morrone, @baallan is this ready to go?

baallan:
I owe running some tests based on Chris and my last comments. will try to get that done tomorrow.

baallan:
@morrone the following additions to your patch make the logs quieter and the permissions better defaulted.

```
iff --git a/ldms/src/store/store_csv.c b/ldms/src/store/store_csv.c
index 6037df0..0fe6a75 100644
--- a/ldms/src/store/store_csv.c
+++ b/ldms/src/store/store_csv.c
@@ -790,14 +790,14 @@ static int print_header_from_store(struct csv_store_handle *s_handle, ld
 
        /* dump data types header, or whine and continue to other headers. */
        if (s_handle->typeheader > 0 && s_handle->typeheader <= TH_MAX) {
-               fp = fopen(s_handle->typefilename, "w");
+               fp = fopen_perm(s_handle->typefilename, "w", LDMSD_DEFAULT_FILE_PERM);
                if (!fp) {
                        int rc = errno;
                        PG.msglog(LDMSD_LERROR, PNAME ": print_header: %s "
                                "failed to open types file (%d).\n",
                                s_handle->typefilename, rc);
                } else {
-                       ch_output(fp, tmp_path, CSHC(s_handle), &PG);
+                       ch_output(fp, s_handle->typefilename, CSHC(s_handle), &PG);
                        csv_format_types_common(s_handle->typeheader, fp, 
                                s_handle->typefilename, CCSHC(s_handle),
                                s_handle->udata, &PG, set,
diff --git a/ldms/src/store/store_csv_common.c b/ldms/src/store/store_csv_common.c
index d9890be..abb8851 100644
--- a/ldms/src/store/store_csv_common.c
+++ b/ldms/src/store/store_csv_common.c
@@ -913,8 +913,10 @@ void close_store_common(struct csv_store_handle_common *s_handle, struct 
        }
 
        rename_output(s_handle->filename, FTYPE_DATA, s_handle, cps);
-       rename_output(s_handle->headerfilename, FTYPE_HDR, s_handle, cps);
-       rename_output(s_handle->typefilename, FTYPE_KIND, s_handle, cps);
+       if (s_handle->altheader)
+               rename_output(s_handle->headerfilename, FTYPE_HDR, s_handle, cps);
+       if (s_handle->typeheader)
+               rename_output(s_handle->typefilename, FTYPE_KIND, s_handle, cps);
        replace_string(&(s_handle->filename), NULL);
        replace_string(&(s_handle->headerfilename),  NULL);
        replace_string(&(s_handle->typefilename),  NULL);
````

morrone:
> @morrone the following additions to your patch make the logs quieter and the permissions better defaulted.

Feel free to submit a follow-on PR!

baallan:
@tom95858 This patch plus #664 quiets unnecessary file-related store_csv debug messages.

PR #656, opened 2021-03-10
add test of csv store create-time uid/gid

baallan:
In this test, files are born user 99 and closed with change to user 199.
This works on centos 7 local file systems.
It may have issues on some NFS implementations. 
To run the test after a make install, chown privilege is needed.
The invocation is:
	allowroot=1 ldms-static-test.sh uidgidcsv

The output should show something like an end of:

sleep 5 ...done
-rw-------. 1 nobody nobody 576 Mar 10 09:54 /scratch1/baallan/ann-ovis/ldmstest/uidgidcsv/store/node/uidgidcsv.1615395250
sleep 5 ...done
-rw-------. 1 199 nobody 2131 Mar 10 09:54 /scratch1/baallan/ann-ovis/ldmstest/uidgidcsv/store/clemens-c7/uidgidcsv.1615395250
sleep 1 ...done
logs and data stored under /scratch1/baallan/ann-ovis/ldmstest/uidgidcsv
done

### DIFF
--- a/ldms/scripts/examples/.canned
+++ b/ldms/scripts/examples/.canned
@@ -29,6 +29,7 @@ csvcheckattrs # one ok and one failed config_check
 vg_csv
 vg_sos
 shm # store fails because of schema ambiguity
+uidgidcsv
 
 # working with job metrics
 clock.job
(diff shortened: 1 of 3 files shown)


baallan:
@valleydlr This test should always work as described on a local spinny/nvme or RAM (/dev/shm) file system and (afaik) lustre. If it fails on a particular nfs implementation,  we should note that in our man pages.

PR #658, opened 2021-03-11
Fix ldmsctl to provide the correct msg length to response handlers

nichamon:
The patch (5195ba3) that unifies the code to gather records into a whole
LDMSD message introduced the bug. The patch does not set the value of
the msglen variable in the __handle_cmd(). Resulting in __handle_cmd
always pass 0 as the response message length. The response handlers of
the status commands, e.g., prdcr_status, updtr_status, and
daemon_status, use the value provided by the __handle_cmd() when they
parse the string into a JSON object.

### DIFF
--- a/ldms/src/ldmsd/ldmsctl.c
+++ b/ldms/src/ldmsd/ldmsctl.c
@@ -2291,7 +2291,6 @@ static int __handle_cmd(struct ldmsctl_ctrl *ctrl, char *cmd_str)
 		exit(1);
 	}
 
-	size_t msglen = 0;
 	rc = 0;
 	ldmsctl_buffer_t recv_buf;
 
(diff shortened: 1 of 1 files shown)


nichamon:
Ultimately I would like to get rid of the `len` argument in the `cmd->resp` definition and change all response handlers to pass `attr->attr_len` to `json_parse_buffer()` instead of `len`. I didn't do this in the patch to make the patch as small as possible.

narategithub (approved):
The `ldmsd_ctrl_test` in `https://github.com/ovis-hpc/ldms-test` that has previously failed is now a pass.

nichamon:
@tom95858 We should merge this patch; otherwise, prdcr_stauts, updtr_status, #_status in ldmsctl won't work.

PR #659, opened 2021-03-11
Revert "Remove set_info from ldms_lookup message"

narategithub:
This reverts commit d4565971667bea774c515890dfcf84e1ec68fd13.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1052,6 +1052,86 @@ out:
 	return;
 }
 
+/* Caller should hold the set lock */
+static void __copy_set_info_to_lookup_msg(char *buffer, ldms_name_t schema,
+						ldms_name_t inst_name,
+						struct ldms_set *set)
+{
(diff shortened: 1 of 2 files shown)


narategithub:
@tom95858,

I just found out that setgroup has been broken from `setgroup_test` in `ldms-test` repo. With help from @nichamon, we identified that the removal of lookup set_info from d456597 was the culprit. We also confirmed that the set_info won't be set correctly on the aggregator because the set was not yet created when agg received dir_reply, and when the set was created in the lookup path the set_info was not available to set anymore. Hence, the group set on the aggregator won't have member information from the sampler. Given the time constrain, we think that reverting the commit is the best action.

With #658 and this pull request, all test cases in `ldms-test` succeeded.

nichamon:
Without this PR, the LDMS sets on aggregators  won't have any remote set_info until there is a dir_upd that updates remote set_info. This won't affect the auto_update feature because it uses only the hint info in the dir_reply. However, ldmsd group gets the set member list by calling `ldms_set_set_info_get()`.

PR #663, opened 2021-03-13
add needed set swaps for tx2mon sample().

baallan:


baallan:
@Snell1224 additional testing revealed that data sets weren't being marked with updated timestamps in sample() after initial population, even though the actual data was being updated. This should fix that.

@tom95858 can you add Snell1224 to the project so she appears as a possible reviewer going forward?

Snell1224:
@baallan @tom95858 Tested with make check/install and ldms-static-test.sh. Worked as expected.

PR #665, opened 2021-03-15
fixes #657 which includes per message memory leaks

### LINKED ISSUE #657: leaks in need of repair
ldmsd_request.c: 
prdcr_subscribe_regex_handler leaks stream_name variable on each call.
stream_subscribe_handler leaks stream_name on each call.

baallan:


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -90,6 +90,8 @@ void ldmsd_stream_deliver(const char *stream_name, ldmsd_stream_type_t stream_ty
 			&& entity == NULL	/* data hasn't been parsed yet */
 			&& parser == NULL)	/* we haven't tried and failed already */
 		{
+			if (parser)
```
tom95858, line 93: parser has to be NULL to enter this path. This is just dead code I think.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -1707,6 +1708,7 @@ static int prdcr_unsubscribe_regex_handler(ldmsd_req_ctxt_t reqc)
 	ldmsd_send_req_response(reqc, reqc->line_buf);
 	if (prdcr_regex)
```
tom95858, line 1709: This check is unnecessary. free() correctly handles NULL.

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_request.c
baallan: the patch is to fix leaks, not clean up  pre-existing style issues.

ldms/src/ldmsd/ldmsd_stream.c
baallan: 2nd+ trips through the loop parser is not null.

baallan: but, yes, i'll drop this non fix and figure out how else we can get to losing that pointer.

ldms/src/ldmsd/ldmsd_request.c
tom95858: This is not a style issue; the check is unnecessary and the change is right next to yours. As we find these little nits whether or not this change introduced them, we should fix them. Please fix.

baallan: As there is no reasonable limit to "as we find" and "right next to" in unrelated nit fixing, I simply stopped at the file scope.

tom95858:
Never mind, I'll fix it.

PR #666, opened 2021-03-16
Upstream sampler plugins from ldms-plugins-llnl

morrone:
This is a PR to upstream four plugins from the ldms-plugins-llnl project. They are: dcgm (Nvidia GPU monitoring), lustre_client, lustre_ost, and lustre_mdt.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -689,6 +689,23 @@ fi
 OPTION_GITINFO
 OPTION_HOSTINFO
 
+AC_ARG_WITH([dcgm],
+  [AS_HELP_STRING([--with-dcgm],
+    [Include code that depends on Nvidia DCGM @<:@default=check@:>@])],
+  [],
+  [with_dcgm=check])
+LIBDCGM=
+AS_IF([test "x$with_dcgm" != xno],
+      [AC_CHECK_LIB([dcgm], [dcgmInit],
+         [have_dcgm=true],
+         [have_dcgm=false
+          if test "x$with_dcgm" != xcheck; then
+            AC_MSG_FAILURE([--with-dcgm was given, but test for dcgm failed])
+          fi
+         ])],
+      [have_dcgm=false])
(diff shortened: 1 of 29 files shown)


tom95858:
Hi @morrone, this looks great. I will attempt to do a thorough review here shortly. Right off the top, could you please edit the top level LICENSE file and add the license/keys to which your header file comments refer.

morrone:
> Hi @morrone, this looks great. I will attempt to do a thorough review here shortly. Right off the top, could you please edit the top level LICENSE file and add the license/keys to which your header file comments refer.

The top level LICENSE file already contains the licenses to which my headers refer, which is why I am referring to that. I purposely choose licensing that matched ovis's licensing. Although I do now notice a couple of issues: The top level LICENSE file refers to a file "COPYING" as containing the GPLv2, but that is missing at the top level. And frankly, LICENSE should probably be called COPYING since everything refers to that file. We wouldn't want to have conflicting LICENSE and COPYING files (one that says dual-licensing, and the other just containing the GPLv2). I think I'll submit a PR to clean up that situation. And then I can update this PR to reflect that if it looks good.

baallan:
@valleydlr a subtle difference from the old wide format lustre samplers is these new ones don't provide the local rate computation on read/write bytes, which was found to be handy at snl and ncsa, but might be considered redundant with a robust analysis pipeline. 

If precomputed rates are a must have, we'll need to make a subsequent feature request; I wouldn't hold up merging this for it.

morrone:
There are more than subtle differences too. This are not reworkings of the existing plugins, they are completely new ones. So anyone considering them should consider all of the differences.

morrone:
I updated the copyright comments on the assumption that #669 will land too.

baallan:
I believe at some point we deliberately took away 'copying' and several other typical files  at the top level. At present they seem to live under lib/ and ldms/. I don't recall the rationale for most of that.

PR #667, opened 2021-03-16
Adding/documenting free() semantics for various object free functions.

baallan:
For headered functions with names containing _free, this patch documents
and in some cases adds, the key semantics of free():
	do nothing if given a NULL pointer
	release resources if given a valid pointer.

Why?
 * This allows us to stop using syntax of the sort: if (x) free_operator(x), making for easier to read and write code.
 * This prevents data-corrupting random walks through memory that start by
dereferencing a NULL argument in destructor code.

Special cases:
 * where assert is in use, assertions are checked after the NULL check,

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1975,6 +1975,9 @@ static void process_push_reply(struct ldms_xprt *x, struct ldms_reply *reply,
 
 void ldms_xprt_dir_free(ldms_t t, ldms_dir_t dir)
 {
+	(void)t;
+	if (!dir)
+		return;
 	int i, j;
 	for (i = 0; i < dir->set_count; i++) {
 		free(dir->set_data[i].inst_name);
--- a/ldms/src/ldmsd/ldmsd_request_util.c
+++ b/ldms/src/ldmsd/ldmsd_request_util.c
@@ -1024,6 +1024,8 @@ struct ldmsd_msg_buf *ldmsd_msg_buf_new(size_t len)
 
 void ldmsd_msg_buf_free(struct ldmsd_msg_buf *buf)
 {
+	if (!buf)
+		return;
 	free(buf->buf);
 	free(buf);
 }
--- a/lib/src/ovis_event/ovis_event.c
+++ b/lib/src/ovis_event/ovis_event.c
@@ -81,8 +81,8 @@ void ovis_scheduler_ref_put(ovis_scheduler_t m)
 {
 	if (!m)
 		return;
-	pthread_mutex_lock(&m->mutex);
 	assert(m->refcount > 0);
+	pthread_mutex_lock(&m->mutex);
 	m->refcount--;
 	if (m->refcount == 0) {
 		pthread_mutex_unlock(&m->mutex);
@@ -225,7 +225,7 @@ loop:
 	if (rc < 0) {
 		if (errno == EAGAIN || errno == EWOULDBLOCK)
 			return;
-		assert(0 && errno);
+		assert(0 == "__ovis_event_pipe_cb: unexpected logic" && errno);
 	}
 	goto loop;
 }
@@ -615,7 +615,7 @@ int ovis_event_term_check(ovis_scheduler_t m)
 	pthread_mutex_lock(&m->mutex);
 	switch (m->state) {
 	case OVIS_EVENT_MANAGER_INIT:
-		assert(0);
+		assert(0 == "unexepected OVIS_EVENT_MANAGER_INIT in ovis_event_term_check");
 		break;
 	case OVIS_EVENT_MANAGER_RUNNING:
 	case OVIS_EVENT_MANAGER_WAITING:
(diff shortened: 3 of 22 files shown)


tom95858 (commented):
@baallan, since this changes the behavior of the free functions to check for NULL and the check for NULL is presumed by the caller, should the asserts just be removed?

ldms/src/core/ldms_xprt.c
```diff
@@ -1975,6 +1975,9 @@ static void process_push_reply(struct ldms_xprt *x, struct ldms_reply *reply,
 
 void ldms_xprt_dir_free(ldms_t t, ldms_dir_t dir)
 {
+	(void)t;
```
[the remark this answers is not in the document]
baallan: to quiet unused argument warnings.

ldms/src/ldmsd/ldmsd_request_util.c
```diff
@@ -1021,6 +1021,8 @@ struct ldmsd_msg_buf *ldmsd_msg_buf_new(size_t len)
 
 void ldmsd_msg_buf_free(struct ldmsd_msg_buf *buf)
 {
+	if (!buf)
+		return;
```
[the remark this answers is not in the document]
baallan: Are you saying a null pointer check will crash? or that we might get lucky and a free of a chunk of memory arbitrarily pointed at by data at an offset in the 0 page will crash? It's been my experience that free(0->offset) is failure prone but inconsistently so.

baallan:
Re removing assertions: some of them are checking api behavior/use contracts that I am not in a good position to say no longer need checking. The additional null checks should (where added) keep us from entering the realm of undefined behaviors when compiled with -DNDEBUG.

lib/src/ovis_json/ovis_json.c
```diff
@@ -541,7 +541,9 @@ int json_attr_rem(json_entity_t d, char *name)
 static void json_list_free(json_list_t a)
 {
 	json_entity_t i;
-	assert(a->base.type == JSON_LIST_VALUE);
+	assert(a && a->base.type == JSON_LIST_VALUE);
```
tom95858, line 544: Please

lib/src/ovis_json/ovis_json.c
```diff
@@ -541,7 +541,9 @@ int json_attr_rem(json_entity_t d, char *name)
 static void json_list_free(json_list_t a)
 {
 	json_entity_t i;
-	assert(a->base.type == JSON_LIST_VALUE);
```
tom95858, line 544: Please move this to after the NULL pointer check.

lib/src/ovis_json/ovis_json.c
```diff
@@ -552,14 +554,18 @@ static void json_list_free(json_list_t a)
 
 static void json_str_free(json_str_t s)
 {
-	assert(s->base.type == JSON_STRING_VALUE);
```
tom95858, line 555: Please move this an similar asserts to after the NULL pointer check.

lib/src/ovis_util/big_dstring.h
```diff
@@ -104,6 +104,8 @@ static inline const char *bdstrval(const big_dstring_t *dsPtr); \
 static void bdstr_free(big_dstring_t *dsPtr) \
 { \
 	assert(NULL != dsPtr); \
```
tom95858, line 106: Given the proposed change, callers to bdstr_free() may now assume that dsPtr may be NULL. If we add this assert, these apps will break.

tom95858 (changes requested):

lib/src/ovis_json/ovis_json.c
baallan: do i understand correctly that you want the null check split into a statement separate and before the type assertion, rather than combined? makes sense, just want to double check.

baallan:
@tom95858 I believe i've fixed the things you requested. I rebased on ovis-4 as well.

lib/src/ovis_util/dstring.c
```diff
@@ -292,6 +292,8 @@ void dstr_trunc(dstring_t * dsPtr, int length)
 void dstr_free(dstring_t * dsPtr)
 {
 	assert(NULL != dsPtr);
```
tom95858, line 294: @baallan, I think this assert needs to be removed.

lib/src/ovis_event/ovis_event.c
```diff
@@ -79,6 +79,8 @@ void ovis_scheduler_ref_get(ovis_scheduler_t m)
 static inline
 void ovis_scheduler_ref_put(ovis_scheduler_t m)
 {
+	if (!m)
+		return;
 	pthread_mutex_lock(&m->mutex);
 	assert(m->refcount > 0);
```
tom95858, line 85: I know this bug is unrelated to your fix, but this assert belongs before the pthread_mutex_lock() call.

tom95858 (changes requested):

lib/src/ovis_util/dstring.c
baallan: ok

baallan:
@tom95858 I did those, and also a few unadorned assert(0) expanded to include a clue string in the same files.

tom95858:
Hi @baallan, where are we on this? I'd like to get it merged.

baallan:
its ready; fixed what you requested yesterday.

PR #668, opened 2021-03-17
Configure simplification

morrone:
OVIS_INCLUDE_ABS and OVIS_LIB_ABS are autoconf variables used
throughout many Makefile.am files in the tree. However, as it
stands it is less than easy to figureout how and where those
variables are set. One needs to make some educated guesses
and search the code to figure out that they are constructed
by the OPTION_LIB_FLAGS and OPTION_INCLUDE_FLAGS macros.

In practice these macros do more to obfuscate the code than to
improve it. Also, there is no reuse of these macros, so that
is not a good reason to make these macros.

Additionally, the macros set far more include paths than it
really should be.

We turn the OVIS_INCLUDE_ABS and OVIS_LIB_ABS variables into
simple lists.

This may be a first step in trying to fix ovis's header issues.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,51 +63,45 @@ test "$sharedstatedir" = '${prefix}/com' && sharedstatedir='${prefix}/var/lib'
 #
 # link and include flags for build to ensure self-consistency
 #
-
-# ideally, this should yield from lib/src ldms/src
-# "-I$(top_builddir)/lib/src -I$(top_builddir)/ldms/src -I$(top_srcdir)/lib/src -I$(top_srcdir)/ldms/src"
-# when all the includes are fixed.
 dnl contrib should never be in this list
-OPTION_INCLUDE_FLAGS([OVIS],[ \
-	. \
-	lib/src \
-	lib/src/coll \
-	lib/src/third \
-	lib/src/ovis_util \
-	lib/src/ovis_auth \
-	lib/src/ovis_ctrl \
-	lib/src/ovis_ev \
-	lib/src/ovis_event \
-	lib/src/ovis_json \
-	lib/src/zap \
-	lib/src/mmalloc \
-	ldms/src \
-	ldms/src/core \
-	ldms/src/auth \
-	ldms/src/ldmsd \
-	ldms/src/sampler \
-	ldms/src/store \
-])
+OVIS_INCLUDE_ABS="\
+	-I\${abs_top_srcdir}/lib/src \
+	-I\${abs_top_srcdir}/lib/src/coll \
+	-I\${abs_top_srcdir}/lib/src/third \
+	-I\${abs_top_srcdir}/lib/src/ovis_util \
+	-I\${abs_top_srcdir}/lib/src/ovis_auth \
+	-I\${abs_top_srcdir}/lib/src/ovis_ctrl \
+	-I\${abs_top_srcdir}/lib/src/ovis_ev \
+	-I\${abs_top_srcdir}/lib/src/ovis_event \
+	-I\${abs_top_srcdir}/lib/src/ovis_json \
+	-I\${abs_top_srcdir}/lib/src/zap \
+	-I\${abs_top_srcdir}/lib/src/mmalloc \
+	-I\${abs_top_srcdir}/ldms/src \
+	-I\${abs_top_srcdir}/ldms/src/core \
+	-I\${abs_top_srcdir}/ldms/src/auth \
+	-I\${abs_top_srcdir}/ldms/src/ldmsd \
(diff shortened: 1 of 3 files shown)


baallan:
I disagree with all the justifications of this proposed change:

>  it is less than easy to figure out how and where those variables are set.

It's all set in configure.ac, as with every autotools project.
A more reasonable approach to solving this aspect is to place a comment in configure.ac adjacent to the macro calls that says "compute OVIS_INCLUDE_ABS" etc. 

> In practice these macros do more to obfuscate the code than to improve it.

No, the 'knowledge added' is in the directory list, not in the expansion by wrapping every list element in the baroque libtoolisms/cppisms. 

> Also, there is no reuse of these macros, so that is not a good reason to make these macros.

Reuse is not a primary goal in configure scripts. Maintainability is. The core information to be maintained in this case is the list of directories that matter, not the current gnu option syntax provided by the macro expansion. m4 macros in configure.ac function as 'includes' very often in many projects.

> Additionally, the macros set far more include paths than it really should be.

If we have truly dead code directories, then the correct solution is remove them from the list.

> We turn the OVIS_INCLUDE_ABS and OVIS_LIB_ABS variables into simple lists.

The proposed lists are not simpler than the lists you are replacing.

> This may be a first step in trying to fix ovis's header issues.

This patch is a no-op in trying to fix ovis' header issues. 

Please open (or expand on) an actual issue so we can discuss and get to a solution. Maybe something along the lines of:
 * topic: LDMS source include statements and header locations need refactoring to make builds easier to specify
 * comment: Both ldms build before install and third party build after install should function correctly with use of single include option -I$ovis_include_dir.

morrone:
This patch is not a no-op. This patch literally halves the number variables we are setting, and in the case of the includes further halves the number of -I values that appear on the command line.

Did you miss that? Why did you miss that? Is it perhaps because the current system obscures what is really going on?

It looks like since we are obscuring what is going on, people just keep adding directories to the list rather than figuring out a scheme for managing the header files. This patch airs our dirty laundry. It is a first step in figuring out what to clean up.

And in the case of libraries, why is any of this even being done? By moving this out of the function, we have this staring us in the face. and will be reminded that we need to take a look at it. I imagine there was some rationale behind _always_ adding an rpath link to the .libs directories (ignoring the (--disable-rpath configure option, by the way). Maybe it has something to do with running in-tree? But I have a suspicion that this is not the normal way to accomplish whatever it is that was trying to accomplish.

baallan:
@morrone halving a long list which is normally unseen (unless you're in the habit of 'make V=1')  isn't progress. That's why I don't make a note of it here. Unrolling a loop of m4 output with light pruning into configure.ac that we hardly ever look at isn't going to provoke anyone to do anything.

Proposing something which would eliminate the need for more than one -I would be progress.
We need the dirty laundry you refer to burned, not aired.

Re libtoolisms-- yeah, that's a correctness thing Narate discovered. We also need to be very careful that omission of something doesn't leave us pulling a library or header from system locations at build or run time or install. Of course things run in the build tree: that's what 'make check' does.

tom95858:
@morrone, @baallan Did this git resolved?

baallan:
@tom95858 I think it's fair to say that Chris and I are still fully opposed on this specific patch. We seem to be agreed that the real underlying problem is hyperdecomposed header and library locations and/or include statement styles which prevent us from achieving build with a simple
-I$ldms_include_dir -L$ldms_libdir

tom95858:
@baallan, @morrone, do either of you have suggestions for combining libraries and/or headers into a smaller number of larger name spaces?

baallan:
I would start by picking what we want our includes to look like in third party code (and work it so that we use the same #includes in ours). I propose (@morrone may have a better suggestion) (because I'm a little phobic about giant header directories)

```
-I$prefix/include
#include "ldms/zap/thing1.h"
#include "ldms/util/thing2.h"
#include "ldms/ldmsd/ldmsd.h"
```

reducing lib/* to lib/zap, lib/third, and lib/util (all the other subdirss go to util) and then move subdirs zap, third, and ovis_util into ovis/ldms/src and get rid of ovis/lib entirely.

Another alternative is to jump in with both feet on the  $top/[src,include,] organization and have a single include directory for everything except plugins and plugin-type-specific apis. I've always hated the idea that x.h lives in a different directory than x.c, but I think I'm getting over it.
so

```
#include "ldms/ldmsd.h"
#include "ldms/store-api.h"
#include "ldms/csv_plugins_util/format.h"
```

baallan:
A thing that worries me (unrelated to dir layout), however, is we presently have the same symbols being defined in ldms,lib and sos implementations (e.g. rbt).

PR #669, opened 2021-03-17
Add missing COPYING file at top level

morrone:
This update renames the LICENSE to COPYING, because most of the
code refers to a "COPYING" file at the top level. This way
COPYING will now correctly state the dual-license nature of
ovis's licensing. The COPYING file refers to a the gpl v2
in a file named "gpl-2.0.txt" at the top level for further
clarity.


morrone:
If this looks reasonable, I'll update PR #666 to reflect this change.

PR #671, opened 2021-03-18
replace sys_errlist which finally is gone from gnu libc.

baallan:
Looks like gnu libc drops sys_errlist from headers and libraries on or before 2.32.

### DIFF
--- a/lib/src/ovis_util/util.c
+++ b/lib/src/ovis_util/util.c
@@ -990,6 +990,41 @@ const char* ovis_errno_abbvr(int e)
 	return "UNKNOWN_ERRNO";
 }
 
+/** should be bigger than any expected strerror string.
+ * In 2020 ~50 was the upper limit.
+ */
+#define OVIS_SYS_ERRLEN 80
+
+/** should be bigger than any expected errno value.
+ * In 2020 ~140 was the upper limit.
+ */
+#define OVIS_SYS_NERR 256
+
+static char ovis_sys_errlist[OVIS_SYS_NERR][OVIS_SYS_ERRLEN];
+static int ovis_sys_errlist_initialized;
+static int ovis_sys_nerr;
+/* glibc has dropped support for sys_errlist.
+ * we copy it via strerror on first call.
+ * Make an unused call to ovis_strerror from main to avoid races.
+ */
+static void ovis_sys_errlist_init() {
+	int i = 0;
+	char unknown[OVIS_SYS_ERRLEN];
+	strerror_r(OVIS_SYS_NERR, unknown, sizeof(unknown));
+	while (!isdigit(unknown[i]) && unknown[i] != '\0')
+		i++;
+	unknown[i] = '\0';
+	size_t ulen = strlen(unknown);
+
+	for (i = 0 ; i < OVIS_SYS_NERR ; i++) {
+		strerror_r(i, ovis_sys_errlist[i], OVIS_SYS_ERRLEN);
+	}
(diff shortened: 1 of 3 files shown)


lib/src/ovis_util/util.c
```diff
@@ -990,21 +990,53 @@ const char* ovis_errno_abbvr(int e)
...
+ * In 2020 ~140 was the upper limit.
+ */
+#define OVIS_SYS_NERR 256
+
+static char ovis_sys_errlist[OVIS_SYS_NERR][OVIS_SYS_ERRLEN];
+static int ovis_sys_errlist_initialized;
+static int ovis_sys_nerr;
+/* glibc has dropped support for sys_errlist.
+ * we copy it via strerror on first call.
+ * Make an unused call to ovis_strerror from main to avoid races.
```
morrone, line 1008: I think I would prefer to see an explicit initializer exported and employed. Have you followed this instruction everywhere that ovis_strerror is employed?

lib/src/ovis_util/util.c
```diff
@@ -990,21 +990,53 @@ const char* ovis_errno_abbvr(int e)
...
+ */
+static void ovis_sys_errlist_init() {
+	int i = 0;
+	char unknown[OVIS_SYS_ERRLEN];
+	strerror_r(OVIS_SYS_NERR, unknown, sizeof(unknown));
+	while (!isdigit(unknown[i]) && unknown[i] != '\0')
```
morrone, line 1014: Why do we want to strip off the rest of the string beginning at any digit? Needs a code comment if this is the right thing to do.

lib/src/ovis_util/util.c
```diff
@@ -990,21 +990,53 @@ const char* ovis_errno_abbvr(int e)
...
+/** should be bigger than any expected errno value.
+ * In 2020 ~140 was the upper limit.
+ */
+#define OVIS_SYS_NERR 256
+
+static char ovis_sys_errlist[OVIS_SYS_NERR][OVIS_SYS_ERRLEN];
+static int ovis_sys_errlist_initialized;
+static int ovis_sys_nerr;
+/* glibc has dropped support for sys_errlist.
+ * we copy it via strerror on first call.
```
morrone, line 1007: This could use a better explanation. We can't copy something that isn't supported. Instead describe what it actually does, which is build an array of error strings by calling strerror_r() for all possible error values.

And "on first call" doesn't apply to this function, it applies to ovis_strerror().

morrone (changes requested):

lib/src/ovis_util/util.c
baallan: The current gnu libc uses a format for unknown errno values of "Unknown error %d". This code assumes that format will persist and allow us to identify the actual value of sys_nerr (the first occurance of the common prefix, whatever it is).
If the assumption fails, we fall back to OVIS_SYS_NERR -1 and everything will still run 'correctly' but with different string formatting for the 'unknown' error cases.

lib/src/ovis_util/util.c
baallan: added exported init function. it's use is optional.

lib/src/ovis_util/util.c
baallan: this function is externally uncallable in the previous case. I just now made it callable as ovis_sys_errlist_init, but its use is optional. At least the name should now be clear when calling it from main()s.

baallan:
superceded by 672.

Issue #670, opened 2021-03-18
build fails on AArch64, Fedora 33 


LutzWeischerFujitsu:
[jw@cn06 build]$ make VERBOSE=1 -j$(nproc)
make  all-recursive
make[1]: Entering directory '/data/jw/ovis/build'
Making all in lib
make[2]: Entering directory '/data/jw/ovis/build/lib'
Making all in src
make[3]: Entering directory '/data/jw/ovis/build/lib/src'
Making all in third
make[4]: Entering directory '/data/jw/ovis/build/lib/src/third'
make[4]: Nothing to be done for 'all'.
make[4]: Leaving directory '/data/jw/ovis/build/lib/src/third'
Making all in ovis_util
make[4]: Entering directory '/data/jw/ovis/build/lib/src/ovis_util'
  CC       util.lo
../../../../lib/src/ovis_util/util.c: In function 'ovis_strerror':
../../../../lib/src/ovis_util/util.c:1000:19: error: 'sys_nerr' undeclared (first use in this function)
 1000 |  if (e >=0 && e < sys_nerr)
      |                   ^~~~~~~~
../../../../lib/src/ovis_util/util.c:1000:19: note: each undeclared identifier is reported only once for each function it appears in
../../../../lib/src/ovis_util/util.c:1001:10: error: 'sys_errlist' undeclared (first use in this function)
 1001 |   return sys_errlist[e];
      |          ^~~~~~~~~~~
make[4]: *** [Makefile:672: util.lo] Error 1
make[4]: Leaving directory '/data/jw/ovis/build/lib/src/ovis_util'
make[3]: *** [Makefile:576: all-recursive] Error 1
make[3]: Leaving directory '/data/jw/ovis/build/lib/src'
make[2]: *** [Makefile:505: all-recursive] Error 1
make[2]: Leaving directory '/data/jw/ovis/build/lib'
make[1]: *** [Makefile:624: all-recursive] Error 1
make[1]: Leaving directory '/data/jw/ovis/build'
make: *** [Makefile:532: all] Error 2
[jw@cn06 build]$

baallan:
@LutzWeischerFujitsu  please also provide the version of gcc you are using.

baallan:
It appears this is likely related to:
https://bugzilla.redhat.com/show_bug.cgi?id=1856765

baallan:
I have created a patch #671 which resolves this (and the warning from ld in older versions about using sys_errlist).

[...]

PR #672, opened 2021-03-18
compensate for gnu libc 2.32 removal of sys_errlist and use alternate strerror thread safely

baallan:
This solves reported issue #670, by switching everywhere to STRERROR macro defined in ovis_util/util.h.

### DIFF
--- a/ldms/src/store/store_csv_common.c
+++ b/ldms/src/store/store_csv_common.c
@@ -49,6 +49,7 @@
 
 
 #define PLUGNAME 0
+#define _GNU_SOURCE
 #define store_csv_common_lib
 #include <stdlib.h>
 #include <stdio.h>
@@ -161,8 +162,6 @@ static int validate_env(const char *var, const char *val, struct csv_plugin_stat
 
 int create_outdir(const char *path, struct csv_store_handle_common *s_handle,
 	struct csv_plugin_static *cps) {
-#define EBSIZE 512
-	char errbuf[EBSIZE];
 	if (!cps) {
 		return EINVAL;
 	}
@@ -194,23 +193,19 @@ int create_outdir(const char *path, struct csv_store_handle_common *s_handle,
 		case EEXIST:
 			break;
 		default:
-			strerror_r(err, errbuf, EBSIZE);
 			cps->msglog(LDMSD_LERROR,"create_outdir: failed to create directory for %s: %s\n",
-				path, errbuf);
+				path, STRERROR(err));
 			return err;
 		}
 	}
 
 	/* cps->msglog(LDMSD_LDEBUG,"create_outdir: f_mkdir+p(%s, %o)\n", path, mode); */
 	return 0;
-#undef EBSIZE
 }
 
 void rename_output(const char *name,
 	const char *ftype, struct csv_store_handle_common *s_handle,
 	struct csv_plugin_static *cps) {
-#define EBSIZE 512
-	char errbuf[EBSIZE];
 	if (!cps) {
 		return;
 	}
@@ -240,9 +235,9 @@ void rename_output(const char *name,
 		int merr = chmod(name, mode);
 		int rc = errno;
 		if (merr) {
-			strerror_r(rc, errbuf, EBSIZE);
 			cps->msglog(LDMSD_LERROR,"%s: rename_output: unable to chmod(%s,%o): %s.\n",
-				cps->pname, name, s_handle->rename_perm, errbuf);
+				cps->pname, name, s_handle->rename_perm,
+				STRERROR(rc));
 		}
 	}
 
@@ -254,9 +249,8 @@ void rename_output(const char *name,
 		int merr = chown(name, newuid, newgid);
 		int rc = errno;
 		if (merr) {
-			strerror_r(rc, errbuf, EBSIZE);
 			cps->msglog(LDMSD_LERROR,"%s: rename_output: unable to chown(%s, %u, %u): %s.\n",
-				cps->pname, name, newuid, newgid, errbuf);
+				cps->pname, name, newuid, newgid, STRERROR(rc));
 		}
 	}
 
@@ -407,9 +401,8 @@ void rename_output(const char *name,
 		case EEXIST:
 			break;
 		default:
-			strerror_r(err, errbuf, EBSIZE);
 			cps->msglog(LDMSD_LERROR, "%s: rename_output: failed to create directory for %s: %s\n",
-				cps->pname, newname, errbuf);
+				cps->pname, newname, STRERROR(err));
 			free(newname);
 			return;
 		}
@@ -421,19 +414,15 @@ void rename_output(const char *name,
 	err = rename(name, newname);
 	if (err) {
 		int ec = errno;
-		strerror_r(ec, errbuf, EBSIZE);
 		cps->msglog(LDMSD_LERROR,"%s: rename_output: failed rename(%s, %s): %s\n",
-			    cps->pname, name, newname, errbuf);
+			    cps->pname, name, newname, STRERROR(ec));
 	}
 	free(newname);
-#undef EBSIZE
 }
 
 void ch_output(FILE *f, const char *name,
 	struct csv_store_handle_common *s_handle,
 	struct csv_plugin_static *cps) {
-#define EBSIZE 512
-	char errbuf[EBSIZE];
 	if (!cps) {
 		return;
 	}
@@ -455,9 +444,8 @@ void ch_output(FILE *f, const char *name,
 		int merr = fchmod(fd, mode);
 		int rc = errno;
 		if (merr) {
-			strerror_r(rc, errbuf, EBSIZE);
 			cps->msglog(LDMSD_LERROR,"ch_output: unable to chmod(%s,%o): %s.\n",
-				name, s_handle->create_perm, errbuf);
+				name, s_handle->create_perm, STRERROR(rc));
 		}
 	}
 
@@ -469,12 +457,11 @@ void ch_output(FILE *f, const char *name,
 		int merr = fchown(fd, newuid, newgid);
 		int rc = errno;
 		if (merr) {
-			strerror_r(rc, errbuf, EBSIZE);
 			cps->msglog(LDMSD_LERROR,"ch_output: unable to fchown(%d, (%s),%lu, %lu): %s.\n",
-				fd, name, newuid, newgid, errbuf);
+				fd, name, newuid, newgid, STRERROR(rc));
 		}
 		cps->msglog(LDMSD_LDEBUG,"ch_output: fchown(%d, (%s),%lu, %lu): %s.\n",
-			fd, name, newuid, newgid, errbuf);
+			fd, name, newuid, newgid, STRERROR(rc));
 	}
 }
 
--- a/lib/src/ovis_auth/auth.c
+++ b/lib/src/ovis_auth/auth.c
@@ -1,8 +1,8 @@
 /* -*- c-basic-offset: 8 -*-
- * Copyright (c) 2015-2016,2018 National Technology & Engineering Solutions
+ * Copyright (c) 2015-2021 National Technology & Engineering Solutions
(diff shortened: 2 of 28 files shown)


baallan:
The compatibility fail here seems to be an upload failure on the server side after the test passes.

tom95858:
@baallan, I don't see why this is necessary or advisable. Adding additional serialization into the daemon when the potential race artifact is a corrupted log message is bad practice.

baallan:
@tom95858 this is actually removing strerror_r serialization from the daemon in most locations by switching to ovis_strerror which requires no locking. Any place that strerror_r is still in use, it can be replaced if we agree to changed the dependencies of that place to include util.h/libutil. I can add that to this patch if need be.

lib/src/ovis_auth/auth.c
```diff
@@ -65,6 +65,7 @@
 #include <stdio.h>
 #include <arpa/inet.h>
 #include "auth.h"
+#include "ovis_util/util.h"
```
morrone, line 68: Why include ovis_util/util.h as part of this commit?

ldms/src/store/store_csv_common.c
```diff
@@ -913,10 +899,8 @@ void close_store_common(struct csv_store_handle_common *s_handle, struct csv_plu
 	}
 
 	rename_output(s_handle->filename, FTYPE_DATA, s_handle, cps);
-	if (s_handle->altheader)
-		rename_output(s_handle->headerfilename, FTYPE_HDR, s_handle, cps);
-	if (s_handle->typeheader)
-		rename_output(s_handle->typefilename, FTYPE_KIND, s_handle, cps);
+	rename_output(s_handle->headerfilename, FTYPE_HDR, s_handle, cps);
+	rename_output(s_handle->typefilename, FTYPE_KIND, s_handle, cps);
```
morrone, line 903: This hunk does not appear to be related to this PR.

lib/src/ovis_auth/auth.c
```diff
@@ -141,8 +142,10 @@ char *ovis_auth_get_secretword(const char *path, ovis_auth_log_fn_t log)
 	struct stat pstat;
 	if (stat(path, &pstat)) {
 		ret = errno;
+		char ebuf[80];
+		strerror_r(errno, ebuf, sizeof(ebuf));
 		log("Auth error: %s while trying to stat %s\n",
-			strerror(errno), path);
+			ebuf, path);
```
morrone, line 148: Why convert this to strerror_r() when the rest of this commit is changing everything to ovis_strerror()?

morrone (changes requested):

ldms/src/store/store_csv_common.c
baallan: good catch. that's a reversal of a previous PR, and I have no idea how it got there. Must have been an error in rebasing. will fix.

lib/src/ovis_auth/auth.c
baallan: This was the start of getting rid of the strerror_r/strerror from auth, and then I paused because it adds util to the link line of auth.  In the interest of getting rid of all strerror usage, I'll go ahead and finish adding util to the link.

lib/src/ovis_auth/auth.c
baallan: to at least have it be thread safe and the log messages not potentially corrupted. switching to ovis_strerror would be my preference.

morrone:
@tom95858 , unless I am missing something, "thread safe" (or "reentrant") here does not imply locking. It just means employing a reentrant model: using a user-provided buffer to store the error string (as in strerr_r) rather than employing an internal buffer that would collide between reentrant calls (as in strerr). So I don't think this is introducing serialization.

@baallan, it would be helpful if the one new commit in this PR was actually rebased on the commits in PR #671, since that was apparently your intent. Having it based on copies of the commits #671 makes things harder on reviewers because we would have to go in and actually compare the commits manually to determine that the code is correct. If you base this commit on that one, we can just review #671 and then review the one additional commit in this PR.

tom95858:
@morrone, @baallan, I have done a poor job explaining my concern. The only time that strerror returns anything other than a static const char * is when the errno that's passed in is not recognized. IOW, the errno value is invalid. In that case, it sprintf's an error into a static char * and returns it to you. In short, that's the *only* time that there is any kind of thread-safety issue; and if that happens you have another bug.

So to resolve this "problem", we've added a whole new interface (ovis_strerror) that copies all of the existing static const char * into another array of static const char * and resolves the thread issue by returning "unknown error" instead of a string that includes the offending error.

So in short, we've changed 25+ files, countless lines of code, duplicated memory and resolved the potential race of two threads providing sterror with a bad errno and having the resulting error string corrupted. I apologize, but I find this ridiculous.

baallan:
@tom95858 This is not the first time the libc developers have taken away a non-posix feature and made us do "ridiculous" things. 

We need this fix to run on a new machine at SNL and elsewhere (#670).
The lines of code changed (excluding adding 2 small functions in util.c) is < 100.

And it gets rid of a whole class of repetitive warnings when thread analyzing ldmsd, as a bonus.

morrone:
> @morrone, @baallan, I have done a poor job explaining my concern. The only time that strerror returns anything other than a static const char * is when the errno that's passed in is not recognized. IOW, the errno value is invalid. In that case, it sprintf's an error into a static char * and returns it to you. In short, that's the _only_ time that there is any kind of thread-safety issue; and if that happens you have another bug.

You are making assumptions about strerror() implementation, which isn't the best practice. 

I'm not going to advocate this PR's approach of copying all of the strings or using ovis_strerror() is the best approach. I would probably prefer to just use strerror_r(), and deal with the extra lines needed for the temporary buffers. And I agree with your complaint about losing the error number for the unknown errror case.

morrone:
The way I would probably do this is build the strerror_r() call into a logging function.

tom95858:
@baallan, @morrone there is a "%m" format specifier (non-Posix GNU extension). It would get rid of all calls to strerror(errno).

There are also a lot of patterns like this:

```C
{
    int xxx = errno;
    ... strerror(xxx);
}
```
Not sure why that pattern exists, perhaps paranoia about printf/strerror futzing with errno? But that pattern could also be removed with "%m".

Finally, for this kind of thing: 

```C
    ...strerror(ERRNO);
```

Something super simple like this:

```
#define STERROR(_rc_) ({ char *msg=alloca(32); strerror_r(_rc_, msg, 32); msg; })
```
and change strerror(...) --> STRERROR(...)

For those remaining call sites (~22 or so)

baallan:
@tom95858 The use of alloca as proposed here results in an invalid stack pointer being passed into the format machinery. The buffer size needed (for LANG C) is at least 60, and accounting for the internationalization library under strerror probably 80. We need to do either (a) strerror_r and it's syntactic overheads everywhere {as @morrone suggests) or (b) ovis_strerror to be portable to current gnu libc and posix standards. 

Strerror is documented as changing errno sometimes, as may other calls that may be inlined to create the printf arguments, so for many situations where one might want to preserve the errno value (debug, further return code, etc) the x=errno pattern before print is often justified. Certainly not claiming it's always needed in every case where it may have been pasted in ldms development.

morrone:
A third option is to make a new logging function (or set of logging functions) that always append strerror_r() to whatever message the caller wants printed. Basically what we want is perror(), but with format string and variable number of arguments like (f)printf().

We might also consider why we are sending error messages to stdout rather than stderr.

baallan:
The aim of this patch was to get a thread-safe change in place to resolve the build failure on libc 2.32.
A proper refactor/expansion of the logging api to include an option that appends/inserts the strerror strings into selected messages would be a fine thing but be out of scope. Also out of scope is a change I think we all want to expand the api to something like:
LOG(facility, level, format, ...) where we can change the log level by the facility tag.

morrone:
In that case, we could provide sys_nerr and anything else that is missing through autoconf tests and replacement until such point as we have better APIs that no longer use them. That keeps the work-arounds outside of the code proper.

baallan:
@morrone If you have a 'compat' solution such as you suggest, please do so we can close issue 670.

tom95858:
> @tom95858 The use of alloca as proposed here results in an invalid stack pointer being passed into the format machinery.

I'm fairly certain that this claim is false. From the documentation:

```
The  alloca()  function  allocates  size  bytes  of space in the stack frame of the caller.  This temporary space is automatically freed when the **function** that called alloca() returns to its caller.
```
That said, I feared your claim as well, i.e. the { } block would dealloc the stack buffer on exit. However, I reviewed the assembly code and that was not the case. The documentation seems to confirm that the memory is not freed until the "...**function** that called alloca()..." returns.

> The buffer size needed (for LANG C) is at least 60, and accounting for the internationalization library under strerror probably 80.

The 32B was just an example. Make it larger.

> We need to do either (a) strerror_r and it's syntactic overheads everywhere {as @morrone suggests) or (b) ovis_strerror to be portable to current gnu libc and posix standards.
> 

The macro is what @morrone suggests without the coding of the same sequence 22 times.

> Strerror is documented as changing errno sometimes, 

errno is passed in as a parameter, whether strerror changes errno does not affect the parameter value.

> ... as may other calls that may be inlined to create the printf arguments, so for many situations where one might want to preserve the errno value (debug, further return code, etc) the x=errno pattern before print is often justified. Certainly not claiming it's always needed in every case where it may have been pasted in ldms development.

Sure, but the ones I looked at had none of these conditions.

All that said, I would like to avoid changing/re-changing this logic over and over again. The use of str_errlist was reluctantly accepted in the first place after a similarly long exchange in June of last year and here we are again.

If alloca behaves like I believe it does; are we good to go or is there some other concern?

tom95858:
PS: is anyone worried about the "%m" format specifier? I checked it with clang, gcc, and PrgEnv-intel on Cray.

baallan:
re %m, if it's something you can check for at configure and provide an alternative for, then fine. We have quite a bit of gnu C-ism in our code base already-- I wouldn't choose now to get strict posix religion, or c99 religion or whatever.

tom95858:
@baallan, I think we would have to build a specific config test. But that said, If we don't like it, let's just not use it. I'm actually not a fan of the fact that it doesn't take an argument, but rather uses errno directly.

baallan:
@tom95858 With a little testing I'm starting to like the STRERROR macro you proposed instead of a function, but it does appear that the ({}) syntax is still a gnu extension on c99/c11. Does something like the following work on your clang?

Yes, the _msg_ below may be seat belt and suspenders, but it would make reading processed code more clear when debugging in the presence of a msg variable. I would still like to name the macro ovis_strerror to avoid typing in all caps (and since the macro acts like a function).

```
#include "stdlib.h"
#include "stdio.h"
#include "string.h"
#define STRERROR(_rc_) ({ char *_msg_ = alloca(80); strerror_r(_rc_, _msg_, 80); _msg_; })
void foo() {
	printf("p1: %d %s\n", 11, STRERROR(11));
	printf("p2: %d %s\n", 132, STRERROR(132));
	printf("p3: %s %s\n", STRERROR(8), STRERROR(133));
}
int main() {
	foo();
	return 0;
}
```

tom95858:
Hi @baallan,
I'm pretty sure the ({}) syntax is all the way back to K&R, but I'm certainly willing to be informed otherwise. See old implementations of container_of before this kind of thing was added as a GNU extension

baallan:
@tom95858 I agree it's likely to be a universal extension, but I can't actually find it in the recent standards docs. gnu documents it as extension 'statement exprs'. if we're happy requiring gcc on hypothetical embedded platforms, or if at least clang allows it (some arm dev teams are clang based), I think that's good enough for me.

baallan:
Per last week's phone call w/ @tom95858 I will update this to use the suggested macro approach with statement expression style. apparently this works across gcc/clang/intel.

baallan:
@tom95858 assuming this passes all the build checks still running, it is ready to go unless someone wants a prefix (OVIS_, O_, L_?) on STRERROR.

PR #673, opened 2021-03-29
Fix buffer reallocation in ldmsd_request

narategithub:
The size provided to the `realloc()` call was not the size of the entire
buffer. Instead, the remaining size (`sz`) was given. This caused a
buffer overrun on `ldmsd` (e.g. when processing `thread_stats` request).
This patch corrects the size of the reallocation to be the size of the
entire buffer (`off + sz`).

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -5503,10 +5503,10 @@ struct op_summary {
 	int len = snprintf(s, sz, __VA_ARGS__);			\
 	if (len >= sz) {					\
 		uint64_t off = (uint64_t)s - (uint64_t)buff;	\
-		sz += __APPEND_SZ;				\
-		s = realloc(buff, sz);				\
+		sz += LDMS_ROUNDUP(len-sz, __APPEND_SZ);	\
+		s = realloc(buff, off + sz);			\
 		if (!s) {					\
-			goto __APPEND_ERR;				\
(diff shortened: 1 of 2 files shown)


narategithub:
@tom95858 I can reproduce the buffer overrun detected by valgrind on the `thread_stats` command processing. What happened was that when the buffer was reallocated, the remaining size was supplied to `realloc()` instead of the size of the entire buffer. With this patch, the buffer overrun is not seen anymore.

nichamon (approved):

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5498,15 +5498,16 @@ struct op_summary {
 	uint64_t op_mean_us;
 };
 
+#define __ROUND(X) ((((X)-1)|0xFFF)+1) /* round-up to multiples of 4K */
```
tom95858, line 5501: @narategithub,

Coudl we just add this to ldms.h or something so we can use it everywhere.
```C
#define LDMS_ROUNDUP(_sz_, _align_) (((_sz_) + (_align_) - 1) & ~((_align_)-1))
```

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_request.c
narategithub: OK.

PR #678, opened 2021-04-02
backport rdc_sampler from amd on master to OVIS-4

baallan:
Tested against a rhel7 ROCM 4.0.0 installation.

Notable minor differences from master rdc_sampler on which this is base:

- Uses the 4.3.6 sampler_base api from OVIS-4 instead of the draft multi-instance API from master.
- The sampler instance can be configured multiple times, but each call wipes out all previous configuration.
- Man pages supplied.
- All options are defaulted if not given.
- Allows selection of any available field, rather than only the 10 or so default ones; the default list is used
if metrics=xxx is not given. As the RDC header evolves with more fields, the plugin should not need to.
- The metrics argument accepts values that redefine the default list: 'base' and 'xgmi', which may be extended as hardware evolves.
- The default metric list is 'base', which exclude xgmi communication metrics.
- Requested metrics are checked for type compatibility (int64_t).
- New option 'shape' allows the user to select set geometry: 0-  a single wide set of the selected metrics across each gpu; 1- N per-device sets, with a 'device' column included. 2- (reserved, but not supported) a single narrow set with a vector indexed by gpu for each metric. Some unfortunate poking through the abstraction boundary of sampler_base.h results; I have tried to make it obvious where this occurs. Getting rid of the 'shape' option is an anticipated improvement.
- New option 'warmup' controls the delay between starting the amd library and attempting to sample. Do-nothing sample calls instead of sleep() are used to implement the delay.
- The configuration schema= given is used as the base of a derived schema name, where the suffix is determined automatically by the shape= and metrics= options and number of gpus. For a given configuration, the schema name can be determined by running the ldms_rdc_schema_name utility supplied before configuring other plugins, if needed.
- The "business logic" is isolated to rdcinfo.c and the plugin boilerplate is in rdc_plugin.c.
- The obvious 'singleton' pattern is used, and lock management is confined to rdc_plugin.c so the author/maintainer of rdcinfo.c can view it as single-threaded.
- Any failure in querying metrics causes the plugin to rdc_shutdown the gpu query library, otherwise the rdc library spews to log regularly, annoying sys admins.
- Configure with this sampler on AMD platforms looks like:
   ./configure --enable-rdc RDC_CFLAGS=-I/opt/rocm/rdc/include RDC_LIBS=-L/opt/rocm/rdc/lib ...
- Once built, a quick test is 'ldms-static-test.sh rdc_sampler' if ldms is in your PATH.

Open issues:
- Is there any sort of "block query" command to get a batch of counters all at once instead of Nmetrics*Pgpus individual calls?
- There is a probable race at ldmsd exit with the embedded amd library unless the rdc plugin is stopped and terminated. In some circumstances, this leads to a seg fault during process cleanup.


baallan:
Amended to resolve trailing spaces and disable rdc_ function stubs.

morrone:
Perhaps we also should rework this plugin to use a single-GPU schema and one metric set per GPU (all sharing the same schema). We see to be moving that direction with many of our plugins, so it might be best to do that up front on new plugins rather than dealing with the change over later.

configure.ac
```diff
@@ -123,6 +123,7 @@ OPTION_DEFAULT_ENABLE([ovis_ctrl], [ENABLE_OVIS_CTRL])
 OPTION_DEFAULT_ENABLE([ovis_util], [ENABLE_OVIS_UTIL])
 OPTION_DEFAULT_ENABLE([ovis_auth], [ENABLE_OVIS_AUTH])
 OPTION_DEFAULT_ENABLE([zap], [ENABLE_ZAP])
+OPTION_DEFAULT_DISABLE([rdc], [ENABLE_RDC])
```
morrone, line 126: I believe you alluded to this in your PR comment about "fixing" configure.ac. We really shouldn't be using OPTION_DEFAULT_DISABLE for this. We instead need a proper library test for the rdc library. This plugin would just be enabled when the required libraries are included.

baallan: agree. just don't have a valid environment yet to develop that.

baallan:
See 'shape' mentioned in the initial comment that I made when posting the PR. Does exactly what you want. The default is the wide set, so that this implementation will work the same way AMD's original did (from a backend storage point of view).
> 
> Perhaps we also should rework this plugin to use a single-GPU schema ...

ldms/src/sampler/rdc_sampler/Makefile.am
```diff
@@ -0,0 +1,45 @@
...
+COMMON_LIBADD = -lsampler_base -lldms -lovis_util -lcoll \
+                @LDFLAGS_GETTIME@
+
+# with proper rdc present:
+#rdcinc = @ROCM_RDC_INC_INCDIR_FLAG@
+#rdclib = -lrdc_bootstrap
+#rdcldflags = @ROCM_RDC_INC_LIBDIR_FLAG@ @ROCM_SMI_INC_LIBDIR_FLAG@
+
+# with just headers, no runtime
+rdcinc = -I/scratch1/baallan/rdc/include
```
morrone, line 21: "baallan" hard coded.

ldms/src/sampler/rdc_sampler/rdc.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
morrone, line 1: I am not sure why this script exists. All it does is output a static rdc.conf file. We should probably just include an rdc.conf example file rather than having a script. (Update README if this is changed).

morrone (changes requested):

morrone:
This PR should be marked "Draft" since it is still a work in progress.

morrone:
> See 'shape' mentioned in the initial comment that I made when posting the PR. Does exactly what you want. The default is the wide set, so that this implementation will work the same way AMD's original did (from a backend storage point of view).

I don't think we need to default to wide. AMD's plugin hasn't made it into any production LDMS release yet. I think per-gpu metrics sets should be the default.

I also think we should consider making that the _only_ supported "shape". The extra shape code significantly complicates the sampler code. I don't think we want to go down the path of making plugin authors individually implement all of the same additional complexity over and over again. If we genuinely think we need different shapes for plugins, then maybe we need to take a step back and consider offering that functionality to plugin authors through new APIs.

ldms/src/sampler/rdc_sampler/README
```diff
@@ -0,0 +1,15 @@
+
+rdc.sh generate rdc.conf
+
+rdc.conf config script example
+
+old/
```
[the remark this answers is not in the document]
baallan: correct

ldms/src/sampler/rdc_sampler/rdc.sh
baallan: agree

baallan:
@morrone I suspect we will defer to AMD's desires here, if they have some backend processing commitment already regarding the wide set.  If not, I agree a single narrow shape with multiple sets would be much easier to maintain long term.

tom95858:
@baallan, @morrone: I think the single set per GPU is the way to go. Otherwise it becomes really difficult to decode all of this at the back-end. I believe that AMD would welcome our guidance here as to why this is the preferred approach.

miketxli:
@baallan 
Should we expect it to work on older hardware: Vega 10 [Radeon Instinct MI25]?  Yes, It should work with Vega10 and later.

Are there rhel-derived packages for RDC, or are rdc libraries always a local source build? Yes, see https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html. It's not included in rcom-dkms. Try to do "sudo yum install rdc". It works in ubuntu like "sudo apt-get -y install rdc". If it doesn't work that way, you can get the rpm from

miketxli:
@baallan @morrone @tom95858 
The single set per GPU is fine for AMD.

miketxli:
@baallan  Does it work in a real gpu environment, as rewritten? I will try to build and test it on my test machine.

miketxli:
@baallan I think we need a fix in configure.ac and Makefile.am in order to build this correctly.

baallan:
@miketxli  trying to catch up on the conversations:
 - I have not yet tested this against real hardware. I'm working on getting a local hardware/software environment set up that provides a gpu I could test this against. This may take some time. If you happen to have a test node someplace I can access quickly, I'd be happy to work on it there with you instead. 
 - I can remove the wide set option easily and go with a set per device if you are sure it is ok with your team going forward.
 - Because I do not have a full understanding yet of the rdc build/runtime environment, I haven't ported in the patches to configure.ac; my first glance left me thinking maybe there were more flags added to configure than needed.
 - Can you tell me if I need a specific version of gcc just to use rdc with real hardware? Is there some 'must match gcc version=Y' or 'must match gcc version used to build RPM named X"?
- Please let me know how a build/test of the sampler configured with shape=1 goes after you've hacked around the missing bits in configure.ac and removed the baallan gunk in your makefile.am.

miketxli:
- I have test machines, but they are all located in AMD's internal network. I don't have machines that are accessible from outside. 
- We can keep shape configuration for now. It's bit easy for testing
- Yes, RDC itself is based on rocm_smi library, which is a low level library and included in rocm-dkms. You really need get the rocm frame work and RDC installed + HW in order to test it.
- RDC require cmake 3.15
- I will try to hack configure.ac and makefile.am and just to get it compiled and tested.

baallan:
@miketxli It looks like I may have a complete rdc package environment now on a local gpu host.  Is there a simple command line check of the environment adequateness which you can recommend, perhaps a utility showing something about the GPUs discovered?

miketxli:
> @miketxli It looks like I may have a complete rdc package environment now on a local gpu host. Is there a simple command line check of the environment adequateness which you can recommend, perhaps a utility showing something about the GPUs discovered?

run /opt/rocm/bin/rocm-smi -a

ldms/src/sampler/rdc_sampler/rdcinfo.c
```diff
@@ -81,7 +82,39 @@ static rdc_field_t default_field_ids[] = {
...
+	"RDC_FI_PCIE_TX,"
+	"RDC_FI_PCIE_RX,"
+	"RDC_FI_GPU_UTIL,"
+	"RDC_FI_GPU_MEMORY_USAGE,"
+	"RDC_FI_GPU_MEMORY_TOTAL,"
+	"RDC_FI_ECC_CORRECT_TOTAL,"
+	"RDC_FI_ECC_UNCORRECT_TOTAL";
+
+/* expansion for metrics=v20 option */
+const char *default_metrics_20 =
```
miketxli: XGMI is a fast bus between GPUs. There is a physic bus related to this and user can choose to use it or not even for Vega20+ cards. I think to name this as default_metrics-xgmi is better.

baallan: @miketxli  can you suggest appropriate alternatives to both the abbreviations accepted by the metrics= argument? currently it's v10, v20. Sounds to me like maybe it should be (something_else_for_v10),xgmi instead. I can then make the c code match appropriately.

miketxli: I would say with_xgmi and without_xgmi, or xgmi and non_xgmi something like that. XGMI is introduced since Vega20. But not all new cards do support XGMI. Even for Vega20, there is an optional piece of HW to interconnect GPUs. If user choose to not connect. There will still be errors with XGMI.

baallan:
@miketxli I think this is all ready to go. Please let us know if it runs for you (notice the change in configure arguments and the test case provided in the PR main comment above).

@valleydlr I believe we have reached 'good enough' in the amd sampler. A separate PR addresses memory bugs turned up while testing this. @tom95858 can merge at his convenience or wait for test feedback from @miketxli.

miketxli:
> @miketxli I think this is all ready to go. Please let us know if it runs for you (notice the change in configure arguments and the test case provided in the PR main comment above).
> 
> @valleydlr I believe we have reached 'good enough' in the amd sampler. A separate PR addresses memory bugs turned up while testing this. @tom95858 can merge at his convenience or wait for test feedback from @miketxli.

Great! I'm good with this, have tested with Vega10 and new Navi card. It works as expected.

baallan:
@miketxli Thanks for your help in getting this capability added!

morrone:
@tom95858 I thought that we we dropping the shape stuff before landing? At the very least I think we should switch the default to per-device  before we do a release. Better would be to drop the extra code.

tom95858:
@morrone, I apologize if this was a review requirement that I missed. I think @miketxli and @baallan were of the 'good enough' mind set. Given that one needs to install rocm etc... to even get this thing to compile, it seems relatively low risk to get this in so that it is more convenient for AMD, etc... to test and improve.

That said, I prefer switching the default to per-device if that is a currently implemented configuration option.  I don't see any benefit to combining everything into one set (the current default). Could you or someone else submit a pull request so that we can litigate this...soon.

tom95858:
Thanks @baallan for getting this all refactored. This is great.

I hope this doesn't offend anyone, particularly, @miketxli, but if we feel that separate sets is a condition of acceptance, we simply need to say so and explain why. I also think that having this as a per-sampler configuration option is not a good idea and should be discouraged; note that developers adding new samplers will look at this latest one for guidance.

I feel that separate sets per resource instance (GPU) should be the preferred approach  going forward. Requiring analysis modules (and/or stores) to separate device instances from a metric set is inherently hard to do and per resource type dependent.  We have resolved the set delete issue (to handle resources coming and going). It seems appropriate and preferred to move forward with a set-per-resource approach unless there is a very good reason to do otherwise.

miketxli:
@tom95858  There isn't any particular reason we have to keep per-sampler configuration. We are good with per-GPU configuration only.

baallan:
@miketxli @tom95858 @morrone I will remove the wide set option in another PR.

tom95858:
@baallan, @miketxli: thanks very much for the quick replies.

PR #680, opened 2021-04-06
Fix ldmsd_msg_gather to set the msg length to the total msg length

nichamon:
…egnth

ldmsd_msg_gather does not update the request header object's message
length to include the subsequent records' legnths after the first
record. Therefore, the message length may be shorter than the actual
message length. At this time, the bug is asymptotic because no
ldmsd_msg_gather’s callers use the message length. The fix is to prevent
future pitfalls.


nichamon:
@nick-enoent Tom let me know that you found a bug in `ldmsd_msg_gather`. Here is the fix. For some reason, I cannot add you as a reviewer.

nick-enoent:
Thanks for looking at this @nichamon. I've been having issues adding reviewers to my pull requests as well. Not sure if that's something to do with ovis repo privileges or what.

PR #682, opened 2021-04-08
rdc devices only version

baallan:
This removes the shape option, fixes a bug in delivery of the device metric, adds a debug option to ldms_rdc_schema_name, and updates manual pages.


baallan:
@tom95858 the wide set option has been removed and the code retested.

baallan:
Please squash when merging.

I assume it is expected behavior that meta metric changes within a base_sample_being/end cycle are not delivered on the immediately following set pull but appear to an aggregator after the set with the MD GN change has been delivered. For this reason I shifted the device name from meta to metric.

PR #685, opened 2021-04-12
fix usage of pkg_check_modules

baallan:
pkg_check_modules has some m4 foo that requires careful handling per comments in pkg.m4. This adds that care.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,6 +130,11 @@ OPTION_DEFAULT_DISABLE([ugni], [ENABLE_UGNI])
 dnl dnl OPTION_DEFAULT_DISABLE([sos], [ENABLE_SOS])
 dnl dnl OPTION_WITH_OR_BUILD([sos],[../sos],[sos/src sos/include ods/src ods/include])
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -130,15 +130,18 @@ OPTION_DEFAULT_DISABLE([ugni], [ENABLE_UGNI])
...
+PKG_PROG_PKG_CONFIG
+
 dnl RDC defaults to no instead of search because amd currently recommends
 dnl multiple non-default path installations (/opt).
 AC_ARG_ENABLE([rdc],
 	[AS_HELP_STRING([--enable-rdc], [attempt components that depend upon AMD rdc tooling @<:@default=no@:>@ for GPUs. Set RDC_CFLAGS=-I$rdcinclude and RDC_LIBS=-L$rdclibdir])],
 	[],
 	[enable_rdc="no"])
 if test "$enable_rdc" != "no"; then
-    m4_ifndef([PKG_CHECK_MODULES],
```
tom95858, line 140: @baallan, could you please remove this line from the cray-ugni section that follows?

tom95858 (changes requested):

configure.ac
baallan: done

PR #690, opened 2021-04-15
Fix ldmsd_request:stream_xprt_term()

narategithub:
When LDMS xprt terminated in ldmsd, ldmsd went through the collection
of stream clients to clean up the stream clients associated with the
transport. The collection of clients used `<remote_addr,stream_name>`
for rbt keys. However, on rdma transport, the remote addresses of two
transports from the same remote host got resolved to the same address.
This resulted in premature stream client cleanup due to the address
mixed-up. In addition, on the socket transport, `ldms_xprt_sockaddr()`
call in `stream_xprt_term()` returned ENOTCONN the
`ldms_xprt_sockaddr()` exited immediately, leaking the stream client
and the associated LDMS xprt.

This patch modifies the key to of the stream collections to use
`xprt` pointer instead.

NOTE: This has been tested on rdma and sock transports with GDB to
verify the correct find/removal.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -6080,10 +6080,8 @@ static int stream_republish_cb(ldmsd_stream_client_t c, void *ctxt,
 
 /* RSE: remote stream entry */
 struct __RSE_key_s {
-	/* remote addr */
-	struct sockaddr addr;
-	/* remote addr len */
-	uint32_t addr_len;
+	/* xprt ref */
+	ldms_t xprt;
 	/* stream name */
 	char name[];
 };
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 could you please give this a try?

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6348,33 +6311,24 @@ static int stream_client_dump_handler(ldmsd_req_ctxt_t reqc)
 
 void stream_xprt_term(ldms_t x)
 {
-	struct sockaddr local_sa;
 	__RSE_t ent;
 	struct rbn *rbn;
 	char _buff[sizeof(struct __RSE_key_s) + 256] = {};
```
tom95858, line 6316: The _RSE_key_s is not correct. The stream name is fixed at 256B, but the stream name size is arbitrary. We either need to clamp the stream name length or we need to make _name_ a pointer to a dup'd stream name string. We need this patch now, so I will accept it, but I'd like you to fix this and submit another pull request. I think the simplest thing is to make _name_ point to a dup'd string and handle the free of _name_ in the _RSE_key_s cleanup  logic.

PR #694, opened 2021-04-16
Temporary fix the xprt connection establishment delay

aazizo:


aazizo:
@tom95858 I added the check and delay code to check if the xprt is actually connected before returning a successfuk flag from the function, because when we tested with the kokkos connector we were losing the first few messages due to an initial connection problem.

aazizo:
BTW, this is the not the best fix and I just made it to make things work with kokkos in the mean time and show you the problem.

ldms/src/core/ldms_xprt.c
```diff
@@ -3701,6 +3701,18 @@ int ldms_xprt_connect_by_name(ldms_t x, const char *host, const char *port,
 	} else {
 		rc = ldms_xprt_connect(x, ai->ai_addr, ai->ai_addrlen, cb, cb_arg);
 	}
+
```
tom95858, line 3704: Are you specifying a callback function, i.e. the _cb_ argument? If you are, this function is not synchronous. If you want a synchronous connection, then you need to specify NULL for the _cb_ argument.

tom95858 (changes requested):
If you are specifying NULL for the _cb_ argument, then there is a bug in the internal sync_connect_cb function that needs to be fixed.

ldms/src/core/ldms_xprt.c
aazizo: We specify a callback function not NULL. I will check and get back to you.

PR #695, opened 2021-04-19
label origin of appended debug config lines in gender generated config file

baallan:


baallan:
@tom95858 please include this before 4.3.7.

tom95858:
@baallan, 4.3.7 is already tagged. For the purposes of testing, you should host this in your own branch. Regarding genders in general, per the users group meeting, we need to have a discussion about where we go with genders in the future.

PR #697, opened 2021-04-23
Fix ldmsd message with no attributes

narategithub:
For an ldmsd message that has no attributes (e.g.
LDMSD_FAILOVER_PEERCFG_REQ message), the flag of the message was
incorrectly set to only EOM instead of `SOM|EOM`. In addition, the
message length would be exactly 4 bytes. This patch addresses the
message length assertion and the message flags to correctly handle the
message with no attributes.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -3050,7 +3050,7 @@ int ldms_xprt_send(ldms_t _x, char *msg_buf, size_t msg_len)
 	if (!ldms_xprt_connected(x))
 		return ENOTCONN;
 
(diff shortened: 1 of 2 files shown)


narategithub:
Thanks @Snell1224 for testing out failover and discover that it wasn't working, and thanks @nichamon for helping in identifying the root cause.

This patch shall address the failover issue in @Snell1224 test.

tom95858:
@Snell1224 can you confirm that this fixes your issue?

Snell1224:
@tom95858 yes, this fixed the original
issue I had. Sorry for the late response.

And thank you @narategithub and @nichamon for making these fixes.

PR #700, opened 2021-04-26
add compid option to lustre_client

baallan:
this adds a compid helper and applies it to lustre_client and adds a test for it and jobid in lustre_client.

### DIFF
--- a/ldms/src/sampler/Makefile.am
+++ b/ldms/src/sampler/Makefile.am
@@ -3,6 +3,7 @@ ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = .
 lib_LTLIBRARIES =
 pkglib_LTLIBRARIES =
+noinst_LTLIBRARIES =
 
 AM_CPPFLAGS = @OVIS_INCLUDE_ABS@
 AM_LDFLAGS = @OVIS_LIB_ABS@
@@ -13,7 +14,11 @@ CORE_LIBADD = $(top_builddir)/ldms/src/core/libldms.la \
 COMMON_LIBADD = libsampler_base.la $(CORE_LIBADD) \
 		$(top_builddir)/lib/src/coll/libcoll.la
 
-noinst_LTLIBRARIES = libjobid_helper.la
+noinst_LTLIBRARIES += libldms_compid_helper.la
+libldms_compid_helper_la_SOURCES = comp_id_helper.c comp_id_helper.h
+libldms_compid_helper_la_LIBADD = $(CORE_LIBADD)
+
+noinst_LTLIBRARIES += libjobid_helper.la
 libjobid_helper_la_SOURCES = jobid_helper.c jobid_helper.h
 libjobid_helper_la_LIBADD = $(CORE_LIBADD) $(top_builddir)/lib/src/coll/libcoll.la
 
--- a/ldms/src/sampler/comp_id_helper.c
+++ b/ldms/src/sampler/comp_id_helper.c
@@ -1,0 +1,57 @@
+#include "comp_id_helper.h"
+
+int comp_id_helper_schema_add(ldms_schema_t schema, comp_id_t cid)
+{
+	int rc;
+	if (!cid || ! schema)
+		return EINVAL;
+	if (cid->defined) {
+		rc = ldms_schema_meta_add(schema, LDMSD_COMPID, LDMS_V_U64);
+		if (rc < 0) {
+			return -rc;
+		}
+		cid->comp_id_index = rc;
+	}
+	return 0;
+}
+
+int comp_id_helper_metric_update(ldms_set_t set, comp_id_t cid)
+{
+	if (!set || !cid)
+		return EINVAL;
+	if (cid->defined)
+		ldms_metric_set_u64(set, cid->comp_id_index, cid->component_id);
+	return 0;
+}
+
+int comp_id_helper_config_optional(struct attr_value_list *avl, comp_id_t cid)
+{
+	if (!cid || !avl)
+		return EINVAL;
+	char *tmp;
+	char *endp = NULL;
+	errno = 0;
+	cid->defined = false;
+	tmp = av_value(avl, LDMSD_COMPID);
+	if (!tmp)
+		return 0;
+	unsigned long long int j = strtoull(tmp, &endp, 0);
+	if (endp == tmp || errno) {
+		return EINVAL;
+	}
+	cid->component_id = j;
+	cid->defined = true;
+	return 0;
+}
+
+int comp_id_helper_config(struct attr_value_list *avl, comp_id_t cid)
+{
+	int rc = comp_id_helper_config_optional(avl, cid);
+	if (rc)
+		return rc;
+	if (!cid->defined) {
+		cid->defined = true;
+		cid->component_id = 0;
+	}
+	return 0;
+}
--- a/ldms/src/sampler/lustre_client/lustre_client_general.c
+++ b/ldms/src/sampler/lustre_client/lustre_client_general.c
@@ -124,7 +124,7 @@ ldms_set_t llite_general_create(const char *producer_name,
 {
         ldms_set_t set;
         int index;
-        char instance_name[256];
+        char instance_name[LDMS_PRODUCER_NAME_MAX+64];
 
         log_fn(LDMSD_LDEBUG, SAMP" llite_general_create()\n");
         snprintf(instance_name, sizeof(instance_name), "%s/%s",
--- a/ldms/src/sampler/lustre_client/lustre_client_general.c
+++ b/ldms/src/sampler/lustre_client/lustre_client_general.c
@@ -65,46 +65,64 @@ int llite_general_schema_is_initialized()
                 return -1;
 }
 
-int llite_general_schema_init(const char *producer_name)
+int llite_general_schema_init(comp_id_t cid)
 {
         ldms_schema_t sch;
         int rc;
         int i;
 
-        log_fn(LDMSD_LDEBUG, SAMP" llite_general_schema_init()\n");
+        log_fn(LDMSD_LDEBUG, SAMP ": llite_general_schema_init()\n");
         sch = ldms_schema_new("lustre_client");
-        if (sch == NULL)
+        if (sch == NULL) {
+		log_fn(LDMSD_LERROR, SAMP ": lustre_llite_general schema new failed (%s).\n",
+			"out of memory");
                 goto err1;
-	rc = jobid_helper_schema_add(sch);
-	if (rc <0)
+	}
+	const char *field;
+	field = "component_id";
+	rc = comp_id_helper_schema_add(sch, cid);
+	if (rc) {
+		rc = -rc;
 		goto err2;
-        rc = ldms_schema_meta_array_add(sch, "fs_name", LDMS_V_CHAR_ARRAY, 64);
-        if (rc < 0)
+	}
+	field = "job data";
+	rc = jobid_helper_schema_add(sch);
+	if (rc <0) {
+		goto err2;
+	}
+	field = "fs_name";
+        rc = ldms_schema_meta_array_add(sch, field, LDMS_V_CHAR_ARRAY, 64);
+        if (rc < 0) {
                 goto err2;
-        rc = ldms_schema_meta_array_add(sch, "llite", LDMS_V_CHAR_ARRAY, 64);
-        if (rc < 0)
+	}
+	field = "llite";
+        rc = ldms_schema_meta_array_add(sch, field, LDMS_V_CHAR_ARRAY, 64);
+        if (rc < 0) {
                 goto err2;
+	}
         /* add llite stats entries */
         for (i = 0; llite_stats_uint64_t_entries[i] != NULL; i++) {
-                rc = ldms_schema_metric_add(sch, llite_stats_uint64_t_entries[i],
-                                            LDMS_V_U64);
-                if (rc < 0)
+		field = llite_stats_uint64_t_entries[i];
+                rc = ldms_schema_metric_add(sch, field, LDMS_V_U64);
+                if (rc < 0) {
                         goto err2;
+		}
         }
 
         llite_general_schema = sch;
 
         return 0;
 err2:
+	log_fn(LDMSD_LERROR, SAMP ": lustre_llite_general schema creation failed to add %s. (%s)\n",
+		field, STRERROR(-rc));
         ldms_schema_delete(sch);
 err1:
-        log_fn(LDMSD_LERROR, SAMP" lustre_llite_general schema creation failed\n");
         return -1;
 }
 
 void llite_general_schema_fini()
 {
-        log_fn(LDMSD_LDEBUG, SAMP" llite_general_schema_fini()\n");
+        log_fn(LDMSD_LDEBUG, SAMP ": llite_general_schema_fini()\n");
         if (llite_general_schema != NULL) {
                 ldms_schema_delete(llite_general_schema);
                 llite_general_schema = NULL;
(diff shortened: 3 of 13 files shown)


baallan:
Post user-group meeting feedback:
* avoid generating ambiguous schema names, which optional comp-id metric does.
   * not clear if the desired solution is to make comp-id unconditional or to add _c to schema name if compid is present.
* not clear if job info should be optional or required, particularly going forward when node multitenancy gets involved.

ldms/src/sampler/Makefile.am
```diff
@@ -13,6 +13,10 @@ CORE_LIBADD = $(top_builddir)/ldms/src/core/libldms.la \
 COMMON_LIBADD = libsampler_base.la $(CORE_LIBADD) \
 		$(top_builddir)/lib/src/coll/libcoll.la
 
+libldms_compid_helper_la_SOURCES = comp_id_helper.c comp_id_helper.h
+libldms_compid_helper_la_LIBADD = $(CORE_LIBADD)
+lib_LTLIBRARIES += libldms_compid_helper.la
```
morrone, line 18: I am leery of adding these as exported interfaces until we do a little more design work. jobid_helper was intended as a stop gap implementation until we figure out a consistent API to offer external developers. I think this should be noinst_LTLIBARIES like libjobid_helper, no lib_LTLIBRARIES.

ldms/src/sampler/lustre_client/lustre_client_general.c
```diff
@@ -98,7 +116,8 @@ int llite_general_schema_init(const char *producer_name)
 err2:
         ldms_schema_delete(sch);
 err1:
-        log_fn(LDMSD_LERROR, SAMP" lustre_llite_general schema creation failed\n");
+        log_fn(LDMSD_LERROR, SAMP" lustre_llite_general schema creation failed with: %s\n",
```
morrone, line 119: If every conditional has its own error message, then I don't think we should print a second one at the end.

morrone (changes requested):

ldms/src/sampler/Makefile.am
baallan: Whether provided as inst or noist, it's still an API. Why take the binary bloat? We're about to add a whole new API anyway, so I'm not seeing the harm.

ldms/src/sampler/lustre_client/lustre_client_general.c
baallan: The second one at the end centralizes the decoding of STRERROR.

morrone: Yes, but lets just do that in the new log messages. I am strongly opposed to splitting an error message across multiple log entries unless absolutely necessary. In this case it is not at all necessary.

baallan: fixed.

ldms/src/sampler/Makefile.am
baallan: In the past, we had another solution to the jobid helper problem.  The problem is either you look up the job data set *every time*, or you have an awful assumption that the jobid set pointer is not going to expire underneath you. The alternate solution was the (inevitable?) global registry of interesting data items. A plugin/instance can in theory come and go, independent of the registry. The registry can contain named values, such as the stuff we're doing with jobid.   The counter-argument can be made that the local list of set instance is is just another registry, if a slower one because you must lookup the set and then find its element index and then read that value.

baallan:
@morrone still need a choice on how to resolve the schema ambiguity: generate alternate schema name(s) or just include compid always.

ldms/src/sampler/Makefile.am
morrone: > Whether provided as inst or noist, it's still an API. Why take the binary bloat? We're about to add a whole new API anyway, so I'm not seeing the harm.

The increase to binary sizes is insignificant. Adding exporting a public interface imposes maintenance burdens on us as a code team (and if we ignore those burdens, then we produce less-than-pleasant to work with software). Internal-only interfaces can be changed as we like without breaking anything. External facing APIs and ABIs cannot be changed without carefully publishing and orchestrating change.

I do not think we should be exporting these tiny helper libraries. A better approach would be to wrap them up into a larger consistent API and single library. So until we want to go the work of designing a good externally facing API, lets just keep it internal.

morrone: > In the past, we had another solution to the jobid helper problem.

I honestly can't tell what this has to do with this PR.

morrone:
> @morrone still need a choice on how to resolve the schema ambiguity: generate alternate schema name(s) or just include compid always.

Having compid always is fine with me.

ldms/src/sampler/Makefile.am
baallan: @morrone you mentioned jobid_helper as a stopgap.  It would have been nice (and maybe in the new api still could be) if we put component_id and all the job stuff into a registry instead of dragging it all through config lines.

baallan:
@morrone I've extended the patch to make the component id 'always on' for lustre_client.
I'm not at all sure that binary size is insignificant if we approach this as noinst; 15k here and 15k there can add up fast.

But if @tom95858 agrees with you that linking as 'noinst' somehow makes the new helper invisible to other developers, I'll change it.

I suppose one could argue that with the addition of component_id and producer as parameters, we might as well derive this sampler from sampler_base. :-)

ldms/src/sampler/comp_id_helper.c
```diff
@@ -0,0 +1,57 @@
+#include "comp_id_helper.h"
+
+int comp_id_helper_schema_add(ldms_schema_t schema, comp_id_t cid)
+{
+	int rc;
+	if (!cid || ! schema)
+		return EINVAL;
+	if (cid->defined) {
+		rc = ldms_schema_metric_add(schema, LDMSD_COMPID, LDMS_V_U64);
```
morrone, line 9: Unlike jobid, the component ID is not expected to change for the life of the metric set. So this should be "lmds_schema_meta_", should it not?

ldms/src/sampler/lustre_client/lustre_client_general.c
```diff
@@ -135,6 +154,7 @@ ldms_set_t llite_general_create(const char *producer_name,
         ldms_metric_array_set_str(set, index, fs_name);
         index = ldms_metric_by_name(set, "llite");
         ldms_metric_array_set_str(set, index, llite_name);
+	comp_id_helper_metric_update(set, cid);
```
morrone, line 157: It looks to me like you are giving every filesystem the same component id. I suspect that people who care about component ids will want them to be different. This may also implies that your configuration handling function need to take an array or something. Keep in mind though that his plugin is fully dynamic, and we don't necessarily know which filesystem will appear in which order. So probably you need some way to specify a mapping between filesystem names and component IDs in the configuration. Technically a filesystem can be mounted twice and then you would have to somehow figure out which mount gets which component ID. Or may just accept that they are going to have the same component ID. But we at least need to handle the unique multiple fileystems use case, because that it a normal operating mode for this plugin.

ldms/src/sampler/Makefile.am
morrone: Still need to change to noinst.

morrone (changes requested):

morrone:
> I suppose one could argue that with the addition of component_id and producer as parameters, we might as well derive this sampler from sampler_base. :-)

That would not work until sampler_base undergoes more reworking. sampler_base has an in-built assumption that there is a one-to-one mapping of schema to metric set (each metric set has its own schema structure). The API and internal structures don't really allow for the workings of many of my plugins where there is one schema that is shared amongst many metric sets. Its notion of how instance handling works is still too rigid to, for instance, work with this plugin. And sampler_base's handling of component IDs won't map to this plugin's model either.

I would have used sampler_base in my plugins if I could, but it just doesn't work for them in its current form. And when I first started I probably didn't have enough knowledge of the space to overhaul sampler_base and make for a broader swath of sampler plugin use cases. These days I have probably built up enough of the sampler functionality to make some informed suggestions for the next design that will follow on after sampler_base. I am hoping that the next plugin API will incorporate that sort of thing.

ldms/src/sampler/lustre_client/lustre_client_general.c
baallan: The network and file systems components (particularly in the case of this sampler) carry their own device identifier metrics. Component id as practiced is an enumerable (range() is well-defined) alias of hostnames (which being strings don't have a range()) for easy analysis. This allows collectives computation over related hosts. 

As with jobids, there is probably some ideal world in which compids could have been defined as arbitrarily long dotted string notation or both simply belong in a separate data set with a join performed at the database. Years back we chose to go the integer route for the analysis optimization reason.

morrone: > The network and file systems components (particularly in the case of this sampler) carry their own device identifier metrics. Component id as practiced is an enumerable (range() is well-defined) alias of hostnames (which being strings don't have a range()) for easy analysis. This allows collectives computation over related hosts.
> 
> As with jobids, there is probably some ideal world in which compids could have been defined as arbitrarily long dotted string notation or both simply belong in a separate data set with a join performed at the database. Years back we chose to go the integer route for the analysis optimization reason.

That sounds awfully arbitrary and inconsistent to me. What metric sets that are expected to have multiple instances on a node do _not_ contain their own device identifier metrics? gpus contain identifier metrics. Does that mean that every gpu inside of a node gets the same component ID? CPUs are numbered. Does that mean they use the same component ID when they are in the same node?

baallan: It's awfully arbitrary.  :-) It isn't inconsistent. The 'problem' is just that instead of being named component_id in the wayback (which it was when we didn't quite know what to do with it) the evolution of practice has been to something that might be reasonably called 'host_id'.  Everything inside the node gets the same component_id, just like everything inside the cluster gets the same cluster name in a slurm database.  The fact that the 'component_id' maps 1:1 with hostname is an artifact of producer being added to our schemas years after component_id practice was established.  

There was a prototype used of component_ids with hierarchical structures embedded in the single u64, where some bits were the host, some bits were a core number or gpu number or network device number or dimm number or what have you, but it was all pretty complicated and you had to play mask games and noone but the inventor understood it. Makes a lot more sense to just have device name metrics in device-type specific schema.

At this point, the weight of downstream tools assuming compid maps to hostname will, I expect, keep the usage the way it is.

baallan: If you want to really pick at a legacy item (that doubles our data size), take a look at udata, which to my knowledge is no longer used for anything except making store code complicated. 
I would really like to remove *that*.

ldms/src/sampler/lustre_client/lustre_client_general.c
```diff
@@ -65,46 +65,64 @@ int llite_general_schema_is_initialized()
...
         ldms_schema_t sch;
         int rc;
         int i;
 
-        log_fn(LDMSD_LDEBUG, SAMP" llite_general_schema_init()\n");
+        log_fn(LDMSD_LDEBUG, SAMP ": llite_general_schema_init()\n");
         sch = ldms_schema_new("lustre_client");
-        if (sch == NULL)
+        if (sch == NULL) {
+		log_fn(LDMSD_LERROR, SAMP ": lustre_llite_general schema new failed (%s).\n",
```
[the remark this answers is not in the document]
baallan: we have a bias against >80char lines.

ldms/src/sampler/lustre_client/lustre_client_general.c
morrone: Ah, that didn't sound like how they described it in the last call. But if that is the case, then so be it. But I think it would be good to add a comment to the man page under the component id option that explicitly states "all metric sets in this producer will use this same component id" or something to that effect.

baallan: Since the user has no way of specifying alternate values per set instance, isn't that a given? But perhaps a bit of explanation of component_id added to the sampler_base page wouldn't be out of order.

tom95858: I am struggling with why adding a single uint64_t metric to a set requires touching 11 files, adding two new files and introducing a handful of new API.

If we want to hold court on how component_id should be managed, then let's do that in an issue or on the wiki, but not shoe horn API design issues into a change that would otherwise require about 3 lines of code.

morrone: No, I don't think it necessarily is. The plugin might try to disambiguate them in some way, like it does for instances. I've had repeated lack of clarity over comp ID through the past year, so explicit comments like I requested would be very helpful.

baallan:
@morrone man page update and noinst change added.

ldms/src/sampler/lustre_client/lustre_client_general.c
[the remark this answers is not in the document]
morrone: Yes, but there is consistency and there is foolish consistency. Sometimes a longer line is reasonable. But if you really want to avoid it, C concatenates string literals, so using that to split the message across lines would be preferable to using programmatic substitution.

ldms/src/sampler/lustre_client/lustre_client_general.c
baallan: added compid explanation to man page.

morrone (changes requested):

morrone:
Burried in one of the code comment @tom95858 questioned why there are so many files touched and the new comp id API. Here's my take on the situation:

Tom, you know how you want plugins to be written with a near magical amount of minimal coding the new plugin API? This is the sort of thing that we will need to offer to plugin authors through an API to make that happen. Otherwise this is one of dozens of tedious tasks that all the plugins authors need to replicate when they write their plugin.

Now, I am not claiming that this is the long term approach we want to take for that API. I think multiple of these related items need to be combined into a single well-formed library (with good consistent function name prefixes) rather than dozens of tiny libraries. And possibly one set of functions could handle many different options, allowing the plugin author to select a subset of configuration options through a bit mask or something. But as long as we understand that and keep these tiny libraries internal to the tree (noinst, not published), then I am less concerned about having some tiny helper libraries for now. These demonstrate that there is an need that is unmet in the current sampler_base implementation. We can address that need in a temporary fashion now, while planning for something better in association with the new plugin API.

I suspect that folks will probably look at other of my plugins and decide that those need comp ID added as well. This helper library would make that easier and consistent across multiple plugins.

I would perhaps suggest that this PR could have its commits arranged a little better. I would probably do a series of logically separated commits ordered like so:

1) A commit that just introduces the comp ID helper library
2) A commit that adds comp ID to the lustre_client plugin
3) A commit that contains the unrelated cleanup (e.g. introducing colons to log messages)
4) A commit that introduces tests

Ultimately I wasn't planning to enforce that organization. But would that perhaps help make things more digestible for you, @tom95858?

morrone:
@baallan, do we really need those three empty files in the examples directory? That might be a way to shave a few files off the PR too.

baallan:
closing and reopening from another branch, as rebasing got messed up eventually.

morrone:
When you get that branch working locally, you could just force push it to "baallan:compid_helper" on github. There isn't necessarily a need to start a new PR.

morrone:
I can help with git-fu if you have any questions.

baallan:
I took it as an opportunity to add files in different commits, which (splitting an old commit along file lines) is something i've never found a happy write-up about. 
I believe the new version has all the things you asked about in #700 (except the 3 empty files which are a requirement of the testing framework in use-- nother nit to be fixed) addressed.

morrone:
@baallan

Here is one set of instructions on how to split up a commit in your history (look in the "Splitting a Commit" section):

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Here is another explanation of the process:

https://embeddedartistry.com/blog/2018/02/19/code-cleanup-splitting-up-git-commits-in-the-middle-of-a-branch/

If those don't do it for you, I can probably help fill in more detail.

PR #701, opened 2021-04-26
fixes missing support for instance in lustre_client

baallan:
The producer parameter is now used (instead of ignored) as the
prefix of the instance names generated by appending /$llite_name.
This prevents breakage (redundant instance names) during scale testing
with multiple samplers per node. 

If producer is not specified, default of $hostname remains.
Man page updated accordingly.

### DIFF
--- a/ldms/src/sampler/lustre_client/Plugin_lustre_client.man
+++ b/ldms/src/sampler/lustre_client/Plugin_lustre_client.man
@@ -14,9 +14,10 @@ or a configuration file. The lustre_client plugin provide a metric set for each 
 mounts found on a node.  The schema is named "lustre_client".  The data for the metric sets is
 generally found in /proc/fs/lustre/llite/*/stats.
 
-This plugin currently employs zero configuration.  The producer name is set to the hostname, and the metric set instance names are
-derived from the llite instance name.  Any user-supplied configuration values will be ignored.  Future versions may add
-configuration options.
+This plugin currently employs zero configuration.  The producer name is set to the hostname
+by default, and the metric set instance names are
+derived from the llite instance name. Any user-supplied configuration values not
+documented here will be ignored.
 
 This plugin should work with at least Lustre versions 2.8, 2.10, and 2.12.
 
@@ -33,13 +34,18 @@ name=<plugin_name>
 .br
 This MUST be lustre_client.
 .TP
-job_set=<metric set name>
+job_set=<job metric set name>
 .br
 The name of the metric set that contains the job id information (default=job_id)
+.TP
+producer=<alternate host name>
+.br
+The default used for producer (if not provided) is the result of gethostname().
+The set instance names will be $producer/$llite_name.
 .RE
 
-.SH BUGS
-No known bugs.
+.SH NOTES
+Improperly spelled option names are not trapped as configuration errors.
 
 .SH EXAMPLES
 .PP
@@ -51,4 +57,4 @@ start name=lustre_client interval=1000000
 .fi
 
 .SH SEE ALSO
-ldmsd(8), ldms_quickstart(7), ldmsd_controller(8), ldms_sampler_base(7)
+ldmsd(8), ldms_quickstart(7), ldmsd_controller(8), ldms_sampler_base(7), gethostname(2)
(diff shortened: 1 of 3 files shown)


ldms/src/sampler/lustre_client/Plugin_lustre_client.man
```diff
@@ -33,9 +33,14 @@ name=<plugin_name>
...
 .TP
-job_set=<metric set name>
+job_set=<job metric set name>
 .br
 The name of the metric set that contains the job id information (default=job_id)
+.TP
+instance=<metric set prefix>
+.br
+The prefix used to construct the set instance names. If not given, this defaults to
+the result of gethostname(). The set instance names will be $instance/$llite_name.
```
morrone, line 43: The default is hostname/llite_name, not just the hostname.

baallan: You seem to be confusing (default) value of the prefix with the constructed set name. If you think it would be more clear, this could be reworded as:
```
The set prefix used to construct the set names. If not given, this defaults to the result of gethostname(). The set names will be $prefix/$llite_name.
```

ldms/src/sampler/lustre_client/lustre_client.c
```diff
@@ -227,6 +227,15 @@ static int config(struct ldmsd_plugin *self,
                   struct attr_value_list *kwl, struct attr_value_list *avl)
 {
         log_fn(LDMSD_LDEBUG, SAMP" config() called\n");
+	char *ival = av_value(avl, "instance");
+	if (ival) {
+		if (strlen(ival) < sizeof(producer_name)) {
+			strncpy(producer_name, ival, sizeof(producer_name));
```
morrone, line 233: Why are you putting the "instance" name in the "producer name" variable?  This would also change thre producer when the instance is set. I think that would be unexpected. I think you need to add an instance_name variable, add that to the parameters passed to llite_general_create(), and merge things there.

tom95858:
I think the convention has been that the producer name was the prefix for the instance name. This is also required when running multiple samplers on the same host since gethostname() would create a collision.

ldms/src/sampler/lustre_client/lustre_client.c
baallan: The "custom" of ldms deployment is that for a given schema, the set instance name is scripted as $producer/$schema (or at least $producer/$plugin).  
Where there are multiple sets per plugin (as this case, or ibnet, or ibmad), prescribing the entire instance name in the configuration script is nonsensical, so we use the 'instance' as a prefix instead of the entire name: $instance/$device_name.
The convention for $producer is $hostname, but for scale testing where many sampler daemons run on the same node, the sets must be distinguishable so "producer_name[]" must be some extension/replacement on hostname (e.g. localhost[1-n] or $hostname/$k). 

This all fits well with the C handling of the "instance" parameter as a prefix instead of inventing a new configuration item (instance_prefix=) and pushing around extra arguments.
The same prefix applies to all the set instances, unless you want to call config twice.

morrone: I don't understand what that explanation had to do with my question. I am fine with you having an "instance" configurable, but the "instance" configurable should not be changing the "producer" field. The "producer" configurable should change the producer field.

And something seems a little fishy here to me about the currently implementation breaking a test environment. The code is currently setting the instance name to "producer/llite_name", so the only way those would conflict in a testing environment is if the testing environment is failing to correctly configure different producer names for different producers on the same node.

baallan: ah. so you want 's/instance/producer/g' in this change afa the config input. I'll do that.

morrone: I just don't want the instance configurable to change the producer name. :)

But, yes, it sounds like making the producer name configurable is probably what you really needed rather than making the instance name directly configurable. One could potential make both configurable, but it sounds like producer might be sufficient for your needs.

baallan: revised to producer option.

ldms/src/sampler/lustre_client/Plugin_lustre_client.man
```diff
@@ -33,13 +34,18 @@ name=<plugin_name>
...
+producer=<alternate host name>
+.br
+The default used for producer (if not provided) is the result of gethostname().
+The set instance names will be $producer/$llite_name.
 .RE
 
-.SH BUGS
-No known bugs.
+.SH NOTES
+Improperly spelled option names are not trapped as configuration errors.
```
morrone, line 48: I.E., just like all of the other ldms plugins. :) It would be great to improve the plugin API so every plugin author didn't need to repeat those checks over and over again.

morrone (approved):
Looks good now.

PR #702, opened 2021-04-26
Ldmsd request validation

nick-enoent:
Validate length of request headers
Validate attr doesn't reference beyond end of request
Check attr string is null terminated

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -209,6 +209,7 @@ static int xprt_stats_handler(ldmsd_req_ctxt_t req_ctxt);
 static int prdcr_stats_handler(ldmsd_req_ctxt_t req_ctxt);
 static int thread_stats_handler(ldmsd_req_ctxt_t req_ctxt);
 static int set_stats_handler(ldmsd_req_ctxt_t req_ctxt);
+static int bad_req_handler(ldmsd_req_ctxt_t req_ctxt);
 static int unimplemented_handler(ldmsd_req_ctxt_t req_ctxt);
 static int eperm_handler(ldmsd_req_ctxt_t req_ctxt);
 static int ebusy_handler(ldmsd_req_ctxt_t reqc);
@@ -766,6 +767,42 @@ err:
 	return rc;
 }
 
+int validate_ldmsd_req(ldmsd_req_ctxt_t reqc, ldmsd_req_hdr_t rh)
+{
+	ldmsd_req_attr_t attr;
+	unsigned int off;
+	/* Request consists of only the header and terminating discriminator */
+	if (rh->rec_len == sizeof(*rh) + sizeof(attr->discrim)) {
+		attr = (ldmsd_req_attr_t)(rh + 1);
+		if (attr->discrim != 0)
+			return 0;
+		else
+			return 1;
+	}
+	/* Request contains, at least, one attribute*/
+	else if (rh->rec_len < sizeof(*rh) + sizeof(*attr))
+		return 0;
+
+	off = sizeof(*rh);
+	attr = (ldmsd_req_attr_t)(rh+1);
+
+	while (attr->discrim) {
+		off += sizeof(*attr) + attr->attr_len;
+		/* Check attr doesn't reference beyond end of request buffer */
+		if (off + sizeof(attr->discrim) > rh->rec_len)
+			return 0;
+		/* Check null termination of each attr_value */
+		else if (attr->attr_value[attr->attr_len-1] != '\0')
+			return 0;
+		attr = (ldmsd_req_attr_t)&attr->attr_value[attr->attr_len];
+		if (attr->discrim) {
+			if (off + sizeof(*attr) + sizeof(attr->discrim) > rh->rec_len)
+				return 0;
+		}
+	}
+	return 1;
+}
+
 int ldmsd_handle_request(ldmsd_req_ctxt_t reqc)
 {
 	struct request_handler_entry *ent;
@@ -774,6 +811,7 @@ int ldmsd_handle_request(ldmsd_req_ctxt_t reqc)
 	uid_t luid;
 	gid_t lgid;
 	mode_t mask;
+	int rc;
 
 	__dlog("handling req %s\n", ldmsd_req_id2str(reqc->req_id));
 
@@ -784,6 +822,9 @@ int ldmsd_handle_request(ldmsd_req_ctxt_t reqc)
 
 	ent = &request_handler[request->req_id];
 
+	rc = validate_ldmsd_req(reqc, request);
+	if (!rc)
+		return bad_req_handler(reqc);
 	/* Check for unimplemented request */
 	if (!ent->handler)
 		return unimplemented_handler(reqc);
@@ -5031,6 +5072,16 @@ static int greeting_handler(ldmsd_req_ctxt_t reqc)
 out:
 	free(rep_len_str);
 	free(num_rec_str);
+	return 0;
+}
+
+static int bad_req_handler(ldmsd_req_ctxt_t reqc)
+{
+	size_t cnt;
+	reqc->errcode = EBADRQC;
+	cnt = Snprintf(&reqc->line_buf, &reqc->line_len,
+			"LDMSD received bad request.");
+	ldmsd_send_req_response(reqc, reqc->line_buf);
 	return 0;
 }
 
--- a/ldms/src/ldmsd/ldmsd_request.h
+++ b/ldms/src/ldmsd/ldmsd_request.h
@@ -508,11 +508,10 @@ int ldmsd_append_reply(struct ldmsd_req_ctxt *reqc, const char *data, size_t dat
 void ldmsd_send_error_reply(ldmsd_cfg_xprt_t xprt, uint32_t msg_no,
 			    uint32_t error, char *data, size_t data_len);
 void ldmsd_send_req_response(ldmsd_req_ctxt_t reqc, const char *msg);
+int validate_ldmsd_req(ldmsd_req_ctxt_t reqc, ldmsd_req_hdr_t rh);
 int ldmsd_handle_request(ldmsd_req_ctxt_t reqc);
 static inline ldmsd_req_attr_t ldmsd_first_attr(ldmsd_req_hdr_t rh)
 {
-	if (rh->rec_len <= sizeof(*rh))
-		return NULL;
 	return (ldmsd_req_attr_t)(rh + 1);
 }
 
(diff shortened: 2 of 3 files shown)


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -774,16 +797,20 @@ int ldmsd_handle_request(ldmsd_req_ctxt_t reqc)
...
 	gid_t lgid;
 	mode_t mask;
+	int rc;
 
 	__dlog("handling req %s\n", ldmsd_req_id2str(reqc->req_id));
 
 	/* Check for request id outside of range */
 	if ((int)request->req_id < 0 ||
-	    request->req_id >= (sizeof(request_handler)/sizeof(request_handler[0])))
+		request->req_id >= (sizeof(request_handler)/sizeof(request_handler[0])))
```
morrone, line 806: I'm not a fan of this indenting change.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5034,6 +5061,16 @@ static int greeting_handler(ldmsd_req_ctxt_t reqc)
 	return 0;
 }
 
+static int bad_req_handler(ldmsd_req_ctxt_t reqc)
+{
+	size_t cnt;
+	reqc->errcode = EBADRQC;
+	cnt = Snprintf(&reqc->line_buf, &reqc->line_len,
+			"LDMSD received bad request.");
```
morrone, line 5069: Maybe "malformed" instead of bad? It maybe gives more of a hint about what was wrong with the request.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,28 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
+{
+	ldmsd_req_attr_t attr;
+	unsigned int off;
+	/* Validate length of request header and request. */
+	if (rh->rec_len <= sizeof(*rh))
+		return 0;
+	attr = (ldmsd_req_attr_t)(rh+1);
+	while (attr->discrim)
+	{
+		off = sizeof(*rh);
```
tom95858, line 780: This needs to be moved outside the loop and then updated inside the loop, for example:
```c
off = sizeof(*rh)
while (attr->discrim) {
    off += sizeof(*attr) + attr->attr_len
    if (off > rh->rec_len)
        return 0;
    attr = (ldmsd_req_attr_t)&attr->attr_value[attr->attr_len];
}
```
As it is, it's only checking that the 1st attribute + the length of the longest attribute is not > rh->rec_len.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,28 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
+		/* Check attr doesn't reference beyond end of request buffer */
+		if (off + attr->attr_len > rh->rec_len)
+			return 0;
+		attr = (ldmsd_req_attr_t)&attr->attr_value[attr->attr_len];
+	}
+	/* Check null termination of last attr string */
```
tom95858, line 786: This check is unnecessary because the while(attr->discrim) already checks this. Also '\0' is the character 0, not an unsigned long.

ldms/src/ldmsd/ldmsd_request.c
tom95858: I think this an accidental change, please remove.

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_request.c
morrone, line 780: I think the math is broken here. I imagine that you meant to set off to "sizeof(*rh)" before the loop, and then increment it by the of the attribute inside of the loop just before updating attr to point to the next attr.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,28 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
+	if (attr->attr_value[attr->attr_len] != '\0')
```
morrone, line 787: We should really check all strings, not just the last one.

ldms/src/ldmsd/ldmsd_request.c
nick-enoent: Ahh, ok that makes sense. I'll make the relevant change

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,28 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
 	return rc;
 }
 
+int validate_ldmsd_req(ldmsd_req_ctxt_t reqc, ldmsd_req_hdr_t rh)
+{
+	ldmsd_req_attr_t attr;
+	unsigned int off;
+	/* Validate length of request header and request. */
+	if (rh->rec_len <= sizeof(*rh))
```
morrone, line 775: Why is it an error for them to be equal? Perhaps the header is all that is needed.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,28 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
+		return 0;
+	attr = (ldmsd_req_attr_t)(rh+1);
+	while (attr->discrim)
```
morrone, line 778: We have to verify that the request buffer is large enough to hold this attr struct _before_ we access anything in the attr struct (i.e before accessing attr->discrim).

morrone (changes requested):

ldms/src/ldmsd/ldmsd_request.c
tom95858: If we are certain that all attribute values are strings, then the check would be:
```c
if (attr->attr_value[attr->attr_len-1] != '\0')
    return 0;
```
attr->attr_value[attr->attr_len] points to the 1st byte of the next attribute, i.e. the 1st byte the next attr->discrim

ldms/src/ldmsd/ldmsd_request.c
nick-enoent: That's a good point, I'll update the commit to reflect this.

ldms/src/ldmsd/ldmsd_request.c
nick-enoent: I think this is the only change left to address. I'm inclined to keep "bad", as the error returned is literally EBADRQC, and malformed could limit the function's perceived use in the future. However, if the general thought is that this error reporting function will only ever be used for this particular use case, then I don't have any qualms changing it to malformed.

ldms/src/ldmsd/ldmsd_request.c
morrone: The code still need this. Must check that "off+sizeof(*attr) >= rh->rec_len" _before_ accessing things in the attr.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,29 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
+	{
+		off += sizeof(*attr) + attr->attr_len;
+		/* Check attr doesn't reference beyond end of request buffer */
+		if (off > rh->rec_len)
+			return 0;
+		/* Check null termination of each attr_value */
+		else if (attr->attr_value[attr->attr_len-1] != '\0')
+			return 0;
+		attr = (ldmsd_req_attr_t)&attr->attr_value[attr->attr_len];
```
morrone, line 788: And here again we need to make sure that this new "attr" fits in the rh->rec_len before we go accessing attr->discrim in the while() statement. It may make sense to rework this as a do-while or for loop so you don't need to put that check in two different places.

morrone (changes requested):

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -766,6 +767,29 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
 	return rc;
 }
 
+int validate_ldmsd_req(ldmsd_req_ctxt_t reqc, ldmsd_req_hdr_t rh)
+{
+	ldmsd_req_attr_t attr;
+	unsigned int off;
+	/* Validate length of request header and request. */
+	if (rh->rec_len < sizeof(*rh))
+		return 0;
```
tom95858, line 776:
```
if (rh->rec_len == sizeof(*rh))
    return 1;
```
Then add code to ldmsd_attr_first() to return NULL in this case.

tom95858: Also,
```
else if (rh->rec_len < sizeof($rh) + sizeof(attr->discrim))
    return 0;
```
At that point you know that the message includes attr->discrim and it can be safely checked.

morrone: Wait, why wouldn't we just check sizeof(*attr)? Are we really allowing partial attr structs?? If we don't check the full attr size here, then we have to do anther size check later before we access attr->attr_len. That seems like madness. I hope we don't allow that in the protocol.

tom95858: Yes, the last 'attr' is just the discriminator which is the 'end of attribute list marker' No it's not madness. I believe you use NFS and RPC works exactly this way.

morrone: Could you point me to where it works like that?

In any event, a quirk like that in the protocol needs to be clearly documented, and LDMS is certainly missing that.

But my original point stands that we have to check that the attr->attr_len fits in the overall message size before we access attr->attr_len. Knowing that attr->discrim fits is not sufficient to know that the rest of the attr can be safely checked.

ldms/src/ldmsd/ldmsd_request.c
tom95858: You're right, I think after line 788
```
if (attr->discrim && off + sizeof(*attr) > rh->rec_len)
    return 0;
```
At that point attr_len can be safely checked next time around the loop. And yes, perhaps refactoring the logic as a do-while could make things simpler.

ldms/src/ldmsd/ldmsd_request.c
tom95858: > Could you point me to where it works like that?
> 

RFC 8687 is the latest spec. linux/net/sunrpc/xprtrdma/rpc_rdma.h is the header file.

> In any event, a quirk like that in the protocol needs to be clearly documented, and LDMS is certainly missing that.
> 

There's a comment in the ldmsd_request.h file that documents this, perhaps not well enough.

> But my original point stands that we have to check that the attr->attr_len fits in the overall message size before we access attr->attr_len. Knowing that attr->discrim fits is not sufficient to know that the rest of the attr can be safely checked.

... and it's a good point.

nick-enoent:
@morrone No problem

PR #703, opened 2021-04-27
Report authentication not found error

tom95858:
Previously, a missing authentication configuration
object was reported as ENOMEM.


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -544,6 +544,7 @@ char * ldmsd_stream_client_dump()
 					first_client?"":",",
 					sym?sym[0]:_pbuf,
 					c->c_ctxt);
+			free(sym);
```
morrone, line 547: This looks like a good fix, but maybe snuck into the wrong commit?

tom95858: ok, i pulled it out

PR #707, opened 2021-04-28
Use unique connection identifier in message keys

tom95858:
A [pointer:msg_no] key was being used to match
requests with responses. However, if an ldms transport
were disconnected, and then reconnected, the pointer
could be reused in a subsequent connection resulting
in matching a response with the wrong request. This
change generates a unique uint64_t identifier
that is used as part of the key instead of a pointer
to the transport instance.


narategithub:
@tom95858 I'll test this against ldms-test and will get back to you.

narategithub:
@tom95858 This passed all tests in ldms-test repo.

nichamon (approved):

PR #710, opened 2021-05-05
Add an attribute to stream publish requests to not receiving the ack

nichamon:
Stream publishers could add an optional LDMSD_ATTR_TYPE attribute
to tell LDMSD's stream_publish_handler not to send an acknowledgement
back.

The stream republish path uses the attribute so that it would free the
command request right after it sends the request. This is to reduce the
memory footprint in the stream republish/forward path.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -5933,7 +5933,13 @@ static int stream_publish_handler(ldmsd_req_ctxt_t reqc)
 	}
 
 	reqc->errcode = 0;
-	ldmsd_send_req_response(reqc, "ACK");
+	/*
+	 * If a LDMSD_ATTR_TYPE attribute exists, the publisher does not want
+	 * an acknowledge response.
+	 */
+	attr = ldmsd_req_attr_get_by_id(reqc->req_buf, LDMSD_ATTR_TYPE);
+	if (!attr)
+		ldmsd_send_req_response(reqc, "ACK");
 
 	if (!ldmsd_stream_subscriber_count(stream_name))
 		/* There are no subscribers, ignore the data */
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 This is an approach to reduce the memory footprint in the stream republish path and the number of request contexts in the tree. It is an alternative of #711 .

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5983,6 +5985,9 @@ static int stream_republish_cb(ldmsd_stream_client_t c, void *ctxt,
 	ldmsd_req_cmd_t rcmd = ldmsd_req_cmd_new(ldms, LDMSD_STREAM_PUBLISH_REQ,
```
tom95858, line 5985: Should this check if rcmd == NULL?

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5933,7 +5933,9 @@ static int stream_publish_handler(ldmsd_req_ctxt_t reqc)
 	}
 
 	reqc->errcode = 0;
-	ldmsd_send_req_response(reqc, "ACK");
+	attr = ldmsd_req_attr_get_by_id(reqc->req_buf, LDMSD_ATTR_TYPE);
```
tom95858, line 5936: I'm guessing that the presence of this attribute means that this is the republish path.

tom95858 (changes requested):
I don't see how #711 and #710 are alternatives to one another. Can't they both be done?

ldms/src/ldmsd/ldmsd_request.c
nichamon: Yes, only the republish path adds an LDMSD_ATTR_TYPE attribute in the stream_publish request.

ldms/src/ldmsd/ldmsd_request.c
nichamon: I just added the check.

nichamon:
> I don't see how #711 and #710 are alternatives to one another. Can't they both be done?

nichamon:
> I don't see how #711 and #710 are alternatives to one another. Can't they both be done?

Yes, we can merge both of them.

tom95858:
@nichamon Can you please combine these two, rebase and post a pull request? Thanks.

nichamon:
#712 contains the patch. I'm closing this.

PR #711, opened 2021-05-05
Make LDMSD free the send buffer after sending a request

nichamon:
LDMSD dose not use request_context's send buffer
(ldmsd_req_ctxt->rep_buf) after it sends the content in the buffer. The
patch applies only on ldmsd_req_cmd's request context.

We made the patch in order to reduce the memory footprint in the stream
republish path. However, the patch reduces LDMSD's memory usage when
sending any request to another LDMSD.


nichamon:
@tom95858 This is an alternative of #710 .

nichamon:
#712 includes the patch. I'm closing this.

PR #712, opened 2021-05-06
Stream republish optimization

nichamon:


nichamon:
@tom95858 Here is the pull request that contains both patches in #710  and #711 .

tom95858 (changes requested):
This commit message is misleading. The req_buf is not used, but the comment says rep_buf (reply buffer). The context is not sent, the buffer is. Could you please change the commit message to the following:

Free the request buffer after send

This reduces the memory footprint of the requests held in the msg_tree awaiting
replies.

PR #713, opened 2021-05-06
Fix a key collision problem between sent requests and received requests

nichamon:
Without the patch, an LDMSD’s request context key consists of a message
number and a connection ID. The host generates the message number of
to-be-sent requests. In contrast, the peer generates the message number
of received requests. LDMSD has a single tree that contains the contexts
of both sent requests and received requests. The connection IDs in the
request sent to a peer and in the request received from the same peer
will be the same.
Moreover, the message numbers of the two requests could be the same
because the host and the peer generate them. When this happens, we would
see error messages “Duplicate message number …” or “Cannot find the
original request …”.

The patch adds a flag in request context keys to differentiate between
to-be-sent requests and received requests. It prevents the key collision
problem above.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -92,16 +92,6 @@ pthread_mutex_t sp_list_lock = PTHREAD_MUTEX_INITIALIZER;
 #define LDMSD_PLUGIN_LIBPATH_MAX	1024
 struct plugin_list plugin_list;
 
-void ldmsd_cfg_unix_cleanup(ldmsd_cfg_xprt_t xprt)
-{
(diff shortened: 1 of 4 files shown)


nichamon:
I found the bug when I tested OVIS-4 with failover. Sometimes failover works and sometimes it doesn't. I saw the error messages "Cannot find the original request context" or "Duplicate message number .." in the log file of the LDMSD pair. We see this now because we use a generated connection ID instead of memory addresses. Before the patch that use unique connection ID in message keys, the connection ID of requests allocated by `alloc_req_cmd_ctxt` is the memory address of `ldms_t`, but the connection Id of requests received in `ldmsd_process_config_request` is the memory address of the `ldmsd_cfg_xprt_t`.

tom95858:
This problem exists regardless of whether the  connection portion of the key is an ID or a pointer, correct?

PR #715, opened 2021-05-07
Add failover for prdcr_subscribe

nichamon:
Without the patch, the failover pair does not share the prdcr_subscribe
configuration. As a result, the failover buddy will not send a
stream_subscribe request to the producers that it has taken over from
its failed buddy. The patch addresses this, and the failover buddy sends
a stream_subscribe request to the producers it has taken over. If there
is another aggregation level, the patch assumes that the failover buddy
has received a stream_subscibe request from its aggregator before taking
over the failed daemon. In other words, the failover buddy does not
automatically republish the stream to the failed daemon’s aggregator.


nichamon:
I found this problem while I was testing #712 and #713. The symptom I saw is as follows. Agg11 and agg22 were a failover pair. Agg11 received a `prdcr_subscribe` request, and it has sent `stream_subscribe` requests to its producers. When agg11 died, and agg12 took over, agg12 did not receive any streams from agg11's producers.

PR #717, opened 2021-05-12
Fix wrong config transport type

nichamon:
LDMSD does not set the type of LDMS-endpoint configuration transports,
so the type is the configuration file. As a result, the connection ID is
the configuration transport memory address instead of the LDMS transport
ID.


nichamon:
I introduced this bug in #713.

PR #718, opened 2021-05-12
Make ldmsd's stream data NULL-terminated

nichamon:


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -281,7 +281,13 @@ int ldmsd_stream_publish(ldms_t xprt,
 
 	rc = stream_send(xprt, buf, msg_no, 0, (char *)data, data_len);
```
narategithub, line 282: I think in the case of `data[data_len-1] != '\0'`, the `data_len` here should be `data_len+1`.

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -281,7 +281,13 @@ int ldmsd_stream_publish(ldms_t xprt,
...
 	if (rc)
-		goto err;
+			goto err;
+
+	if (data[data_len - 1] != '\0') {
+		rc = stream_send(xprt, buf, msg_no, 0, "\0", 1);
+		if (rc)
+			goto err;
+	}
```
narategithub, line 290: I cannot add a comment at line 278 so I have to add it here :(

At line 278, the `data_len` should be `data_len + 1` if `data[data_len-1] == '\0'`.

tom95858 (commented):
All the call sites except ldmsd_stream_publish_file use strlen(content)+1 for the data len which includes the termination '\0'. I have a separate change for this that I will submit in a
bit.

narategithub:
@tom95858 please don't forget spank_notifier. It calls `ldmsd_steam_publish()`, and it did not include '\0' in the data.

ldms/src/ldmsd/ldmsd_stream.c
tom95858: The point is that if the caller terminated the data as required this doesn't need to be called. The caller is providing data_len, I don't think we can assume that data[data_len] is valid.

nichamon: It sounds like you want to force the caller to null-terminate the data. If my understanding is correct, I will change the function to return an error if `data[data_len - 1] != '\0\'`.

I'm aware that `data[data_len]` may not be valid, but I think I missed your point here. We check `data[ data_len - 1 ]` not `data[data_len]`.

By the way, all current callers terminate the `data`, so this is actually asymptotic. However, I think it is a good idea to check.

PR #719, opened 2021-05-13
Validate requests before converting the network-endian to the host endian

nichamon:


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -774,36 +773,41 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
 			return 0;
 		/* Check null termination of each attr_value */
-		else if (attr->attr_value[attr->attr_len-1] != '\0')
+		else if (attr->attr_value[attr_len-1] != '\0')
 			return 0;
-		attr = (ldmsd_req_attr_t)&attr->attr_value[attr->attr_len];
-		if (attr->discrim) {
-			if (off + sizeof(*attr) + sizeof(attr->discrim) > rh->rec_len)
+		attr = (ldmsd_req_attr_t)&attr->attr_value[attr_len];
+		if (ntohl(attr->discrim)) {
```
tom95858, line 809: Byte swapping zero is a no-op

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -774,36 +773,41 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
...
+		attr_len = ntohl(attr->attr_len);
+
+		off += sizeof(*attr) + attr_len;
 		/* Check attr doesn't reference beyond end of request buffer */
-		if (off + sizeof(attr->discrim) > rh->rec_len)
+		if (off + sizeof(attr->discrim) > rec_len)
 			return 0;
 		/* Check null termination of each attr_value */
-		else if (attr->attr_value[attr->attr_len-1] != '\0')
+		else if (attr->attr_value[attr_len-1] != '\0')
```
tom95858, line 806: This will break if attr_len is zero. it should be something like:
```
else if (attr_len && attr->attr_value[attr_len-1] != '\0')
```

tom95858 (changes requested):
Please fix that one bug with attr_len and this is good to go I think. Did you run this through the configuration regression testing? If so, we're missing tests for attributes like 'reset' which have zero length.

ldms/src/ldmsd/ldmsd_request.c
nichamon: `attr->discrim` can be a non-zero value.

ldms/src/ldmsd/ldmsd_request.c
nichamon: Thanks for catching this. I just pushed the change.

nichamon:
I tested the change with stream and failover, and both work. When I tested stream, I included #718 .

ldms/src/ldmsd/ldmsd_request.c
tom95858: You're only checking if it's zero or not. Zero is zero regardless of the byte order. If you were checking if attr->discrim equaled some value other than zero or -1 (0xff....) then you would have to byte swap it.

ldms/src/ldmsd/ldmsd_failover.c
```diff
@@ -2355,6 +2356,7 @@ int failover_ping_handler(ldmsd_req_ctxt_t req)
 	data.state = f->state;
 	data.conn_state = f->conn_state;
 	data.flags = f->flags;
+	data.zero = 0;
```
tom95858, line 2359: Seems like this is a change from another fix?

nichamon: The failover's ping responses are not null-terminated. With the validation patch, failover won't work. To fix this, @narategithub appends the responses with zero, and I included the fix to `ldmsd_failover.c` in this pull request.

tom95858:
@nichamon can you please rebase this change?

nichamon:
@tom95858 I just rebased the patch as requested.

narategithub:
@nichamon I think you (or Tom) need to mark the requested change "resolved" in order for this PR to be merged. Can you also rebase this on top of OVIS-4 too?

narategithub (changes requested):
ignore me .. just testing

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -774,36 +773,41 @@ static int string2attr_list(char *str, struct attr_value_list **__av_list,
 	return rc;
 }
 
-int validate_ldmsd_req(ldmsd_req_ctxt_t reqc, ldmsd_req_hdr_t rh)
+int validate_ldmsd_req(ldmsd_req_hdr_t rh)
```
narategithub, line 776: ignore me. I'm just testing ...

narategithub (changes requested):
ignore me again ... just testing.

narategithub (approved):

nichamon:
@tom95858 I rebased this on top of the OVIS-4 branch. If you still cannot merge it, please approve the changes. On the 'File changed' tab, click 'review changes' and then 'approve'.

PR #728, opened 2021-05-27
Cleanup up contexts in msg_tree on disconnect

narategithub:
The message tree maintains configuration requests and replies.
If a connection were to reset while there were outstanding configuration
requests, these context would remain in the tree forever.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -6302,7 +6302,7 @@ static int stream_client_dump_handler(ldmsd_req_ctxt_t reqc)
 	return rc;
 }
 
-void stream_xprt_term(ldms_t x)
+void ldmsd_xprt_term(ldms_t x)
 {
 	__RSE_t ent;
 	struct rbn *rbn;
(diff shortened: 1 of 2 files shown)


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6324,6 +6324,20 @@ void stream_xprt_term(ldms_t x)
...
 	}
 	__RSE_rbt_unlock();
+
+	/* Free outstanding configuration requests */
+	req_ctxt_tree_lock();
+	ldmsd_req_ctxt_t reqc;
+	RBT_FOREACH(rbn, &msg_tree) {
+		struct rbn *succ;
+		reqc = container_of(rbn, struct ldmsd_req_ctxt, rbn);
+		if (reqc->xprt->ldms.ldms == x) {
```
tom95858, line 6334: @nichamon did you say that this check is incorrect or could collide with multiple transports?

nichamon: No, I didn't say that. What I mentioned was the `rbn` at line 6337 is not evaluated because the macro `RBT_FOREACH` will make `rbn` pointing to its successor in the next iteration. We may replace the macro with 
```
rbn = rbt_min(&msg_tree);
while (rbn) {
     succ = rbn_succ(rbn);
     /* Cleanup the request context if it corresponds to x. */     
     rbn = succ;
}
```

nichamon: Another thing is that only the passive side of the connection calls `ldmsd_xprt_term()`. The active side could leak request contexts in `msg_tree` that it has not received the EOF of the request yet. This will occur when a sampler daemon republished a large stream to an aggregator, but the aggregator has not received all the records.

tom95858: Ah right! Thanks, I remembered that you had issues with this but I couldn't remember what they were.

tom95858:
This is replaced with #733

PR #733, opened 2021-05-27
Cleanup up contexts in msg_tree on disconnect

tom95858:
The message tree maintains configuration requests and replies. If a connection were to reset while there were outstanding configuration requests, these context would remain in the tree forever.

I have added the changes recommended by Nichamon. I have tested this at 3 levels of aggregation, but would appreciate any testing that you have.

PR #738, opened 2021-05-28
Make reqc->xprt point to memory on the heap instead of the stack

nichamon:
The LDMSD config transport’s memory addresses passed by ldmsd_process_msg_request()’s callers are on the stack. This is not an issue if the requests come from configuration files. In contrast, the memory address of an LDMS config transport could be invalid while the request context is in the message tree because ldmsd_recv_msg() had been returned already. Moreover, LDMSD could access the invalid address in the disconnected path when ldmsd_xprt_term() iterates through the message tree.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -555,7 +555,7 @@ int __process_config_file(const char *path, int *lno, int trust,
 	ssize_t off = 0;
 	ssize_t cnt;
 	size_t buf_len = 0;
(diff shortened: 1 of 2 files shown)


nichamon:
@tom95858 @narategithub I ran into problems when I test Narate's 733-plus branch. I develop the patch on top of his branch. However, the patch depends on #733 .

narategithub:
@nichamon can you please rebase this on top of OVIS-4?

nichamon:
We won't take this patch.

PR #742, opened 2021-06-02
Don't take LDMS transport's reference in stream_republish_cb()

nichamon:
stream_republish_cb() takes an LDMS transport's reference but does not
put back. Sampler daemons would leak LDMS transports when it republishes
stream to the aggregator. ldmsd_req_cmd_new() already takes a reference
when it initializes the configuration transport.


nichamon:
When I tested #741 , I found that sampler daemons had dangling LDMS transports in `xprt_list` after the sampler daemons disconnected from the aggregator.

PR #745, opened 2021-06-10
Add a periodic flush interval to storage policies

tom95858:
This change adds a periodic call to a storage plugin's flush method.
Previously, this method was never called. The optional 'flush' keyword
is added to the strgp_add command and is used to specify the flush
interval. If the keyword is not specified, the default interval is zero
and the flush method is never called.


tom95858:
@nichamon could you please add the new strgp_add option to the ldmsdctl command?

narategithub (approved):

PR #748, opened 2021-06-11
Update strgp_status to report the flush interval

nichamon:


nichamon:
@tom95858 Please feel free to close this PR if we don't wanna add the flush interval to the strgp_status report.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -2586,12 +2586,15 @@ int __strgp_status_json_obj(ldmsd_req_ctxt_t reqc, ldmsd_strgp_t strgp,
...
 		       "\"schema\":\"%s\","
 		       "\"plugin\":\"%s\","
+		       "\"flush\":\"%ld.%06ld\","
 		       "\"state\":\"%s\","
 		       "\"producers\":[",
 		       strgp->obj.name,
 		       strgp->container,
 		       strgp->schema,
 		       strgp->plugin_name,
+		       strgp->flush_interval.tv_sec,
```
baallan, line 2596: @nichamon Don't we need %.09ld for flush since you're printing tv_nsec and not tv_usec? or do you want to divide by 1000 first?

tom95858: @nichamon please fix this and I will accept it. Should we just convert this to a double and print a formatted double? I know that it's not precise, but @baallan is correct, the format implies usecs, not nsecs.

PR #750, opened 2021-06-17
Make ldmsd_stream_publish_file() support Piped stdin

nichamon:


nichamon:
I have tested the patch with.
```
echo -n pause | ldmsd_stream_publish ....
```
and
```
ldmsd_stream_publish ... -f <file>
```

LDMSD receives the stream successfully.

nichamon:
I also think about limiting the size of a stream if `pipe` is used. The limit could be the max message length, or it could be a constant number, not varied by transports. Limiting the size would prevent the out-of-memory error. What do you guys think?

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -427,23 +427,60 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
...
+
+	} else {
+		if (ESPIPE != errno) {
+			msglog("The given file is invalid.\n");
+			goto close_xprt;
+		}
+
+		/* a non-seekable FILE */
+		char *s;
+		char b[max_msg_len];
```
tom95858, line 460: This buffer needs to be malloc'd to avoid blowing out the stack.

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_stream.c
nichamon: Got it!

PR #752, opened 2021-06-18
Make ldmsd_stream_publish_file() null terminated for non-seekable stream

nichamon:


nichamon:
@tom95858 This morning I tested the top of OVIS-4 after you merged the pull requests. On the container I used, I set the MALLOC_PERTURB_ env so that the system will initialize allocated memory to a non-zero value, and I found that `ldmsd_stream_publish_file()` doesn't null-terminated the stream data if pipe is used. I also tested the code on the same container when I developed the patch and confirmed that if I don't set the env, the code works fine. To make the function always work, I created this pull request.

PR #753, opened 2021-06-18
Add a stream blob writing plugin.

baallan:
This plugin subscribes to any number of streams and creates a message-indexed
dump of each stream. There is no parsing of the messages, so this is fast.
Converting blobs to any particular database format is left to other tools.
Test/demo script and man page included


baallan:
@tom95858 Ready to review/merge.

PR #755, opened 2021-06-21
Fix zap_rdma get_name() local/remote mix up

narategithub:


tom95858 (commented):
I noticed this too, however, when I changed it like you have a bunch of tests broke. So I think there is a bunch of logic that 'corrects' this error down stream.

tom95858:
@narategithub I didn't mean test code in, I meant functionality. When I changed it, things like prdcr_subscribe, etc... stopped working.

narategithub:
Ah. I C. I have not tested those functionalities on rdma. I'll test them and let you know (as well as find all places that call zap_get_name()).

tom95858:
@narategithub:
1. Do we have prdcr_subscribe, etc... tests in TADA?
2. Have we run them?
3. Can you add them if they are not already there?

narategithub:
@tom95858 we have `ldms-test/prdcr_subscribe_test`, but that is container-based which run only on socket transport. I'll make a new one that run on host (like `ldms-test/scalability`) so that `rdma` can be used.

narategithub:
@tom95858 

`ldms-test/direct_prdcr_subscribe_test` which test prdcr_subscribe ( publish -> 2 samplers -> 1x L1 -> 1x L2 -> dump + verify ) directly on the hosts (non-containerized) has been pushed to `ldms-test` repo. This pull request has been tested against it over rdma transport on cygnus cluster.

PR #757, opened 2021-06-22
Delete the unused test code in the main() function

nichamon:


tom95858 (commented):
somehow my comment about removing the -T, etc... options from the FMT macro didn't make it into the review.

PR #758, opened 2021-06-26
add kernel/netlink based pid notifier & simple publish stream api

baallan:
Program netlink-notifier does similar work to slurm notifier without
a spank/slurm dependency. 

The ldms_sps library provides a simple-to-use library wrapping the
ldmsd_stream and ldms_xprt libraries.
In particular it lets the user specify the target daemon/port (or many)
and then delivers message to any available port and drops messages when
no ports are available. It transparently handles reconnections, timeouts,
threads, etc.
It supports netlink-notifier and is proposed to support kokkos and other
long-lived stream publishers.


tom95858:
@baallan , it seems like given the turn on this that it should be a WIP, not a pull request.

baallan:
The churn has been on documentation and extra options, largely. But i'll mark it WIP for now. Expecting to finish up testing tomorrow. The ldms_sps, which is the part that most needs inspecting, is stable afaict.

baallan:
@tom95858 ready to review/merge

baallan:
@tom95858 changed back to work-in-progress, as a unit-ish test is in development and testing with darshan extension found some important nits.

tom95858:
Hi @baallan is this still WIP?

baallan:
@tom95858 It is now back to work in progress because we turned up some issues in testing it as part of darshan.
Also, the issue on lingering set names ( #771 ) needs to be resolved, I suppose, before we merge this.

tom95858:
@baallan, this is good to go yes?

tom95858:
Hi @baallan, the only think we might consider is that "netlink" is a generic interface to Linux kernel services. This sampler is using one instance of that to get notified of new processes. We might want to consider renaming this sampler/configuration option to something more specific. Right now, I think it's fine, but it is certainly a likelihood that we will use the netlink interface in future samplers fr other purposes.

baallan:
I believe it is good enough, unless there is a new "stop publishing/stop subscribing" api I have missed. I too can imagine we might find additional uses for netlink, though at the moment the one I want most (trace file operations at kernel level) seems to be out of scope for netlink (but existing in several elsewheres).

PR #761, opened 2021-07-01
minor patches for json, htbl, and rbt 

baallan:
json needs an value_str_str function, htbl needs a test like rbt, and rbt needs format string patches to pass -Wall -Werr.

### DIFF
--- a/lib/src/ovis_json/ovis_json.c
+++ b/lib/src/ovis_json/ovis_json.c
@@ -6,6 +6,7 @@
 
 #define JSON_BUF_START_LEN 8192
 
+
 const char *json_type_name(enum json_value_e typ)
 {
 	static const char *json_type_names[] = {
@@ -120,6 +121,10 @@ json_entity_t json_attr_find(json_entity_t d, const char *name)
 	assert (d->type == JSON_DICT_VALUE);
 	ent = htbl_find(d->value.dict_->attr_table, name, strlen(name));
 	if (ent) {
+#ifdef JDEBUG
+		fprintf(stderr, "json found attr %s while searching for %s\n",
+			(char *)ent->key, name);
+#endif
 		a = container_of(ent, struct json_attr_s, attr_ent);
 		return &a->base;
 	}
@@ -135,6 +140,10 @@ int json_attr_count(json_entity_t d)
 
 int attr_cmp(const void *a, const void *b, size_t key_len)
 {
+#ifdef JDEBUG
+	fprintf(stderr, "attr_cmp( %s, %s, %zu)\n",
+		(char *)a, (char *)b, key_len);
+#endif
 	return strncmp(a, b, key_len);
 }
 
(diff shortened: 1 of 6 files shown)


baallan:
@narategithub @tom95858 The patches here came up while developing stream data plugins and debugging htbl & rbt usage.

baallan:
extra push because I forgot to delete the #define JDEBUG inside the json C.

lib/src/ovis_json/ovis_json.c
```diff
@@ -654,6 +668,12 @@ json_str_t json_value_str(json_entity_t value)
 	return value->value.str_;
 }
 
+const char * json_value_str_str(json_entity_t value)
```
tom95858, line 671: I'm not sure I see the value of this interface. A json_str_t is a counted string and this interface hides that.

baallan: The value in this is that we stop writing instances of

value->value.str_->str

any place we need to print the entity string value or compare the entity string value to another non-json string.
Why would we want to keep breaking the entity encapsulation specifically for strings when we have similar operators for json_value_int, etc?

tom95858: You're breaking the entity encapsulation, not preserving it. There is already a json_value_str(e), the counted string is accessed as json_value_str(e)->str and json_value_str(e)->len.

tom95858:
@baallan, I do think it is a legitimate question as to why it needs to be a counted string as opposed to just a null terminated char *. I _think_ I just did it that way to avoid lots of strlen(s) when the length is known a-priori.

lib/src/ovis_json/ovis_json.c
baallan: Why would I want to be writing 
```
json_value_str(e)->str
instead of
json_value_str_str(e)
```
Seems like every other string in the c/c++ world provides some form of "just give me the pointer".

Actually, what I'd really like to see is
const char *json_value_str(e);
and
struct json_str_s *json_value_jstr(e);
since I find only a single instance of json_value_str in the ldms code which is not used as json_value_str(e)->str (in calling papi_process_config_data) and 144 instances of jvs(e)->str.

If that's an acceptable change, I'll volunteer to make it.

tom95858:
I think I like your suggestion.

baallan:
> 
> 
> I think I like your suggestion.

I've made several. Which one?

tom95858:
@baallan, add a convenience interface that returns the str_.str portion of  the json counted string. To avoid changing the current interfaces and the associated code, you could call the interface json_value_cstr(e) and it would return the "C-string" associated with the counted string 'e'.

baallan:
> 
> 
> @baallan, add a convenience interface that returns the str_.str portion of the json counted string. To avoid changing the current interfaces and the associated code, you could call the interface json_value_cstr(e) and it would return the "C-string" associated with the counted string 'e'.

will do.

baallan:
@tom95858 updated to  json_value_cstr(e) and included it in ovis_json_test.

PR #762, opened 2021-07-07
fix missing v_char case in csv, whitespace and err msg.

baallan:


baallan:
Without the char value type case added here, the store rejects all data sets containing a char metric.

PR #763, opened 2021-07-07
Demote ENOENT message to LINFO on update callback

baallan:
Samplers removing set instances (and this propagating upstream)
is now a routine event (lustre mount points, app_sampler processes).
This demotes that event to an info from an error, leaving the other
non-specific messages as LERROR.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_updtr.c
+++ b/ldms/src/ldmsd/ldmsd_updtr.c
@@ -550,11 +550,16 @@ void __ldmsd_prdset_lookup_cb(ldms_t xprt, enum ldms_lookup_status status,
 					"It is likely that there are multiple "
 					"producers providing a set with the same instance name.\n",
 					prd_set->prdcr->obj.name, prd_set->inst_name);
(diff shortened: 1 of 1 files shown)


tom95858:
@baallan I have no issue with demoting the error, but this is the lookup path, not the update path which means it happens once when the agg becomes aware of the set via dir_add. Are you seeing this error a lot? If you are it seems like it may be indicative of either a configuration error or some deeper bug.

baallan:
@tom95858 I can send you a log of the nuisance messages if you like. These happen in testing the linux_proc_sampler (similar to app_sampler in terms of sets appear/disappear).

baallan:
This is fine as-is, but additional work will likely touch the same function in resolving #771.

nichamon:
See my comments in #771 .

tom95858:
@baallan could you please close this. This has been fixed with the SET_DELETE message change.

baallan:
@tom95858 I was out friday. I will rerun my tests that provoke it and see if there's anything that still needs revising.

PR #764, opened 2021-07-07
Add env var LDMSD_LOG_TIME_SEC to alter log timestamp format.

baallan:
The default behavior is to log with date strings. If the env var
LDMSD_LOG_TIME_SEC is set to anything, epoch time (%lu.%06lu) is
used instead.


baallan:
@tom95858 I tried that first (since slurm has that feature), but it got me into a lot of 'undefined behavior' territory if the format string is bogus and it also did NOT get me the sub-second field I was after. So if we want arbitrary strftime support, that would be a second feature for managing time format.

PR #767, opened 2021-07-16
Add linux_proc_sampler and related doc derived from app_sampler.

baallan:
This new sampler consumes pids from a stream compatible with it or the app_sampler and supplied by slurm_notifier or netlink-notifier pending in #758.
It has been tested on CTS-1 hardware/toss-3 and centos.
It provides reference tests for app_sampler and linux_proc_sampler driven by ldms-static-test.sh and man page.
It is partially derived (with attribution) from forkstat.

It is disabled by default; configure with '--enable-app-sampler' to test; changing the default to enabled for app_sampler and this would be a good thing but is separable.
It will not pass a build bot test until #761 is merged for json_value_str_str.

### DIFF
--- a/ldms/scripts/examples/app_sampler
+++ b/ldms/scripts/examples/app_sampler
@@ -1,0 +1,20 @@
+export plugname=app_sampler
+export dsname=$(ldms_dstat_schema_name mmalloc=1 io=1 fd=1 auto-schema=1)
+export dstat_schema=$dsname
+portbase=61060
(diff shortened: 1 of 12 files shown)


baallan:
@tom95858 this can be easily tested after you merge #758, #761, #763.

baallan:
@tom95858 this sampler is ready for review.

I  propose changing the default of app-sampler to OPTION_DEFAULT_ENABLE before merging, as the code in this directory is not dependent on anything uncommon.

PR #770, opened 2021-07-20
Fix the failover's ping logic

nichamon:
The patch fixes two bugs in the failover’s ping logic. First, the ping
response does not include PEERCFG_ACTIVATED in the flag when the peercfg
has been activated. Second, the ping response handler checks whether the
failover buddy has activated the peercfg or not with the wrong logic. It
verifies whether the buddy’s failover has been configured or not instead
of checking if the buddy has started the peercfg. If the buddy has
started the peercfg, it should not start its own cfg yet.


nichamon:
I found the bugs by code inspection and git history. The patch passed the failover_test in the ldms-test repo.

PR #774, opened 2021-07-27
fix #765 lnet_stats  retry needed

### LINKED ISSUE #765: lnet_stats retry needed
sampler lnet_stats, unlike lustre_client and other recent lustre samplers, does not cope with changing lustre mount status. in particular the sampler fails if no mount points exist.

We need to have lnet_stats retry periodically and not emit errors if there is no stats file because lustre is temporarily inactive.

baallan:
fixes #765
includes updated man page and test driver

### DIFF
--- a/ldms/man/Plugin_lnet_stats.man
+++ b/ldms/man/Plugin_lnet_stats.man
@@ -41,7 +41,7 @@ Optional full path name of stats file to use. If not supplied, the default searc
 .SH NOTES
 .PP
 The default search path followed for LNET stats is:
-/sys/kernel/debug/lnet/stats:/proc/sys/lnet/stats. Which file will exist, if either, depends on the 
+/sys/kernel/debug/lnet/stats:/proc/sys/lnet/stats. Which file will exist, if either, depends on the
 Lustre version and how many volumes are currently mounted. Be aware that /sys/kernel/debug normally is only readable by privileged users.
 .PP
 The stats file disappears when all mounts are unmounted or not yet mounted. While it is missing, the data set is not updated.
--- a/ldms/man/Plugin_lnet_stats.man
+++ b/ldms/man/Plugin_lnet_stats.man
@@ -36,12 +36,14 @@ different schema. If not specified, will default to `lnet_stats`.
 .TP
 file=<stats_path>
 .br
-Optional full path name of stats file if not in default location.
+Optional full path name of stats file if not in a default location.
 
 .SH NOTES
 .PP
 The default location for the lnet stats path is:
 /sys/kernel/debug/lnet/stats, and /proc/sys/lnet/stats depends on the Lustre version. Be aware that /sys/kernel/debug is only readable by privileged users.
+.PP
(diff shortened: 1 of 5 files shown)


ldms/man/Plugin_lnet_stats.man
```diff
@@ -36,12 +36,14 @@ different schema. If not specified, will default to `lnet_stats`.
 .TP
 file=<stats_path>
 .br
-Optional full path name of stats file if not in default location.
+Optional full path name of stats file if not in a default location.
 
 .SH NOTES
 .PP
 The default location for the lnet stats path is:
 /sys/kernel/debug/lnet/stats, and /proc/sys/lnet/stats depends on the Lustre version. Be aware that /sys/kernel/debug is only readable by privileged users.
```
tom95858, line 44: depends --> depending?

ldms/man/Plugin_lnet_stats.man
```diff
@@ -50,7 +52,6 @@ The default location for the lnet stats path is:
 .SH BUGS
 Assumes file name instead of looking it up from the Lustre runtime libraries,
```
tom95858, line 53: When you say assumes, isn't it trying the two different locations? What is the order. Does the the 'file' argument override this?

ldms/src/sampler/lnet_stats.c
```diff
@@ -300,15 +309,15 @@ static int config(struct ldmsd_plugin *self, struct attr_value_list *kwl, struct
 
 	rc  = create_metric_set(base);
 	if (rc) {
-		msglog(LDMSD_LERROR, SAMP ": failed to create a metric set.\n");
+		msglog(LDMSD_LWARNING, SAMP ": failed to create a metric set.\n");
```
tom95858, line 312: Why is this only a warning?

ldms/man/Plugin_lnet_stats.man
baallan: I will try to rephrase the man page details you point out in some satisfactory way.

ldms/src/sampler/lnet_stats.c
baallan: This is a warning because it is an expected and likely recoverable failure without admin intervention; should that be demoted further to info? ldmsd may be coming up while lustre is temporarily not providing lnet data in any location; this occurs if there are no active mount points. I will review the messages that will precede or follow this one during the failure cases and see if they could be made more informative.

tom95858: My reading of the code may be incorrect, but it looks like if the source file (i.e. /proc/...) could not be opened, it fails before trying to call create_metric_set. Maybe the file is there but empty?

baallan: No, you are correct. The change to warning is incorrect, now that the file isn't actually checked as part of create_metric_set.
I will undo that.

baallan:
@tom95858 this should be good to go now.

PR #775, opened 2021-07-29
fix #773 unreleased lock in __send_lookup_reply

### LINKED ISSUE #773: Potential error(e.g., deadlock ) due to the unreleased lock &set->lock (CWE-667)
Dear developers, thank you for your checking! It seems that the lock `&set->lock` may not be released correctly if the method `__send_lookup_reply` when the` !msg` is satisfied and the control flow goes to  `err_0` that does not release the lock &set->lock. Should it be a real bug? I think the fix is to insert `pthread_mutex_unlock(&set->lock);` in the branch of` !msg`.

https://github.com/ovis-hpc/ovis/blob/84513760731014c170104ff11608867b62096b58/ldms/src/core/ldms_xprt.c#L1274

The process is detailed as below.
```
static int __send_lookup_reply(struct ldms_xprt *x, struct ldms_set *set,
			       uint64_t xid, int more)
{
	...;
	pthread_mutex_lock(&x->lock); // acquire the lock x->lock first
	
	...;

	pthread_mutex_lock(&set->lock);  // acquire the lock set->lock
	...;
	if (!msg) {  // the condition may be satisfied
		rc = ENOMEM;
		goto err_0; // goto err_o mistakenly with unrelased the lock set->lock
	}

	...;
	pthread_mutex_unlock(&set->lock);
	...;
	if (zerr != ZAP_ERR_OK) {
		...;
		goto err_0; // correctly since the lock set->lock has been released
	}
	pthread_mutex_unlock(&x->lock);
	...;
	return 0;
 err_0:
	pthread_mutex_unlock(&x->lock); 
	return rc;
}
```

Best,

baallan:
fixes related error logging in caller also.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1271,6 +1271,7 @@ static int __send_lookup_reply(struct ldms_xprt *x, struct ldms_set *set,
 	msg = calloc(1, msg_len);
 	if (!msg) {
 		rc = ENOMEM;
+		pthread_mutex_unlock(&set->lock);
 		goto err_0;
 	}
 
@@ -1423,7 +1424,9 @@ static void process_lookup_request_re(struct ldms_xprt *x, struct ldms_request *
 	if (zerr != ZAP_ERR_OK) {
 		x->zerrno = zerr;
 		x->log("%s: x %p: zap_send synchronously errors '%s'\n",
-				__func__, x, zap_err_str(rc));
+				__func__, x, zap_err_str(zerr));
+		x->log("%s: local error code %d (%s)\n",
+				__func__, rc, STRERROR(rc));
 		ldms_xprt_close(x);
 	}
 }
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1423,10 +1423,9 @@ static void process_lookup_request_re(struct ldms_xprt *x, struct ldms_request *
 	zap_err_t zerr = zap_send(x->zap_ep, &hdr, sizeof(hdr));
 	if (zerr != ZAP_ERR_OK) {
 		x->zerrno = zerr;
-		x->log("%s: x %p: zap_send synchronously errors '%s'\n",
-				__func__, x, zap_err_str(zerr));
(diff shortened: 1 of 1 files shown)


tom95858:
@nichamon could you please review this

nichamon:
@tom95858 The fixes are legit.

@baallan Thanks for catching the missing unlock and the wrong error code variable. What do think about merging the two error messages, the `zap_send` synchronously errors and the local error code? The two messages are related. If `zap_send()` doesn't fail, the client will receive the error. In this case, the `zap_send()` failed so you logged the error that triggers the `zap_send()`. Shall we log the error something like 
```
x->log("%s: x %p: zap_send synchronously errors '%s' while sending the error %d (%s)\n",
				__func__, x, zap_err_str(zerr), rc, STRERROR(rc));
```

baallan:
@nichamon done combining the messages.

baallan:
@nichamon looks like the build fail on ubuntu is provoked by github and not the code change.

nichamon:
> @nichamon looks like the build fail on ubuntu is provoked by github and not the code change.

@baallan Github didn't even run ./autogen.sh yet. It failed when it tried to fetch the code. Could you try to re-run the test again?

nichamon (approved):

baallan:
@nichamon I tried, and there is no page which presents me with an option to rerun the check or the entire workflow; I presume this is due to my role in the project being 'outside collaborator'.

nichamon:
@baallan What works for me in this situation is to do `git commit --amend` without changing the code and the commit message. Could you please give it a try?

On your local machine,
```
> git checkout <branch name>
> git commit --amend # Save and close the editor without changing anything
> git push -f <remote> HEAD
```
You would need to _force_ push the HEAD to the remote branch because `git commit --amend` will always change the TIP of the branch locally.

baallan:
@nichamon the re-run checks hack that seems to work is to close and immediately reopen the issue, doable with  UI buttons, though undocumented.

baallan:
@nichamon is there anything else you would like changed before you can approve this one?

nichamon (approved):

nichamon:
@baallan I approved it yesterday before you resolved the Github action fail. I just approved it again just in case.

PR #776, opened 2021-07-29
fixes #772 missing unlock in process_dir_set_info

### LINKED ISSUE #772: Potential error(e.g., deadlock ) due to the unreleased lock (CWE-667)
Dear developers, thank you for your checking! It seems that the lock `&lset->lock` may not be released correctly if the method  `int __process_dir_set_info` returns in the below two branches. Should it be a real bug? I think the fix is to insert `pthread_mutex_unlock(&lset->lock);` before returning.

https://github.com/ovis-hpc/ovis/blob/84513760731014c170104ff11608867b62096b58/ldms/src/core/ldms_xprt.c#L1686
https://github.com/ovis-hpc/ovis/blob/84513760731014c170104ff11608867b62096b58/ldms/src/core/ldms_xprt.c#L1692

```
static int __process_dir_set_info(struct ldms_set *lset, enum ldms_dir_type type,
				ldms_dir_set_t dset, json_entity_t info_list)
{
	...;

	if (lset)
		pthread_mutex_lock(&lset->lock);
	for (j = 0, info_entity = json_item_first(info_list); info_entity;
	         ...;
		if (!dset->info[j].key)
			return ENOMEM;
		...;
		if (!dset->info[j].value)
			return ENOMEM;

		...;
	}
	if (lset) {
		pthread_mutex_unlock(&lset->lock);
		...;
	}
	return 0;
}
```

Best,

baallan:


tom95858:
@nichamon could you please review this

ldms/src/core/ldms_xprt.c
```diff
@@ -1678,7 +1678,7 @@ static int __process_dir_set_info(struct ldms_set *lset, enum ldms_dir_type type
 				ldms_dir_set_t dset, json_entity_t info_list)
 {
 	json_entity_t info_entity, e;
-	int j, rc;
+	int j, rc = 0;
```
nichamon, line 1681: @baallan I think we don't need to initialize the `rc` here. The function will assign a value in case of an error before returning it.

ldms/src/core/ldms_xprt.c
```diff
@@ -1723,8 +1727,11 @@ static int __process_dir_set_info(struct ldms_set *lset, enum ldms_dir_type type
 		}
 		pair = nxt_pair;
 	}
+err:
 	if (lset) {
 		pthread_mutex_unlock(&lset->lock);
+		if (rc)
```
nichamon, line 1733: @baallan I think this is not correct because if `lset` is NULL, the function will always return 0, regardless of the error. Consider the `goto err;` on line 1697. We should separate the error out and the normal out.

```
out:
    if (lset) {
        pthread_mutex_unlock(&lset->lock);
        if (type == ....) {
          ....
        }
    }
    return 0;
err:
    if (lset)
        pthread_mutex_unlock(&lset->lock);
    return rc;
```

nichamon (changes requested):

tom95858:
@nichamon please approve when it looks correct.

ldms/src/core/ldms_xprt.c
baallan: @nichamon if lset is not null but info_list is empty, rc is uninitialized unless we set it.
At the scope of this function, there is no assurance that info_list is populated, and gcc warns that rc can be used uninit at 1733.
If compiling such that warnings are promoted to error, this bombs the build.

ldms/src/core/ldms_xprt.c
baallan: @nichamon you are correct that in the lset==null case, the rc gets lost. We can keep simplify and keep the out-path code unified by simply changing the final 'return 0' to 'return rc'. I've pushed a revision that does this.

nichamon (approved):

PR #779, opened 2021-08-24
Get rid of the log file lock

nichamon:
Logging to the log file becomes a performance bottleneck when a burst of
errors occur, especially on aggregators. We use event-driven programming
to get rid of the log file's lock. The ldmsd_log() function publishes an
event to a worker exclusively responsible for printing data to
the log file or the syslog. Thus, there is only one thread that accesses the log file.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -267,31 +267,18 @@ err:
 	return rc;
 }
 
-int __log(enum ldmsd_loglevel level, char *msg)
+int __log(enum ldmsd_loglevel level, char *msg, struct timeval *tv, struct tm *tm)
 {
 	if (log_fp == LDMSD_LOG_SYSLOG) {
 		syslog(ldmsd_loglevel_to_syslog(level), "%s", msg);
 		return 0;
 	}
 
-	if (log_time_sec == -1) {
-		char * lt = getenv("LDMSD_LOG_TIME_SEC");
-		if (lt)
-			log_time_sec = 1;
-		else
-			log_time_sec = 0;
-	}
 	if (log_time_sec) {
-		struct timeval tv;
-		gettimeofday(&tv, NULL);
-		fprintf(log_fp, "%lu.%06lu: ", tv.tv_sec, tv.tv_usec);
+		fprintf(log_fp, "%lu.%06lu: ", tv->tv_sec, tv->tv_usec);
 	} else {
-		time_t t;
 		char dtsz[200];
-		struct tm tm;
-		t = time(NULL);
-		localtime_r(&t, &tm);
-		if (strftime(dtsz, sizeof(dtsz), "%a %b %d %H:%M:%S %Y", &tm))
+		if (strftime(dtsz, sizeof(dtsz), "%a %b %d %H:%M:%S %Y", tm))
 			fprintf(log_fp, "%s: ", dtsz);
 	}
 
@@ -309,16 +296,18 @@ int log_actor(ev_worker_t src, ev_worker_t dst, ev_status_t status, ev_t ev)
 	enum ldmsd_loglevel level = EV_DATA(ev, struct log_data)->level;
 	char *msg = EV_DATA(ev, struct log_data)->msg;
 	uint8_t is_rotate = EV_DATA(ev, struct log_data)->is_rotate;
+	struct timeval *tv = &EV_DATA(ev, struct log_data)->tv;
+	struct tm *tm = &EV_DATA(ev, struct log_data)->tm;
 	int rc;
 
 	if (is_rotate) {
 		rc = __logrotate();
 	} else {
-		rc = __log(level, msg);
+		rc = __log(level, msg, tv, tm);
 		if (0 == ev_pending(logger_w))
 			fflush(log_fp);
+		free(msg);
 	}
-	free(msg);
 	ev_put(ev);
 	return rc;
 }
@@ -328,6 +317,9 @@ void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap)
 	ev_t log_ev;
 	char *msg;
 	int rc;
+	struct timeval *tv;
+	struct tm *tm;
+	time_t t;
 
 	log_ev = ev_new(log_type);
 	if (!log_ev)
@@ -339,6 +331,22 @@ void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap)
 	EV_DATA(log_ev, struct log_data)->level = level;
 	EV_DATA(log_ev, struct log_data)->is_rotate = 0;
 
+	if (log_time_sec == -1) {
+		char * lt = getenv("LDMSD_LOG_TIME_SEC");
+		if (lt)
+			log_time_sec = 1;
+		else
+			log_time_sec = 0;
+	}
+	if (log_time_sec) {
+		tv = &EV_DATA(log_ev, struct log_data)->tv;
+		gettimeofday(tv, NULL);
+
+	} else {
+		tm = &EV_DATA(log_ev, struct log_data)->tm;
+		t = time(NULL);
+		localtime_r(&t, tm);
+	}
 	ev_post(NULL, logger_w, log_ev, NULL);
 }
 
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -300,7 +300,7 @@ int __log(enum ldmsd_loglevel level, char *msg)
 	}
 
 	fprintf(log_fp, "%s", msg);
-	fflush(log_fp);
+
 	return 0;
 }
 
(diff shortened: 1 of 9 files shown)


baallan:
This is a very undesirable change as implemented. In particular:
* We need logging to work precisely when malloc has failed, so starting every message with a data allocation for an event is  a non-starter.
* Adding a call to free (which can be a huge performance bottleneck) for the event processing to every log message can be a huge time sink and memory fragmenter.

For the cited case of aggregators, if logging is a problem, just configure to log less. This patch turns a "slows down aggregation" problem into a "blows up memory" problem which is likely to be substantially worse.

It seems like this is a band-aid instead of solving the problem of allowing logging to be precisely targeted as in:
https://github.com/ovis-hpc/ovis-wiki/wiki/Proposal-5

tom95858:
> This is a very undesirable change as implemented. In particular:
> 
>     * We need logging to work precisely when malloc has failed, so starting every message with a data allocation for an event is  a non-starter.

Ensuring ENOMEM messages get to the log file is no small challenge. The current implementation does not guarantee this either.

> 
>     * Adding a call to free (which can be a huge performance bottleneck) for the event processing to every log message can be a huge time sink and memory fragmenter.

I believe this is an overstatement. Malloc/free are quite efficient.

That said, it would be _more_ efficient to have a pool of pre-allocated log-event buffers and then vsprintf into those buffers for logging. By the way, printf/vsprintf may call malloc().

Also, @nichamon could you please compare the CPU utilization of this with the existing logging. I'm not asking that you instrument the logging path, just look at the CPU utilization of the daemon as a whole with and without this patch.

> 
> 
> For the cited case of aggregators, if logging is a problem, just configure to log less. This patch turns a "slows down aggregation" problem into a "blows up memory" problem which is likely to be substantially worse.
>

You measured this? Can you share the results?
 
> It seems like this is a band-aid instead of solving the problem of allowing logging to be precisely targeted as in:
> https://github.com/ovis-hpc/ovis-wiki/wiki/Proposal-5

This removes a very hot lock from the daemon while ensuring the ordering of messages. Proposal-5 is orthogonal to this patch I believe.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -262,9 +301,45 @@ void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap)
...
+void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap)
+{
+	ev_t log_ev;
+	char *msg;
+	int rc;
+
+	log_ev = ev_new(log_type);
+	if (!log_ev)
+		return;
+	rc = vasprintf(&msg, fmt, ap);
```
tom95858, line 335: Perhaps a queue of 1024B message buffers that we can vsprintf into? This will reduce the overhead by avoiding the malloc/free.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -262,9 +301,45 @@ void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap)
...
+	return rc;
+}
+
+void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap)
+{
+	ev_t log_ev;
+	char *msg;
+	int rc;
+
+	log_ev = ev_new(log_type);
```
tom95858, line 332: Does ev_new() call malloc in the current implementation?

baallan:
@tom95858 Re the memory blows up: lets take the advertised (JB) 10000 clients per aggregator, 1Hz sampling, 10 log lines of 80 bytes per set update, 10 sets per client. This comes to 80MB/s. It's very easy to picture that kind of data creation rate (across 2 aggregators) outstripping a local log disk or remote nfs link.  Granted in most hpc systems with 10000 clients, hardware will be better provisioned at the L1, but maybe not at the L2.

Re the lock: I'm not sure how moving the hot lock down into malloc/free really helps us. Seems like we're throwing a lot of thought into just moving a lock around. You still have to have a lock to inject/remove messages into the log event list, don't you?

Why would *vsnprintf* be calling malloc? Wouldn't a preallocated (and perhaps growable) ring buffer with variable record size be the thing to do here to stay out of malloc/free?

tom95858:
The pre-allocated buffer is what has already been suggested in a review. With regard to speculation about malloc/free and 'moving the lock around', @nichamon 's testing showed that with the new implementation and a log setting of INFO we were able to reliably sample and store a sizable cluster configuration. Without the change, the daemon backed up and stopped sampling and storing altogether. I think @nichamon has demonstrated that an evidence-based approach to design choices results in better outcomes.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -103,6 +103,7 @@
 #define LDMSD_MEM_SIZE_ENV "LDMSD_MEM_SZ"
 #define LDMSD_MEM_SIZE_STR "512kB"
 #define LDMSD_MEM_SIZE_DEFAULT 512L * 1024L
+#define LDMSD_FLUSH_LOG_TIME 10 * 60 /* 10 mins */
```
tom95858, line 106: @nichamon, I don't see why we need this capability. If the log message rates are low, the log buffer will get flushed because the log event queue is often empty. If the log message rates are high, the log buffer (4096B) will fill up and get flushed anyway by libc. Am I missing something?

nichamon: @tom95858 No, you didn't miss anything. I added the code to make a comparison without the patch. However,  OVIS-4 doesn't use the event-driven paradigm, so the only way is to flush periodically. I just tried running the current OVIS-4 with the INFO log level on Voltrino. With the OVIS-4 code case, Zap spawned tens of threads because the threads were stuck in flushing the log file.

nichamon:
@tom95858  I just tried running the current OVIS-4 with the INFO log level on Voltrino. With the OVIS-4 code base (without this patch), Zap spawned tens of threads because the threads were stuck in flushing the log file. The results from this run and the run without multi-thread zap strongly suggest that we should not have Zap threads or LDMSD worker threads responsible for writing to the log file.

What is the objective of having the pre-allocated buffers? Is it to avoid spending CPU time to do malloc/free at every log message? Having the pre-allocated buffer pools will need 1) a lock and 2) a strategy that handles when the number of log messages overwhelms the pre-allocated buffers. Could we try without the pre-allocated buffers and go from there when we witness issues?

ldms/src/ldmsd/ldmsd.c
nichamon: Moreover, I have a concern that the condition `0 == ev_pending(logger_worker)` may flush too often. I would like to compare the memory usages of the condition and the fixed interval strategy. The strategy that flushes when the queue is empty may have worse performance than the fixed interval in the situation that most log messages enqueue right after the logger worker processes the previous log message.

nichamon:
After discussing with @tom95858, we conclude that we will address the pre-allocated buffer in the future. I have removed the periodically flush logic from the patch, so LDMSD will call `fflush()` when there are no messages waiting in the log queue.

baallan:
@nichamon the major reason for preexisting buffers is to ensure (as best one can) that a message about ENOMEM gets delivered to the log. Which is hard to do if messages cause malloc.

tom95858:
@nichamon 
> 
> What is the objective of having the pre-allocated buffers? Is it to avoid spending CPU time to do malloc/free at every log message?

That is a part of the objective, but not entirely.

> Having the pre-allocated buffer pools will need 1) a lock

We can use a lock-less producer/consumer data structure where only the producer updates the head, and only the consumer updates the tail. But even without that, the issue is that we're holding a global lock while we flush to the log. Even with a lock, the time that the lock is held is infinitesimal as compared to the time it is held now.

The 2nd reason is that this will naturally throttle the logger. Right now the logger backlog potentially grows unbounded. With a pool of pre-allocated buffers, when these buffers fill up, the logging thread must wait until there is space to log additional messages. 

The number of these pre-allocated buffers could be scaled with the debug level. Given that the log messages are naturally bursty, e.g. when an aggregator connects to thousands of producers and then looks up every metric set, or after each update cycle; if we scale to a limit that supports the bursts, we don't hold up the daemon while it is trying to get work done in the presence of a large number of log messages.

> and 2) a strategy that handles when the number of log messages overwhelms the pre-allocated buffers.

This is true, in that case the logging thread would have to block until there is space to log the new message. Right now, however, competing log clients block on every message.

> Could we try without the pre-allocated buffers and go from there when we witness issues?

I think so, it is better than the current approach, but perhaps could be made better. With the current serial logging, let's move the lock above the fflush(). I don't think it will help actually because internally fflush() locks the output file descriptor anyway, but it would be interesting to try.

I'll look at the event subsystem to see if it has a similar issue, i.e. overrunning the worker thread.

nichamon:
>  With the current serial logging, let's move the lock above the flush().

FYI, with this patch, LDMSD has no global lock for logging. 

How logging works in this patch.

- The logging thread creates an event and allocates a string for the log message to log a message.
- Then, the logging thread posts the event to the logger worker.
- The logger writes to the file descriptor.
- If the logger has no other events in the queue, it calls `fflush()`.

tom95858:
@nichamon,

> > With the current serial logging, let's move the lock above the flush().
> 
> FYI, with this patch, LDMSD has no global lock for logging.
> 

I realize that, I'm talking about the _old_ locking without this patch. That's the lock I'm talking about moving, for the purposes of comparison with the logging _with_ this patch.


BTW, even  with the patch there _is_ locking inside fprintf, fflush, etc... on the log file pointer.

tom95858:
@nichamon can you please let me know when you are ready to have this merged? It's not a WIP or draft, but changes keep being made.

tom95858 (approved):

nichamon:
@tom95858 It is ready. The changes 8 days ago addressed the flush call. The change today was a rebased on the top of the latest OVIS-4.

PR #780, opened 2021-08-25
Re-initialize ldms/zap threads at first endpoint creation

narategithub:
The new process has only one thread after fork (see `fork(2)`), and
hence the need of `pthread_atfork()` in `ldms.c`, `zap.c` and
`zap_sock.c` to re-initialize the set delete thread, zap worker
threads and zap sock io thread respectively. However, the Slurm
job_container feature failed to run a task because it called
`setns()` that did not support multi-threading. Basically, Slurm
job_container feature could not be used with slurm_notifier.

To work around the Slurm job_container `setns()` issue, this patch
modified `ldms.c`, `zap.c` and `zap_sock.c` so that they will not
re-create the threads right away after fork. Instead, they will just
reset the initialization state at fork, and the transport-related
threads will be created at the creation of the first endpoint after
fork (the set-related thread will be created at the first set
creation).

This has been tested on top of OVIS-4.3.7 and slurm 20.11.8 on OGC
cygnus cluster.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -953,6 +953,40 @@ uint32_t __ldms_set_size_get(struct ldms_set *s)
 
 #define LDMS_GRAIN_MMALLOC 1024
 
+static int delete_thread_initialized = 0;
+static pthread_t delete_thread;
+
+void ldms_atfork()
+{
+	__atomic_store_n(&delete_thread_initialized, 0, __ATOMIC_SEQ_CST);
+}
+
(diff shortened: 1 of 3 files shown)


tom95858:
@eric-roman can you confirm that this fixes the issue and that you are happy with the fix?

eric-roman:
I was able to test this and I am not seeing the problem.  Before closing the issue I want our container expert, who provided the original test case, to take a look at it, but I think it's all set.

PR #783, opened 2021-08-31
V4 staging

narategithub:
This pull requests contain (a lot of) commits from `v4-staging` branch which includes the following features:
- metric unit support
- `zap_send_mapped()` (similar to RDMA send)
- zap shared map: a zap memory map that is created independent of zap endpoint and can be used with many endpoints (regardless of transport types)
- ref library
- zap send completion / ldms send completion events
- new zap thread model

This has been tested with:
- containerized test cases in `ldms-test` (sock only)
- non-containerized scalability test in `ldms-test/scalability` (`scale` branch) on `sock`, `rdma`, and `ugni`
- `zap_test` on `sock`, `rdma`, and `ugni`


narategithub:
@tom95858 I'll look into the compatibility test issue tomorrow.

narategithub:
@tom95858 This is ready to review/merge.

Issue #771, opened 2021-07-22
set lifecycle and L1 problem


baallan:
I have a sampler (pr 767) where  L0 sets have the following lifecycle (driven by events in /proc/$pid):
```
ldms_set_new
ldms_set_publish
ldmsd_set_register
(L1 discovers)
ldmsd_set_deregister
ldms_set_unpublish
ldms_set_delete
```

And at the L1, there seems to be some issue with it never getting properly eliminated; the updtr keeps looking for it forever, as logged below. Notice that the set has an existence shorter than the updater interval (1 sec). I am stumped as to why L1 is lingering on it. Help!

1626913802.534716: INFO      : Adding the metric set 'localhost1/0/1626913802.492302/19019'
1626913803.103245: DEBUG     : updtr_task sched '1000000': set 'localhost1/0/1626913802.492302/19019'
1626913803.145117: INFO      : prdcr localhost1: disappeared (No such file or directory) set in lookup callback of set 'localhost1/0/1626913802.492302/19019'
1626913804.104459: DEBUG     : updtr_task sched '1000000': set 'localhost1/0/1626913802.492302/19019'
1626913804.147541: INFO      : prdcr localhost1: disappeared (No such file or directory) set in lookup callback of set 'localhost1/0/1626913802.492302/19019'
1626913805.101622: DEBUG     : updtr_task sched '1000000': set 'localhost1/0/1626913802.492302/19019'
1626913805.144678: INFO      : prdcr localhost1: disappeared (No such file or directory) set in lookup callback of set 'localhost1/0/1626913802.492302/19019'
1626913806.103705: DEBUG     : updtr_task sched '1000000': set 'localhost1/0/1626913802.492302/19019'
1626913806.146866: INFO      : prdcr localhost1: disappeared (No such file or directory) set in lookup callback of set 'localhost1/0/1626913802.492302/19019'
1626913807.103955: DEBUG     : updtr_task sched '1000000': set 'localhost1/0/1626913802.492302/19019'
1626913807.143999: INFO      : prdcr localhost1: disappeared (No such file or directory) set in lookup callback of set 'localhost1/0/1626913802.492302/19019'
1626913808.102939: DEBUG     : updtr_task sched '1000000': set 'localhost1/0/1626913802.492302/19019'
1626913808.142982: INFO      : prdcr localhost1: disappeared (No such file or directory) set in lookup callback of set 'localhost1/0/1626913802.492302/19019'
[...]

PR #786, opened 2021-09-01
Advertise LDMS set delete event to peers

narategithub:
Previously, LDMS advertised the set delete event only to peers that has
successfully looked up the set. In the case of quick set add-delete, the
peer will only get DIR ADD event but won't get LDMS set delete
notification because the lookup process has not completed in time. This
patch modifies LDMS so that it advertises set delete event to downstream
LDMS peers so that the peer can undo what it has done from DIR ADD
event.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -573,7 +573,6 @@ int __ldms_set_unpublish(struct ldms_set *set)
 		return ENOENT;
 
 	set->flags &= ~LDMS_SET_F_PUBLISHED;
-	__ldms_dir_del_set(set);
 	return 0;
 }
 
@@ -751,12 +750,6 @@ static void __destroy_set(void *v)
 	pthread_mutex_unlock(&__del_tree_lock);
 }
 
-static void __set_delete_cb(ldms_t xprt, int status, ldms_set_t rbd, void *cb_arg)
-{
-	struct ldms_set *set = cb_arg;
-	ref_put(&set->ref, "xprt_set_coll");
-}
-
 void ldms_set_delete(ldms_set_t s)
 {
 	ldms_t x;
@@ -803,6 +796,9 @@ void ldms_set_delete(ldms_set_t s)
 		free(pp);
 	}
 
+	/* Notify downstream transports about the set deletion. */
+	__ldms_dir_del_set(s);
+
 	/*
 	 * Clean up the lookup peer collection and
 	 * remove set from the transport's set collection.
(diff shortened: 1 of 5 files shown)


narategithub:
@tom95858 @nichamon 

The code passed this test: https://github.com/ovis-hpc/ldms-test/blob/quick_add_rm/quick_set_add_rm_test .

baallan:
@narategithub is this expected to be a complete fix for #771 ?

baallan:
@narategithub In trying this patch, my screen is drowned in messages:
"ERR: Level-3: <time>: (<ptr>) (<pid>) event_cb:186 Received invalid event type"
Don't know if this is an error or just a debug that slipped in.

I no longer see the lingering sets on the L1, as mentioned in #771; thank you.

In testing (valgrind drd), i'm seeing a bunch of new data race errors in zap and zap sock that i didn't see in earlier testing (~2 weeks ago). I don't think this is related to the patch here.

In testing (valgrind memcheck) I'm seeing a new message (below) which indicates the message len checking is in some case getting an uninitialized value.  This might be related to the patch here if the delete message is on a new code path. The same message occurs on the L1 and L2.
```
==32411== Thread 4:
==32411== Conditional jump or move depends on uninitialised value(s)
==32411==    at 0x4E46C38: ldms_xprt_sockaddr (ldms_xprt.c:3811)
==32411==    by 0x4E47064: ldms_zap_cb (ldms_xprt.c:2744)
==32411==    by 0x1204BDF2: process_sep_msg_accepted (zap_sock.c:518)
==32411==    by 0x1204D0F0: sock_read (zap_sock.c:1355)
==32411==    by 0x1204D0F0: sock_ev_cb (zap_sock.c:1150)
==32411==    by 0x1204DC24: io_thread_proc (zap_sock.c:1406)
==32411==    by 0x70D5EA4: start_thread (in /usr/lib64/libpthread-2.17.so)
==32411==    by 0x63909FC: clone (in /usr/lib64/libc-2.17.so)
==32411==  Uninitialised value was created by a stack allocation
==32411==    at 0x4E46CB0: ldms_zap_cb (ldms_xprt.c:2693)
==32411== 
```

tom95858:
@narategithub Did you confirm/disconfirm that this issue is not the cause of the error messages @baallan is seeing?

tom95858:
@narategithub it looks to me like ldms_xprt_sockaddr is all balled up. It looks like it's trying to avoid calling zap_get_name() if it has already been called, but there is other logic that tests that the address buffer is big enough to contain the result addresses. A look at the call sites shows that many of the input buffer size is not initialized. Finally the initialization that calls zap_get_name() sets the input length incorrectly and won't work with IPv6

PR #787, opened 2021-09-02
Apply the event-driven paradigm to LDMSD's routines

nichamon:


nichamon:
Please expect constant code changes as I'm debugging the patch and I'll rebase it on top of #779 tonight.

PR #788, opened 2021-09-02
ldms list type (heap)

narategithub:


ldms/src/sampler/examples/list_sampler/list_sampler.c
```diff
@@ -0,0 +1,410 @@
...
+#include <stdio.h>
+#include <stdarg.h>
+#include <string.h>
+#include <sys/types.h>
+#include <time.h>
+#include <pthread.h>
+#include <assert.h>
+#include "ldms.h"
+#include "ldmsd.h"
+#include "../../sampler_base.h"
```
baallan, line 67: why do we need ../../sampler_base.h? we seem to use unqualified in several other samplers.

narategithub:
@tom95858 

This is currently a draft. The patches were ported from @tom `ovis-list.tgz`. It built, but I have not tested it yet.

With its current stage, it will break the compatibility with OVIS-4 HEAD. In order to preserve the compatibility, I think we need the following:

1. the `ldms_set_hdr->heap_sz` must use the `reserved` space, In other words:
```c
struct ldms_set_hdr {
	/* The unique metric set producer name */
	char producer_name[LDMS_PRODUCER_NAME_MAX];
	uint64_t meta_gn;	/* Meta-data generation number */
	struct ldms_version version;	/* LDMS version */
	uint8_t flags;	/* Set format flags */
	uint8_t pad1;	/* data pad */
	uint8_t pad2;	/* data pad */
	uint8_t pad3;	/* data pad */
	uint32_t card;		/* Size of dictionary */
	uint32_t meta_sz;	/* size of meta data in bytes */
	uint32_t data_sz;	/* size of metric values in bytes */
	uint32_t heap_sz;	/* size of heap in bytes */
	uint32_t uid;           /* UID */
	uint32_t gid;           /* GID */
	uint32_t perm;          /* permission */
	uint32_t array_card;    /* number of sets in the set array */
	uint32_t reserved[8];	/* area reserved for compatible core updates */
	uint32_t dict[OVIS_FLEX];/* The attr/metric dictionary */
};
```
should be 
```c
	/* The unique metric set producer name */
	char producer_name[LDMS_PRODUCER_NAME_MAX];
	uint64_t meta_gn;	/* Meta-data generation number */
	struct ldms_version version;	/* LDMS version */
	uint8_t flags;	/* Set format flags */
	uint8_t pad1;	/* data pad */
	uint8_t pad2;	/* data pad */
	uint8_t pad3;	/* data pad */
	uint32_t card;		/* Size of dictionary */
	uint32_t meta_sz;	/* size of meta data in bytes */
	uint32_t data_sz;	/* size of metric values in bytes */
	uint32_t uid;           /* UID */
	uint32_t gid;           /* GID */
	uint32_t perm;          /* permission */
	uint32_t array_card;    /* number of sets in the set array */
	uint32_t heap_sz;	/* size of heap in bytes */
	uint32_t reserved[7];	/* area reserved for compatible core updates */
	uint32_t dict[OVIS_FLEX];/* The attr/metric dictionary */
```
( In the earlier call with you, I was mistakenly looking at the pad1, pad2, pad3. That's why I said we have only 24 more bits for the `heap_sz`).


2. The `ldms_set_hdr->data_sz` should be `actual_data_sz + heap_sz`. This is because OVIS-4 uses `ldms_set_hdr->data_sz` to calculate the location of the set data (in set array). In the case of the current set data index being non-zero, if the `data_sz` does not include `heap_sz`, the OVIS-4 client (e.g. `ldms_ls`) will calculate wrong set data address if it lookup the set from this patch. See: https://github.com/narategithub/ovis/blob/2729bab9fb554fd1d72073a85dace2c8097b4e79/ldms/src/core/ldms_private.h#L182

REMARK: Since OVIS-4 HEAD doesn't know `LDMS_V_LIST` and `LDMS_V_LIST_ENTRY`, it is obvious that the OVIS-4 client won't be able to 100% work with the new ldmsd that uses `LDMS_V_LIST`. The OVIS-4 ldmsd collecting data from the new ldmsd would probably be OK ... depending on how storage plugins handle unknown datatype. OVIS-4 `ldms_ls` that query the new ldmsd that uses `LDMS_V_LIST` will print `"Unknown metric type"` for that particular metric.

@tom95858 what do you think? Any comments / suggestions? I'll proceed with 1 and 2 and test it while waiting for your input.

tom95858:
@narategithub, yeah, that's why I was a bit confused. I thought I had made much space for future features. You're use of that 1st word for the heap size is the appropriate use of that reserved space.

tom95858:
@narategithub I don't think it is a necessary backward compatibility claim that ldms_ls from an old LDMS will work with a new LDMS. I think it's that a new aggregator will continue to work with older LDMS samplers, i.e. the aggregators can be newer, but the samplers can be older.

But I think we should socialize this claim. We discussed this at the 2019 LDMSCONN, but that was a while ago.

narategithub:
@tom95858 got it.

baallan:
Within a single series like v4, ls should keep working from old version hosts to new aggregators.
in particular, a cluster on 4.x (well the admin nodes on it hosting 4.x L1s) should be able to ls 4.y L2s at the analysis servers.  At the very least, the user of 4.x -> 4.y ls should get back a well formed error message describing the problem, not a crash.

narategithub:
@tom95858 This is ready for your review. It passed the compatibility test, the [ldms_list_test](https://github.com/narategithub/ldms-test/blob/ldms_list/ldms_list_test) in `narategithub/ldms-list` (will merge to `ovis-hpc/ldms-list` after this PR is merged), and other test cases in `ldms-list` -- except `setgroup` which is a known bug that shall be addressed in another PR.

tom95858:
> Within a single series like v4, ls should keep working from old version hosts to new aggregators.

This is specifically not a goal per the compatibility guidelines we discussed at LDMSCONN

> in particular, a cluster on 4.x (well the admin nodes on it hosting 4.x L1s) should be able to ls 4.y L2s at the analysis servers. At the very least, the user of 4.x -> 4.y ls should get back a well formed error message describing the problem, not a crash.

It should certainly not crash. We should test this actually. @narategithub, would you please add a forward "does not crash" test to the auto-tests? Specifically 4.3.3 --> pull request.

ldms/src/core/ldms.h
```diff
@@ -1336,6 +1337,36 @@ uint64_t ldms_set_meta_gn_get(ldms_set_t s);
...
+ */
+uint64_t ldms_set_heap_gn_get(ldms_set_t s);
+
+/**
+ * \brief Tell LDMS to copy previous data in the set array on transaction begin.
+ *
+ * When \c ldms_transaction_begin() is called, the set data points to the next
+ * data slot in the set array and the metric modification by the application
+ * will be applied to the new data slow. By default, the new data slot is left
+ * as-is is.
```
tom95858, line 1352: May add "This presumes that the sample() function will update all list entry data at each invocation."

ldms/src/core/ldms.h
```diff
@@ -1336,6 +1337,36 @@ uint64_t ldms_set_meta_gn_get(ldms_set_t s);
...
+ *
+ * By calling \c ldms_set_data_copy_set(s, 1), LDMS will copy the data from the
+ * previous slot into the new slot at \c ldms_transaction_begin().
+ *
+ * REMARK: For the the heap data (in which ldms_list and its elements reside),
```
tom95858, line 1357: Please remove REMARK:

ldms/src/core/ldms.h
```diff
@@ -1336,6 +1337,36 @@ uint64_t ldms_set_meta_gn_get(ldms_set_t s);
...
+ * if the heap structure has changed (i.e. `ldms_list_append()` or
+ * `ldms_list_del()` was called), the heap section will be copied over to the
+ * new slot regardless of the data copy flag to preserve the heap structure.
+ * Note that in the case of data manipulation w/o heap structure changes (no
+ * calling to `ldms_list_append()` nor `ldms_list_del()`), the data won't be
+ * copied over if the copy flag is not set to on.
+ *
+ * \param s  The \c ldms_set_t handle.
+ * \param on \c 1 for turing data copy flag on, or \c 0 for turing it off.
```
tom95858, line 1366: "turing" --> "turning"

ldms/src/core/ldms.h
```diff
@@ -1336,6 +1337,36 @@ uint64_t ldms_set_meta_gn_get(ldms_set_t s);
...
+ */
+void ldms_set_data_copy_set(ldms_set_t s, int on);
```
tom95858, line 1368: Perhaps rename the parameter 'on' to 'on_n_off'

ldms/src/core/ldms_heap.h
```diff
@@ -0,0 +1,157 @@
+/* -*- c-basic-offset: 8 -*-
+ * Copyright (c) 2019 National Technology & Engineering Solutions
+ * of Sandia, LLC (NTESS). Under the terms of Contract DE-NA0003525 with
+ * NTESS, the U.S. Government retains certain rights in this software.
+ * Copyright (c) 2013,2018 Open Grid Computing, Inc. All rights reserved.
```
tom95858, line 5: The dates should be updated here to 2021

ldms/src/core/ldms_heap.h
```diff
@@ -0,0 +1,157 @@
...
+	rrbt_t addr_tree;
+	struct rrbt_instance size_inst;
+	struct rrbt_instance addr_inst;
+} *ldms_heap_t;
+
+struct ldms_heap_base {
+	char signature[8];	/* Identifies start of heap and
+				   prevents 0 from ever being returned
+				   by an allocation */
+	uint8_t start[0];
```
tom95858, line 87: should [0] be OVIS_FLEX or is this really *start?

ldms/src/core/ldms_heap.h
```diff
@@ -0,0 +1,157 @@
...
+};
+
+/**
+ * \brief Get information about the heap configuration
+ *
+ * \param mmi	Pointer to the mm_info structure to be filled in.
```
tom95858, line 93: This parameter list is incomplete/incorrect.

ldms/src/core/ldms_heap.h
```diff
@@ -0,0 +1,157 @@
...
+ */
+void ldms_heap_get_info(ldms_heap_t h, ldms_heap_info_t i);
+
+/**
+ * \brief Initialize the heap.
+ *
+ * Initializes the heap data at teh specified base address
+ *
+ * \param size	The size of the heap in bytes.
```
tom95858, line 102: This is missing the _heap_ and _base_ parameters

ldms/src/core/ldms_heap.h
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * \brief Gets a handle to the heap at the specified base address
+ *
+ * Returns a handle to the heap at the specified address.
+ *
+ * \param base	The address where the heap is located
+ * \returns A handle to the heap or NULL if there is an error
+ */
+ldms_heap_t ldms_heap_get(ldms_heap_t h, struct ldms_heap *heap, void *base);
```
tom95858, line 115: Again the doc string doesn't match the signature. Please just fix these up everywhere.

ldms/src/core/ldms_core.h
```diff
@@ -193,15 +195,31 @@ struct ldms_set_hdr {
 	uint8_t pad3;	/* data pad */
 	uint32_t card;		/* Size of dictionary */
 	uint32_t meta_sz;	/* size of meta data in bytes */
-	uint32_t data_sz;	/* size of metric values in bytes */
+	uint32_t data_heap_sz;	/* size of metric values in bytes */
```
tom95858, line 198: If __data_sz__ includes both the static + heap data size, then I think __data_sz__ is the correct name. Otherwise, I think that __data_heap_sz__ is confusing.

ldms/src/core/ldms_core.h
```diff
@@ -193,15 +195,31 @@ struct ldms_set_hdr {
...
 	uint32_t uid;           /* UID */
 	uint32_t gid;           /* GID */
 	uint32_t perm;          /* permission */
 	uint32_t array_card;    /* number of sets in the set array */
-	uint32_t reserved[8];	/* area reserved for compatible core updates */
```
tom95858, line 201: Did you determine that you didn't need to consume any of the reserved space? If so, please remove this field.

tom95858 (changes requested):

ldms/src/core/ldms_core.h
narategithub: It is already removed. The `red` part is the removed stuff.

ldms/src/core/ldms_heap.h
narategithub: It should be OVIS_FLEX. I'm addressing it.

ldms/src/core/ldms_heap.h
narategithub: Got it. Thanks for catching these.

ldms/src/sampler/examples/list_sampler/list_sampler.c
narategithub: no need. I did this to hush the warning from the autocompletion tool I used with Vim and forgot to change back. Changing it back now.

narategithub:
> 
> 
> > Within a single series like v4, ls should keep working from old version hosts to new aggregators.
> 
> This is specifically not a goal per the compatibility guidelines we discussed at LDMSCONN
> 
> > in particular, a cluster on 4.x (well the admin nodes on it hosting 4.x L1s) should be able to ls 4.y L2s at the analysis servers. At the very least, the user of 4.x -> 4.y ls should get back a well formed error message describing the problem, not a crash.
> 
> It should certainly not crash. We should test this actually. @narategithub, would you please add a forward "does not crash" test to the auto-tests? Specifically 4.3.3 --> pull request.

Sure. I'm adding it with this PR.

narategithub:
@tom95858 ready for you to re-review.

ldms/src/core/ldms.h
```diff
@@ -1882,6 +1937,35 @@ int32_t ldms_metric_array_get_s32(ldms_set_t s, int id, int idx);
...
+ * \brief Append a new value to a list
+ *
+ * Append a new value entry to a list metric.
+ *
+ * \param s	The set handle
+ * \param i	The metric index
+ * \param typ	The value type.
+ * \param count	The element count if the type is an array.
+ */
+ldms_mval_t ldms_list_append(ldms_set_t s, ldms_mval_t l, enum ldms_value_type typ, size_t count);
```
tom95858, line 1951: parameter names don't match the doc string

ldms/src/core/ldms.h
```diff
@@ -1336,6 +1337,37 @@ uint64_t ldms_set_meta_gn_get(ldms_set_t s);
  * \returns	The 64bit data generation number.
  */
 uint64_t ldms_set_data_gn_get(ldms_set_t s);
+
+/**
+ * \brief Get the heap generation number.
+ */
+uint64_t ldms_set_heap_gn_get(ldms_set_t s);
```
tom95858, line 1344: Missing parameter specification.

ldms/src/core/ldms.h
```diff
@@ -1882,6 +1937,35 @@ int32_t ldms_metric_array_get_s32(ldms_set_t s, int id, int idx);
...
+ */
+ldms_mval_t ldms_list_append(ldms_set_t s, ldms_mval_t l, enum ldms_value_type typ, size_t count);
+ldms_mval_t ldms_list_first(ldms_set_t s, ldms_mval_t l, enum ldms_value_type *typ, size_t *count);
+ldms_mval_t ldms_list_next(ldms_set_t s, ldms_mval_t v, enum ldms_value_type *typ, size_t *count);
+size_t ldms_list_len(ldms_set_t s, ldms_mval_t l);
+
+/**
+ * \brief Remove \c v from the list \c list.
+ *
+ * The handle \c v is not valid after the call. If \c v is an \c LDMS_V_LIST,
```
tom95858, line 1959: More doc string issues...

ldms/src/core/ldms.h
```diff
@@ -1882,6 +1937,35 @@ int32_t ldms_metric_array_get_s32(ldms_set_t s, int id, int idx);
...
+ * the members of \c v will be recursively deleted.
+ *
+ */
+int ldms_list_del(ldms_set_t s, ldms_mval_t lh, ldms_mval_t v);
+
+/**
+ * Purge all elements from the list \c lh.
```
tom95858, line 1966: and here too

tom95858 (changes requested):
Other than the doc stuff, I think it's ready to merge...thanks @narategithub

narategithub:
> 
> 
> Other than the doc stuff, I think it's ready to merge...thanks @narategithub

@tom95858 oops. sorry about that. I'd just pushed the changes to those doc strings. Could you please take a look once more?

PR #790, opened 2021-09-03
store_csv: Add time_format option

morrone:
The time format used by LDMS in CSV files can be a bit difficult to use in practice. A great many existing tools use milliseconds-since-epoch as an internal or natively understood time format (Kafka, Cassandra, Elasticsearch, etc.). This PR introduces a store_csv option that keeps the current time format as the default, but allows the use of and alternate time format (time_format=1), where:

A value of 1 chooses an alternate format where the first field (Time_msec) is <milliseconds since epoch>, and the
second field (Time_usec) is just the additional <microseconds> since the epoch in excess of the milliseconds since
epoch. In other words, there is no overlap of the values in the first and seconds fields, which is in contrast to the
repetition employed by format 0.

### DIFF
--- a/ldms/man/Plugin_store_csv.man
+++ b/ldms/man/Plugin_store_csv.man
@@ -42,7 +42,7 @@ take on the value configured in the default config line or opt_file.
 .SH STORE_CSV CONFIGURATION ATTRIBUTE SYNTAX
 .TP
 .BR config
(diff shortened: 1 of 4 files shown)


baallan:
I think this feature would be better put into a set transform plugin so that all stores could see the change in data, but I don't have a strong objection to it being added to store_csv this way. Wouldn't the better solution for Kafka, Cassandra, Elasticsearch be to have their own store plugins?

morrone:
> I think this feature would be better put into a set transform plugin so that all stores could see the change in data, but I don't have a strong objection to it being added to store_csv this way. Wouldn't the better solution for Kafka, Cassandra, Elasticsearch be to have their own store plugins?

I'm not transforming any data in the metric set, so I don't think transforming sets is the way to go. I am just changing how the store_csv plugin chooses to represent time.

Yes, a Kafka store would certainly address our needs (because that in turn feeds the database(s)). A Kafka store is certainly on the roadmap, but could be several months from fruition.

Further, the changes that we need to make to our infrastructure to allow direct data feeds to Kafka (rather than our current all-CSV-files-all-the-time approach).  This new time_format option lets us get around the time problem today.

tom95858:
@oceandlr, @valleydlr is this change ok with you?

valleydlr:
+1

PR #796, opened 2021-09-10
Fix auth_munge compatiblity issue over rdma xprt

narategithub:
auth_munge uses local address obtained from zap as munge data payload.
The peer then unmunge the payload and compare with its remote address
(which is our local address). If the payload matches the expected
address, the connection is authenticated.

zap_rdma had a bug that mixed up remote/local addresses, and the bug
stayed in the tree for quite some time (since v4.3.3). After v4.3.7,
commit 891da2b addressed the local/remote address mixed-up bug, but
this introduced auth_munge compatibility issue over rdma xprt because
v4.3.7 expected the mixed-up remote address, but 891da2b would send
the correct one, causing auth_munge to invalidate the connection.

This patch is a work-around in auth_munge to use the mixed-up
addresses for the rdma transport.

### DIFF
--- a/ldms/src/auth/ldms_auth_munge.c
+++ b/ldms/src/auth/ldms_auth_munge.c
@@ -174,7 +174,16 @@ int __auth_munge_xprt_begin(ldms_auth_t auth, ldms_t xprt)
 				(void*)&a->rsin, &a->sin_len);
 	if (rc)
 		return rc;
-	merr = munge_encode(&a->local_cred, a->mctx, &a->lsin, a->sin_len);
+	/*
+	 * zap_rdma from OVIS-4.3.7 and earlier has a bug that swaps
+	 * local/remote addresses. Since the old peers expect to receive the
+	 * swapped address, in order to be compatible with them, we have to send
+	 * the swapped address in the case of rdma transport.
+	 */
+	merr = (strncmp(xprt->name, "rdma", 4) == 0)?
+		munge_encode(&a->local_cred, a->mctx, &a->rsin, a->sin_len):
+		munge_encode(&a->local_cred, a->mctx, &a->lsin, a->sin_len);
+
 	if (merr)
 		return EBADR; /* bad request */
 	len = strlen(a->local_cred);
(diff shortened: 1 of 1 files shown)


narategithub:
This patch has been tested with the following test scenario:

- [x] v4.3.7 sampler --rdma/munge--> 796 ldms_ls
- [x] v4.3.7 sampler --rdma/none--> 796 ldms_ls
- [x] v4.3.7 sampler --sock/munge--> 796 ldms_ls
- [x] v4.3.7 sampler --sock/none--> 796 ldms_ls

- [x] 796 sampler --rdma/munge--> 796 ldms_ls
- [x] 796 sampler --rdma/none--> 796 ldms_ls
- [x] 796 sampler --sock/munge--> 796 ldms_ls
- [x] 796 sampler --sock/none--> 796 ldms_ls

- [x] 796 sampler --rdma/munge--> v4.3.7 ldms_ls
- [x] 796 sampler --rdma/none--> v4.3.7 ldms_ls
- [x] 796 sampler --sock/munge--> v4.3.7ldms_ls
- [x] 796 sampler --sock/none--> v4.3.7ldms_ls

ldms/src/auth/ldms_auth_munge.c
```diff
@@ -174,7 +174,11 @@ int __auth_munge_xprt_begin(ldms_auth_t auth, ldms_t xprt)
 				(void*)&a->rsin, &a->sin_len);
 	if (rc)
 		return rc;
-	merr = munge_encode(&a->local_cred, a->mctx, &a->lsin, a->sin_len);
+	/* zap_rdma compatibility work-around */
```
morrone, line 177: This comment should probably detail what we are working around a little more. I would also like to know what our approach should be to eliminating this work around in future releases, so we don't need to keep this around forever.

tom95858: @narategithub please add a bit more detail about how the original RDMA provider swapped the local/remote addresses and these addresses are passed on the wire as part of munge authentication. This comment doesn't say what we're 'working around.'

narategithub: done.

tom95858:
@narategithub There's a reason why we _fixed_ this. In other words, there was some other issue that this swapping caused. Do you remember what that was? Because swapping these back is going to reintroduce that issue.

PR #800, opened 2021-09-20
correct use of uninitialized len value in xprt.

baallan:


ldms/src/core/ldms_xprt.c
```diff
@@ -2747,7 +2747,7 @@ static void ldms_zap_cb(zap_ep_t zep, zap_event_t ev)
 		__sync_fetch_and_add(&xprt_connect_count, 1);
 		/* initialize/cache addr since connected */
 		struct sockaddr l, r;
-		socklen_t len;
+		socklen_t len = 0;
```
tom95858, line 2750: I think this should be sizeof () or something like that. I think the ldms_xprt_sockaddr() function is incorrect. The logic in ldms_xprt_sockaddr() is wrong I think. It checks for len == 0 and then if it is sets the length of struct sockaddr which is incorrect. This will break for IPv6.

baallan: @narategithub @nichamon please have a look at this initialization and the ZAP_EVENT_CONNECTED in the callback and tell us what you meant to happen. Based on the comment internal to ldms_xprt_sockaddr I proposed to init len here to 0.

tom95858: This is fixed by #804

PR #802, opened 2021-09-21
Fix __process_dir_set_info() handling wrong key names

nichamon:


nichamon:
This is the bug I found in the process_dir_reply() path that causes the auto-update to fail. With the patch, the auto-update works. I'm writing a test for the auto-update mode.

@narategithub The patch does not fix the bug in the set_group logic, though.

nichamon:
> This is the bug I found in the process_dir_reply() path that causes the auto-update to fail. With the patch, the auto-update works. I'm writing a test for the auto-update mode.
> 
> @narategithub The patch does not fix the bug in the set_group logic, though.

I found another bug in the same path. I have updated the patch. With the patch, OVIS-4 passes the `setgroup_test` in the `ldms-test` repo.

PR #803, opened 2021-09-21
Fix ldms update never completed bug

narategithub:
When the ldms consumer (e.g. ldms_ls or agg) updated the looked-up set
before the producer (e.g. sampler) sampled a new data since previous
update, the consumer side won't get update complete callback at all.
What happened was that the zap_read completion was delivered but the
logic in ldms xprt detected that the data was old and didn't call update
complete. This patch modified the code to call the update callback in
the case of receiving the old data.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -2203,15 +2203,23 @@ static void __handle_update_data(ldms_t x, struct ldms_context *ctxt,
 		goto cleanup;
 	}
 	n = __le32_to_cpu(set->meta->array_card);
-	/* update current index from the update */
+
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 This passed all existing test cases in [ldms-test](https://github.com/ovis-hpc/ldms-test), including `setgroup_test` (fixed by @nichamon #802) and the new test for this corner case in my tree [here](https://github.com/narategithub/ldms-test/blob/set_array_hang/set_array_hang_test). I'll merge the test case into `ldms-test/master` when this PR is merged.

narategithub:
This broke `ldms_list_test` .. I forgot to add `ldms_list_test` into `test-all.sh`.  I'll re-open this after it really passed all the tests.

PR #804, opened 2021-09-22
Fix ldms_xprt_sockaddr() calls to not assume ipv4 addresses

narategithub:
Also add compatibility test with munge authentication in github actions since the patch touches `ldms_auth_munge.c`

### DIFF
--- a/.github/workflows/4.3.3-compat-test.yaml
+++ b/.github/workflows/4.3.3-compat-test.yaml
@@ -42,7 +42,7 @@ jobs:
         set -e
         mkdir -p build
         pushd build
-        ../configure --prefix=/opt/ovis-4 --enable-etc
+        ../configure --prefix=/opt/ovis-4 --enable-etc --enable-munge
         make
         make install
     - name: test preparation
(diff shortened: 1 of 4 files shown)


narategithub:
@tom95858 I think `ldms_xprt_sockaddr()` itself works correctly. If the caller provide smaller output buffer, it returns `E2BIG` to let the caller know that the provided buffer size is not enough, and `*sa_len` is set to the required size. As for the caching ... I have a vague memory that I put it in because the debugging code keep calling `ldms_xprt_sockaddr()` and it was so slow when calling `zap_get_name()` repeatedly, and caching helped.

https://github.com/ovis-hpc/ovis/blob/24da2a3bccb2032fe9379abc512ff98411c45205/ldms/src/core/ldms_xprt.c#L3808-L3836

tom95858:
This is missing the fix to ldms_xprt_sockaddr that checks for a 0 length and sets the length to sizeof(sockaddr) if so.

narategithub:
> 
> 
> This is missing the fix to ldms_xprt_sockaddr that checks for a 0 length and sets the length to sizeof(sockaddr) if so.

I think `ldms_xprt_sockaddr()` already covered this case. Let's assume that `*sa_len` is 0. If this is the first call, `x->sa_len` is also 0. Then, the following code will get and cache the local and remote addresses with the length.

https://github.com/ovis-hpc/ovis/blob/24da2a3bccb2032fe9379abc512ff98411c45205/ldms/src/core/ldms_xprt.c#L3815-L3825

If this call is not the first call, the local and remote addresses with length are already cached.

Then, because `*sa_len` is 0, which will always be less than the `x->sa_len` that has been cached, the following code will be executed and the function returned with `E2BIG`.

https://github.com/ovis-hpc/ovis/blob/24da2a3bccb2032fe9379abc512ff98411c45205/ldms/src/core/ldms_xprt.c#L3827-L3830

😃

tom95858:
This "optimization" presumes that every call to this function provides the same input buffer and size. That is a bug. If we need to cache the address we should do so in the Zap transport, not in ldms_xprt_sockaddr(). I am curious as to what performance problem you are seeing and on which transport(s).

narategithub:
TL; DR: I'm removing the caching + running the test in ldms-test.

> 
> 
> This "optimization" presumes that every call to this function provides the same input buffer and size. That is a bug. 
It doesn't care about input buffer size as long as it fits output size. It behaves the same as `getsockname(2)` except for that `ldms_xprt_sockaddr()` doesn't copy addresses at all if the size is not big enough. The cached addresses are `struct sockaddr_storage` which is 128 bytes. If we have a zap transport that will return address bigger than 128 bytes, then we have a bug in this case.

>If we need to cache the address we should do so in the Zap transport, not in ldms_xprt_sockaddr(). I am curious as to what performance problem you are seeing and on which transport(s).
OK. I'll remove the caching from ldms xprt. Regarding the performance issue, I think my memory got mixed up with something else. Weeding through gitlog I found 1405c161b57bd4d974ef9762f1136258066b4964 that introduce address caching in ldms_xprt. The following is a snippet from the commit message:

```
        [x] ldms_xprt_sockaddr() caching: cache the remote/local addresses
            on the success zap_get_name() so that the remote/local addresses
            are still available even after the endpoint became disconnected
            (but not yet destroyed), which is required for the republish
            client key construction. The caching is valid because the
            local/remote addresses are not changed after the endpoint
            becomes connected.
```

Someone must be calling `ldms_xprt_sockaddr()` after the endpoint became disconnected but still wanted the addresses when it was connected. Anyway, in our current OVIS-4 head (b0b8b7fb), there are only two places that call `ldms_xprt_sockaddr()`: in `ldms_auth_munge.c` and the call to initial the cache after becoming connected in `ldms_xprt.c`. Both places are not the case that the quoted commit message referred to.

narategithub:
@tom95858 the caching is now removed.

narategithub:
also passed all container-based test cases in `ldms-test`.

tom95858:
> TL; DR: I'm removing the caching + running the test in ldms-test.
> 
> > This "optimization" presumes that every call to this function provides the same input buffer and size. That is a bug.
> 
> It doesn't care about input buffer size as long as it fits output size. It behaves the same as `getsockname(2)` except for that `ldms_xprt_sockaddr()` doesn't copy addresses at all if the size is not big enough. 

I'm failing to make the point. There is no such thing as 'input' vs. 'output' buffer, it's the same buffer. The __caller__ sets the buffer size because it is the only one that knows what was provided. 

> ... The cached addresses are `struct sockaddr_storage` which is 128 bytes. If we have a zap transport that will return address bigger than 128 bytes, then we have a bug in this case.
> 

It doesn't matter how big the cache buffer is. You're stuffing it into the buffer the caller provided regardless; the size of which was set to zero when this caching logic was introduced.

> > If we need to cache the address we should do so in the Zap transport, not in ldms_xprt_sockaddr(). I am curious as to what performance problem you are seeing and on which transport(s).
> 
> OK. I'll remove the caching from ldms xprt. Regarding the performance issue, I think my memory got mixed up with something else. Weeding through gitlog I found [1405c16](https://github.com/ovis-hpc/ovis/commit/1405c161b57bd4d974ef9762f1136258066b4964) that introduce address caching in ldms_xprt. The following is a snippet from the commit message:
> 
> ```
>         [x] ldms_xprt_sockaddr() caching: cache the remote/local addresses
>             on the success zap_get_name() so that the remote/local addresses
>             are still available even after the endpoint became disconnected
>             (but not yet destroyed), which is required for the republish
>             client key construction. The caching is valid because the
>             local/remote addresses are not changed after the endpoint
>             becomes connected.
> ```
> 
> Someone must be calling `ldms_xprt_sockaddr()` after the endpoint became disconnected but still wanted the addresses when it was connected. Anyway, in our current OVIS-4 head ([b0b8b7f](https://github.com/ovis-hpc/ovis/commit/b0b8b7fb10c37f7e9cfa8dfc9212cd47db8c4609)), there are only two places that call `ldms_xprt_sockaddr()`: in `ldms_auth_munge.c` and the call to initial the cache after becoming connected in `ldms_xprt.c`. Both places are not the case that the quoted commit message referred to.

I'm not following you here. You say you're removing the caching logic and then arguing that it's required by the streams logic.

I would like the interface to behave reliably and consistently. Right now it assumes that the caller provided struct sockaddr_storage *. If this is the requirement of the interface, then change the parameter types to struct sockaddr_storage * so that the function declaration makes this obvious. The compiler will then check that this condition is met. Then you can leave the "caching" logic in there.

In any case the check for size needs to be removed because it's set by the ldms_xprt_sockaddr function internally and does not represent the size of the buffer provided by the caller.

narategithub:
@tom95858 

Sorry for the confusion, here's the status of this PR:
- Caching has been removed. I agree with you that if it is requited, it should be done by the underlying transport.
- I think right now we don't need caching. There is currently only 1 place that calls `ldms_xprt_sockaddr()`, which is in `ldms_auth_munge.c`, which previously assumed ipv4 `struct sockaddr_in` address buffers. It is changed to `struct sockaddr_storage` to accommodate other kinds of addresses as well. According to `socket.h`, `struct sockaddr_stroage` is "Structure large enough to hold any socket address (with the historical exception of AF_UNIX)."
- I think the caching removal is safe. It also passed all of the container-based test cases in `ldms-test` and the compatibility test with munge introduced in this pull request.

My opinion regarding `struct sockaddr *` vs `struct sockaddr_storage *`, I think we should keep `struct sockaddr *` to mimic `getsockname(2)`. 

-------------
P.S. The caching `x->sa_len` was set to `sizeof( struct x->local_sa )` before calling `zap_get_name()`. `zap_get_name()` never received buffer lenght 0 (@nichamon fixed it in a916c4f3ef71f6c8fe19fa33d8909a7ff80c465d).

PR #806, opened 2021-09-23
Fix ldms update never completed bug

narategithub:
When the ldms consumer (e.g. ldms_ls or agg) updated the looked-up set
before the producer (e.g. sampler) sampled a new data since previous
update, the consumer side won't get update complete callback at all.
What happened was that the zap_read completion was delivered but the
logic in ldms xprt detected that the data was old and didn't call update
complete. This patch modified the code to call the update callback in
the case of receiving the old data.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -2203,15 +2203,24 @@ static void __handle_update_data(ldms_t x, struct ldms_context *ctxt,
 		goto cleanup;
 	}
 	n = __le32_to_cpu(set->meta->array_card);
-	/* update current index from the update */
+
 	data = __ldms_set_array_get(set, ctxt->update.idx_from);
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 This time it passed all of the container-based test cases in `ldms-test`, including the new one `set_array_hang_test` that specifically covering the issue this PR is solving. I'll merge `set_array_hang_test` into `ldms-test/master` after this PR is merged.

You can see the test results here: https://gist.github.com/narategithub/8b2d35cb7a74b2a9678a367da0220c6e#file-test-all-log

narategithub:
p.s. `github` does not allow me to re-open #803 due to a forced push.

tom95858:
@narategithub is this good to go?

narategithub:
@tom95858 I believe it is.

PR #807, opened 2021-09-23
Fix ldms_xprt_put() and ldms_xprt_get()

nichamon:


nichamon:
While I was investigating the cause of a deadlock on L1 in a 2-level aggregation test with flapped sampler daemons, I found that the `ldms_xprt_put()` always takes the `xprt_list_lock`, even though the function does not remove the endpoint from the `xprt_list`. This was a part of a deadlock between a thread processing a SET_DELETE reply from L2 and another thread sending a SET_DELETE request to L2.

Thread processing a SET_DELETE reply from L2
- `process_set_delete_reply()`
- `pthread_mutex_lock(&x->lock)`
- `__ldms_free_ctxt()`
- `ldms_xprt_put(x)`
- `pthread_mutex_lock(&xprt_list_lock)` regardless of ref_count

Thread sending a SET_DELETE request to L2
- `ldms_set_delete()`
- `__ldms_dir_del_set()`
- `pthread_mutex_lock(&xprt_list_lock)` to iterate send the SET_DELETE request to any clients that have sent a dir_request.
- `ldms_xprt_set_delete()`
- `pthread_mutex_lock(&x->lock)` to create a SET_DELETE request context

In this patch, I changed `ldms_xprt_put()` so that it would take the `xprt_list_lock` only when removing the endpoint from the `xprt_list`. I was the one who made `ldms_xprt_put()` to always take the `xprt_list_lock`. I think it was the wrong fix to the problem that `ldms_xprt` took a reference after it has reached 0. If this happens, it means there is a bug somewhere.  The change in `ldms_xprt_get()` makes sure that we would know it happens.

PR #809, opened 2021-09-27
Fix SIGSEGV in zap_rdma __ep_flush() path

narategithub:
`__ep_flush()` may be called before cq_channel is created. Hence, it
shall check if `cq_channel` is not NULL before trying to reap the cq
event from the channel in the flush path. In addition, `__ep_flush()`
should reap the event from the channel directly instead of calling
`z_rdma_handle_cq_event()` which will re-arm the channel that we are
flushing.

### DIFF
--- a/lib/src/zap/rdma/zap_rdma.c
+++ b/lib/src/zap/rdma/zap_rdma.c
@@ -404,12 +404,19 @@ static int __ep_flush(struct z_rdma_ep *rep)
 	 * endpoint is in a bad state (rejected, connection errors, or
 	 * disconnected). */
 	int ret;
+	struct ibv_cq *ev_cq;
+	void *ev_ctx;
 
 	/* disable cq events */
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 

The change passed the end-to-end test https://github.com/ovis-hpc/ldms-test/tree/master/scalability with moderate scale setting:
40 samplers over 4 hosts (10 samplers/host) ==> 4 L1 on 2 hosts (2/host) ==> 2 L2 on 1 host ==> 1 L3 w/ SOS.

I also manually tested `ldms_ls -x rdma -p 10000 -h cygnus-01-iw` but with no `ldmsd` running on cygnus-01. `ldms_ls` yielded `Connection failed/rejected.` as expected. FYI, it did have segmentation fault before this PR.

narategithub:
Forgot to mention that I'm adding a script in `ldms-test` to cover this.

PR #810, opened 2021-09-28
Get rid of markdown characters in the github action scripts

nichamon:


tom95858:
What is the reason for this change?

narategithub:
@tom95858

In Markdown, the text between backticks "\`" are formatted as fixed width (e.g. `term`). I accidentally put them into the string in the test script to emphasize some terms. However, backticks has a special meaning in `bash`. It is equivalent to `$()`, which takes the text between backticks or parentheses and execute it as a command. For example,
```sh
# These two are equivalent
$ echo "I am `whoami`"
$ echo "I am $(whoami)"
```

As such, the text meant to be emphasized in the error message are evaluated as a command, which is a bug. We have not seen this before because the tests passed and the error string was not evaluated. @nichamon found this out because one of the test script failed and the error message got evaluated.

PR #821, opened 2021-10-14
add man page for ldmsd_stream_publish and -r/-i options

baallan:
Options -r count -i delay allow the publisher to repeat file input
to the publishing target count times with a delay between repeats.


baallan:
@tom95858 this adds the stream publish iterator feature used to drive tests that turn up the deadlock in #777.
This will probably need rebasing after #777 is merged, since it also adds the man page.

PR #823, opened 2021-10-19
fix error messages for truncate and listen functions in ldmsd.c

baallan:


baallan:
@tom95858 please review this.

PR #824, opened 2021-10-19
fix multiple -x handling and listener setup messages on ldmsd

baallan:
To make -x options handle multiple listeners independently, the optional fields in a listener specification are extended to:
XPRT:PORT[:[host]:[AUTH[:authopts]]]

Without this patch, all -x defined listeners are tied to the -a (default_auth) configuration.
Where the AUTH field is omitted, the default auth method (as possibly redefined with the -a argument) will be used.

Implementation: where AUTH is the name of an authentication plugin specified explicitly with -x, it will cause the equivalent of
```
auth_add name=AUTH plugin=AUTH 
```
with authopts also included if provided.

This enables listeners to be defined in the systemd service file (or elsewhere) such as:
```
ldmsd -x rdma:411:host-ib0:ovis:conf=/file -x sock:412::munge
```
so that a daemon can be started and listening without any input file needed to define listeners and authentications.

### DIFF
--- a/ldms/man/ldmsd.man
+++ b/ldms/man/ldmsd.man
@@ -129,13 +129,15 @@ Display the usage for named plugin. Special names all, sampler, and store match 
 .SS
 Communication Options:
 .TP
-.BI -x " XPRT:PORT:HOST"
+.BI -x " XPRT:PORT[:HOST:AUTH:OPTS]"
 .br
-Specifies the transport type to listen on. May be specified more than once for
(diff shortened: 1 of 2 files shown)


baallan:
@tom95858 after #825 is merged this would pass all the checks.
if both 825 and 826 are merged, I will  need to refactor this slightly to account for 826.

baallan:
creating a new version of this shortly that handles ipv6 syntax.

PR #826, opened 2021-10-22
Support LDMSD cmd-line options in configuration files

nichamon:
To specify a cmd-line option in a configuration file, use the command
'option'. For example,

option l=/var/log/ldmsd.log v=INFO
option m=2gb

A configuration file example is at ldms/examples/cmdline-opts/ldmsd_w_cmdline_opts.conf.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -1506,18 +1506,18 @@ class LdmsdCmdParser(cmd.Cmd):
         Specify the command-line options
 
         Parameters:
-            [defulat_auth_args=]    (-A) Arguments of the default authentication method
-            [default_auth=]         (-a) Default authentication method
-            [banner=]               (-B) Banner mode: 0=do banner file, 1=auto-delete at exit, 2=keep the file
-            [host_name=]            (-H) Host/producer name used by the kernel metric sets
-            [publish_kernel=]       (-k) Publish kernel: true or false
-            [log_file=]             (-l) Path to the log file or 'syslog'
-            [set_memory=]           (-m) Set memory, e.g., 2GB, 512mB
-            [daemon_name=]          (-n) Daemon name
-            [worker_threads=]       (-P) Number of worker threads
-            [pid_file=]             (-r) Path to the PID file
-            [kernel_set_path=]      (-s) Path to a text file containing the kernel metric values
-            [log_level=]            (-v) Log level: DEBUG, INFO, ERROR, CRITICAL, QUIET
+            -A, -- defulat_auth_args    Arguments of the default authentication method
+            -a, -- default_auth         Default authentication method
+            -B, --banner                Banner mode: 0=do banner file, 1=auto-delete at exit, 2=keep the file
+            -H, --host_name             Host/producer name used by the kernel metric sets
+            -k, --publish_kernel        Publish kernel: true or false
+            -l, --log_file              Path to the log file or 'syslog'
+            -m, --set_memory            Set memory, e.g., 2GB, 512mB
+            -n, --daemon_name           Daemon name
+            -P, --worker_threads        Number of worker threads
+            -r, --pid_file              Path to the PID file
+            -s, --kernel_set_path       Path to a text file containing the kernel metric values
+            -v, --log_level             Log level: DEBUG, INFO, ERROR, CRITICAL, QUIET
 
         Use 'listen' instead of the -x option.
 
@@ -1526,6 +1526,17 @@ class LdmsdCmdParser(cmd.Cmd):
             -F        Foreground mode
             -u        List named plugins
             -V        Print LDMS version
+
+        You would specify the command-line options as if you specify them at
+        the command line.
+
+        Examples:
+           # Specify a short option
+           option -a ovis
+           # Specify multiple short options
+           option -m 2GB -P 16
+           # Specify multiple long options
+           option --log_file /path/to/logfile --log_level ERROR
         """
         print("Not supported! The 'option' command is only supported in a configuration file.")
 
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -1500,6 +1500,35 @@ class LdmsdCmdParser(cmd.Cmd):
     def complete_auth_add(self, text, line, begidx, endidx):
         return self.__complete_attr_list('auth_add', text)
 
+    def do_option(self, arg):
+        """
+        ONLY SUPPORTED IN CONFIGURATION FILES
+        Specify the command-line options
+
+        Parameters:
+            [defulat_auth_args=]    (-A) Arguments of the default authentication method
+            [default_auth=]         (-a) Default authentication method
+            [banner=]               (-B) Banner mode: 0=do banner file, 1=auto-delete at exit, 2=keep the file
+            [host_name=]            (-H) Host/producer name used by the kernel metric sets
+            [publish_kernel=]       (-k) Publish kernel: true or false
+            [log_file=]             (-l) Path to the log file or 'syslog'
+            [set_memory=]           (-m) Set memory, e.g., 2GB, 512mB
+            [daemon_name=]          (-n) Daemon name
+            [worker_threads=]       (-P) Number of worker threads
+            [pid_file=]             (-r) Path to the PID file
+            [kernel_set_path=]      (-s) Path to a text file containing the kernel metric values
+            [log_level=]            (-v) Log level: DEBUG, INFO, ERROR, CRITICAL, QUIET
+
+        Use 'listen' instead of the -x option.
+
+        The following options must specified only at the command line.
+            -c        Path to a configuration file
+            -F        Foreground mode
+            -u        List named plugins
+            -V        Print LDMS version
+        """
+        print("Not supported! The 'option' command is only supported in a configuration file.")
+
     def parseline(self, line):
         """Parse the line into a command name and a string containing
         the arguments.  Returns a tuple containing (command, args, line).
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -59,6 +59,8 @@
 #include <sys/socket.h>
 #include <coll/rbt.h>
 #include <pthread.h>
+#include <unistd.h>
+#include <getopt.h>
 #include <pwd.h>
 #include <grp.h>
 #include <ovis_util/util.h>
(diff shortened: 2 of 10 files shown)


tom95858:
> This is a bizarre feature. What is the use-case?

This feature allows for the configuration of listeners, log file locations, log file levels, etc... from a configuration file, and by extension ldmsd_controller, and maestro. This feature was requested by Sandia in 2019/2020. It has been back-ported from master to OVIS-4 as part of our efforts to retain backward compatibility while continuing to incorporate new features.

WRT to "bizarre", please refrain from demeaning language directed at the work product of individuals posting pull requests.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6680,3 +6699,131 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+		['m'] = { "set_memory"            },
+		['n'] = { "daemon_name"    },
+		['P'] = { "worker_threads"    },
+		['r'] = { "pid_file"        },
+		['s'] = { "kernel_file"    },
+		['v'] = { "log_level"       },
+};
+
+extern int ldmsd_process_cmd_line_arg(char opt, char *value);
+static int cmd_line_arg_set_handler(ldmsd_req_ctxt_t reqc)
```
tom95858, line 6722: @nichamon if we're going to have this option syntax, what do you think about converting the input string to an argv and an argc and then using getopt to parse it? This will remove all that strcmp code you have in there and allow you to use the getopts #define and getopt_long structure definition you already have.

nichamon: @tom I like the idea to provide the long options not only in configuration files but also at the command line. Converting a string to argc and argv should be trivial and doesn't need any complicated parsing, so I'll make the change.

tom95858:
Hi @nichamon is this ready to go?

nichamon:
> Hi @nichamon is this ready to go?

@tom95858 Yes, it is ready.

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -1500,6 +1500,35 @@ class LdmsdCmdParser(cmd.Cmd):
     def complete_auth_add(self, text, line, begidx, endidx):
         return self.__complete_attr_list('auth_add', text)
 
+    def do_option(self, arg):
+        """
+        ONLY SUPPORTED IN CONFIGURATION FILES
+        Specify the command-line options
+
+        Parameters:
+            [defulat_auth_args=]    (-A) Arguments of the default authentication method
```
morrone, line 1509: Typo: "defulat"

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -1500,6 +1500,35 @@ class LdmsdCmdParser(cmd.Cmd):
     def complete_auth_add(self, text, line, begidx, endidx):
         return self.__complete_attr_list('auth_add', text)
 
+    def do_option(self, arg):
+        """
+        ONLY SUPPORTED IN CONFIGURATION FILES
+        Specify the command-line options
+
+        Parameters:
```
morrone, line 1508: Should we be documenting all of these options in a place where they are not supported? What is the likelihood that we will remember to to keep this comment up to date when any code changes don't have any effect on this location?

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6680,3 +6701,89 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+	return 0;
+}
+
+static int cmd_line_arg_set_handler(ldmsd_req_ctxt_t reqc)
+{
+	char *s, *token, *ptr1, *lval;
+	int rc = 0;
+	int argc = 1;
+	int i;
+	char *argv[11];
```
morrone, line 6752: What is the rationale for hardcoding this as "11"? Probably needs a code comment at the very least.

nichamon:
@tom95858 I changed the option command so that it takes the options the same format as the command line.

For example, option -l /path/to/logfile -v ERROR`

tom95858:
@nichamon, did you address @morrone concern about the hard coding of the array length?

nichamon:
> @nichamon, did you address @morrone concern about the hard coding of the array length?
> 

I put a comment explaining why I picked 11, which is the total number of cmd-line options supported in config files.

I also rebased the code and resolved the conflicts.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+
+static int cmd_line_arg_set_handler(ldmsd_req_ctxt_t reqc)
+{
+	char *s, *token, *ptr1;
+	int rc = 0;
+	int argc = 1;
+	int max_opt = 11;
+	char *argv[11]; /* 11 is the total number of supported cmd-line options in config files */
+
+	s = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_STRING);
```
baallan, line 6816: doesn't s need to be freed at some point?
why not dynamically allocate argv and realloc if it proves to small? this code is far away from the code driving the size needed for argv here.

morrone:
> > @nichamon, did you address @morrone concern about the hard coding of the array length?
> 
> I put a comment explaining why I picked 11, which is the total number of cmd-line options supported in config files.

I am not sure that the comment resolved things. I am a bit confused about a few things in the __cmdline_options function. I'll comment inline.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+	}
+	return 0;
+}
+
+static int cmd_line_arg_set_handler(ldmsd_req_ctxt_t reqc)
+{
+	char *s, *token, *ptr1;
+	int rc = 0;
+	int argc = 1;
+	int max_opt = 11;
```
morrone, line 6813: First of all, this should be marked "const", or be a #define.

By my count, there are 17 short opts. Even subtracting the 4 unallowed short options mentioned in __cmdline_options, that still leaves 13 "supported" options.

Next, this type of hard-coded value has a tendency to cause problems in the future. At the very least, you would also need to add a comment near the definition of short_opts and long_opts in ldmsd.c saying something like "If you add any command line options here, you will need to increment the hard-coded max_opt in ldmsd_request.c:cmd_line_arg_set_handler()".

Better yet would probably be to just write a function that walks both short_opts and long_opts, and returns the total number of options.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+	char *argv[11]; /* 11 is the total number of supported cmd-line options in config files */
```
morrone, line 6814: Using the same number in the array length as "max_opt" doesn't seem correct. At the very least, we need to double the max_opt to leave room for the argument values, correct? And then this function needs another slot for "option" in argv[0]. And we need another at the end for a "NULL". So by my count, we need this to be (max_opt*2)+2.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+
+	s = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_STRING);
+	opterr = 0;
+
+	argv[0] = "option";
+	for (token = strtok_r(s, " \t\n", &ptr1); token;
+			token = strtok_r(NULL, " \t\n", &ptr1)) {
+		if (argc > max_opt - 2) {
```
morrone, line 6822: argv should be big enough to include "option" at the beginning and NULL at the end. Then this can just be "argc >= max_opt". And when that happens, we can just return a "too many options" error.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+			rc = __cmdline_options(reqc, argc, argv);
```
morrone, line 6823: We really shoudn't call this here. Just make the array long enough that we can call __cmdline_options() once.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+			if (rc) {
+				/* Reset rc to 0 because the error is caused by user */
+				rc = 0;
+				goto send_reply;
+			}
+			argc = 1;
+		}
+		argv[argc++] = token;
+		argv[argc++] = strtok_r(NULL, " \t\n", &ptr1);
```
morrone, line 6832: I think this strtok_r is meant to grab the value of the option, but there are two problems with this. First, there are some (at least one) options for which the value is optional. So this winds up tokenizing the next option rather than the value for the current option. Second, this seems to ignore the possibility that the user used "--longoption=blah" format, where the option and value would wind up in the same token.

Ultimately this function should not care about options and values, only about breaking up the tokens at whitespace. So we should only fill in one argv array value per loop.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+		if (argv[argc-1] == NULL)
```
morrone, line 6833: If you follow my suggestions, this check is eliminated.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6742,3 +6763,80 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
+			argc--;
+	}
+
```
morrone, line 6836: If you follow my suggestions, you will need to set "argv[argc] = NULL" after the loop.

morrone (changes requested):

nichamon:
@tom95858 @morrone @baallan I got rid of the hard-coded array size.  I count the number of cmd-line tokens and allocate the array accordingly. Parsing through the option string twice shouldn't affect LDMSD setup time significantly.

@morrone  Thanks for the detailed review. The current logic supports  "--longoption=blah", and it has been tested. Please let me know if it doesn't work for you.

@baallan 
> doesn't s need to be freed at some point?
Thanks for catching this.

morrone:
> @morrone Thanks for the detailed review. The current logic supports "--longoption=blah", and it has been tested. Please let me know if it doesn't work for you.

It is OK with me. Thanks.

PR #835, opened 2021-11-12
Report errors when interval/offset/regex of updtr_# is invalid

nichamon:


nichamon:
The patch addresses minor problems in various updtr_* config commands, which were found by the new test scripts in https://github.com/ovis-hpc/ldms-test/pull/15.

PR #838, opened 2021-11-25
Register lustre's job_stats and slurm sets to LDMSD's set_tree

nichamon:


nichamon:
@tom95858 I found additional sets (job_stats) that the Lustre plugins do not register to LDMSD's set tree. Slurm plugin also doesn't register the sets to LDMSD's set tree. I created the previous pull request #837  prematurely. Sorry!.

PR #846, opened 2021-12-14
add @auth:authopts to ldmsd -x option syntax

baallan:
This allows multiple listener specifications with independent authentication
methods from the command line to simplify startup via systemd service scripts
or test cases.


baallan:
something off here w/git usage.

PR #847, opened 2021-12-15
fix ldmsd -a order dependence bug and enable independent -x listeners

baallan:
In addition to fixing sequence dependence (-x before -a yields wrong default
authentication setting) of listener options, this patch extends the
-x option syntax to -x xprt:port[:host][@auth[:authopts]]
This allows multiple listener specifications with independent authentication
methods from the command line to simplify startup via systemd service scripts
or test cases. Previously all -x options were tied to the same (-a) authentication method.

Internal documentation of cleanup also added.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -469,6 +469,15 @@ double ldmsd_timeval_diff(struct timeval *start, struct timeval *end)
 
 extern void ldmsd_strgp_close();
 
+/*
+ * \param x number of a functionality or location (not an errno).
+ *  	This value will be the exit code which can be looked up
+ *  	in the automatically generated ldmsd_exits man page tabulating
+ *  	x and reason value.
+ * \param admin/user-oriented reason string describing failed functionality.
+ * Note: Log messages issued before cleanup is called should provide
+ * details like file:line:function_name if needed and helpful.
+ */
 void cleanup(int x, const char *reason)
 {
 	int llevel = LDMSD_LINFO;
@@ -1756,24 +1765,41 @@ int ldmsd_listen_start(ldmsd_listen_t listen)
 	return rc;
 }
 
-static int __create_default_auth()
+static int __create_auth(const char *n, const char *pname, struct attr_value_list *opt)
 {
 	ldmsd_auth_t auth_dom;
 	int rc = 0;
-
-	auth_dom = ldmsd_auth_find(DEFAULT_AUTH);
-	if (auth_dom)
+	auth_dom = ldmsd_auth_find(n);
+	char *sopt = av_to_string(opt, AV_EXPAND);
+	if (auth_dom) {
+		if (opt && opt->count) {
+			ldmsd_log(LDMSD_LERROR, "Repeated auth method %s"
+				" must not have options given again:%s.\n",
+				n, sopt);
+			free(sopt);
+			return E2BIG;
+		}
+		free(sopt);
 		return 0;
-
-	auth_dom = ldmsd_auth_new_with_auth(DEFAULT_AUTH, auth_name, auth_opt,
+	}
+	ldmsd_log(LDMSD_LINFO, "Configuring auth method %s with plugin %s "
+		"and options %s\n", n, pname, sopt ? sopt : "(none)");
+	free(sopt);
+	auth_dom = ldmsd_auth_new_with_auth(n, pname, opt,
 					geteuid(), getegid(), 0600);
 	if (!auth_dom) {
-		ldmsd_log(LDMSD_LCRITICAL, "Failed to set the default "
-				"authentication method, errno %d\n", errno);
+		ldmsd_log(LDMSD_LCRITICAL, "Failed to set the %s "
+			"authentication method, errno %d\n", n, errno);
 		rc = errno;
 	}
 	return rc;
 }
+
+static int __create_default_auth()
+{
+	return __create_auth(DEFAULT_AUTH, auth_name, auth_opt);
+}
+
 
 struct ldmsd_str_ent *ldmsd_str_ent_new(char *s)
 {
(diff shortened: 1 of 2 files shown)


baallan:
looking into what it is about the test that causes this fail. there's been no protocol changes other than the command line, so if fixing a bug has unmasked another bug then fixing the test script will be the solution.

tom95858:
@nichamon can you review this whenever you get a chance?

baallan:
I don't propose this patch for the pending release if we are sticking to the tomorrow-release schedule.

Alternatively to this redesign, we accept order dependence of -a -A -x, but allow multiple specification of the -a -A and -x, with the last values seen for -a,-A becoming the defaults. This would require a different patch that unifies the -x processing with the -a/-A parsing but allow the redefinition of -a/-A until all cmdline options have processed.

tom95858:
@baallan, should we revert your previous patch? I had erroneously assumed that it had been tested.

baallan:
the previous patch wasn't applied. i closed it and opened this one after porting the previous patch to Mons rearranged option handling.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -1804,11 +1830,111 @@ void ldmsd_str_list_destroy(struct ldmsd_str_list *list)
...
+				}
+				t++;
+			}
+		}
+		struct attr_value_list *xav = av_new(xosize);
+		struct attr_value_list *xkv = av_new(xosize);
+		if (x_auth_opts) {
+			tokenize(x_auth_opts, xkv, xav);
+		}
+		int xrc = __create_auth(_auth, _auth, xav);
```
nichamon, line 1908: Using the `_auth` as the authentication domain main posts a limitation that users can use an authentication method once at the command line, regardless of the authentication method arguments. LDMSD won't support the following example.
```
ldmsd -x sock:10001@ovis:conf=~/ldmsauth.conf -x sock:10002@ovis:conf=~/ldmsauth2.conf
```

ldms/src/ldmsd/ldmsd.c
```diff
@@ -1804,11 +1830,111 @@ void ldmsd_str_list_destroy(struct ldmsd_str_list *list)
...
+	}
+	if (_host && !strlen(_host)) {
+		_host = NULL;
+	}
+	if (_auth && !strlen(_auth)) {
+		_auth = NULL;
+	}
+	/* Use the default auth domain unless specified */
+	if (!_auth) {
+		_auth = DEFAULT_AUTH_NAME;
```
nichamon, line 1887: This should be
```
_auth = DEFAULT_AUTH
```
`DEFAULT_AUTH` is the name of the default authentication domain, which uses the authentication method `DEFAULT_AUTH_NAME`. Currently, `DEFAULT_AUTH_NAME` is `none`.

This results in users cannot start LDMSD with a listening endpoint using the default authentication method. LDMSD won't start with the following command, for example,
```
ldmsd -x sock:10001
```

Moreover, any user-specific default authentication method won't work.
```
ldmsd -x sock:10001 -a munge
```
This is the cause that the patch fails the GitHub test.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -1804,11 +1830,111 @@ void ldmsd_str_list_destroy(struct ldmsd_str_list *list)
...
+		if (xrc) {
+			ldmsd_log(LDMSD_LERROR, "%s: failed to add auth"
+				" method %s\n", STRERROR(xrc), _auth);
+			cleanup(13, "add auth on command-line failed.");
+		}
+	}
+	ldmsd_log(LDMSD_LINFO, "main() Trying to listen on "
+		"%s:%s:%s@%s.\n", _xprt, _port,
+		_host ? _host : "NULL_HOST", _auth);
+	ldmsd_listen_t listen = ldmsd_listen_new(_xprt, _port, _host, _auth );
```
nichamon, line 1920: If you make the authentication domain name unique, the new name will be given instead of `_auth` when creating a new listen object.

nichamon (changes requested):

tom95858:
@baallan, @nichamon why are we doing this exactly? What is wrong with the expanded auth configuration supported in the configuration files.

nichamon:
> @baallan, @nichamon why are we doing this exactly? What is wrong with the expanded auth configuration supported in the configuration files.

@tom95858 Yes, the configuration commands in the configuration files support specifying a specific authentication for each listening endpoint. As @baallan was the one who made the changes, I'll wait for him to answer your questions.

In case someone isn't familiar with how to do this in the configuration files, here are some examples. You can find examples of configuration files with such usages here https://github.com/ovis-hpc/ovis/tree/OVIS-4/ldms/examples/cmdline-opts.

```
auth_add name=my_munge plugin=munge
listen xprt=sock port=10002 auth=my_munge # NOTE: auth=<auth_domain_name>
```

```
auth_add name=munge # If 'plugin' is not given, the plugin name and the authentication domain name are the same.
listen xprt=sock port=10003 auth=munge
```

baallan:
> @baallan, @nichamon why are we doing this exactly? What is wrong with the expanded auth configuration supported in the configuration files.

@tom95858 this allows us to start a daemon from a command line in  a systemd script with multiple listeners with distinct auth methods and no dependence on another configuration file. Such a daemon can be ready for immediate contact by user applications or by maestro configuration.

tom95858:
@nichamon, @baallan, where are we on this. This patch seems to break things correct?

baallan:
i need to respond yet to Mon's latest comments. converted to draft for now.

baallan:
@tom95858 why are you closing something marked as draft?

tom95858:
@baallan because it's been in this state since December 14?

tom95858:
@baallan can we please close this?

baallan:
Given the changes in adjacent parts of the code, I will need to redo from scratch enabling multiple listeners from the command line.

PR #848, opened 2021-12-16
fix validations of interval/offset on updtr_start and uninit count leading to arbitrarily large calloc

baallan:
fixes ldmsd_request.c uninit variable that can lead to allocation failure and incomplete input validation that can lead to updtr start failure. The commits are split by topic.

in the updtr case, user can specify an invalid offset while defaulting the interval, so the half-range validation is needed in two places.


nichamon (approved):

baallan:
@nichamon thanks for reviewing. @tom95858 let me know if you need something else changed here before merging.

PR #849, opened 2021-12-16
update build to support rocm 4.5 rdc device headers on gcc4-gcc8

baallan:
working on reporting the issue upstream. With this patch, the sampler builds and works.


baallan:
I loathe having to make compat headers like this, but it seems there's no other workaround until upstream fixes their header or our platforms catch up to their unstated compiler requirements. I wonder if maybe they build with clang or latest gcc instead of widely deployed gccs.

tom95858:
Could you please state the problem that this patch solves. Thanks.

baallan:
https://github.com/RadeonOpenCompute/ROCm/issues/1641
documents the problem.

PR #852, opened 2021-12-17
update return value to prevent nuisance warnings.

baallan:
this stops spurious warnings  in the log about
": create_hfi_port error "


tom95858:
Hi @baallan, please convert this from a draft or close it.

tom95858:
@baallan, still WIP...what's up with this pull request?

baallan:
converted to an issue assigned to me.
the code here is small enough to throw away since it doesn't entirely solve the problem.

PR #854, opened 2021-12-17
add auth support to ibmad_sampler

baallan:
this add auth configuration and register/deregister to the ibmad sampler.


baallan:
this somehow does not address the  problem that the updater hint is never set.

PR #859, opened 2022-01-06
Fix store issues with SOS-5

narategithub:


ldms/src/store/slurm/store_slurm.c
```diff
@@ -564,7 +566,9 @@ _open_store(struct sos_instance *si, ldms_set_t set)
 	}
 
  add_schema:
+	uuid_generate(schema_uuid);
 	slurm_schema_template.name = si->schema_name;
+	slurm_schema_template.uuid = (char*)schema_uuid;
```
tom95858, line 571: why is this cast necessary?

ldms/src/store/store_app/store_app.c
```diff
@@ -332,6 +334,8 @@ __get_sos_schema(store_app_cont_t cont, const char *name,
 	if (!schema) {
 		/* Create a new schema for 'name' */
 		struct sos_schema_template *tmp = &app_schema_template;
+		uuid_generate_random(uuid);
```
tom95858, line 337: If we're generating random uuid, then it's not really useful for ensuring that schema across containers and partitions are the same. If we have multiple aggregators storing data for the same schema, they will have different UUID which means that attempting to combine these partitions later into a single container for analysis will fail...

tom95858 (commented):
I think we need to have a discussion about how to handle this. The PAPI sampler comes to mind as a particular challenge.

ldms/src/store/slurm/store_slurm.c
narategithub: Without the casting, GCC warned about "converts between pointers to integer types with different sign".

ldms/src/store/store_app/store_app.c
narategithub: You're right. This should be the LDMS schema digest, and this patch must be on top of the LDMS schema digest patch. I'll close this pull request put this change on top of the schema digest commit.

narategithub:
The `sos_schema.uuid` should be the LDMS schema digest. This PR is closed and the commit will be a part of #858.

Issue #860, opened 2022-01-06
Sampler changes register metric sets, but fail to deregister


morrone:
There are a few recent sampler changes that appear to add ldmsd_set_register() calls (a good thing) but fail to include the matching ldmsd_set_dereigster() which (unless deregister happens implicitly and isn't documented), which cause memory leaks.

The commits are at least:

* 51e28d16 2021-11-25 Register slurm_sampler sets to LDMSD's set tree
* 8cbe866c 2021-11-25 Register lustre's job stats sets to LDMSD's set tree
* 9c29c371 2021-11-24 Register papi_sampler sets to the LDMSD's set tree

I'm not sure how oftern papi_sampler metric sets come and go, but this is especially a big problem for the job stats sets, which will come and go with the rate of system jobs, which can be relatively high.

PR #861, opened 2022-01-07
Add missing ldmsd_set_deregister() calls, plus other minor fixes

morrone:
Add missing ldmsd_set_deregister() calls to several samplers, as mentioned in issue #860.

Additionally, add the missing ldms_set_unpublish() call to the slurm_sampler and add both ldmsd_set_register() and ldmsd_set_unregister() calls to the dcgm_sampler.


nichamon (approved):
Thanks for catching this.

PR #865, opened 2022-01-10
Change ibmad and dcgm sampler names to match new file name

morrone:
The new plugin names for ibmad and dcgm are ibmad_sampler
and dcgm_sampler, respectively. Symlinks to the previous
names still exist for the old names to aid transition.

Unfortunately, the API makes it such that only one of these
name (practically speaking) can have working metric
set hints. This can be addressed in the future in a new
plugin API.

We make the plugins report the same name as the new
plugin file name. The plugins probably aren't in too
wide use yet, so we err on the side of making this
work for future users.

### DIFF
--- a/ldms/src/sampler/dcgm_sampler/dcgm_sampler.c
+++ b/ldms/src/sampler/dcgm_sampler/dcgm_sampler.c
@@ -21,7 +21,7 @@
 
 #define _GNU_SOURCE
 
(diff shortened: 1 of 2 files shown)


baallan:
+1

baallan:
@morrone will/did you follow up with removing the symlink creation from makefile.am? it keeps breaking my build/debug process anyway.

PR #866, opened 2022-01-11
add register/deregister for ibnet sets so hints should populate

baallan:
the ibnet sampler developed before register/deregister needs updating.
this patch updates it.


baallan:
@tom95858 this sampler patch is needed for the pending release to support auto interval on the ibnet sampler.

PR #867, opened 2022-01-12
add linux_proc_sampler option to collect sys_clk_tck for set which may report ticks

baallan:


baallan:
@tom95858 this adds ticks/sec metadata to the linux_proc_sampler as we did for procstat and others.

baallan:
@tom95858 looks like github test platform is back and this pull passed everything as expected. ready to merge.

PR #869, opened 2022-01-13
fix bugs in  auth arg  and instance name handling for genders-based configuration

baallan:
-A must be included in the command line when an auth_ovis conf or auth_munge socket option is included in the configuration and the default instance name must be overridden if instance is included in the user-defined configuration.


baallan:
@tom95858 @nichamon none of the files in this PR have anything to do with the ldms/src tree, so they cannot break 4.3.3 compatibility. (all changes are shell scripting or man page except for a standalone c++ file generator)
I trust the fail occurred due to 4.3.3 test revision work that appears to be going on.

nichamon:
@baallan @tom95858 I'm aware of the problem. I have been digging into the problem this afternoon, and a patch to fix the problem should be ready tonight.

baallan:
@tom95858 i reran the checks after the workflow updates and it passed/is ready.

PR #873, opened 2022-01-14
fix build warnings in ipmi sampler

baallan:
-Werr build will fail if ipmi is enabled. this fixes that.


baallan:
@tom95858 please merge this before release if possible.

PR #877, opened 2022-01-29
fix build for boost-dependence when not in /usr

baallan:
This fixes makefile.am insufficiency when boost is not installed with prefix /usr. Affects only the c++ genders config file processor.


baallan:
need this for the 4.8 release @tom95858

Issue #15, opened 2019-01-17
LDMS Users group


oceandlr:
Topic: LDMS Users Group
Schedule below. For call in info and notes, please see the [ovis-hpc/ovis wiki](https://github.com/ovis-hpc/ovis/wiki/User-Group-Meeting-Notes)

=======================================================

Time: Jan 21, 2019 12:00 PM Mountain Time (US and Canada)

Every 2 weeks on Mon, until May 27, 2019, 10 occurrence(s)

Jan 21, 2019 12:00 PM

Feb 4, 2019 12:00 PM

Feb 18, 2019 12:00 PM

Mar 4, 2019 12:00 PM

Mar 18, 2019 12:00 PM

Apr 1, 2019 12:00 PM

Apr 15, 2019 12:00 PM

Apr 29, 2019 12:00 PM

May 13, 2019 12:00 PM

May 27, 2019 12:00 PM

https://zoom.us/j/599135580

One tap mobile

+16468769923,,599135580# US (New York)

+16699006833,,599135580# US (San Jose)

Dial by your location

+1 646 876 9923 US (New York)

+1 669 900 6833 US (San Jose)

Meeting ID: 599 135 580

oceandlr:
quick start video from users group meeting has been posted

ovis-hpc:
closing issue since the links are in the sidebar

PR #882, opened 2022-02-22
add value expansion by other keys in av_list

baallan:
This allows common usage patterns to be expressed without change across plugins, for example:
```
config name=foo job_set=${host}/jobid component_id=${COMPONENT_ID} schema=~{name} instance=${host}/~{schema}
```
might expand to
```
config name=foo job_set=nid0001/jobid component_id=123 schema=foo instance=nid0001/foo
```
and the equivalent maestro yaml line count shrinks by 75% and reduces the chance of typos and misapplication of a common patterns.

### DIFF
--- a/lib/src/ovis_util/util.c
+++ b/lib/src/ovis_util/util.c
@@ -90,6 +90,27 @@ static const char *get_env_var(const char *src, size_t start_off, size_t end_off
 	return name;
 }
 
+static const char *get_tilde_var(const char *src, size_t start_off, size_t end_off)
+{
+	static char name[100];
+	size_t name_len;
+
+	/* Strip ~, {, and } from the name */
+	if (src[start_off] == '~')
+		start_off += 1;
+	if (src[start_off] == '{')
+		start_off += 1;
+	if (src[end_off-1] == '}')
+		end_off -= 1;
+
+	name_len = end_off - start_off;
+	assert(name_len < sizeof(name));
+
+	strncpy(name,  &src[start_off], name_len);
+	name[name_len] = '\0';
+	return name;
+}
+
 int _scpy(char **buff, size_t *slen, size_t *alen,
 	    const char *str, size_t len)
 {
@@ -221,8 +242,117 @@ err:
 	return NULL;
 }
 
+static char *str_repl_tilde(struct attr_value_list *avl, const char *str, int depth);
+static const char *av_value_tilde(struct attr_value_list *avl, const char *name, int depth)
+{
+	if (depth > avl->count) {
+		errno = EBADSLT;
+		return NULL;
+	}
+	if (!avl)
+		return NULL;
+	int i;
+	for (i = 0; i < avl->count; i++) {
+		if (strcmp(name, avl->list[i].name))
+			continue;
+		char *tilsub = str_repl_tilde(avl, avl->list[i].value, depth);
+		if (tilsub) {
+			string_ref_t ref = malloc(sizeof(*ref));
+			if (!ref) {
+				free(tilsub);
+				return NULL;
+			}
+			ref->str = tilsub;
+			LIST_INSERT_HEAD(&avl->strings, ref, entry);
+			return tilsub;
+		} else
+			break;
+	}
+	return "";
+}
+
+/* expand and return ~{key} replacements */
+static char *str_repl_tilde(struct attr_value_list *avl, const char *str, int depth)
+{
+	if (!str)
+		return NULL;
+	const char *name;
+	const char *value;
+	const char *values[100];
+	char *res, *res_ptr;
+	size_t res_len;
+	regex_t regex;
+	char *expr = "\\~\\{[[:alnum:]_]+\\}";
+	int rc, i;
+	regmatch_t match[100];
+	size_t name_total, value_total;
+
+	name_total = value_total = 0;
+
+	rc = regcomp(&regex, expr, REG_EXTENDED);
+	assert(rc == 0);
+	int pos;
+	rc = 0;
+	for (i = pos = 0; !rc; pos = match[i].rm_eo, i++) {
+		match[i].rm_so = -1;
+		rc = regexec(&regex, &str[pos], 1, &match[i], 0);
+		if (rc)
+			break;
+		res_len = match[i].rm_eo - match[i].rm_so;
+		name_total += res_len;
+		name = get_tilde_var(&str[pos], match[i].rm_so, match[i].rm_eo);
+		value = av_value_tilde(avl, name, depth + 1);
+		if (value) {
+			value_total += strlen(value);
+			values[i] = value;
+		} else {
+			regfree(&regex);
+			return NULL; /* recursion */
+		}
+		match[i].rm_eo += pos;
+		match[i].rm_so += pos;
+	}
+	regfree(&regex);
+	if (!i) {
+		res = strdup(str);
+		return res;
+	}
+
+	/* Allocate the result string */
+	res_len = strlen(str) - name_total + value_total + 1;
+	res = malloc(res_len);
+	if (!res) {
+		errno = ENOMEM;
+		return NULL;
+	}
+
+	size_t so = 0;
+
+	res_ptr = res;
+	res[0] = '\0';
+	for (i = 0; i < sizeof(match) / sizeof(match[0]); i++) {
+		if (match[i].rm_so < 0)
+			break;
+		size_t len = match[i].rm_so - so;
+		memcpy(res_ptr, &str[so], len);
+		res_ptr[len] = '\0';
+		res_ptr = &res_ptr[len];
+		so = match[i].rm_eo;
+		strcpy(res_ptr, values[i]);
+		res_ptr = &res[strlen(res)];
+	}
+	/* Copy the remainder of str into the result */
+	strcpy(res_ptr, &str[so]);
+	assert(res_len == (strlen(res) + 1));
+	return res;
+}
+
 /*
- * This function returns a value that must be freed
+ * This function returns a value that must be freed.
+ * It expands ~{} substitutions from other keys
+ * before expanding environment options.
+ * The value of ~{key_undefined} is the empty string, not an error.
+ * A recursion detected (a:~{b}, b:~{a}) returns EBADSLT.
  */
 char *str_repl_env_vars(const char *str)
 {
@@ -264,8 +394,10 @@ char *str_repl_env_vars(const char *str)
 		match[i].rm_so += pos;
 	}
 	regfree(&regex);
-	if (!i)
-		return strdup(str);
+	if (!i) {
+		res = strdup(str);
+		return res;
+	}
 
 	/* Allocate the result string */
 	res_len = strlen(str) - name_total + value_total + 1;
@@ -313,7 +445,11 @@ char *av_value(struct attr_value_list *av_list, const char *name)
 	for (i = 0; i < av_list->count; i++) {
 		if (strcmp(name, av_list->list[i].name))
 			continue;
-		char *str = str_repl_env_vars(av_list->list[i].value);
+		char *tilsub = str_repl_tilde(av_list, av_list->list[i].value, 0);
+		if (!tilsub)
+			return NULL;
+		char *str = str_repl_env_vars(tilsub);
+		free(tilsub);
 		if (str) {
 			string_ref_t ref = malloc(sizeof(*ref));
 			if (!ref) {
--- a/lib/src/ovis_util/util.h
+++ b/lib/src/ovis_util/util.h
@@ -116,6 +116,11 @@ struct attr_value_list {
 
 /**
  * \brief Get the value of attribute \c name
+ * Values will be fully substituted by str_repl_env_vars before
+ * being returned.
+ * \param name the key to search for in av_list.
+ * \return the first value with key name, or NULL if there is no
+ * key with the name or if the value cannot be expanded.
  */
 char *av_value(struct attr_value_list *av_list, const char *name);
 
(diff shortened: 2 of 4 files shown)


baallan:
@tom95858 This patch addresses https://github.com/ovis-hpc/maestro/issues/15 and allows the hard-coding of patterns that do no always apply to be removed from maestro.

tom95858:
@nichamon could you please review this? Thanks.

tom95858:
@baallan it seems like this only replaces values with other values on the same config line. In other words, there's no dictionary of 'vars' that you are referring to, just other strings in the same line.  Can you please provide an example of how this reduces the maestro config line count by %75? It seems like this belongs in the YAML parsing in maestro where one can refer to any value in the config file as opposed  to just a single line in the v4 config file.

baallan:
> @baallan it seems like this only replaces values with other values on the same config line. In other words, there's no dictionary of 'vars' that you are referring to, just other strings in the same line. 

The dictionary here is indeed other things in the same line, which is a limitation of this implementation. Supporting ~{} syntax at the yaml parse level would be highly preferable.
The anchor syntax lets us reduce repetitive config lines that introduce the possibility of typos, a very good thing.

The anchor substitution syntax in yaml suffers from the following well-known limitations, over which the ~{} interpretation *for our application* provides easy-to-understand improvements:
 * Anchor names are global and cannot be redefined, so they do not provide the context-sensitivity this solution provides.
 * Lists cannot be composed from multiple anchor substitutions
 
These deficiencies have been discussed in the yaml community and where it is firmly concluded (by the developers, not by the numerous users that bring it up every year in frustration on stackoverflow) that all the various proposed solutions belong at the scope of the application interpretation and not in the yaml spec.

> Can you please provide an example of how this reduces the maestro config line count by %75? It seems like this belongs in the YAML parsing in maestro where one can refer to any value in the config file as opposed to just a single line in the v4 config file.

yes. The sampler config syntax has some incorrectly hardwired attributes in the present maestro implementation (maestro bug #15). The clean yaml solution is to put those attributes into the yaml anchor for sampler defaults and let them be overridden in the usual yaml way. But the yaml anchor needs to include ~{name} (the context) to be useful.

For an example from our arm cluster (sampler plugin configuration)
wc -l $NO TILDE_VERSION
122
wc -l $TILDE_VERSION
54
(yaml blocks below)

The tilde version of sampler configs:
Incidentally this would be even smaller if we put the ~{} processing into the yaml instead of ldms because then we need only one anchor and one <<: in most sampler cases (reducing most plugins to two lines only instead of 4, lowering the potential line count to ~30).
```
    plugins :
      - <<: *sampler-time-defaults
        name : jobid
        config :
          <<: *sampler-config-defaults
          file : /var/run/ldms.slurm.jobinfo
      - <<: *sampler-time-defaults
        name : lustre_client
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : loadavg
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : tx2mon
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : meminfo
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : lnet_stats
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : vmstat
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : procnfs
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : procstat
        config :
          <<: *sampler-config-defaults
          maxcpu : 0
          schema : procstat_0
      - <<: *sampler-time-defaults
        name : procnet
        config :
          <<: *sampler-config-defaults
      - <<: *sampler-time-defaults
        name : sysclassib
        config :
          <<: *sampler-config-defaults
          ports : "mlx5_0.1,mlx5_1.1"
          schema : sysclassib_lserver
      - <<: *sampler-time-defaults
        name : dstat
        config :
          <<: *sampler-config-defaults
```

the no-tilde version
```
  - name :  jobid 
    interval :  1s 
    offset :  0s 
    config :
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      perm :  0644 
      schema :  ~{name}
      file :  /var/run/ldms.slurm.jobinfo 
  - name :  lustre_client 
    interval :  1s 
    offset :  0s
    config :
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name}
  - name :  loadavg 
    interval :  1s 
    offset :  0s
    config :
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name} 
  - name :  tx2mon 
    interval :  1s 
    offset :  0s 
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name} 
  - name :  meminfo 
    interval :  1s 
    offset :  0s 
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name} 
  - name :  lnet_stats 
    interval :  1s 
    offset :  0s 
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name}
  - name :  vmstat 
    interval :  1s 
    offset :  0s 
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name}
  - name :  procnfs 
    interval :  1s 
    offset :  0s 
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name}
  - name :  procstat 
    interval :  1s 
    offset :  0s
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  procstat_0
      maxcpu : 0
  - name :  procnet 
    interval :  1s 
    offset :  0s
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  ~{name}
  - name :  sysclassib 
    interval :  1s 
    offset :  0s
    config :  
      producer : ${host}
      component_id :  ${COMPONENT_ID} 
      instance :  ${host}/~{schema} 
      job_set :  ${host}/jobid 
      perm :  0644 
      schema :  sysclassib_lserver
      ports :  mlx5_0.1,mlx5_1.1 
    name :  dstat 
    interval :  1s 
    offset :  0s
    config : 
      producer : ${host}
      component_id :  ${COMPONENT_ID}  
      job_set :  ${host}/jobid 
      schema :  ~{name}
      perm : "0644"
```

for completeness, the anchors used in the tilde version are:
```
sampler_time_defaults : &sampler-time-defaults
  interval : "1s"
  offset : "0s"

sampler_config_defaults : &sampler-config-defaults
  job_set : ${host}/jobid
  component_id: ${COMPONENT_ID}
  schema : ~{name}
  instance : ${host}/~{schema}
  job_set : ${host}/jobid
  perm : "0644"
```

baallan:
To implement ~{} support based on pyyaml/maestro takes a tilde substitution class (~50 lines python) and a tree walk performing substitutions on the parsed yaml config (~30 lines python).   These were done last year. If our ultimate goal is to eliminate python and have ldmsd.c consume yaml, then these two bits of python would have to be translated.

tom95858:
I'm bit confused because both examples use ~{} substitution. The key difference seems to be that one version uses a merge key to set the defaults which also includes ~{}.

In any event. I am inclined add the ~{} substitution to the yaml parsing in maestro.

WRT ldmsd consuming YAML directly, the maestro --local mode is intended for this purpose.

lib/src/ovis_util/util.c
```diff
@@ -90,6 +90,27 @@ static const char *get_env_var(const char *src, size_t start_off, size_t end_off
 	return name;
 }
 
+static const char *get_tilde_var(const char *src, size_t start_off, size_t end_off)
+{
+	static char name[100];
```
nichamon, line 95: This will make `str_repl_env_vars()` not thread-safe. I think we don't have any places that would access the path concurrently. However, we could make this thread-safe, e.g., make the function receiving the `char * name` and its size.

baallan:
Apologies: The where the longer (no-tilde label) example has ~{name} or ~{schema} they would have to be replaced with the context-specific values. As the point was to demonstrate the line count, i didn't manually do all the string substitutions. You can clearly see, though, that a yaml anchor such as ' *name or *schema ' wouldn't work since these are global.

I have the ~{} substitution coded for maestro level, if you accept that it is useful from this example.
That would let me close this PR.

lib/src/ovis_util/util.c
```diff
@@ -313,7 +445,11 @@ char *av_value(struct attr_value_list *av_list, const char *name)
 	for (i = 0; i < av_list->count; i++) {
 		if (strcmp(name, av_list->list[i].name))
 			continue;
-		char *str = str_repl_env_vars(av_list->list[i].value);
+		char *tilsub = str_repl_tilde(av_list, av_list->list[i].value, 0);
```
nichamon, line 448: Please make the same change in `av_value_at_idx()`.

lib/src/ovis_util/util.h
```diff
@@ -116,6 +116,11 @@ struct attr_value_list {
 
 /**
  * \brief Get the value of attribute \c name
+ * Values will be fully substituted by str_repl_env_vars before
+ * being returned.
+ * \param name the key to search for in av_list.
+ * \return the first value with key name, or NULL if there is no
```
nichamon, line 122: According to the code, `av_value()` returns an empty string if the key doesn't exist or it cannot expand the value.

nichamon (changes requested):

nichamon:
@baallan @tom95858 I reviewed the code in detail before I saw your conversation. 

The change doesn't affect the maestro YAML config file line count, which is also shown by the examples Ben gave above. The changes' benefit is that users don't have to type the same strings multiple times on the same line in LDMSD config files. Ben showed this in his first comment. I'm not sure about its impact on the production environment as we use Maestro instead of writing LDMSD config files directly.

morrone:
> @baallan @tom95858 I reviewed the code in detail before I saw your conversation.
> 
> The change doesn't affect the maestro YAML config file line count, which is also shown by the examples Ben gave above. The changes' benefit is that users don't have to type the same strings multiple times on the same line in LDMSD config files.

I think if you reread Ben's example, you will see that is shows a maestro YAML example where it reduces the line count. Reducing duplicated typing on the same line very likely would not be sufficient reason to introduce this change.

I think I am also currently leaning towards preferring this support in maestro.

lib/src/ovis_util/util.c
baallan: good catch.

lib/src/ovis_util/util.c
baallan: get_env_var is also thread un-safe in the same way.
Whether we take this pull request or not, the use of the static name variable needs to be converted to a caller-supplied buffer.

Can fix both misuses with an additional commit on this pr if we decide to take this pr.

lib/src/ovis_util/util.h
```diff
@@ -268,13 +273,20 @@ int f_mkdir_p(const char *path, __mode_t mode);
...
  * The function returns memory that was allocated with malloc()
  *
  * \param str The input string.
- * \retval NULL There was insufficient memory to allocate the output string.
- * \retval Ptr to a string with the environment variable values replaced.
+ * \retval NULL There was insufficient memory to allocate the output string
+ * or recursion was detected; errno will be ENOMEM or EBADSLT respectively.
+ * \retval Ptr to a string with the environment variable and ~{} values replaced.
  */
 char *str_repl_env_vars(const char *str);
```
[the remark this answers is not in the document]
baallan: i'm not picky about the name. But the current name has been part of the public API for a while so the rename could break code we don't know about.  The extension of the semantics is unlikely to affect anyone since ~{xxx} is not a common pattern of naming things likely to be on the right side of attr=value.

lib/src/ovis_util/util.h
baallan: util.c lines 450 and 465 disagree with you. can you explain?

baallan:
> I think if you reread Ben's example, you will see that is shows a maestro YAML example where it reduces the line count. Reducing duplicated typing on the same line very likely would not be sufficient reason to introduce this change.

In the ldmsd config language, the point of the ~{} is not to simplify a single line but to let the same line be used over and over for different plugins with only the name= attribute having to be redefined.

> I think I am also currently leaning towards preferring this support in maestro.

That's where I want to put it too, because if I put it in maestro/yaml, i can do substitutions not only within the line but from the nearest upward scope in the yaml tree.

lib/src/ovis_util/util.h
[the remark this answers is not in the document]
tom95858: @nichamon, @baallan should I accept this pull request or do we still have design issues to resolve?

baallan:
@tom95858  I believe the conversation above suggests the right design solution is to put the ~{} expansion feature into maestro, so this can be closed.

PR #884, opened 2022-02-22
Bug fix for ldmsd_listen_new error handling

nick-enoent:
Fix error where ldmsd hardcodes error to ENOMEM if
ldmsd_listen_new fails to create listen object

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -1643,7 +1643,6 @@ ldmsd_listen_t ldmsd_listen_new(char *xprt, char *port, char *host, char *auth)
 				getuid(), getgid(), 0550); /* No one can alter it */
 	free(name);
 	if (!listen) {
-		errno = ENOMEM;
 		return NULL;
 	}
 


ldms/src/ldmsd/ldmsd.c
```diff
@@ -1643,7 +1643,6 @@ ldmsd_listen_t ldmsd_listen_new(char *xprt, char *port, char *host, char *auth)
 				getuid(), getgid(), 0550); /* No one can alter it */
 	free(name);
 	if (!listen) {
-		errno = ENOMEM;
```
tom95858, line 1646: Did you check to see that the errno is already set?

I _think_ this is what the errno should be, depending on the reason for the failure.

* auth _name_ not found, ENOENT
* address in use, EADDRINUSE
* memory allocation failure, ENOMEM
* other, EINVAL

PS: these errors would be set inside the ldmsd_cfgobj_new_with_auth() function.

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd.c
nick-enoent: I did. This error is set in ldmsd_cfgob_new_with_auth, however it will return a null object regardless of whether the error is ENOMEM, or EEXIST. Currently, if the address is in use, ldmsd_cfgobj_new_with_auth will set the errno to EEXIST, but this can be trivially changed if you'd prefer EDDARINUSE. The removed line in ldmsd_listen_new will set the error to ENOMEM if the listen object is NULL, regardless if the error has already been assigned.

PR #885, opened 2022-02-23
Addressing passive zap endpoint thread starvation

narategithub:
When a passive endpoint was assigned to a thread that has an active
endpoint with high activity, the routine of the passive endpoint (e.g.
accepting a connection) could be starved. The routines of the passive
endpoints are quite light-weighted. So, this patch assigns the passive
endpoints into its own thread to avoid the starvation.


narategithub:
@tom95858,

I've tested this patch against `ldms-test`, and also manually gdb the sampler and the aggregator to see that the passive endpoints are being assigned to the `passive_ep_thread`.

PR #890, opened 2022-03-03
Do not keep unused zap sock IO entries

narategithub:
`zap_sock` IO entries were kept in the `free_q` so that it can be used
immediately for the next IO alloc. This patch `free()` the IO entries
instead of keeping it around.


narategithub:
@tom95858 This is a part of the stream "leak" you asked me to look into. Please do not merge this yet. I think the OOM on L1 you described was caused by IO queue to L2 growing w/o bound due to a lot of incoming stream (from L0 / samplers) and just 1 output pipe to L2.

tom95858:
@narategithub Oops, I merged this...

PR #908, opened 2022-04-13
Add LDMS's ldms_mval_t setter/getter APIs

nichamon:


ldms/src/core/ldms.c
```diff
@@ -2875,6 +2875,276 @@ double ldms_metric_array_get_double(ldms_set_t s, int mid, int idx)
...
+}
+
+void ldms_mval_set_s64(ldms_mval_t mv, int64_t v)
+{
+	mv->v_s64 = __cpu_to_le64(v);
+}
+
+void ldms_mval_set_float(ldms_mval_t mv, float v)
+{
+#if LDMS_SETH_F_LCLBYTEORDER == LDMS_SETH_F_LE
```
tom95858, line 2925: @nichamon, I think __cpu_to_le32 reduces to a "no-op" on LE platforms. Was there an issue with the forced type cast in the __cpu_to_le32? It seems like if there were, then it would break on BE systems for the same reason.

ldms/src/core/ldms.c
```diff
@@ -2875,6 +2875,276 @@ double ldms_metric_array_get_double(ldms_set_t s, int mid, int idx)
...
+	mv->a_f[idx] = v;
+#else
+	/* This type abuse is necessary to avoid the integer cast
+	 * that will strip the factional portion of the float value
+	 */
+	*(uint32_t *)&mv->a_f[idx] = __cpu_to_le32(*(uint32_t *)&v);
+#endif
+}
+
+void ldms_mval_array_set_double(ldms_mval_t mv, int idx, double v)
```
tom95858, line 3003: @nichamon, same thing here only for doubles.

ldms/src/core/ldms.c
```diff
@@ -2875,6 +2875,276 @@ double ldms_metric_array_get_double(ldms_set_t s, int mid, int idx)
...
+	mv->a_s64[idx] = __cpu_to_le64(v);
+}
+
+void ldms_mval_array_set_float(ldms_mval_t mv, int idx, float v)
+{
+#if LDMS_SETH_F_LCLBYTEORDER == LDMS_SETH_F_LE
+	mv->a_f[idx] = v;
+#else
+	/* This type abuse is necessary to avoid the integer cast
+	 * that will strip the factional portion of the float value
```
tom95858, line 2997: *fractional

tom95858 (changes requested):

ldms/src/core/ldms.c
nichamon: No, there was no issue. I have tested it by removing the `#if` condition, and it works fine on my little-endian machine. I think it's safe to remove the `#if` condition. I copied the code from `ldms_metric_set_float()` and pasted it here.

ldms/src/core/ldms.c
nichamon: Thanks for catching this.

ldms/src/core/ldms.c
nichamon: Got it!

ldms/src/core/ldms.c
nichamon: What I meant by removing the `#if` condition is to replace 
```
#if LDMS_SETH_F_LCLBYTEORDER ...
     mv->v_f = v;
#else
     *(uint32_t *)&mv->v_f = __cpu_to_le32(*(uint32_t *)&v);
#endif
```
by
```
*(uint32_t *)&mv->v_f = __cpu_to_le32(*(uint32_t *)&v);
```

PR #909, opened 2022-04-13
Make LDMSD collect and report stream information

nichamon:


nichamon:
@tom95858 @narategithub @nick-enoent LDMSD sends a stream update to clients when a new stream is available. It only reports the information about the new stream. The information isn't helpful for the load/balancing because we don't have the rate of the new stream yet. I need to think more about how/when LDMSD will send a stream update with useful information for both new streams and existing streams. For your information, with this patch, clients can ask for information on the available streams periodically (multiple times).

baallan:
@nichamon please explain "Purge stream publishers if it's inactive". In particular, it's perfectly reasonable for a long lived hpc publisher (e.g. something monitoring process creation events) to be idle for many hours/days. Under what conditions is something publishing intermittently "inactive"? If its underlying connection to the first receiving daemon is gone? If the connection is there but just quiet?

nichamon:
> @nichamon please explain "Purge stream publishers if it's inactive". In particular, it's perfectly reasonable for a long lived hpc publisher (e.g. something monitoring process creation events) to be idle for many hours/days. Under what conditions is something publishing intermittently "inactive"? If its underlying connection to the first receiving daemon is gone? If the connection is there but just quiet?

This code purges a publisher after not sending any messages to LDMSD for a specific duration. The duration is configurable. I have thought about your case. I also consider how slurm_notifier works, which creates a new connection for each message. I decided not to use connected/disconnected events as a measurement.

baallan:
For bursty clients (which vary between frequent events and long quiet periods and are a common if not the dominant pattern in hpc) assuming a new connection with each event forces poor performance.  Since the duration is configurable, I suggest the duration be configurable by the calling client (which is the only place where behavior expectations are actually known). A global default which is only globally tunable cannot satisfy two use cases simultaneously. It is likely we will always have at least two use cases: application data (probably minutes timeout desired) and system events (process start/end- timeout desired in days), and a third use case of an IB fabric sampler rebalancing its load periodically via stream messages (timeout in hours).

PR #913, opened 2022-04-20
Add an API to create a set with a heap of a specific size

nichamon:


tom95858:
@nichamon sorry for letting this sit so long. I think we discussed adding a new API to set an attribute in the schema to set the heap size to override the default (e.g. ldms_schema_heap_size_set). We also discussed adding this new API. For some reason we chose the later. Do you recall why?

nichamon:
> @nichamon sorry for letting this sit so long. I think we discussed adding a new API to set an attribute in the schema to set the heap size to override the default (e.g. ldms_schema_heap_size_set). We also discussed adding this new API. For some reason we chose the later. Do you recall why?

We wanted to solve the problem that LDMSD may fail to add a new list item either because the heap is too small or the free heap memory is fragmented. An easy thing to do is delete the current set and create a new set with a bigger heap. If LDMS provides only ```ldms_set_new_with_heap(...., size_t heap_sz)```, authors can create a set in one call. Compared to LDMS providing only ```ldms_schema_heap_sz_get()```, authors need to call the API and then ```ldms_set_new()```. It is more convenient to have ```ldms_set_new_with_heap()```. Another reason is that the heap is per set, not per schema, so IMO it is natural to provide the heap size when creating a new set.

I saw another reason we cannot provide ```ldms_schema_heap_size_set()``` while I implemented the change. ```ldms_schema``` does not cache the total heap size. It caches the heap size for each metric list. The total heap size is calculated at set creation.

ldms/src/test/test_ldms_set_new.c
```diff
@@ -0,0 +1,97 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <assert.h>
+#include "ldms.h"
+
+#define UID 1000
+#define GID 1000
+#define PERM 777
+#define HEAP_SZ 512 x 1024L
```
tom95858, line 9: Please put #define that contain computations in parenthesis.

tom95858:
Hi @nichamon, there seem to be merge conflicts now. Can you please rebase and push?

tom95858:
@nichamon, could you please fix the merge conflicts in this patch?

nichamon:
> @nichamon, could you please fix the merge conflicts in this patch?

I rebased the patch on top of OVIS-4.

PR #919, opened 2022-05-19
fix bug in ldmsd option processing

baallan:
when -n is not specified and -x is not specified, null pointer
dereference occurs in computing daemon name; this patch fixes that.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -2215,7 +2215,7 @@ int main(int argc, char *argv[])
 			myhostname[0] = '\0';
 		listen = (ldmsd_listen_t)ldmsd_cfgobj_first(LDMSD_CFGOBJ_LISTEN);
 		snprintf(myname, sizeof(myname), "%s:%d",
-				myhostname, listen->port_no);
+				myhostname, listen ? listen->port_no : -1);
 	}
 
 	if (!foreground) {


ldms/src/ldmsd/ldmsd.c
```diff
@@ -2215,7 +2215,7 @@ int main(int argc, char *argv[])
 			myhostname[0] = '\0';
 		listen = (ldmsd_listen_t)ldmsd_cfgobj_first(LDMSD_CFGOBJ_LISTEN);
 		snprintf(myname, sizeof(myname), "%s:%d",
-				myhostname, listen->port_no);
+				myhostname, listen ? listen->port_no : -1);
```
tom95858, line 2218: I've seen this same bug. Should the port_no be zero though?

baallan: Since 0 is ambiguous but still suggests a listener, my first inclination if -1 seems inappropriate is to add a separate print omitting the ':%d', @tom95858.  Happy to do that if you prefer.

tom95858:
@narategithub, @baallan the only place this is used is in the failover logic when informing the peer of your name. It seems like this approach is rife for misconfiguration. @narategithub can we get rid of this code and change ldmsd_myname_get() to construct the name by constructing the name from the local address in  f->ax (see ldmsd_failover.c:995.) If we can, then I propose we get rid of the global variable 'myname' altogether.

That said, is there any reason we need the LDMSD_ATTR_PEER_NAME at all? The failover peer could figure this out from the remote address in the transport on which it receives the LDMSD_FAILOVER_PAIR_REQ.

narategithub:
@tom95858 ldmsd may listen on multiple transports (e.g. ugni on 411 and sock on 412). In such case, what should be the name? 

Or, since the failover is destined to be removed, in favor of miestro, shall we remove both failover and myname now? I'm inclined to remove failover + myname now :)

I believe the 'myname' is only used by the failover code.

tom95858:
Hi @narategithub, I get all that, but the peer already __knows__ the name. It's sitting in the transport handle on which it receives the PEER config request. All the myname does it make it possible to mess it up with the -n option; or am I missing something?

baallan:
@tom95858 fine w/me to remove failover, etc but I don't know if Jim has any non-sandia sites using it.

baallan:
@narategithub seems odd to have -n but nothing else failover related on the command line. looks like all other failover aspects are in the config file, so why not retire -n (and add to config file if not already covered)?

narategithub:
@tom95858 

On `ldmsd_1`, if `failover_config` is configured with `peer_name=ldmsd_2`, the `ldmsd_1` only accept failover pair request from the peer that declare its name as `ldmsd_2`. 
https://github.com/ovis-hpc/ovis/blob/b7e7158e9ebe72e156d2f930d60b141c62702e30/ldms/src/ldmsd/ldmsd_failover.c#L1944

If the peer is named other than `ldmsd_2`, the peer will get an error in the response and the connection is terminated:
https://github.com/ovis-hpc/ovis/blob/b7e7158e9ebe72e156d2f930d60b141c62702e30/ldms/src/ldmsd/ldmsd_failover.c#L963-L967

If the peer has the correct name `ldmsd_2`, the peer (`ldmsd_2`) will then proceed to request configurations from `ldmsd_1`:
https://github.com/ovis-hpc/ovis/blob/b7e7158e9ebe72e156d2f930d60b141c62702e30/ldms/src/ldmsd/ldmsd_failover.c#L957-L960

tom95858:
As we've discussed, this pull request is supplanted by #922

PR #920, opened 2022-05-19
Format byte strings when returning DirSet info

nick-enoent:
Return DirSet info dict key:values formatted as strings, rather than byte strings

### DIFF
--- a/ldms/python/ldms.pyx
+++ b/ldms/python/ldms.pyx
@@ -1410,7 +1410,7 @@ cdef class DirSet(object):
         self.data_gn = ds.meta_gn
         self.timestamp = (ds.timestamp.sec, ds.timestamp.usec)
         self.duration = (ds.duration.sec, ds.timestamp.usec)
-        self.info = { ds.info[i].key : ds.info[i].value \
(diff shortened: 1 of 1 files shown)


ldms/python/ldms.pyx
```diff
@@ -1410,7 +1410,7 @@ cdef class DirSet(object):
         self.data_gn = ds.meta_gn
         self.timestamp = (ds.timestamp.sec, ds.timestamp.usec)
         self.duration = (ds.duration.sec, ds.timestamp.usec)
-        self.info = { ds.info[i].key : ds.info[i].value \
+        self.info = { ds.info[i].key.decode('utf-8') : ds.info[i].value.decode('utf-8') \
```
tom95858, line 1413: This is a nit, but the decode should probably be xx.decode(encoding='utf-8'). It works like it currently does because encoding happens to be the 1st keyword parameter. If they change that in the future, this could break.

tom95858 (changes requested):

ldms/python/ldms.pyx
nick-enoent: Good point, I just force pushed an update.

PR #921, opened 2022-05-20
fix compid, producer, log level in lustre dynamic plugins

baallan:
reduced log level to 'debug' for checks of the top directory
that will be missing when each lustre service is off (an expected condition
in some deployments).

added component_id and producer option processing from attributes
for consistency with other plugins and updated man pages.

for mdt,ost which do not run on compute nodes, leaving out jobid metric.


baallan:
Operation of these changes was verified on stria at snl.

tom95858:
@morrone this looks fine to me. Could you please review?

tom95858:
I guess this is OK with @morrone, so I will merge it.

PR #924, opened 2022-05-23
Make ldmsd_controller receive a single config command at the cmd-line

nichamon:
With the patch, users can give a single config command at the command line.
--cmd <cfg> is a new ldmsd_controller's command-line option. For example,

ldmsd_controller -x <xprt> -p <port> -h <host> --cmd 'prdcr_status'
ldmsd_controller -x <xprt> -p <port> -h <host> --cmd 'prdcr_start name=node0001'

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -1609,6 +1609,8 @@ if __name__ == "__main__":
         parser.add_argument('--script',
                             help = "Execute the script and send the output \
                             commands to the connected ldmsd")
+        parser.add_argument('--cmd',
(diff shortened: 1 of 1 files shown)


tom95858:
@nichamon, how is this better than just
    echo prdcr_status | ldmsd_controller --...

nichamon:
> @nichamon, how is this better than just echo prdcr_status | ldmsd_controller --...

I don't know whether it's better than echo and pipe method besides the change allowing users to send a single config command without invoking 3 commands. The echo and pipe overhead are insignificant.

I don't do echo and then pipe (for an ancient reason which I forgot what it was), and that's why I added the new command-line option.

PR #925, opened 2022-05-24
Fix a config transport memory corruption

nichamon:
Without the patch, the config transport referenced by ldmsd_req_ctxt objects
points to memory on the stack. LDMSD may access the config transport's
memory after it has been reused.


nichamon:
@tom95858 I found the bug when I implemented the stream info logic.

PR #927, opened 2022-05-24
Allow applications to set LDMS transport's context

nichamon:


nichamon:
@tom95858 The stream_info patch uses the API to verify if an LDMS transport endpoint connects to a producer.

PR #928, opened 2022-05-24
Allow LDMSD interfaces to query stream information

nichamon:


nichamon:
The pull request is on top of #927 and #925.

nichamon:
The patch provides two configuration commands to query the stream information: _stream_dir_ and _prdcr_stream_dir_.

* The _stream_dir_ config message is to query the stream information of an LDSMD.
* The _prdcr_stream_dir_ config message is to query the stream information of the producers matched by a regular expression.

**Examples of the _stream_dir_ response**
* Sending a _stream_dir_ request to a sampler daemon (samplerd2)
image
The sampler daemon received the data of two streams: foo and bar. It subscribed to stream 'foo', but not stream 'bar'. _first msg_ and _last msg_ are the timestamps the sampler daemon has received the first and last messages of the streams. Obviously, there are no producers on the sampler daemons.

   image
   This sampler daemon (samplerd1) received the data from stream 'foo', and it subscribed to it.

* Sending a _stream_dir_ request to an aggregator
image
  The aggregator received the data from stream 'foo' that samplerd1 and samplerd2 republished. The first line contains the overall information of stream 'foo'. The information in subsection 'producers' is the information per producer. For example, the aggregator has received 6 messages: 2 from samplerd1 and 4 from samplerd2.

**An Example of the _prdcr_stream_dir_ response**
* Sending a _prdcr_stream_dir_ request to an aggregator
image
  The aggregator acted as a proxy that forwarded the _stream_dir_ request to the matched producers and aggregated the result. Since samplerd2 didn't subscribe to stream 'bar', it didn't republish the stream 'bar' data to the aggregator. The _stream_dir_ response didn't include stream 'bar'.

nichamon:
In the beginning, Maestro could send the _prdcr_stream_dir_ request to an L1 to get the stream information of each producer. Then, it has the stream names to construct the _prdcr_subscribe_ requests. Later on, it could use _stream_dir_ to get the stream information of an LDMSD or _prdcr_stream_dir_ to the stream information of a set of producers.

PR #930, opened 2022-05-25
Add a thermal sampler that monitors coretemp

nichamon:


nichamon:
@tom95858 I didn't see the pull request so I created it from your branch.

tom95858:
@nichmon, I've modified this to use a list of records for each core. I'll update the branch shortly. One issue is that ldms_ls does not report the metric units in the output. It also formats the list data as JSON. Not sure if that's what we want.

nichmon:
Wrong nichmon!

nichamon:
The pull request was replaced by #931.

PR #931, opened 2022-05-27
Add sampler to collect from the coretemp module

tom95858:
This sampler monitors data from the /sys/dev entries
collected by the coretemp module.

### DIFF
--- a/ldms/src/sampler/coretemp/coretemp.c
+++ b/ldms/src/sampler/coretemp/coretemp.c
@@ -1,0 +1,440 @@
+/* -*- c-basic-offset: 8 -*-
+ * Copyright (c) 2022 National Technology & Engineering Solutions
+ * of Sandia, LLC (NTESS). Under the terms of Contract DE-NA0003525 with
+ * NTESS, the U.S. Government retains certain rights in this software.
+ * Copyright (c) 2022 Open Grid Computing, Inc. All rights
+ * reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the BSD-type
+ * license below:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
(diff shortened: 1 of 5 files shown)


ldms/src/sampler/coretemp/coretemp.c
```diff
@@ -0,0 +1,424 @@
...
+		if (!sens->temp_crit) {
+			fp = fopen(sens->temp_crit_path, "r");
+			rc = fread(fbuf, 1, sizeof(fbuf), fp);
+			fclose(fp);
+			fbuf[rc] = '\0';
+			sens->temp_crit = (float)strtol(fbuf, NULL, 0) / 1000;
+		}
+		/* temp_crit_alarm */
+		if (!sens->temp_crit_alarm_fp)
+			sens->temp_crit_alarm_fp = fopen(sens->temp_crit_alarm_path, "r");
```
nichamon, line 239: Similar to `sens->temp_max_fp`, should we fclose `sens->temp_crit_alarm_fp` after the loop?

tom95858: This is the same as the other one. The reason some files are closed is because their contents never change, e.g. temp_label; which is the text label for the temperature sensor and it never changes. So that file is opened once, read, and then closed. It is never read again. By contrast temp_crit_alarm_fp may change, so it is opened once, and then subsequently fseek(0) before reading.

baallan (changes requested):

ldms/src/sampler/coretemp/coretemp.c
```diff
@@ -0,0 +1,424 @@
...
+		msglog(LDMSD_LERROR, SAMP ": Set already created.\n");
+		return EINVAL;
+	}
+
+	base = base_config(avl, SAMP, SAMP, msglog);
+	if (!base){
+		rc = EINVAL;
+		goto err;
+	}
+
```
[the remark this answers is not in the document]
tom95858: @baallan it seems to me that you would want to know what the max temperature has been and whether or not there has been a critical temperature warning. We are only talking a few metrics here, I don't see the value of adding the additional complexity and associated schema variability.

PR #932, opened 2022-05-27
add 'set_load' to set_stats response

nick-enoent:
calculated by taking the frequency of updater set intervals in seconds
multiplied by the data size of a set

Example Response:

Name                 Count
-------------------- ----------------
active_count                        6
deleting_count                      0
mem_total_kb                      512
mem_free_kb                       478
mem_used_kb                        34
set_load                         3000

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -6352,6 +6352,12 @@ static char * __set_stats_as_json(size_t *json_sz)
 	char *buff, *s;
 	size_t sz = __APPEND_SZ;
 	struct mm_stat stats;
+	int rc;
+	double freq;
+	double set_load = 0;
+	uint32_t data_sz;
+	ldmsd_name_match_t match;
+	ldmsd_updtr_t updtr = NULL;
 
 	(void)clock_gettime(CLOCK_REALTIME, &start);
 	mm_stats(&stats);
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6361,12 +6367,57 @@ static char * __set_stats_as_json(size_t *json_sz)
...
+				for (ref = ldmsd_updtr_prdcr_first(updtr); ref;
+						ref = ldmsd_updtr_prdcr_next(ref)) {
+					ldmsd_prdcr_lock(ref->prdcr);
+					ldmsd_prdcr_set_t prd_set;
+					for (prd_set = ldmsd_prdcr_set_first(ref->prdcr); prd_set;
+							prd_set = ldmsd_prdcr_set_next(prd_set)) {
+						rc = regexec(&match->regex, prd_set->inst_name, 0, NULL, 0);
+						if (rc)
+							continue;
+						freq = 1000000 / (double)prd_set->updt_hint.intrvl_us;
```
nichamon, line 6385: `prd_set->updt_interval` is the current update interval. `prd_set->updt_hint.intervl_us` is the update hint received from the sampler daemon. I guess you want to calculate the current set load. `prd_set->updt_interval` is the value you want to use.

I think both values are the same because we usually set the sample interval and the update interval to 1 second.  The values will be different when the sample interval differs from the update interval.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6361,12 +6367,57 @@ static char * __set_stats_as_json(size_t *json_sz)
...
+			}
+		} else {
+			ldmsd_prdcr_ref_t ref;
+			for (ref = ldmsd_updtr_prdcr_first(updtr); ref;
+					ref = ldmsd_updtr_prdcr_next(ref)) {
+				ldmsd_prdcr_lock(ref->prdcr);
+				ldmsd_prdcr_set_t prd_set;
+				for (prd_set = ldmsd_prdcr_set_first(ref->prdcr); prd_set;
+						prd_set = ldmsd_prdcr_set_next(prd_set)) {
+					freq = 1000000 / (double)prd_set->updt_hint.intrvl_us;
```
nichamon, line 6402: The same here.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -6361,12 +6367,57 @@ static char * __set_stats_as_json(size_t *json_sz)
...
+	}
+	ldmsd_cfg_unlock(LDMSD_CFGOBJ_UPDTR);
+
 	__APPEND("{");
 	__APPEND(" \"active_count\": %d,\n", ldms_set_count());
 	__APPEND(" \"deleting_count\": %d,\n", ldms_set_deleting_count());
 	__APPEND(" \"mem_total_kb\": %g,\n", (double)stats.size / 1024.0);
 	__APPEND(" \"mem_free_kb\": %g,\n", (double)(stats.bytes * stats.grain) / 1024.0);
 	__APPEND(" \"mem_used_kb\": %g,\n", (double)(stats.size - (stats.bytes * stats.grain)) / 1024.0);
+	__APPEND(" \"set_load\": %g,\n", set_load);
```
nichamon, line 6420: Let's add the unit of set_load. Is it bytes/sec? Do you want to make it to be kb/sec?

tom95858: The other numbers are kb, but maybe the set_load name should imply the units. Also, seems like we mix kb as 1024 and 1000. We should be consistent.

PR #933, opened 2022-05-31
Decomposition

narategithub:
- Implement `decomposition` feature (set-row conversion)
- Modify `store_sos` and `store_csv` to support decomposition
- Implement `store_kafka` (require decomposition / no traditional config support)


narategithub:
@tom95858 This has passed the decomposition test in https://github.com/narategithub/ldms-test/blob/decomp/ldmsd_decomp_test and all other tests, except `setgroup_test`. `OVIS-4` also failed `setgroup_test`, so I think the bug is already in the `OVIS-4` branch and not directly related to this patch. I'll debug `setgrup_test` and submit a separate patch (I'm guessing a change related to setinfo might have broken the setgroup).

PR #934, opened 2022-06-01
Reformat ldms_ls's lists of record and record array outputs

nichamon:


nichamon:
With the change, ldms_ls output for the coretemp set is as follows. It's an example of a list of records.
image

softirq_list is an example of a list of a primitive type.
image

nichamon:
@tom95858 The record entry unit is printed at the header line.

tom95858:
This looks awesome...

PR #935, opened 2022-06-01
Fix store_csv SIGSEGV due to missing mval assignment

narategithub:


narategithub:
@tom95858 this patch fix a store_csv bug discovered by [ldms-test/store_list_record_test](https://github.com/ovis-hpc/ldms-test/blob/master/store_list_record_test) (just merged into the ldms-test repo). The bug was introduced by the decomposition changes.

PR #940, opened 2022-06-06
fix double cleanup() of whatever source

baallan:
At least one way to invoke double cleanup and assorted memory
errors is to invoke daemon_exit in a script.
This protects cleanup with a mutex and a 'cleaned' flag.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -469,8 +469,15 @@ double ldmsd_timeval_diff(struct timeval *start, struct timeval *end)
 
 extern void ldmsd_strgp_close();
 
+static pthread_mutex_t cleanup_lock = PTHREAD_MUTEX_INITIALIZER;
+static int cleaned;
 void cleanup(int x, const char *reason)
 {
+	pthread_mutex_lock(&cleanup_lock);
+	if (cleaned) {
+		pthread_mutex_unlock(&cleanup_lock);
+		exit(x);
+	}
 	int llevel = LDMSD_LINFO;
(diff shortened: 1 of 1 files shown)


nichamon:
@baallan Could you explain the symptom? Did multiple threads call cleanup() due to a SIGTERM/SIGKILL?

baallan:
send the daemon_exit command with controller or ldmsctl and look at the daemon with valgrind.

tom95858:
@baallan, I cannot reproduce this using either valgrind or gdb. I only see it being called once. @nichamon and I looked at the code and the only way it is called is from the main thread or through the signal handler. But even if it could be reproduced, using a mutex when a process is exiting can result in a hang because the memory in which the mutex is located could be unmapped (i.e. contain garbage). Could you be more specific about the symptom?

tom95858:
@baallan, if it's in 'main', then it's probably. fine. But everything in the process is actually mapped using mmap. You can see them if you look in /proc/<pid>/maps. When the process is exiting, all of those maps are going away. The heap memory, etc...all mapped. The order in which those maps are undone is not defined. In main, you are probably fine, but in the shared libraries, perhaps not. But I'm still asking ... what symptom is this actually fixing? Does it SEGV or just give valgrind heartburn?

baallan:
@tom95858 valgrind gets very clear heartburn and often (but of course not always) w/out valgrind we get an assert or seg fault usually around one of the rbt trees..

tom95858:
@baallan confirms that this happens in his test script when he issues the daemon_exit command and then concurrently sends SIGTERM.

PR #944, opened 2022-06-08
add lustre_mdc (client side metadata perf) sampler

baallan:
Includes jobid and set perm support, basic test script.
The sampler can run without root access if timing data is excluded.
The mdc timing stats can be suppressed optionally.

### DIFF
--- a/ldms/src/sampler/lustre_mdc/NOTES
+++ b/ldms/src/sampler/lustre_mdc/NOTES
@@ -1,0 +1,76 @@
+Read data is found in (for lustre 2.12)
+
+FILE (operation counts)
+
+/proc/fs/lustre/mdc/*/md_stats
+snapshot_time             1653661756.400269074 secs.nsecs
+close                     3485 samples [reqs]
+create                    4694 samples [reqs]
+getattr                   34 samples [reqs]
+intent_lock               990613 samples [reqs]
+rename                    13906 samples [reqs]
+setattr                   12082 samples [reqs]
+read_page                 6102 samples [reqs]
+unlink                    3 samples [reqs]
+intent_getattr_async      23075 samples [reqs]
+revalidate_lock           79295 samples [reqs]
+
+and these values are defined in kernel code:
+
+lustre/lustre/obdclass/lprocfs_status.c
+lustre/lustre/include/obd_class.h
+mps_stats[], enum mps_stat_idx
+most recently (2.12 lustre) seen to be:
+	[LPROC_MD_CLOSE]		= "close",
+	[LPROC_MD_CREATE]		= "create",
+	[LPROC_MD_ENQUEUE]		= "enqueue",
+	[LPROC_MD_GETATTR]		= "getattr",
+	[LPROC_MD_INTENT_LOCK]		= "intent_lock",
+	[LPROC_MD_LINK]			= "link",
+	[LPROC_MD_RENAME]		= "rename",
+	[LPROC_MD_SETATTR]		= "setattr",
+	[LPROC_MD_FSYNC]		= "fsync",
+	[LPROC_MD_READ_PAGE]		= "read_page",
+	[LPROC_MD_UNLINK]		= "unlink",
+	[LPROC_MD_SETXATTR]		= "setxattr",
+	[LPROC_MD_GETXATTR]		= "getxattr",
+	[LPROC_MD_INTENT_GETATTR_ASYNC]	= "intent_getattr_async",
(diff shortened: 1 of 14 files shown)


ldms/src/sampler/lustre_mdc/lustre_mdc.c
```diff
@@ -0,0 +1,288 @@
...
+		goto out1;
+	mdc->name = strdup(mdc_name);
+	if (mdc->name == NULL)
+		goto out2;
+	snprintf(path_tmp, PATH_MAX, "%s/%s", basedir, mdc_name);
+	mdc->path = strdup(path_tmp);
+	if (mdc->path == NULL)
+		goto out3;
+	snprintf(path_tmp, PATH_MAX, "%s/md_stats", mdc->path);
+	mdc->stats_path = strdup(path_tmp);
```
morrone, line 71: I think this naming is going to be very confusing in the future. We should not rename "md_stats" to "stats" and rename "stats" to "timing". "stats" is (almost?) always a file with a standardized format.

In other words, I think we would be better off with "mdc->md_stats_path" and "mdc->stats_path".

morrone (commented):
I would like to see you rename this commit/PR to remove the string "client side mds perf". MDS means "MetaData Server". MDC means "MetaData Client". So you might say "client side metadata perf", but "client side mds" is like saying "client side server side".

I am not a big fan of combining the */md_stats data with the */stats data into a single schema. I am even less a fan of that schema changing just because a path is readable or not. I would personally like to see separate schemas and metric sets for each.

Separating them might also be helpful for future development. "stats" is a standardized formatting approach in lustre, and we might (with the help of the new records work) be able to write a sampler that can use a standard schema that would work for _all_ "stats" files. But I'm not 100% sure that is workable yet.

Should this sampler be named "lustre_mdc_sampler"? Did we decided that all new samplers needed to end in "_sampler"?

ldms/src/sampler/lustre_mdc/NOTES
```diff
@@ -0,0 +1,66 @@
+Read data found in
```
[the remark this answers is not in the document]
baallan: I do. It's aimed at maintainers rather than users, so not included in man page. Hoping to speed up things for whoever updates when lustre changes its metrics again.

ldms/src/sampler/lustre_mdc/lustre_mdc.c
baallan: I can split the set, but I did not initially because:
 * there's no way to specify that I want set A at interval one and set B at interval two.
 * it adds transmission overhead
 * both sets carry the same metadata including device for the metrics
 
I can certainly adjust the schema name dependent on its content (mdc_ops or mdc_ops_timing) so there is no backend heartache based on config params.

Until we bump to a major version change and rename all the samplers, i think we only need/want the _sampler suffix on samplers which need it to avoid conflicts by having libxxx_sampler vs libxxx (like dcgm). but that's just my opinion.

baallan:
marked as draft while i work on changes.

ldms/src/sampler/lustre_mdc/lustre_mdc.c
morrone: > stats files seems to be standard on in the sense of a column format rather than a file format, where content varies wildly row to row and there are already examples of abusing the column format.

I'm pretty sure that "stats" is generated by the same function in Lustre regardless of its location. Where do you think you see it changing the column format?

> I will look into the internal renaming of things so that variable names line up with file leaf names in the specific mdc context. At the user level, 'stats' is apparently meaningless in the sense that it's a catch-all file name that means anything the context (lustre subsystem) needs it to mean. To me, 'timing' is much more specific and meaningful.

I'm talking about in the code. Anyone working on the lustre sampler code would presumably need to have some familiarity with lustre naming. And no on familiar with lustre calls will recognized "timing" as a specific in lustre. A "stats" file in lustre is a well-known format in the lustre world.

baallan: I have renamed vars/members  in c to correspond to basename of the file being read and adjusted schema name depending on content.

baallan:
rebased on ovis-4.

baallan:
rebased to current top of tree; no new changes.

tom95858:
@baallan is this ready to merge?

baallan:
@tom95858 it's good enough for me. runs as documented and clean in valgrind.

tom95858:
@morrone, are you happy with this?

morrone:
I don't have any objection.

PR #945, opened 2022-06-08
add option to suppress - in output of ldms-sensors-config

baallan:
updates utility script to optionally suppress - in the labels/names published by the kernel.


baallan:
@tom95858 this pr addresses the metric name '-' problem for the filesingle sampler config generator (only if the user requests it, so as not to break existing schema unintentionally).

Issue #938, opened 2022-06-06
config now allowing load errors


baallan:
Previously, a plugin load failing would stop further interpretation of the input and exit the daemon.
Presently (top of OVIS-4:  02ac5119)  the code runs through sampler plugin load errors.
The prior behavior during misconfiguration needs to be restored.
e.g..

```
Mon Jun 06 12:37:38 2022: INFO      : Listening on sock:61078 using `sock` transport and `none` authentica
tion
Mon Jun 06 12:37:38 2022: DEBUG     : Checking for jobid in /scratch2/baallan/ovis/LDMS_install/lib/ovis-l
dms
Mon Jun 06 12:37:38 2022: ERROR     : Failed to load the plugin 'jobid': dlerror /scratch2/baallan/ovis/LD
MS_install/lib/ovis-ldms/libjobid.so: cannot open shared object file: No such file or directory
Mon Jun 06 12:37:38 2022: DEBUG     : msg_no 15 flags 3 rec_len 204 rsp_err -1
Mon Jun 06 12:37:38 2022: ERROR     : msg_no 15: error -1: Failed to load the plugin 'jobid'. dlerror /scr
atch2/baallan/ovis/LDMS_install/lib/ovis-ldms/libjobid.so: cannot open shared object file: No such file or
 directory
Mon Jun 06 12:37:38 2022: DEBUG     : msg_no 16 flags 3 rec_len 76 rsp_err 2
Mon Jun 06 12:37:38 2022: ERROR     : msg_no 16: error 2: Plugin 'jobid' configuration error.
Mon Jun 06 12:37:38 2022: DEBUG     : msg_no 17 flags 3 rec_len 67 rsp_err 2
Mon Jun 06 12:37:38 2022: ERROR     : msg_no 17: error 2: Sampler 'jobid' not found.
Mon Jun 06 12:37:38 2022: DEBUG     : Checking for meminfo in /scratch2/baallan/ovis/LDMS_install/lib/ovis
-ldms
Mon Jun 06 12:37:38 2022: DEBUG     : msg_no 18 flags 3 rec_len 28 rsp_err 0
Mon Jun 06 12:37:38 2022: DEBUG     : msg_no 19 flags 3 rec_len 28 rsp_err 0
Mon Jun 06 12:37:38 2022: DEBUG     : msg_no 20 flags 3 rec_len 28 rsp_err 0
[...]

PR #949, opened 2022-06-15
Fix ldmsd config error but not exit bug

narategithub:
The commit c4f19c6 allocated a new config transport memory because the
`ldmsd_req_ctxt` can later corrupt the transport that was in the stack.
The subsequent request handling call will modify `xprt->rsp_err` if
there is an error. However, it will be modifying the newly allocated
xprt memory, leaving the config transport unchanged. Consequently, the
`rsp_err` in the config transport (on the stack) stayed 0.

This patch addresses the issue by copying the `xprt` content that is
modified by the subsequent config handler call back to the config
transport.


baallan:
#938 may be fixed by this

nichamon (approved):

PR #950, opened 2022-06-15
remove symlink libdcgm.so

baallan:
It doesn't help back compat, due to the problem fixed by #865,
and it breaks the build in some development workflows
(make install after another make install has happened).


tom95858:
@morrone is this OK with you?

PR #952, opened 2022-06-16
Improve the as_is decomp's error messages

nichamon:


narategithub (approved):
👍

ldms/src/decomp/as_is/decomp_as_is.c
```diff
@@ -256,20 +256,26 @@ __decomp_as_is_config(ldmsd_strgp_t strgp, json_entity_t jcfg,
 	TAILQ_FOREACH(jidx, &jidxs->item_list, item_entry) {
 		didx = &dcfg->idxs[j];
 		if (jidx->type != JSON_DICT_VALUE) {
-			DECOMP_ERR(reqc, EINVAL, "an index must be an object\n");
+			DECOMP_ERR(reqc, EINVAL, "strgp '%s': index '%d': "
```
tom95858, line 259: I think something more like "expected the index list entries to be dictionaries." I don't know that users know what an "object" is in the context of JSON.

nichamon:
@tom95858 I have addressed your comments.

PR #956, opened 2022-06-16
fix races on static buffers in ldms.c json formatting

baallan:
Functions dir_update under ldms_set_publish and process_dir_request
can both be using __ldms_format_set_meta_as_json at the
same time. In this patch, making the format strings for state, perm stack
instead of static makes this safe and the data output correct.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -309,10 +309,21 @@ struct get_set_names_arg {
  * }
  */
 
-static const char *perm_string(uint32_t perm)
+
+#define STATE_BUF_SZ 4
+#define PERM_BUF_SZ 16
+size_t __ldms_format_set_meta_as_json(struct ldms_set *set,
+				      int need_comma,
+				      char *buf, size_t buf_size)
 {
-	static char str[16];
-	char *s = str;
+	size_t cnt;
+	char dbuf[2*LDMS_DIGEST_LENGTH+1];
+	ldms_digest_t digest = ldms_set_digest_get(set);
+
+	/* format perm */
+	uint32_t perm = __le32_to_cpu(set->meta->perm);
+	char perm_str[PERM_BUF_SZ];
+	char *s = perm_str;
 	int i;
 	*s = '-';
 	s++;
@@ -335,37 +346,25 @@ static const char *perm_string(uint32_t perm)
 		s++;
 	}
 	*s = '\0';
-	return str;
-}
 
-char *set_state(struct ldms_set *set)
-{
-	static char str[8];
+	/* format state flags */
+	char state[STATE_BUF_SZ];
 	if (set->data->trans.flags == LDMS_TRANSACTION_END)
-		str[0] = 'C';
+		state[0] = 'C';
 	else
-		str[0] = ' ';
+		state[0] = ' ';
 	if (set->flags & LDMS_SET_F_LOCAL)
-		str[1] = 'L';
+		state[1] = 'L';
 	else if (set->flags & LDMS_SET_F_REMOTE)
-		str[1] = 'R';
+		state[1] = 'R';
 	else
-		str[1] = ' ';
+		state[1] = ' ';
 	if (set->flags & LDMS_SET_F_PUSH_CHANGE)
-		str[2] = 'P';
+		state[2] = 'P';
 	else
-		str[2] = ' ';
-	str[3] = '\0';
-	return str;
-}
+		state[2] = ' ';
+	state[3] = '\0';
 
-size_t __ldms_format_set_meta_as_json(struct ldms_set *set,
-				      int need_comma,
-				      char *buf, size_t buf_size)
-{
-	size_t cnt;
-	char dbuf[2*LDMS_DIGEST_LENGTH+1];
-	ldms_digest_t digest = ldms_set_digest_get(set);
 	cnt = snprintf(buf, buf_size,
 		       "%c{"
 		       "\"name\":\"%s\","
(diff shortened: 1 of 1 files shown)


ldms/src/core/ldms.c
```diff
@@ -338,9 +339,10 @@ static const char *perm_string(uint32_t perm)
 	return str;
 }
 
-char *set_state(struct ldms_set *set)
+#define STATE_BUF(x) char x[8]
```
nichamon, line 342: Similar here.

nichamon, line 342: same here

ldms/src/core/ldms.c
```diff
@@ -366,6 +368,8 @@ size_t __ldms_format_set_meta_as_json(struct ldms_set *set,
 	size_t cnt;
 	char dbuf[2*LDMS_DIGEST_LENGTH+1];
 	ldms_digest_t digest = ldms_set_digest_get(set);
+	STATE_BUF(ss);
+	PERM_BUF(ps);
```
nichamon, line 372: The lines will become.
```
char ss[8];
char ps[16];
```

nichamon, line 372: The lines will become.
```
char ss[8];
char ps[16];
```

nichamon (changes requested):
Thanks for catching this. 

By the way, what was the symptom you saw? I might see the symptom before but I didn't realize that it was caused by the race.

baallan:
> Thanks for catching this.
> 
> By the way, what was the symptom you saw? I might see the symptom before but I didn't realize that it was caused by the race.

The symptom would be very hard to spot visually since the race between two calls of set_state or perm formatting will always be syntactically valid-- it just may not match the either set being printed.
I spotted it with DRD testing.

ldms/src/core/ldms.c
```diff
@@ -309,9 +309,10 @@ struct get_set_names_arg {
  * }
  */
 
-static const char *perm_string(uint32_t perm)
+#define PERM_BUF(x) char x[16]
```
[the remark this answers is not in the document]
baallan: @nichamon It is a common misconception that you can define and have the compiler check string sizes with the syntax you propose. The compiler will happily compile { char x[2]; uint32_t p = 0; perm_string(p, x); }, even in pedantic mode and leave you to enjoy your buffer overrun.

C does, however provide a solution for fixed array sizes: { typedef char(*b_t)[14]; int foo(b_t buf) {...} }. This solution has some annoying side effects (casting where buf is used inside foo), but it does let the compiler trap (warn) about the undersized  buffer in  { char x[2]; uint32_t p = 0; perm_string(p, x); }.  I prefer the macro version, but I will switch to the typedef since it enables compiler checking instead of reading comments or inferring from the function signature.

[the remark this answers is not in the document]
baallan: @nichamon I removed the macro-concealed type declaration.

baallan:
@nichamon while i've fixed the race on the buffers, there's still fundamentally a race between  process_dir_request(PDR) and dir_update(DU) nested inside ldmsd_set_register (Below). Is there a lock on something that should be taken and isn't or a mis-ordering of insertion into some data structures such that PDR is seeing the set before DU is done?
```
==9254==    by 0x4E5E2DD: dir_update (ldms_xprt.c:496)
==9254==    by 0x4E5E538: __ldms_dir_upd_set (ldms_xprt.c:553)
==9254==    by 0x4E576BD: ldms_set_info_set (ldms.c:3497)
==9254==    by 0x40BA92: ldmsd_set_update_hint_set (ldmsd.c:995)
==9254==    by 0x40BC50: ldmsd_set_register (ldmsd.c:937)
```

ldms/src/core/ldms.c
```diff
@@ -309,9 +309,11 @@ struct get_set_names_arg {
  * }
  */
 
-static const char *perm_string(uint32_t perm)
+#define PERM_BUF_SZ 16
+typedef char(*perm_buf)[PERM_BUF_SZ];
```
tom95858, line 313: Just pass in a char *. What's the point of the typedef?

ldms/src/core/ldms.c
```diff
@@ -309,9 +309,11 @@ struct get_set_names_arg {
...
+static const char *perm_string(uint32_t perm, perm_buf pstr)
```
tom95858, line 314: Pass in a char *. This function is called in one place and always consumes exactly 10B.

ldms/src/core/ldms.c
```diff
@@ -338,9 +340,11 @@ static const char *perm_string(uint32_t perm)
 	return str;
 }
 
-char *set_state(struct ldms_set *set)
+#define STATE_BUF_SZ 8
+typedef char(*state_buf)[STATE_BUF_SZ];
+static const char *set_state(struct ldms_set *set, state_buf sstr)
```
tom95858, line 345: Same here pass in the char *, and the return type should not be const. The caller is passing in the buffer, and the function is modifying it.

ldms/src/core/ldms.c
```diff
@@ -366,6 +370,8 @@ size_t __ldms_format_set_meta_as_json(struct ldms_set *set,
 	size_t cnt;
 	char dbuf[2*LDMS_DIGEST_LENGTH+1];
 	ldms_digest_t digest = ldms_set_digest_get(set);
+	state_buf ss = alloca(sizeof *ss);
+	perm_buf ps = alloca(sizeof *ps);
```
tom95858, line 374: This should just be char ss[4], and char ps[16] or whatever you want. But all of this extra syntax is just confusing for other people wondering what all the fuss is about.

ldms/src/core/ldms.c
baallan: @tom95858 You've hit on the core of the problem which is that some simple formatting was extracted into single-use functions; this introduced the races and the opportunity to mismatch buffer sizes.
How about I just move the formatting back where it belongs in __ldms_format_set_meta_as_json?

tom95858: That is fine with me.

PR #958, opened 2022-06-23
add performance benchmark of jbuf operation

baallan:
The overhead for using jbuf to assemble json messages in darshan (relative to sprintf) is roughly 3.7. At high message rates this matters, so I'm adding a benchmark we can investigate and monitor for performance improvement or regression.
The benchmark runs with a message during 'make installcheck'

### DIFF
--- a/lib/src/ovis_json/Makefile.am
+++ b/lib/src/ovis_json/Makefile.am
@@ -9,6 +9,7 @@ AM_YFLAGS = -d
 
 EXTRA_DIST = ovis_json_lexer.l ovis_json_parser.y
 
+
 CLEANFILES = ovis_json_parser.h ovis_json_parser.c ovis_json_lexer.c
 
 ovis_json_parser.c: ovis_json_parser.h ovis_json.h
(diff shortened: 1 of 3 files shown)


tom95858:
@baallan, could you provide the Darshan input files? It seems like we can't run this test without them.

tom95858:
@baallan never mind, I see you generate the data on the fly.

tom95858:
@baallan, if you convert the jbuf_append_attr to jbuf_append_str, the jbuf version runs a bit faster than sprintf. Also, the sprintf version doesn't malloc and free each record and could overflow while the the jbuf API will realloc the buffer if it 'overflows', your sprintf version doesn't do that.

Here's the results with the refactored json buffer side:
```
    1000000 sprintf time us 1.81507e+06
    1000000 jbuf time us    1.78475e+06
```

So I suggest that the Darshan connector be refactored to use the alternate API, specifically:

```c
	jb = jbuf_append_str(jb, "{ \"uid\":%" PRId64 ", \"exe\":\"%s\",\"job_id\":%" PRId64
		",\"rank\":%" PRIu64 ",\"ProducerName\":\"%s\",\"file\":\"%s\",\"record_id\":%"
		PRIu64",\"module\":\"%s\",\"type\":\"%s\",\"max_byte\":%" PRId64
		",\"switches\":%" PRId64 ",\"flushes\":%" PRId64 ",\"cnt\":%" PRId64
		",\"op\":\"%s\",\"seg\":[{\"data_set\":\"%s\",\"pt_sel\":%" PRId64
		",\"irreg_hslab\":%" PRId64 ",\"reg_hslab\":%" PRId64 ",\"ndims\":%" PRId64
		",\"npoints\":%" PRId64 ",\"off\":%" PRId64 ",\"len\":%" PRId64
		",\"dur\":%0.6f,\"timestamp\":%lu.%06lu}]}",
	dC.uid,
	dC.exename,
	dC.jobid,
	dC.rank,
	hname,
	dC.filename,
	dC.record_id,
	mod_name,
	data_type,
	max_byte,
	rw_switch,
	flushes,
	record_count,
	rwo,
	dC.data_set,
	dC.hdf5_data[0],
	dC.hdf5_data[1],
	dC.hdf5_data[2],
	dC.hdf5_data[3],
	dC.hdf5_data[4],
	offset,
	length,
	total_time,
	tspec_end.tv_sec,
	micro_s);
	if (!iter)
		printf("%s\n", jb->buf);
	jbuf_free(jb);
	return 0;
```

baallan:
@tom95858 that's interesting news. @Snell1224 and Omar please take note and try it out.

Re memory buffer allocation effects and a static buffer: that is something to consider. we noted in the darshan application that there is an upper bound and take advantage of it. I do wonder about multithreading issues, but I haven't had time to inspect darshan's thread strategy.

PR #965, opened 2022-06-29
Enable logging for legacy stores where a set schema has different definitions

nichamon:
A storage policy maps to a set schema and assumes all sets of the same schema
have the same metric list. If this is not true, due to misconfiguration or
multiple architectures on the system, LDMSD may crash when it tries to store
sets.

With the patch, LDMSD logs a message when it finds a set that has a different
schema digest from its storage policy's cached digest. Also, LDMSD does not
store such sets. In another word, LDMSD will store sets that a storage policy
has cached its digest.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.h
+++ b/ldms/src/ldmsd/ldmsd.h
@@ -459,6 +459,9 @@ struct ldmsd_strgp {
 	/** Schema name of the metric set on the producer */
 	char *schema;
 
+	/** The digest of the first set used in opening the store */
+	ldms_digest_t digest;
+
 	/** The container name in which the storage backend will place data */
 	char *container;
 
--- a/ldms/src/ldmsd/ldmsd_strgp.c
+++ b/ldms/src/ldmsd/ldmsd_strgp.c
@@ -94,6 +94,7 @@ void ldmsd_strgp___del(ldmsd_cfgobj_t obj)
 		free(strgp->plugin_name);
 	if (strgp->decomp_name)
 		free(strgp->decomp_name);
+	free(strgp->digest);
 	ldmsd_cfgobj___del(obj);
 }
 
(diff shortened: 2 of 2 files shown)


baallan:
As currently implemented, this change has problems:
 * it does not account for storage policy metric lists (which may make incongruent sets congruent for purposes of the store policy)
 * it assumes that all store plugin implementations care about precisely matching the schema as it was digested by the ldmsd strgp framework, causing plugins which have a different matching requirement to lose data and causing spurious log messages about mismatches.

The proper conditions of schema matching can really only be evaluated by each specific store plugin. All we need to implement at the framework level is a way for each store plugin instance to evaluate and store its own digest (accounting for the strgp metric list) on the local set instance (more generally, the store plugin should be able to allocate and register a void pointer cached in each local set instance of concern.  These pointers should not be transmitted to other nodes/daemons. If we believe there is more than one plugin sharing a common digest rule, then common helper functions for each such rule can be defined and shared.

nichamon:
@baallan **This change is mainly to prevent LDMSD from crashing due to storage policies' misconfiguration and report the cause.** The crashes had happened many times in the past when a system consists of different architectures or configurations that lead to different metrics in the proc files, such as meminfo and vmstat. **By the design of legacy storage policies, all sets that are stored using a storage policy must have the same set schema definition.** In other words, a storage policy maps to an LDMS schema digest. The schema digests used in the comparison have one-to-one correspondences to the schema definitions. For example, if `/proc/meminfo/` on the Haswell and KNL machines have different numbers of metrics, you need different storage policies for the meminfo on the Haswell machines and another storage policy for the meminfo on the KNL machines.

The restriction is because LDMSD passes an array of metric indices to store plugins' `store()` interface. The metric indices are used instead of a metric name list because using indices does not require metric lookup, resulting in a better performance than using a metric name list.

I understand that each storage plugin has its own limitations and assumptions. Having two different digests does not mean that all store plugins cannot successfully store the metrics. However, comparing schema digests of two sets will definitely prevent the crashes you have seen so far. Two sets having different digests means that their schema definitions are not the same. Thus, the metric indices of the two schemas may refer to different metrics. Also, the error message is not invalid. I could add a reminder to check the storage policy configuration.

What you want is a new feature for the legacy storage policy which is to relax the restriction, which is not in the scope of the patch. That being said, we just introduced the decomposition storage policy that is more flexible than the legacy one, in the sense that users do not require to define two different storage policies for the meminfo on the Haswell and on the KNL machines.

PR #967, opened 2022-06-30
Add the aggregated stream info to stream_dir

nichamon:


nichamon:
The `stream_dir` response includes the aggregated information with the key '_AGGREGATED_'.
image

nichamon:
The `prdcr_stream_dir` command also reports the aggregated info of each producer. The data under the _AGGREGATED_ name is the aggregated info of each producer.
image
To obtain the aggregated info of an aggregator, users will send the `stream_dir` command to the aggregator.

PR #969, opened 2022-07-01
Fix ldms_set_new_custom() that accesses a NULL pointer

nichamon:
The patch fixes the bug that ldms_set_new_custon() accesses the set
after __record_set() returns NULL.


tom95858:
@nichamon I think we need a new name for this interface.

tom95858:
@nichamon, can we rename this interface ldms_set_create( )? It's not in extensive use...yet and this name feels a bit clumsy.

nichamon:
> @nichamon, can we rename this interface ldms_set_create( )? It's not in extensive use...yet and this name feels a bit clumsy.

@tom95858 I renamed the API as you suggested. Sorry, it took me so long.

PR #971, opened 2022-07-05
fixes #957 race on __ldms_format_set_meta_as_json

### LINKED ISSUE #957: A race in the process_dir_update() and ldmsd_set_register()
@nichamon while i've fixed the race on the buffers, there's still fundamentally a race between  process_dir_request(PDR) and dir_update(DU) nested inside ldmsd_set_register (Below). Is there a lock on something that should be taken and isn't or a mis-ordering of insertion into some data structures such that PDR is seeing the set before DU is done?
```
==9254==    by 0x4E5E2DD: dir_update (ldms_xprt.c:496)
==9254==    by 0x4E5E538: __ldms_dir_upd_set (ldms_xprt.c:553)
==9254==    by 0x4E576BD: ldms_set_info_set (ldms.c:3497)
==9254==    by 0x40BA92: ldmsd_set_update_hint_set (ldmsd.c:995)
==9254==    by 0x40BC50: ldmsd_set_register (ldmsd.c:937)
```

_Originally posted by @baallan in https://github.com/ovis-hpc/ovis/issues/956#issuecomment-1159249695_

baallan:
__ldms_format_set_meta_as_json has an assumption of a lock on the set,
but one of the call paths to __ldms_format_set_meta_as_json was missing
that lock. ldmsd_set_register (descendants) and ldms_transaction_end are
some of the other sites using it.

### DIFF
--- a/ldms/src/core/ldms_private.h
+++ b/ldms/src/core/ldms_private.h
@@ -181,6 +181,10 @@ extern int __ldms_get_local_set_list(struct ldms_name_list *head);
 extern void __ldms_empty_name_list(struct ldms_name_list *name_list);
 
 extern void __ldms_dir_update(ldms_set_t set, enum ldms_dir_type t);
+/* format set meta info (not metrics) into buf.
+ * \return count of characters added.
+ * NOTE: set->lock mutex must be held before this is called.
+ */
 extern size_t __ldms_format_set_meta_as_json(struct ldms_set *set,
 					     int need_comma,
 					     char *buf, size_t buf_size);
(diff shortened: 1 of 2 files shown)


tom95858:
@baallan, I believe the missing lock here is intentional. The set->lock is a hot-lock to the extent that it will hold off set updates and other important latency sensitive activities and the function in question is a bit of a pig.  But it is possible to receive a dir request and get the last transaction time, etc... and/or a duration that is not for the posted transaction time. Given that these values are informational for the purpose of ldms_ls; absent an actual problem or bug, I'm not inclined to add this lock. If anyone can provide a use-case that is a bigger problem than introducing update jitter, I'm happy to reconsider.

PR #972, opened 2022-07-07
Add support for sampling DAOS telemetry

mjmac:
Adds a sampler for DAOS (https://github.com/daos-stack/daos),
a high-performance object storage system.

Signed-off-by: Michael MacDonald <mjmac.macdonald@intel.com>


mjmac:
@baallan, @morrone: Apologies for dropping this out of the blue... Is there a particular process that should be followed for requesting review/landing of this PR?

baallan:
@tom95858 handles the pull requests. If you can add any information here in the comments about what OS versions daos is supported on and what kind of testing has been applied to the sampler (e.g. valgrind and clang or gcc static code audit) to ensure it works in a long-running multithreaded environment without leaking memory, file descriptors, etc., that would be helpful. 
The tricky cases are typically around erroneous config arguments and around stopping and reconfiguring and restarting the sampler without shutting down the ldmsd.

PR #973, opened 2022-07-07
Update sos_begin_x() calls

narategithub:
sos commit ID 1de8aa5 changed `sos_begin_x()` API. This patch modify the
calls to match the changed API.


tom95858:
Hi @narategithub, the new top of tree is SOS-6. I made the mistake of changing sos_begin_x to take a timeout parameter instead of adding a new interface sos_begin_x_wait. So we should close this and add the appropriate docs so people know that the top of tree OVIS-4 needs SOS-6.

narategithub:
@tom95858 Got it.

PR #975, opened 2022-07-08
Not calling assert() in zap_sock:sock_read() in the ZAP_EP_CLOSE case

nichamon:
We can arrive in sock_read() when the endpoint is in the ZAP_EP_CLOSE
state if the EPOLLIN events arrived between the endpoint state changing
to ZAP_EP_CLOSE and the connection shutdown, so we should not call
assert().

### DIFF
--- a/lib/src/zap/sock/zap_sock.c
+++ b/lib/src/zap/sock/zap_sock.c
@@ -1324,8 +1324,20 @@ static void sock_read(struct epoll_event *ev)
 			/* good */
 			break;
 		case ZAP_EP_CLOSE:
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 We looked into the sock_read() path because @bschwal saw that `zap_sock()` asserted() at line 1335 in the ZAP_EP_CLOSE based on v4.3.8. However, the assert() call was protected with the `__zap_assert` flag without the patch. If @bschwal didn't build with -DDEBUG or change the `__zap_assert` value, applying this patch could mask the real problem.

@bschwal Did you build OVIS with -DDEBUG or change the value of `__zap_assert` variable?

PR #979, opened 2022-07-12
fix missing error check & init to prevent uninit pointer use

baallan:
ldmsd_stream_publish when invoked with incomplete/incorrect options can use a null ldms stream pointer.

### DIFF
--- a/ldms/src/ldmsd/test/ldmsd_stream_publish.c
+++ b/ldms/src/ldmsd/test/ldmsd_stream_publish.c
@@ -32,6 +32,7 @@ static struct option long_opts[] = {
 	{0,          0,                 0,  0 }
 };
 
+void usage(int argc, char **argv) __attribute__((noreturn));
 void usage(int argc, char **argv)
 {
 	printf("usage: %s -x <xprt> -h <host> -p <port> "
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/test/ldmsd_stream_publish.c
```diff
@@ -32,6 +32,7 @@ static struct option long_opts[] = {
 	{0,          0,                 0,  0 }
 };
 
+void usage(int argc, char **argv) __attribute__((noreturn));
```
tom95858, line 35: void implies that there is no return value. what is this for?

baallan: noreturn means the code exits, not that it is void.
this enables correct determination at the higher scope that a code path terminates after a usage call rather than continuing an using uninit variables. the gcc analyzer doesn't dive down into usage and discover the exit() or abort or whatever else.

nichamon:
@baallan What symptom does this pull request fix? The patch broke `ldmsd_stream_publish` with the `-f` option.

nichamon:
@tom95858 PR #990 fixes the problem caused by this patch.

PR #986, opened 2022-07-15
Add LDMS CSV data quality checker

baallan:
Detector for interval, duplicates, and dropouts in store_csv output


baallan:
@bschwal I've put the store_cvs dup/hole quality checker into the ovis repo so you can run it as needed on other csv files for comparison to sosdb. It can also be compiled without the full make system as noted in the source code. see ldms/src/store/test/csv_time_drops.cxx.

Issue #981, opened 2022-07-12
ldms/src/decomp/static/decomp_static.c:956: dcol used uninit


baallan:
The error message in decomp_static appears to have been cloned from the make_rows region but dcol is used uninit afaict, likely leading to confusing or seg faulting messages when debugging configuration errors.

tom95858:
@nichamon, @baallan is this resolved?

tom95858:
@baallan please reopen if necessary.

PR #989, opened 2022-07-19
Fix a msg in decomp_stat.c that accesses an uninit var (#981)

nichamon:


nichamon:
@narategithub Could you please check the log message whether it's correct?

ldms/src/decomp/static/decomp_static.c
```diff
@@ -953,13 +953,12 @@ static int __decomp_static_decompose(ldmsd_strgp_t strgp, ldms_set_t set,
...
-				ldmsd_lerror("strgp '%s': row '%d' col[dst] '%s': "
-					     "the value type (%s) is not "
-					     "compatible with the source metric type (%s). "
-					     "Please check the decomposition configuration.\n",
-					     strgp->obj.name, i, dcol->dst,
-					     ldms_metric_type_to_str(dcol->type),
-					     ldms_metric_type_to_str(mcol->mtype));
+				ldmsd_lerror("strgp '%s': the metric type (%s) of "
+					     "row %d:col %d is different from the type (%s) of "
+					     "metric '%s'.\n", strgp->obj.name,
```
narategithub, line 958: @nichamon instead of "metric", could you please change it to "LDMS metric"? I think it is clearer.

narategithub (changes requested):

PR #990, opened 2022-07-26
Fix ldmsd_stream_publish's bug caused by #979

nichamon:
The pull request #979 prevents users from using ldmsd_stream_publish
with the -f option. To be precise, it causes ldmsd_stream_publish to
always return an error if one of the -n, -N, and -l options is not
given. The patch fixes the problem.

### DIFF
--- a/ldms/src/ldmsd/test/ldmsd_stream_publish.c
+++ b/ldms/src/ldmsd/test/ldmsd_stream_publish.c
@@ -193,15 +193,16 @@ int main(int argc, char **argv)
 	if (stream_new || line_mode) {
 		/* Create a transport endpoint */
 		ldms = ldms_xprt_new_with_auth(xprt, NULL, auth, NULL);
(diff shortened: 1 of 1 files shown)


tom95858:
Sorry @nichamon, I assumed this patch was tested.

nichamon:
> Sorry @nichamon, I assumed this patch was tested.

No problem! @narategithub caught it. He is setting up the routine regression test on Cygnus. Good thing he caught it early.

baallan:
@nichamon @narategithub thanks for the catch.

In tracking down the null pointer case, it looks like the first time the ldms xprt pointer is checked when publish is used is at the assert in ldms_xprt_connected. Given that it's a public api prone to incomplete/incorrect use, it would be helpful if the xprt pointer were checked and diagnosed at a scope much nearer the functions in ldmsd_stream.h.

PR #992, opened 2022-07-26
Correct the set's heap size in the set's meta

nichamon:
The heap size value in the set's meta was not rounded up, so the value may
be less than the actual heap size. This causes LDMSD's clients receive
the wrong heap size and then create a set with a too small heap. Thus,
the clients could access the memory beyond the heap region.


nichamon:
@tom95858 The patch fixes the problem I mentioned in the meeting this morning. The heap size in the set's meta wasn't rounded up, so it may be smaller than the actual heap size. `ldms_ls` creates the heap according to the value in the set's meta, so its set's heap could be smaller than what it should be. Thus, it may access memory beyond the heap. The problem affects sets created by calling `ldms_set_new_with_heap` or `ldms_set_create`.

PR #993, opened 2022-07-26
Install ${PREFIX}/etc but not /etc

narategithub:
PR #985 relocated ${PREFIX}/etc to a much deeper location
${PREFIX}/share/doc/ovis-ldms/sample_init_scripts/opt/etc. This may be
fine for example startup scripts (for init.d and systemd), but is
extremely inconvenient for ENV files like
${PREFIX}/etc/profile.d/set-ovis-variables.sh and
${PREFIX}/etc/ld.so.conf.d/ovis-ld-so.conf.  The relocation also broke
tests that use these two files to setup OVIS environment.

This patch partially reverts #985 by installing
${PREFIX}/etc/profile.d/set-ovis-variables.sh and
${PREFIX}/etc/ld.so.conf.d/ovis-ld-so.conf. To prevent installing these
files into "/etc", the installation is guarded by the sysconfdir
checking condition, i.e. if "$(sysconfdir)" is "/etc", they won't be
installed.

### DIFF
--- a/.github/workflows/4.3.3-compat-test.yaml
+++ b/.github/workflows/4.3.3-compat-test.yaml
@@ -45,7 +45,6 @@ jobs:
         ../configure --prefix=/opt/ovis-4 --enable-etc --enable-munge
         make
         make install
-        ln -s /opt/ovis-4/share/doc/ovis-ldms/sample_init_scripts/opt/etc /opt/ovis-4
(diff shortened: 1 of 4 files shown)


baallan:
@tom95858 looks like a reasonable solution. If this breaks anything in my existing rpm scripts, i'll update.

PR #996, opened 2022-08-02
Add a new Slurm sampler that supports arbitrary number of jobs

nichamon:


nichamon:
Here is an example of ldms_ls' s output with the new Slurm set. The set contains two lists: job_list and task_list. 

image

tom95858:
@nichamon, is this still a draft or is it ready to merge?

nichamon:
> @nichamon, is this still a draft or is it ready to merge?

It was tested. It is ready to be merged. The test script accompanying the patch is here. https://github.com/ovis-hpc/ldms-test/pull/25

I made it a draft because I wanted to discuss the set schema and how we should store the metrics using the decomposition and how we should correspond a metric set to the running job(s). We can merge it now, and I will make a patch if we want to change the set schema.

PR #1000, opened 2022-08-12
Variorum LDMS Plugin for vendor-neutral power management

tpatki:
This pull request introduces a vendor-neutral power monitoring plugin that reports node, CPU (socket), memory and GPU power using LLNL's Variorum library (https://variorum.readthedocs.io/).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -351,6 +351,26 @@ dnl options for contributed samplers
 OPTION_DEFAULT_DISABLE([ipmireader], [ENABLE_IPMIREADER])
 OPTION_DEFAULT_DISABLE([tutorial-sampler],[ENABLE_TUTORIAL_SAMPLER])
 
+dnl Variorum dependency support
+AC_ARG_ENABLE([variorum],
+    [AS_HELP_STRING([--enable-variorum], [require components that depend upon libvariorum (and libjansson) @<:@default=check@:>@])],
+    [],
+    [enable_variorum="check"])
+AS_IF([test "x$enable_variorum" != xno],[
+    AC_LIB_HAVE_LINKFLAGS([variorum], [], [#include <variorum.h>
+                           #include <variorum_topology.h>])
+    AS_IF([test "x$enable_variorum" != xcheck],[
+        AS_IF([test "x$HAVE_LIBVARIORUM" = xno],
+            [AC_MSG_ERROR([libvariorum or headers not found])])
+    ])
+    AC_LIB_HAVE_LINKFLAGS([jansson], [], [#include <jansson.h>])
+    AS_IF([test "x$enable_variorum" != xcheck],[
+        AS_IF([test "x$HAVE_LIBJANSSON" = xno],
+            [AC_MSG_ERROR([libjansson or headers not found])])
+    ])
+])
+AM_CONDITIONAL([HAVE_LIBVARIORUM], [test "x$HAVE_LIBVARIORUM" = xyes])
+AM_CONDITIONAL([HAVE_LIBJANSSON], [test "x$HAVE_LIBJANSSON" = xyes])
 
 dnl check for libcurl if influx is configured
 OPTION_DEFAULT_DISABLE([influx], [ENABLE_INFLUX])
@@ -987,6 +1007,7 @@ ldms/src/contrib/sampler/daos/Makefile
 ldms/src/contrib/sampler/daos/test/Makefile
 ldms/src/contrib/sampler/ipmireader/Makefile
 ldms/src/contrib/sampler/tutorial/Makefile
+ldms/src/contrib/sampler/variorum_sampler/Makefile
 ldms/src/contrib/sampler/geopm_sampler/Makefile
 ldms/src/third-plugins/Makefile
 ldms/src/ldmsd/Makefile
--- a/ldms/src/contrib/sampler/variorum_sampler/Makefile.am
+++ b/ldms/src/contrib/sampler/variorum_sampler/Makefile.am
@@ -1,0 +1,16 @@
+pkglib_LTLIBRARIES =
+lib_LTLIBRARIES =
+dist_man7_MANS =
+dist_man1_MANS =
+EXTRA_DIST =
+
+AM_CPPFLAGS = @OVIS_INCLUDE_ABS@
+AM_LDFLAGS = @OVIS_LIB_ABS@
+COMMON_LIBADD = -lsampler_base -lldms -lovis_util -lcoll \
+		@LDFLAGS_GETTIME@
+
+libvariorum_sampler_la_SOURCES = variorum_sampler.c
+libvariorum_sampler_la_LIBADD = $(COMMON_LIBADD) $(LTLIBVARIORUM) $(LTLIBJANSSON)
+pkglib_LTLIBRARIES += libvariorum_sampler.la
+dist_man7_MANS += Plugin_variorum_sampler.man
+EXTRA_DIST += README.md
--- a/ldms/src/contrib/sampler/variorum_sampler/Plugin_variorum_sampler.man
+++ b/ldms/src/contrib/sampler/variorum_sampler/Plugin_variorum_sampler.man
@@ -1,0 +1,66 @@
+.\" Manpage for Plugin_variorum_sampler
+.\" Contact j_hannebert@coloradocollege.edu, patki1@llnl.gov, shoga1@llnl.gov, brink2@llnl.gov or rountree4@llnl.gov to correct errors or typos.
+.TH man 7 "27 Jun 2022" "v4" "LDMS Plugin variorum_sampler man page"
+
+.SH NAME
+Plugin_variorum_sampler - man page for the LDMS Variorum plugin
+
+.SH SYNOPSIS
+Within ldmsd_controller or a configuration file:
+.br
+config name=variorum_sampler [common attributes]
+
+.SH DESCRIPTION
+With LDMS (Lightweight Distributed Metric Service), plugins for the ldmsd (ldms daemon) are configured via ldmsd_controller
+or a configuration file. The variorum_sampler plugin provides power data using the JSON API in Variorum, a vendor-neutral library that provides access to low-level hardware knobs.
+The sampler, when configured, automatically detects the number of sockets on the host machine and then provides, for each socket, an LDMS record containing power data.
+For each socket, the values provided are: node power consumption in Watts (identical across sockets); socket ID number; CPU power consumption in Watts;
+GPU power consumption in Watts (aggregated across all GPUs on the socket, and
+reported as -1 on unsupported platforms); and memory power consumption in Watts.
+
+.PP
+The variorum sampler depends on Variorum 0.6.0 or higher and Jansson. The sampler cannot be built without these libraries. If either library is installed in a non-standard location, paths to the respective install directories should be provided to Autoconf using
+the --with-libjansson-prefix and/or --with-libvariorum-prefix flag.
+
+.SH CONFIGURATION ATTRIBUTE SYNTAX
+The variorum sampler plugin uses the sampler_base base class. This man page covers only the configuration attributes, or those with default values, specific to the this plugin; see ldms_sampler_base.man for the attributes of the base class.
+
+.TP
+.BR config
+name=<plugin_name> exclude_ports=<devs>
+.br
+configuration line
+.RS
+.TP
+name=<plugin_name>
+.br
+This MUST be variorum_sampler.
+.TP
+schema=<schema>
+.br
+Optional schema name. It is intended that the same sampler on different nodes with different metrics have a
+different schema. If not specified, will default to `variorum_sampler`.
+.RE
+
+.SH BUGS
+No known bugs; however, if Variorum cannot access the hardware knobs, the sampler will be unable to access any data. This will result in an error being printed to the log file: "variorum_sampler: unable to obtain JSON object data". This error can be resolved by ensuring that
+hardware knob access is enabled using the requirements here: https://variorum.readthedocs.io/en/latest/HWArchitectures.html
+
+.SH EXAMPLES
+.PP
+Within ldmsd_controller or a configuration file:
+.nf
+load name=variorum_sampler
+config name=variorum_sampler producer=vm1_1 instance=vm1_1/variorum_sampler
+start name=variorum_sampler interval=1000000
+.fi
+
+.SH AUTHORS
+Jessica Hannebert <j_hannebert@coloradocollege.edu> (Colorado College, internship at Lawrence Livermore National Laboratory).
+Tapasya Patki <patki1@llnl.gov> (Lawrence Livermore National Laboratory).
+Kathleen Shoga <shoga1@llnl.gov> (Lawrence Livermore National Laboratory).
+Stephanie Brink <brink2@llnl.gov> (Lawrence Livermore National Laboratory).
+Barry Rountree <rountree4@llnl.gov> (Lawrence Livermore National Laboratory).
+
+.SH SEE ALSO
+ldmsd(8), ldms_quickstart(7), ldmsd_controller(8), ldms_sampler_base(7)
--- a/ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
+++ b/ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
@@ -1,0 +1,249 @@
+/* -*- c-basic-offset: 8 -*- */
+/* Copyright 2022 Lawrence Livermore National Security, LLC
+ * See the top-level COPYING file for details.
+ *
+ * SPDX-License-Identifier: (GPL-2.0 OR BSD-3-Clause)
+ */
+
+#define _GNU_SOURCE
+#include <inttypes.h>
+#include <unistd.h>
+#include <sys/errno.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "ldms.h"
+#include "ldmsd.h"
+#include "sampler_base.h"
+#include "variorum.h"
+#include "jansson.h"
+#include "variorum_topology.h"
+
+#define SAMP "variorum_sampler"
+
+static ldms_set_t set = NULL;
+static ldmsd_msg_log_f msglog;
+static base_data_t base;
+static int nsockets;
+static const char *SOCKET_METRICS[] = {"power_cpu_watts_socket_", "power_gpu_watts_socket_", "power_mem_watts_socket_"};
+static char** metric_names = NULL;
+static int i_node;
+static int i_sock;
+static int i_cpu;
+static int i_gpu;
+static int i_mem;
+static int lh_idx;
+static ldms_mval_t* rec_idxs;
+static char* result_string;
+
+static int create_metric_set(base_data_t base)
+{
+        int rc, i, metric, socket;
+        char metric_name[28];
+        char socket_num[11];
+        ldms_schema_t schema;
+        ldms_mval_t rec_inst;
+
+        // allocate space for metric names
+        if (!metric_names) {
+                metric_names = malloc(3 * nsockets * sizeof(char*));
+        }
+        for (metric = 0; metric < (3 * nsockets); metric++) {
+                metric_names[metric] = malloc(39);
+        }
+
+        // allocate space for record pointers
+        if (!rec_idxs) {
+                rec_idxs = malloc(nsockets * sizeof(ldms_mval_t));
+        }
+
+        schema = base_schema_new(base);
+        if (!schema) {
+                msglog(LDMSD_LERROR,
+                "%s: The schema '%s' could not be created, errno=%d.\n",
+                __FILE__, base->schema_name, errno);
+                rc = errno;
+                goto err;
+        }
+
+        // create record
+        ldms_record_t rec_def = ldms_record_create("variorum_socket_record");
+        i_node = ldms_record_metric_add(rec_def, "node_watts", "watts", LDMS_V_D64, 0);
+        i_sock = ldms_record_metric_add(rec_def, "socket_ID", "number", LDMS_V_U64, 0);
+        i_cpu = ldms_record_metric_add(rec_def, "cpu_watts", "watts", LDMS_V_D64, 0);
+        i_gpu = ldms_record_metric_add(rec_def, "gpu_watts", "watts", LDMS_V_D64, 0);
+        i_mem = ldms_record_metric_add(rec_def, "mem_watts", "watts", LDMS_V_D64, 0);
+
+        // calculate required heap size to support 1 record per socket
+        size_t heap_sz = nsockets * ldms_record_heap_size_get(rec_def);
+
+        // add record definition to the schema
+        int rec_def_idx = ldms_schema_record_add(schema, rec_def);
+
+        // add the metric list to the schema
+        int lh_idx = ldms_schema_metric_list_add(schema, "power", NULL, heap_sz);
+
+        set = base_set_new(base);
+        if (!set) {
+                rc = errno;
+                goto err;
+        }
+
+        ldms_mval_t lh = ldms_metric_get(set, lh_idx);
+
+        for(socket = 0; socket < nsockets; socket++) {
+                // create a new record
+                rec_inst = ldms_record_alloc(set, rec_def_idx);
+                rec_idxs[socket] = rec_inst;
+                // set the socket number
+                ldms_record_set_u64(rec_inst, i_sock, socket);
+                // put the record into the list
+                ldms_list_append_record(set, lh, rec_inst);
+                // create metric name list (for querying json object later on)
+                for(metric = 0; metric < 3; metric++) {
+                        strcpy(metric_name,SOCKET_METRICS[metric]);
+                        sprintf(socket_num,"%d",socket);
+                        strcat(metric_name,socket_num);
+                        strcpy(metric_names[(metric*nsockets)+socket], metric_name);
+                }
+        }
+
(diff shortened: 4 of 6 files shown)


ldms/src/contrib/sampler/variorum_ldms/Makefile.am
```diff
@@ -0,0 +1,16 @@
...
+dist_man7_MANS =
+dist_man1_MANS =
+
+AM_CPPFLAGS = @OVIS_INCLUDE_ABS@
+AM_LDFLAGS = @OVIS_LIB_ABS@
+COMMON_LIBADD = -lsampler_base -lldms -lovis_util -lcoll \
+		@LDFLAGS_GETTIME@
+
+if ENABLE_VARIORUM_LDMS
+libvariorum_ldms_la_SOURCES = variorum_ldms.c
```
morrone, line 12: Please change the name from "libvariorum_ldms" to "libvariorum_sampler". The directory name (and potentially some files) should be changed to match.

ldms/src/contrib/sampler/variorum_ldms/Makefile.am
```diff
@@ -0,0 +1,16 @@
...
+lib_LTLIBRARIES =
+dist_man7_MANS =
+dist_man1_MANS =
+
+AM_CPPFLAGS = @OVIS_INCLUDE_ABS@
+AM_LDFLAGS = @OVIS_LIB_ABS@
+COMMON_LIBADD = -lsampler_base -lldms -lovis_util -lcoll \
+		@LDFLAGS_GETTIME@
+
+if ENABLE_VARIORUM_LDMS
```
morrone, line 11: Please remove this "if ENABLE_VARIORUM_LDMS" conditional. This check is already being made in the Makefile.am one directory above this. This subdirectory will not be entered if ENABLE_VARIORUM_LDMS is not set.

ldms/src/contrib/sampler/variorum_ldms/Makefile.am
```diff
@@ -0,0 +1,16 @@
...
+libvariorum_ldms_la_SOURCES = variorum_ldms.c
+libvariorum_ldms_la_LIBADD = $(COMMON_LIBADD) -lvariorum -ljansson
+pkglib_LTLIBRARIES += libvariorum_ldms.la
+dist_man7_MANS += Plugin_variorum_ldms.man
+endif
```
morrone, line 16: I don't see the README.md mentioned here. At the very lease is should be in EXTRA_DIST.

configure.ac
```diff
@@ -350,7 +350,31 @@ AM_CONDITIONAL([ENABLE_GRPTEST_LDMS_TEST], [test "x$ENABLE_LDMS_TEST_FALSE" = "x
 dnl options for contributed samplers
 OPTION_DEFAULT_DISABLE([ipmireader], [ENABLE_IPMIREADER])
 OPTION_DEFAULT_DISABLE([tutorial-sampler],[ENABLE_TUTORIAL_SAMPLER])
-
+OPTION_DEFAULT_ENABLE([variorum_ldms],[ENABLE_VARIORUM_LDMS])                   
+
+dnl check for Variorum dependencies                                             
+if test "${VARIORUM_DIR+set}" == set; then
```
morrone, line 356: This is going to need to be overhauled. We should not pass paths to library dependencies in environment variables that autoconf doesn't know about. Instead, the autoconf way to handle this is to create a AC_ARG_WITH blocks. People would do "configure --with-jansson=/some/path --with-variorum=/some/other/path".

So here's what I think we need:

First, drop the "OPTION_DEFAULT_ENABLE([variorum_ldms],[ENABLE_VARIORUM_LDMS])"
Next, add an AC_ARG_WITH block for jansson (this checks for the external jansson library)
Then add an AC_ARG_WITH block for variorum (this checks for the external variorum library)
Finally, if you think you need it, add an AC_ARG_ENABLE block for the "variorum_sampler". This would
  just check if the the previously mentioned AC_ARG_WITH checks passed.

Take a look at the AC_ARG_WITH geopm (later in this file) as a basic example of what I am suggesting.

Also note that we shouldn't be setting these paths in the global CFLAGS/LDFLAGS/CPPFLAGS. Instead there should be individual versions of each of those that are used where needed. You can see hos the geopm example saves and restores those global values around the tests for the libraries and headers.

morrone (changes requested):
We'll need to change the approach to autoconf used in this PR. See the code other inline comments for details.

FYI have primarily reviewed the build systems parts of this PR. I only skimmed the C code of the new sampler.

tpatki:
Hi @morrone Thank you so much for getting the process started!

Jessica and I will start working on the comments for the name, Makefile and autoconf soon (early next week), and update this PR accordingly. 

We can also demo the plugin to you internally if you are interested in looking at the functionality across different architectures.

baallan:
Please mark this PR draft while major revisions are on-going.

It would be helpful if the interaction/conflict between the library underlying this samplers (indirect) usage of MSR and the syspapi sampler use of MSRs  were documented.  In particular since in many cases MSRs are scarce and the kernel may attempt to time-slice them if they are oversubscribed, inaccurate data for both samplers might result.

It appears jansson is not directly supported in RHEL7. It would be fantastic if variorum had a build switch to select between jansson and ovis_json (which has all the needed json functionality afaict).

morrone:
I'm not sure that it makes sense to have a build switch to select between jansson or ovis_json. If ovis_json meets their needs, then they should probably just switch to using that. Keeping the extra build complexity around for jansson when it is no longer needed doesn't seem desirable to me.

tpatki:
@baallan The library uses more than just MSRs (MSRs are only available on Intel systems and AMD systems, for IBM we use OPAL firmware, and for Nvidia and AMD GPUs we use NVML and Rocm). This is well documented here: https://variorum.readthedocs.io/en/latest/HWArchitectures.html as what Variorum uses internally will be architecture-specific.

What other documentation would you like to see here? 

Regarding Jansson vs OVIS-JSON -- we don't support that in Variorum yet, and we can in the future in a few months based on our prioritization, but I think that'll be beyond the scope of this PR. I am hoping that can be an enhancement we can work on later. Let me know your thoughts on that.

morrone:
Oh, if that is what baallan was requesting, then I agree. The external variorum package should not use ovis-json at all. There are plenty of json libraries out there in the world. We don't need add ovis's to the pile. Moreover, doing so in variorum would introduce a circular dependency at build time, and we can't have that.

I had assumed though that baallan meant that the variorum _sampler_ in this PR should use ovis-json. That at least would be reasonable.

tpatki:
Sure, updating the sampler to use your OVIS-JSON should be possible. Can you point me to the documentation for that and equivalent of the following or an example that does something similar in your codebase?

`json_loads(string ...)` 

`json_real_value(json_object_get(<obj-name>, "key")) `

baallan:
I would like to eliminate (at least from the ldms config/build) direct dependence on libjansson. If linking libvariorum.so automatically pulls in libjansson that is in theory ok.   If using ovis_json to parse data will still leave us with the explicit link dependency on jansson, then there's no point in using  the ovis_json.h code.

I actually *was* suggesting the radical course of making variorum json handling pluggable; on previous projects with plugins 'json version hell' has ensued when everyone likes their own version of their own third-party favorite json lib.  But I see there's resistance... understandable.


> @morrone Sure, updating just the sampler to use your OVIS-JSON library should be possible easily. I probably misunderstood what @baallan was suggesting there. Can you point me to the documentation for that and equivalent of the following or an example that does something similar in your codebase?

https://github.com/ovis-hpc/ovis/blob/OVIS-4/lib/src/ovis_json/ovis_json.h
> 
> `json_loads(string ...)`

json_parse_buffer

> 
> `json_real_value(json_object_get(<obj-name>, "key")) `

This is some combination of json_value_float and json_value_find

tpatki:
I'm not entirely sure if updating just the LDMS sampler code to parse with OVIS-JSON removes the need for linking in libjansson for the LDMS part. You will still need libjansson for building variorum, of course. 

Pullling @slabasan into this discussion so she can comment here too.

Libjansson is a more common C library for doing JSON stuff, so I don't think we want to move to OVIS-JSON altogether  in Variorum. We export the JSON object externally as a string, so I believe users of Variorum (eg LDMS) can parse it the way they like. 

May be we can at some point support more JSON libraries, but that's not on the roadmap at least at the moment for a few months, and that's a lot of internal code changes across multiple architectures that we can't make time for until we meet other priorities for Variorum. This isn't because of resistance necessarily due to having a favorite JSON library, but because of limited staff (2 main developers) and several other important things to support instead. That being said, I do hope that we can revisit that bit of supporting other JSON libraries in Variorum in the future when we have more time to devote to adding more JSON-related support.

tpatki:
Is there a current norm for other LDMS plugins that use their own JSON libraries?

configure.ac
```diff
@@ -350,31 +350,81 @@ AM_CONDITIONAL([ENABLE_GRPTEST_LDMS_TEST], [test "x$ENABLE_LDMS_TEST_FALSE" = "x
...
+          AC_MSG_ERROR([`variorum_topology.h` not found (required by variorum_ldms). You may need to update variorum.]))
+        AC_CHECK_HEADER(jansson.h, [have_variorum=yes],[have_variorum=no]
+          AC_MSG_ERROR([`jansson.h` not found (required by variorum_ldms). You may need to set JANSSON_DIR if jansson is already installed.]))
+        AC_CHECK_LIB(jansson, json_real_value, [have_variorum=yes],[have_variorum=no]
+          AC_MSG_ERROR([jansson not found (required by variorum_ldms). You may need to set JANSSON_DIR if jansson is already installed.]))
+        AC_CHECK_LIB(variorum, variorum_get_node_power_json, [have_variorum=yes],[have_variorum=no]
+          AC_MSG_ERROR([variorum not found (required by variorum_ldms). You may need to set VARIORUM_DIR if variorum is already installed.]))
+        CPPFLAGS=$save_CPPFLAGS
+        CFLAGS=$save_CFLAGS
+        LDFLAGS=$save_LDFLAGS
```
morrone, line 397: The last thing we do here is have a set of conditionals like this (in pseudo-code, I'll leave it to you to do the correct AS_IF and everything):

if have_varioum is no {
     if with_variorum == "check"
               AC_MSG_WARN("libvariorum not found")
     else
               AC_MSG_ERROR("libvarirorum not found")
}

This is because in the case where no explicit "--with-variorum" or "--without-variorum" was provided on the configure command line, we are just including support if it happens to be found. If not, that is non-fatal and we will just leave out any components that require it.

configure.ac
```diff
@@ -350,31 +350,81 @@ AM_CONDITIONAL([ENABLE_GRPTEST_LDMS_TEST], [test "x$ENABLE_LDMS_TEST_FALSE" = "x
...
+])
+AM_CONDITIONAL([ENABLE_VARIORUM], [test x$have_variorum = xyes])
```
morrone, line 399: We only want to set the ENABLE_VARIORUM AM conditional is _both_ have_variorum is yes and have_jansson is yes. (You'll need to add the test for libjansson as I mentioned, and replicated most of the comments I have mad about the variorum section in the jansson section.)

configure.ac
```diff
@@ -350,31 +350,81 @@ AM_CONDITIONAL([ENABLE_GRPTEST_LDMS_TEST], [test "x$ENABLE_LDMS_TEST_FALSE" = "x
...
+AM_COND_IF([ENABLE_VARIORUM],[
```
morrone, line 400: I don't think we need the AC_DEFINE. We are using the AM conditional in the parent Makefile.am of the variorum sampler to control if that subdir is built. So we can remove this AM_CONF_IF and the AC_DEFINE of ENABLE_VARIORUM.

morrone (changes requested):
OK, thanks! That is a set in the right direction. More comments added inline.

ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
```diff
@@ -0,0 +1,234 @@
+#define _GNU_SOURCE
```
morrone, line 1: We need a copyright statement at the top of the every source code file. I recommend the following comments at the top of this file:
```
/* -*- c-basic-offset: 8 -*- */
/* Copyright 2022 Lawrence Livermore National Security, LLC
 * See the top-level COPYING file for details.
 *
 * SPDX-License-Identifier: (GPL-2.0 OR BSD-3-Clause)
 */
```

ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
```diff
@@ -0,0 +1,234 @@
...
+static int i_cpu;
+static int i_gpu;
+static int i_mem;
+static int lh_idx;
+static ldms_mval_t* rec_idxs;
+static char* result_string;
+
+static int create_metric_set(base_data_t base)
+{
+	int rc, i, metric, socket;
```
morrone, line 34: This file needs to be reformatted to match the rest of ovis-ldms. For instance:

Indent is 8 spaces, no tabs

ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
```diff
@@ -0,0 +1,234 @@
...
+    char metric_name[28];
+    char socket_num[11];
+    ldms_schema_t schema;
+    ldms_mval_t rec_inst;
+
+    //allocate space for metric names
```
morrone, line 40: I would like to see a space after all of //

ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
```diff
@@ -0,0 +1,234 @@
...
+    if (!metric_names) metric_names = malloc(3 * nsockets * sizeof(char*));
```
morrone, line 41: Always put the body of the conditional on the next line. I don't know if we enforce it elsewhere, but I would also like to see braces { } used even with one-line bodies. This tends to avoid a common C error where people decide to add a second line, and they do it at the same indent level, but forget to add the braces.

ldms/src/contrib/sampler/variorum_sampler/Plugin_variorum_sampler.man
```diff
@@ -0,0 +1,65 @@
...
+.SH SYNOPSIS
+Within ldmsd_controller or a configuration file:
+.br
+config name=variorum_sampler [common attributes]
+
+.SH DESCRIPTION
+With LDMS (Lightweight Distributed Metric Service), plugins for the ldmsd (ldms daemon) are configured via ldmsd_controller
+or a configuration file. The variorum_sampler plugin provides power data using the JSON API in Variorum, a vendor-neutral library that provides access to low-level hardware knobs.
+The sampler, when configured, automatically detects the number of sockets on the host machine and then provides, for each socket, an LDMS record containing power data.
+For each socket, the values provided are: node power consumption in Watts (identical across sockets); socket ID number; CPU power consumption in Watts;
```
morrone, line 17: I think we need more of an explanation about what is going on here, because at first read is sounds a little fishy.

What exactly is a "socket" in your nomenclature? My guess would be a CPU socket.

The first thing that I don't necessarily understand is why things like the entire node's power would be grouped and repeated for each socket. A node is just one node. So shouldn't there just be a single "node_power" field in the schema?

And there is absolutely no reason to assume that the number of cpu sockets matches the number of gpus. So why would we list "gpu_power" per socket?

Is gpu_power the aggregate power used by _all_ gpus or just one gpu's power?

At a very least, I would think we need one list for cpus, and antoher list for gpus.
We might need a list for memory too if there isn't a single memory power number (maybe we have per-dimm or per bank memory power numbers?)

morrone (changes requested):
I took a look at the code this time, and besides the minor cosmetic issues, I have a concern about the general approach to the schema. Maybe I'm missing something, but we may need to rethink that.

Jessicat-H:
@morrone Thank you so much for all your suggestions! I apologize for being inattentive; I did not realize my pushes would go straight through to this pull request. I will work my way through your comments, but I know very little about autoconf, so it might take me a bit. Thank you for your patience.

configure.ac
```diff
@@ -359,7 +359,7 @@ AC_ARG_WITH([variorum],
                 [build Variorum sampler @<:@default=check@:>@])],
                 [AS_IF([test "x$withval" != xyes -a "x$withval" != xno],
                         [VARIORUM_CFLAGS="-I$withval/include"
-                         VARIORUM_LDFLAGS="-L$withval/lib"]
+                         VARIORUM_LDFLAGS="-Wl,-rpath -Wl,$withval/lib -L$withval/lib"]
```
morrone, line 361: We really shouldn't have the -W options here. automake/libtool should be using -rpath of its own under the covers when necessary.

Jessicat-H: I thought so too, but this is the only thing I've found so far that gets the library to link properly. Otherwise "ldd libvariorum_sampler.so" says that libvariorum was not found. Perhaps I am doing something wrong when building, but I believe I am following the steps from the wiki. Do you have any idea what might be going wrong?

Jessicat-H:
Pushed some more changes for the weekend; I'll be back to work on this more next week. Thank you again for your patience.

configure.ac
morrone: Well, my first guess is that we should be using _LIBADD rather than LDFLAGS. I'll comment more in the Makefile.am.

ldms/src/contrib/sampler/variorum_sampler/Plugin_variorum_sampler.man
slabasan: Variorum's JSON API has been defined such that it includes all the possible power domains (e.g., CPU, GPU, node, memory) that we may encounter when running variorum on a given HW architecture. Depending on the platform, some of the fields that we have defined in the JSON object will not be populated if the architecture does not provide us with a sensor or register to read the value. As an example, the JSON object when running on lassen will report the node power, the CPU power, and then to answer your question about gpu_power, it is the aggregated power across both GPUs on the socket. 

As of now, Variorum does not yet support two architectures at the same time, so for the lassen case, we cannot report GPU measurements from NVIDIA as well as CPU measurements from IBM. Hope this provides some background on why we have defined the schema in this way, happy to hop on a call with all of us to discuss the details further.

configure.ac
morrone: OH! I am so sorry, it is all coming back to me now and I realize that I should have sent you down a different path. What we want to use (instead of our own manual AC_ARG_WITH) is the very helpful macro AC_LIB_HAVE_LINKFLAGS. There is some documentation on it here:

https://www.gnu.org/software/gnulib/manual/html_node/Searching-for-Libraries.html

But I'll readily admit it can be a little tough to grok that manual.

You'll see that we already use AC_LIB_HAVE_LINKFLAGS in a couple of places. I think that the AC_ARG_ENABLE([papi]) would actually be a perfect example for you. Follow that through to ldms/src/samplers/Makefile.am to see how it uses the results of two AM_CONDITIONALS to decide whether to build a subdirectory. Basically you will wind up doing:

if HAVE_LIBJASSON
if HAVE_LIBVARIORUM
SUBDIR += libvariorum_sampler
endif
endif

And look at ldms/src/samplers/syspapi/Makefile.am. Notice in particular that on the libsyspai_sampler_la_LIBADD line there are "$(LTLIBPAPI) $(LTLIBPFM)".

When you use the AC_LIB_HAVE_LINKFLAGS macro it will automatically create for you the LTLIBVARIORUM and LTLIBJANSSON variables that you can use on the _LIBADD line.

AC_LIB_HAVE_LINKFLAGS([variorum] also auto-magically adds a configure option of the name "--with-libvariorum". It sets a number of other things that we need at well.

I think you would be fine just copying the the AC_ARG_ENABLE([papi] example in the configure.ac file. The only thing that would not do is break out libjansson as an independant test from variorum. But frankly I think it is fine to keep them together at this stage. The ldms team can work out how to split apart the libjansson test at some point in the future when something else wants to use libjansson.

Jessicat-H:
@morrone We tried to switch to using AC_LIB_HAVE_LINKFLAGS, but despite passing in "--with-libvariorum-prefix=<variorum/install/directory>", I am seeing this:
checking for libvariorum... no                                                                                                                                                                            
checking for libvariorum... (cached) no                                                                                                                                                                   
checking for libjansson... yes                                                                                                                                                                            
checking how to link with libjansson... -ljansson
Libjansson is in a standard location; libvariorum is not. Do you have any advice?

morrone:
> @morrone We tried to switch to using AC_LIB_HAVE_LINKFLAGS, but despite passing in "--with-libvariorum-prefix=<variorum/install/directory>"

I believe that the option is named _just_ "--with-libvariorum". You can run "./configure --help" to verify that. (I assume that we at not literally including the greater-than and less-than symbols.)

ldms/src/contrib/sampler/variorum_sampler/Makefile.am
```diff
@@ -4,8 +4,8 @@ dist_man7_MANS =
 dist_man1_MANS =
 EXTRA_DIST =
 
-AM_CPPFLAGS = @OVIS_INCLUDE_ABS@ @VARIORUM_CFLAGS@ @JANSSON_CFLAGS@
-AM_LDFLAGS = @OVIS_LIB_ABS@ @VARIORUM_LDFLAGS@ @JANSSON_LDFLAGS@
+AM_CPPFLAGS = @OVIS_INCLUDE_ABS@
+AM_LDFLAGS = @OVIS_LIB_ABS@
```
morrone, line 8: This looks good, but remember to on the LIBADD line to replace the explicit libraries with the LT* variables that AC_LIB_HAVE_LINKFLAGS creates.

configure.ac
```diff
@@ -351,70 +351,26 @@ dnl options for contributed samplers
...
-])
-AM_CONDITIONAL([ENABLE_VARIORUM_SAMPLER], [test x$have_variorum = xyes -a x$have_jansson = xyes])
-AM_COND_IF([ENABLE_VARIORUM_SAMPLER],[
-        AC_DEFINE([ENABLE_VARIORUM_SAMPLER], [], [Build Variorum sampler])
+AC_ARG_ENABLE([variorum],
+    [AS_HELP_STRING([--enable-variorum], [require components that depend upon libvariorum (and libjansson) @<:@default=check@:>@])],
+    [],
+    [enable_variorum="check"])
+AS_IF([test "x$enable_variorum" != xno],[
+    AC_LIB_HAVE_LINKFLAGS([variorum], [], [#include <variorum.h>])
```
morrone, line 360: There should only be one call to AC_LIB_HAVE_LINKFLAGS for each library. If you want to have multiple include lines in the test, I believe that you can make the [includes] options multiline and have multiple #include statements.

morrone (changes requested):

morrone:
Good news! I think the problem was in the AC_LIB_HAVE_LINKFLAGS macro itself. I updated the macros to the latest from gnulib, and I was able to configure and build ldms against libvariorum. PR #1008 has the updated m4 files.

ldms/src/contrib/sampler/variorum_sampler/Makefile.am
```diff
@@ -10,7 +10,7 @@ COMMON_LIBADD = -lsampler_base -lldms -lovis_util -lcoll \
 		@LDFLAGS_GETTIME@
 
 libvariorum_sampler_la_SOURCES = variorum_sampler.c
-libvariorum_sampler_la_LIBADD = $(COMMON_LIBADD) -lvariorum -ljansson
+libvariorum_sampler_la_LIBADD = $(COMMON_LIBADD) $(LTLIBVARIORUM) $(LTLIBJANSSON) -lvariorum -ljansson
```
morrone, line 13: The LT* variables already contain the "-lvariorum" and "-ljansson" options, so they can be removed from this line.

Jessicat-H:
I am not positive, but I believe these are the last of the changes that needed to be made.

tpatki:
@morrone We have addressed all your comments and tested this thoroughly, so I marked this PR ready for review.

morrone (approved):
Looks good to me! Thanks for working through all of those changes.

configure.ac
```diff
@@ -351,6 +351,26 @@ dnl options for contributed samplers
...
+    [enable_variorum="check"])
+AS_IF([test "x$enable_variorum" != xno],[
+    AC_LIB_HAVE_LINKFLAGS([variorum], [], [#include <variorum.h>
+                           #include <variorum_topology.h>])
+    AS_IF([test "x$enable_variorum" != xcheck],[
+        AS_IF([test "x$HAVE_LIBVARIORUM" = xno],
+            [AC_MSG_ERROR([libvariorum or headers not found])])
+    ])
+    AC_LIB_HAVE_LINKFLAGS([jansson], [], [#include <jansson.h>])
+    AS_IF([test "x$enable_jansson" != xcheck],[
```
morrone, line 367: Ah here "enable_jansson" needs to be changed to "enable_variorum".

morrone (changes requested):

tpatki:
Hi @morrone, I fixed the typo around `enable_variorum`. Please re-run tests when you get a chance, hopefully they will pass correctly this time :)

morrone:
@tom95858, @narategithub, could one of you please approve the workflows to run on this PR?

slabasan:
@tom95858, @narategithub: I've removed the trailing whitespace, which should fix the tests passing. When you get a chance, could you approve the workflows (hopefully last time)?

tpatki:
@narategithub @morrone @tom95858: I believe this PR ready to go now. Do let us know if there's anything else we need to do for this PR.

morrone (approved):

tom95858:
Hi everyone, is this ready to merge?

PR #1003, opened 2022-08-23
Containerized arm64 cython --gdb work around

narategithub:
cython with `--gdb` option resulted in "segmentation fault" when `make`
built ldms python module on arm64 (aarch64) containers. This issue does
not appear on x86_64 containers. This patch adds a work around to
only enable `cython --gdb` option for x86_64 architecture.

### DIFF
--- a/ldms/python/Makefile.am
+++ b/ldms/python/Makefile.am
@@ -22,9 +22,11 @@ ldms_la_CFLAGS = -g -O0  $(PY_INCLUDES) $(PYTHON_CPPFLAGS)
 ldms_la_LDFLAGS = $(PYTHON_LDFLAGS) -module -shared
 ldms_la_LIBADD = $(top_builddir)/ldms/src/core/libldms.la
 
+CYTHON_GDB = $(shell [ "$(uname -p)" = "x86_64" ] && echo --gdb || echo "" )
+
 $(LDMS_PYX_C): ldms.pyx ldms.pxd
 	echo PYTHON_LDFLAGS are "$(PYTHON_LDFLAGS)"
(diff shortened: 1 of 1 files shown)


tom95858:
@narate, can you please post this issue on the cython github page?

narategithub:
@tom95858 sure. Let me make a simple .pyx that reproduce the issue. I think ldms.pyx + ldms code is too complicated to submit to cython.

tom95858:
This was all found on Raspberry Pi, correct? I would like to ensure that this is an issue on Stria, etc... before accepting this.

narategithub:
@tom95858 The problem is found in a containerized Pi4. On containerized x86_64, we don't have this issue. On bare metal machines, both x86_64 and Pi4 do NOT have the issue.

baallan:
Is there a simple-enough reproducer that we could just turn the reproducer into a configure.ac feature check and assume that code for unhappy containers is also going to be built in the unhappy containers?

tom95858:
@narategithub I get it but this will disable --gdb for all platforms except x86_64. So I'd just like to understand what this does and why not just let it fail on those platforms that are broken. This is really a cython bug right?

narategithub:
@tom95858 I think I'll close this pull request. I think my Pi4 might be wonky. When using Pi4 (Ubuntu 22.04), the current https://github.com/ovis-hpc/ldms-containers cannot even successfully build `ovishpc/ldms-dev`. So, I switched to Jetson Nano (another arm64 small machine from Nvidia; the OS is based on Ubuntu 18.04), and everything works without this patch. I'll switch this PR to `draft` to keep the discussion open for now.

tom95858:
@narategithub can you please close this?

PR #1006, opened 2022-08-29
Fix munge auth check

narategithub:
Munge authentication checking must be performed after the context has
been set. Otherwise, we will always check against the default settings
and ignore the options specified by users.


tom95858:
The purpose of the test is to check whether or not the munge daemon is running and auth is even possible. What does the uid/gid of the 'test user' matter? Is it causing a problem?

narategithub:
@tom95858 The previous check ignored "-A socket=PATH_TO_MUNGE_SOCKET" authentication configuration, hence the check will be against an incorrect domain (caught by `ldms-test/ldmsd_auth_test`). The check in this patch uses the `PATH_TO_MUNCH_SOCKET` given by the configuration.

PR #1008, opened 2022-09-06
Update config.rpath and lib-*.m4 files with latest from gnulib.

morrone:
These are the latest files from gnulib. This seems to resolve an issue with the AC_LIB_HAVE_LINKFLAGS() macro that we are seeing in PR #1000.

### DIFF
--- a/config/config.rpath
+++ b/config/config.rpath
@@ -2,7 +2,7 @@
 # Output a system dependent set of variables, describing how to set the
 # run time search path of shared libraries in an executable.
 #
-#   Copyright 1996-2013 Free Software Foundation, Inc.
+#   Copyright 1996-2022 Free Software Foundation, Inc.
 #   Taken from GNU libtool, 2001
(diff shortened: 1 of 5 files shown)


tom95858:
@morrone, @narategithub, the Valiorum pull request fails all tests because a java package is not present in the github container running the pull request tests. Can you please check out the failures and add the necessary packages.

narategithub:
@tom95858 ??? I see "All checks have passed."

image

morrone:
> @morrone, @narategithub, the Valiorum pull request fails all tests because a java package is not present in the github container running the pull request tests. Can you please check out the failures and add the necessary packages.

Maybe instead of java you meant jansson? The libvariorum and libjansson dependencies are now only checks rather than hard requirements in the latest PR #1000. But in any event that should be be discussed in PR #1000 rather than here.

This PR really does address a deficiency in the AC_LIB_HAVE_LINKFLAGS (it was failing to find headers even though they were in the specified path). That was resolved upstream, so updating the m4 files fixed that problem for us.

PR #1010, opened 2022-09-08
Fix the set handle in procstat2 and procnetdev2

nichamon:


narategithub (approved):
👍

nichamon:
@tom95858 I found the bugs when I forced procstat2 to resize by configuring the intr_max to a small number (10-50). I didn't see the problem when I first tested because `base_set_new_heap()` returns the same address as the previous set. It returns a new address after the plugin resized the set multiple times. It's just a chance.

I included the changes in my release-4.3.9-test branch that Jim used to test on Voltrino.

PR #1011, opened 2022-09-08
add -x -p -h args so ldmsd_controller options are like others

baallan:
changes -h (or --help) to -? or --help to remedy collision with our -h host common syntax.

### DIFF
--- a/ldms/man/ldmsd_controller.man
+++ b/ldms/man/ldmsd_controller.man
@@ -33,21 +33,21 @@ PATH
 
 .SH LDMSD_CONTROLLER OPTIONS
 .TP
-.BI --host " HOST"
+.BI -h,--host " HOST"
 Hostname of \fBldmsd\fR to connect to
 .TP
-.BI --port " PORT"
+.BI -p,--port " PORT"
 The port of \fBldmsd\fR to connect to
 .TP
-.BI --xprt " XPRT"
+.BI -x,--xprt " XPRT"
 The transport type (\fBsock\fR, \fBrdma\fR, \fBugni\fR).
 .TP
-.BI -a " AUTH" " | --auth" " AUTH"
+.BI -a,--auth " AUTH"
 The LDMS authentication plugin. Please see
 .BR ldms_authentication (7)
 for more information.
 .TP
-.BI "-A " NAME = VALUE " | --auth-arg " NAME = VALUE
+.BI -A,--auth-arg " NAME=VALUE "
 Options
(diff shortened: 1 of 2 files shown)


ldms/python/ldmsd/ldmsd_controller
```diff
@@ -1712,13 +1712,16 @@ if __name__ == "__main__":
                                          "To connect to an ldmsd, either give " \
                                          "the socket path of the ldmsd or " \
                                          "give both hostname and inet control port. " \
-                                         "If all are given, the sockname takes the priority.")
-        parser.add_argument('--host',
+                                         "If all are given, the sockname takes the priority.",
+                                         add_help=False)
+        parser.add_argument('-H', '--help', action='help', default='==SUPPRESS==',
```
tom95858, line 1717: I think '?' might be better than -H. @nichamon, what do you think?

baallan: I'll update it to any concensus value. I think -? is somewhat less common than -H in the python world, but i can't point to a survey to prove it. Could probably add both, in fact, unless you have a collision with -H in mind for a feature down the road.

nichamon: > I think '?' might be better than -H. @nichamon, what do you think?

If we want to have a shorthand for --help, I vote for -?. I think we won't use -? for any other arguments. Many programs use -H for a particular argument. For example, docker's -H means a path to a file containing a host list. salloc's -H is equivalent to --hold.

tom95858: @baallan, Ok, I like this patch, but can we go with -??

baallan: Will update with the ? instead of H.

ldms/man/ldmsd_controller.man
```diff
@@ -64,7 +64,7 @@ Path to the config file
 .br
 Execute the script and send the output commands to the connected ldmsd
 .TP
-.BR -h
+.BR -H
```
tom95858, line 67: This needs to be -? right?

baallan: yes, fixed.

tom95858:
@baallan could you please squash these 4 commits into one and then force push to your branch. If I squash it here it creates a garbage commit message.

baallan:
@tom95858 squashed as requested.

PR #1019, opened 2022-09-16
Fix procstat.c that doubled frees g.schema

nichamon:
base_del() already frees `schema`. Calling ldms_schema_delete(g.schema) frees the already freed schema.

### DIFF
--- a/ldms/src/sampler/procstat/procstat.c
+++ b/ldms/src/sampler/procstat/procstat.c
@@ -747,20 +747,20 @@ static void term(struct ldmsd_plugin *self)
 		free(g.core_data);
 		g.core_data = NULL;
 	}
-	if (g.base)
+	if (g.base) {
 		base_del(g.base);
+		g.base = NULL;
+	}
 	if (g.set) {
 		ldms_set_delete(g.set);
 		g.set = NULL;
 	}
-	if (g.schema) {
-		ldms_schema_delete(g.schema);
-		g.schema = NULL;
-	}
 	free(g.line);
 	g.line = NULL;
-	if (g.mf)
+	if (g.mf) {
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 I found the bug when I looked at https://github.com/ovis-hpc/ovis/issues/948.

baallan:
actually the patch you need is 👍 ```
diff --git a/ldms/src/sampler/procstat/procstat.c b/ldms/src/sampler/procstat/procstat.c
index a3e5163..a669308 100644
--- a/ldms/src/sampler/procstat/procstat.c
+++ b/ldms/src/sampler/procstat/procstat.c
@@ -747,14 +747,10 @@ static void term(struct ldmsd_plugin *self)
                free(g.core_data);
                g.core_data = NULL;
        }
-       if (g.base)
+       if (g.base) {
                base_del(g.base);
-       if (g.set) {
-               ldms_set_delete(g.set);
+               g.base = NULL;
                g.set = NULL;
-       }
-       if (g.schema) {
-               ldms_schema_delete(g.schema);
                g.schema = NULL;
        }
        free(g.line);
```
Found the same bug tonight as well.

tom95858:
@nichamon it also looks like g.base is not set to NULL after it is freed. Is that also an issue?

nichamon:
> @nichamon it also looks like g.base is not set to NULL after it is freed. Is that also an issue?

Ah, it'll be a problem in the standalone image. Otherwise, it shouldn't be a problem as, after `term()` returns, the plugin library will be unloaded, and the plugin object will be freed. 

Anyway, I'll set `g.base` to NULL so that I can rebase the standalone branch on top of OVIS-4.

PR #1020, opened 2022-09-19
convert __dlog gdb-only controls to option -L control

baallan:
This enables configure replay (independent of source) and dynamic selection of debugging for configuration and failover that is previously  gdb-only accessible.
man page update and test included.
From the log message:
```
    Regardless of message origin (file, controller, maestro),
    this enables us to dynamically specify the recording of config requests,
    failover processing, status query requests in raw or replayable formats.
    It allows absolute and delta time stamps to be recorded.
    It distinguishes delayed start instructions from the rest in replay format.
    It allows the __dlog messages to go to the default log or to a specific file.
    It consolidates __dlog from request and failover into a single function.
    The -L option can be used in file or command line, though like other cmdline options if
    given in file, it is promoted to the front of the file.
```

### DIFF
--- a/ldms/man/ldmsd.man
+++ b/ldms/man/ldmsd.man
@@ -103,7 +103,7 @@ are executed sequentially, except for \fBprdcr_start\fR, \fBupdtr_start\fR,
 the deferred). Then, upon failover pairing success or failure, the other
 deferred configuration objects will be started. Please also note that while
 failover service is in use, prdcr, updtr, and strgp cannot be altered (start,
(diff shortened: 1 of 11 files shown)


baallan:
For a comprehensive open source solution to the log refactoring design (which would still leave us with adding most of the statements in this patch) see also (and improve just the configuration bits and dynamic configuration bits):
https://github.com/HardySimpson/zlog (http://hardysimpson.github.io/zlog/UsersGuide-EN.html

@nichamon @tom95858 @valleydlr

PR #1027, opened 2022-09-30
Fix zap_sock leaking endpoints if servers reject the connect requests

nichamon:
zap_sock moves the endpoint state to ERROR when it receives the rejected message. However, it does not put back the 'connect/accept' reference in sock_event() if the endpoint state is ZAP_EP_ERROR.

The endpoint state move to ERROR from either CONNECTING or ACCEPTING. zap_sock takes the 'connect/accept' reference when it moves the state to CONNECTING, but it does not take the reference when the state moves to ACCEPTING.

The patch makes sure that zap_sock takes the 'connect/accept' reference when the endpoint state moves to CONNECTING or ACCEPTING. Morevoer, it changes zap_sock to return back the reference in sock_event() if the endpoint state is ERROR.

### DIFF
--- a/lib/src/zap/sock/zap_sock.c
+++ b/lib/src/zap/sock/zap_sock.c
@@ -568,7 +568,6 @@ static void process_sep_msg_ack_accepted(struct z_sock_ep *sep)
 		.type = ZAP_EVENT_CONNECTED,
 		.status = ZAP_ERR_OK,
 	};
-	ref_get(&sep->ep.ref, "accept/connect"); /* Release when receive disconnect/error event. */
 	sep->ep.cb(&sep->ep, &ev);
 }
 
@@ -1579,6 +1578,7 @@ static void sock_event(struct epoll_event *ev)
 			zev.type = ZAP_EVENT_CONNECT_ERROR;
 			do_cb = 1;
 		}
+		drop_conn_ref = 1;
 		break;
 	case ZAP_EP_CONNECTING:
 		zev.type = ZAP_EVENT_CONNECT_ERROR;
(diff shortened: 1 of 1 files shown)


nichamon:
The patch is to fix the memory and file descriptor leak that @baallan found.

Here is the scenario.

- L0 listens on a port using auth_ovis.
- L1 sends a connect request without any authentication.
- L0 rejects the connection request due to the unexpected connection message. The auth isn't what L0 expects.
- On L1, zap_sock receives the REJECTED message, In process_sep_msg_rejected(), zap_sock
  - moves the endpoint state from ZAP_EP_CONNECTING to ZAP_EP_ERROR,
  - forwards the REJECTED event to ldms_xprt, and
  - calls shutdown(sep->sock)
- On L1, zap_sock.c:sock_event() gets an epoll event. It does not put back the connect reference because the endpoint state is ZAP_EP_ERROR.
 
I went and looked into zap_sock where it would change the endpoint state to ERROR to see why it doesn't put the 'connect/accept' reference back.

Fours functions in zap_sock move the endpoint state to ZAP_EP_ERROR.

- sock_event() flushes all outstanding I/O and moves the state to ERROR if the state is in CONNECTING, ACCEPTING, or LISTENING.
- process_sep_msg_rejected() moves CONNECTING to ERROR. [client side]
- z_sock_accept() moves ACCEPTING to ERROR if it fails to send the accepted message. [server side]
- z_sock_reject() moves ACCEPTING to ERROR if it fails to send the rejected message. [server side]

baallan:
@nichamon thanks for figuring this out. If you keep the auth mismatch  scenario the same but us the rdma transport instead, does a similar problem occur?

@tom95858 @valleydlr as this is a very significant bug fix; Assuming it passes review) can we do one of the following:
- make a 4.3.9.1 release from top of tree after merging 1027
- make a bug-fixes-only branch OVIS-4.3.9-updates rooted at 4.39 and then cherry pick #1027, #1026, #1020 into that branch?

PR #1030, opened 2022-10-05
linux_proc_sampler add environ, argv, and file access stream message

baallan:
This patch adds the ability to stream the values from environment, cmdline, and files used, since these data don't fit well into the metric fixed string size methods and only need to be reported once.  It also adds labeling for the syscall metric, since the numeric value is meaningless to data end users.

Key data to be obtained from the environment includes SLURM values such as job_id, so that individual processes may be associated with jobs without the need for those processes to be started by srun and notified to ldmsd by a spank plugin.

For the file case, both appearance of a file in use and disappearance of the file in use are logged. 
File logging is on a best-effort basis since it is based on scanning of /proc/pid/fd/* periodically and not any kernel event. This means files only briefly accessed may be missed entirely.
For file name and environment names, administrators may define regular expressions to exclude uninteresting or sensitive data from reporting.
Scanning of fds for newly opened files is performed when the process is first identified and at every Nth metric set sample for the same process, so the overhead of scanning fd/ can be reduced if desired.

### DIFF
--- a/ldms/src/sampler/app_sampler/linux_proc_sampler.c
+++ b/ldms/src/sampler/app_sampler/linux_proc_sampler.c
@@ -651,7 +651,7 @@ static int check_sep(linux_proc_sampler_inst_t inst, const char *sep)
 
 static int cmdline_handler(linux_proc_sampler_inst_t inst, pid_t pid, ldms_set_t set)
 {
-	/* populate `cmdline` and `cmdline_len` */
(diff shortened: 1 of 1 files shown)


baallan:
this is tested in rhel7/TOSS on cts1 hardware and virtual machines with ldms-static-test.sh linux_proc_sampler

baallan:
converting to draft as Schwaller has identified a problem with component_id capture on aarch64.

PR #1031, opened 2022-10-10
LDMSD sos init

tom95858:
This change removes a long deprecated SOS initialization command from 2015 and fixes a warning in the ldms python interface.


narategithub (approved):
It built + passed `agg_slurm_test`.

PR #1034, opened 2022-10-18
Fix the auth_flag if-statement when ldms_xprt receives a send_complete

nichamon:
Before the change, the if-statement uses the bitwise-AND to check whether the authentication has been approved. However, the flags are enumerated. The change makes ldms_xprt forwarding the send_complete events to applications only when the auth flag is equal to LDMS_XPRT_AUTH_APPROVED.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -2907,8 +2907,12 @@ static void ldms_zap_cb(zap_ep_t zep, zap_event_t ev)
 		ldms_xprt_put(x);
 		break;
 	case ZAP_EVENT_SEND_COMPLETE:
-		if (!(x->auth_flag & LDMS_XPRT_AUTH_APPROVED)) {
-			/* Ignore */
+		if (x->auth_flag != LDMS_XPRT_AUTH_APPROVED) {
+			/*
+			 * Do not forward the send_complete to applications
+			 * if the authentication is not approved.
+			 * Applications know only the connection is connecting.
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 Here is the bug I found in `ldms_xprt` that prevents `slurm_notifier` from sending stepd event data to its client. `ldms_xprt` could deliver a SEND_COMPLETE event before delivering the CONNECTED or CONN_ERROR event to clients. `slurm_notifier` interprets the SEND_COMPLETE event as a connection error because its connection state is in connecting. Then, it doesn't publish the step event data to clients. 

The fix fixes most of the missing stepd event data on `slurm_notifier`'s clients.

I tested submitting jobs using `srun` and submitting jobs inside an `salloc` session. `slurm_notifier` sent the stepd event data as expected, and the client received the data successfully. I'm having a problem when submitting a job using sbatch.

Submitting jobs using `sbatch` results in the step ID having a negative value. When this occurs, `slurm_notifier` drops the events. The affected events are `step_init`, `task_init_priv`, `task_exit`, and `step_exit`. The received negative value is 0xfffffffb, which is a special step_id value `SLURM_BATCH_SCRIPT` according to Slurm's source code here https://github.com/SchedMD/slurm/blob/master/slurm/slurm.h#L143. My job script has a single `srun` call, so I expect that the step_id of the task_init_priv should be 0, instead of the special value `SLURM_BATCH_SCRIPT`. I'm investigating this.

PR #1035, opened 2022-10-18
Handle Slurm unloading slurm_notifier before its routine is complete

nichamon:


nichamon:
@tom95858 The pull request is your patch plus some log message modification. I added the word 'slurm_notifier' and the stepd event name to the log messages. I tested the pull request with #1034 and `slurm_notifier` works as expected. Let's test this pull request and #1034 on Voltrino.

PR #1039, opened 2022-10-20
Fix the broken config line parsing logic in ldmsd_request_util.c

nichamon:
ldmsd_parse_config_str() does not add the terminating attribute to the message before converting the message from the host endian to the network endian. LDMSD may crash while processing a long config line.

The change also addresses the logic that chunks the messages into multiple records if the message is larger than the maximum of the transport message size. Before the change, the logic assumes that the message does not contain the terminating attribute.


nichamon:
I made this as a draft because I haven't run it against the test cases that test for long config lines. I manually tested this with the config line that produced the bug.

nichamon:
The patch passed the test in PR https://github.com/ovis-hpc/ldms-test/pull/32.

nichamon:
@tom95858 We need this to test NERSC's configurations.

PR #1040, opened 2022-10-21
Improve the test_sampler plugin

nichamon:
* Remove the assumptions that the compnent_id and job_id are uint64.
* Make the metric attribute delimiter configurable by giving the 'delim' attribute at the action=add_schema line.


nichamon:
@tom95858 We need to test NERSC's configurations.

PR #1041, opened 2022-10-24
Fix linux_proc_sampler build error on Ubuntu 22.04

narategithub:
This patch addressed the following build issues on Ubuntu 22.04
- `readdir_r()` is deprecated. Use `readdir()` is safe in this sampler because the `dir` handle is used only by one thread.
- String truncation warning
- `major()` and `minor()` implicit declaration warnings

### DIFF
--- a/ldms/src/sampler/app_sampler/linux_proc_sampler.c
+++ b/ldms/src/sampler/app_sampler/linux_proc_sampler.c
@@ -60,6 +60,7 @@
 #include <dirent.h>
 #include <ctype.h>
 #include <assert.h>
+#include <sys/sysmacros.h>
 
 #include <coll/rbt.h>
 
@@ -2795,11 +2796,10 @@ static int publish_fd_pid(linux_proc_sampler_inst_t inst, struct linux_proc_samp
 	}
 
 	char path[PROCPID_SZ];
-	char dname[PROCPID_SZ];
+	char dname[PATH_MAX];
 	int blen;
 	char buf[BUFMAX];
 	DIR *dir;
-	struct dirent enttmp;
 	struct dirent *dent;
 	snprintf(path, sizeof(path), "/proc/%" PRId64 "/fd",
 		app_set->key.os_pid);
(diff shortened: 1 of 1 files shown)


ldms/src/sampler/app_sampler/linux_proc_sampler.c
```diff
@@ -2810,14 +2810,12 @@ static int publish_fd_pid(linux_proc_sampler_inst_t inst, struct linux_proc_samp
 	}
 	errno = 0;
 	long n;
-	int rc;
 	char *endptr;
 	struct fentry *fe;
-	while ( (rc = readdir_r(dir, &enttmp, &dent)),
-			(rc == 0 && dent != NULL)) {
```
tom95858, line 2817: @narategithub, I don't think the issue is whether or not more than one thread in the this sampler calls readdir() but whether or not any other code paths calls readdir(). I count eight samplers that call readdir(). Are we saying that all of those samplers call readdir from the same thread? Even if the calls are all in the config path, the recv callback could be on different threads.

This sampler consists primarily of code copied from the app_sampler. It seems we would be better off getting rid of this thing and adding whatever functionality is missing to app_sampler.

All of that said, we need this patch to fix the top of tree. Are the changes to the github build check under version control?

tom95858:
@narategithub could you please remove the conditional config for this sampler. There's no reason that this sampler should be conditionally built.

ldms/src/sampler/app_sampler/linux_proc_sampler.c
narategithub: @tom95858 I understand your concern about `readdir()`. However, in the modern Linux environment, the `readdir_r()` is deprecated and `readdir()` is thread-safe if different threads access different directory stream. The following is a snippet from `readdir()` man page:
```
       However, in modern
       implementations (including the glibc implementation), concurrent
       calls to readdir() that specify different directory streams are
       thread-safe.  In cases where multiple threads must read from the
       same directory stream, using readdir() with external
       synchronization is still preferable to the use of the deprecated
       readdir_r(3) function.
```

In our case, it is safe to use `readdir()`.

narategithub:
@tom95858 `app_sampler` is now enabled by default.

ldms/src/sampler/app_sampler/linux_proc_sampler.c
tom95858: @narategithub OK, fantastic. Because like I said, we call it in many places. That said, can you remove the conditional config logic, I don't see why it is necessary.

baallan: @narategithub dstat and linux_proc_samplers may both be visiting the same proc dirs.  If the opendir returns unique objects scoped to the individual samplers, then it should be thread-safe with readdir as suggested.

baallan: @tom95858 this sampler has substantially more functionality that we are already deploying and fixes several bugs in app_sampler. If you want to remove a sampler, remove app_sampler.c.

narategithub:
@tom95858 it is now ready to merge.

PR #1042, opened 2022-10-24
Add ldms-test-build github workflow

narategithub:
Add ldms-test-build github workflow that mimic OVIS-4 build in ldms-test.


narategithub:
@tom95858 this is an effort to catch build error in the PR before it hits the daily regression ldms-test.

baallan:
@narategithub wonderful addition. Can you also include in it a test or enabled plugin using c++ that causes ldms.h to be ingested by code compiled with c++? We have these __cplusplus markups in our headers that don't seem to be tested.

PR #1043, opened 2022-10-26
Fix a bug introduced by PR 1039

nichamon:
The bug is that ldmsd_parse_config_str() appends the terminating attribute to the request only when there are other attributes.

The patch makes the functions to always append the terminating attribute.


nichamon:
@narategithub The PR is to fix the bug that fails https://github.com/ldms-test/weekly-report/blob/master/test-log/ldmsd_ctrl_test.log.

PR #1045, opened 2022-10-27
Add the metric type to the ldms_metric_template structure

nichamon:
Before the change, struct ldms_metric_template_s does not support META metrics, resulting in that ldms_schema_from_template() API can create a schema that contains only DATA metrics.

The patch adds flag to the ldms_metric_template structure so that applications can specify the metric flag to be META or DATA.


nichamon:
@narategithub Does it make sense that I added the metric flags to the `ldms_metric_template_s` structure? 

It doesn't make sense when the template is for creating a record, but without the flags field, it isn't straightforward to generate a schema containing a meta metric. Specifically, applications must not add any meta metrics in the templates and call `ldms_schema_meta_add()` to add any meta metrics to the schema besides calling `ldms_schema_from_template()`.

nichamon:
@narategithub @tom95858 The core logic change is in `ldms.c` and `ldms.h`. I added the changes in the sampler plugins; otherwise, the patch cannot be built and cannot pass Github's workflow checks.

narategithub (approved):
👍

nichamon:
@tom95858 The change was driven by the testing that uses the test_sampler plugin to simulate metric sets. The test_sampler plugin creates the component_id and job_id metrics in the data part; although, we specified them as meta metrics. After an investigation, I found that the limitation is in the `ldmsd_metric_template_s` structure. Thus, I made the change.

PR #1046, opened 2022-10-27
Fix the test_sampler plugin when initializing the component_id array

nichamon:


nichamon:
@tom95858  Jim found the bug when he was setting up the test on Voltrino.

ldms/src/sampler/examples/test_sampler/test_sampler.c
```diff
@@ -1020,26 +1020,40 @@ static int __init_set(struct test_sampler_set *ts_set)
...
+			v.v_u64 = ts_set->compid;
+			if (ldms_metric_is_array(ts_set->set, mid)) {
+				uint32_t count = ldms_metric_array_get_len(ts_set->set, mid);
+				for (j = 0; j < count; j++)
+					ldms_metric_array_set_u64(ts_set->set, mid, j, ts_set->compid);
+			} else {
+				ldms_metric_set(ts_set->set, mid, &v);
+			}
 		} else {
-			ldms_metric_set(ts_set->set, mid, &v);
```
tom95858, line 1030: @nichamon, what is v set to when the mid < 0? It seems un-initialized to me.

nichamon: If `mid < 0`, `v` isn't set and isn't used. If `mid < 0`, an error message is logged at line 1035.

tom95858: Should we set the mid to some value like -1? Or just leave it uninitialized. Also it looks like this patch needs to be rebased?

nichamon: I think git diff makes this confusing. I actually deleted lines 1023 - 130 in red and replaced them with lines 1023 - 1038 in green. `ldms_metric_set()`  is called only when `mid >= 0` and `v` is set to a value at line 1026. We don't have to assign values to `mid` and `v`. Did I miss anything?

I just rebased the patch.

PR #1047, opened 2022-10-27
Ibmad records sampler

morrone:
Implement the ibmad_records_sampler plugin. Versus the ibmad_sampler, the major change is that we use a single list of records per ib interface rather than a metric set per ib interfaces.

There are a number of other changes and cleanups that fell out of that change. For instance, we can fully use the sampler_base code now that we only have a single metric set.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -976,6 +976,7 @@ ldms/src/sampler/cray_system_sampler/Makefile
 ldms/src/sampler/aries_mmr/Makefile
 ldms/src/sampler/aries_mmr/aries_mmr_set_configs/Makefile
 ldms/src/sampler/ibmad_sampler/Makefile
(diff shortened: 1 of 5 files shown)


tom95858:
@morrone does this break the build if libibmad is not installed?

tom95858:
@morrone, I just didn't see where the checks were done in the new Makefile for whether or not the ibmad sampler was configured. But I may have missed something.

morrone:
The checks are in sampler/Makefile.am:

```
if HAVE_LIBIBMAD
if HAVE_LIBIBUMAD
if HAVE_IBA_IB_TYPES_H
SUBDIRS += ibmad_sampler
SUBDIRS += ibmad_records_sampler
endif
```

PR #1050, opened 2022-11-02
Slingshot samplers

morrone:
Introduce two sampler plugins for slingshot NICs: slibgshot_metrics and slingshot_info. The former is intended for high-frequency detailed counters, and the latter for lower-frequency general information.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1023,6 +1023,7 @@ ldms/src/sampler/ibm_occ/Makefile
 ldms/src/sampler/syspapi/Makefile
 ldms/src/sampler/app_sampler/Makefile
 ldms/src/sampler/slingshot_metrics/Makefile
+ldms/src/sampler/slingshot_info/Makefile
 ldms/src/contrib/sampler/Makefile
 ldms/src/contrib/sampler/daos/Makefile
 ldms/src/contrib/sampler/daos/test/Makefile
--- a/configure.ac
+++ b/configure.ac
@@ -849,6 +849,22 @@ AM_CONDITIONAL([HAVE_DCGM], [test x$have_dcgm = xtrue])
 
 AM_CONDITIONAL([SYSCONFDIR_NOT_ETC], [test "${sysconfdir}" != "/etc"])
 
+AC_LIB_HAVE_LINKFLAGS([cxi], [], [
+#include <stddef.h> /* libcxi.h fails to include this */
+#include <libcxi/libcxi.h>
+])
+AM_CONDITIONAL([HAVE_LIBCXI], [test "x$HAVE_LIBCXI" = xyes])
+
+AC_ARG_ENABLE([slingshot],
+    [AS_HELP_STRING([--enable-slingshot], [require the slinghost related plugins @<:@default=check@:>@])],
+    [],
+    [enable_slingshot="check"])
+AS_IF([test "x$enable_slingshot" = xyes],[
+    AS_IF([test "x$HAVE_LIBCXI" = xno],
+        [AC_MSG_ERROR([libcxi or its headers not found])])
+])
+AM_CONDITIONAL([ENABLE_SLINGSHOT], [test "xenable_slingshot" != xno])
+
 # define substitutions for configvars and other sed-generated files.
 # note carefully the escapes.
 OVIS_DO_SUBST([LDMS_SUBST_RULE], ["sed \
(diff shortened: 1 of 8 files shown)


baallan:
@morrone If there is a public document (other than the header/source du jour) of what these slingshot counters are and what their units (bytes/pages/etc) are, please reference it from the man page. I realize the odds of such documents actually existing may be slim.

Also, is there a test system at LLNL that we can get on and measure the performance (basically, typical sample duration time) of the sampler in various configurations?

morrone:
Other than the documentation in the header file that is already mentioned in the man page (which does give brief descriptions of the counters), I am not aware of additional documentation.

With one slingshot nic and using "group:all", which is probably 1426 counters, ldms_ls reports a duration of around 0.000033s:

```
Schema         Instance                 Flags  Msize  Dsize  Hsize  UID    GID    Perm       Update            Duration          Info
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
slingshot_metrics hetchy12/slingshot_metrics    CL  167872  12928  12800   8126   8126 -r--r----- 1667410750.338158          0.000033 "updt_hint_us"="5000000:0"
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
Total Sets: 1, Meta Data (kB): 167.87, Data (kB) 12.93, Memory (kB): 180.80
```

With four slinghost nics and using "group:all", the duration is 0.000079s:

```
Schema         Instance                 Flags  Msize  Dsize  Hsize  UID    GID    Perm       Update            Duration          Info    
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
slingshot_metrics hetchy28/slingshot_metrics    CL  167872  47232  47104   8126   8126 -r--r----- 1667411159.522041          0.000079 "updt_hint_us"="5000000:0" 
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
Total Sets: 1, Meta Data (kB): 167.87, Data (kB) 47.23, Memory (kB): 215.10
```

So I think we are doing fine for worst-case sample time.

With one nic and using the default set of 39 metrics (that I selected more-or-less at random, because they sounded good at the time), the typical duration is 0.000008s:

```
Schema         Instance                 Flags  Msize  Dsize  Hsize  UID    GID    Perm       Update            Duration          Info    
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
slingshot_metrics hetchy12/slingshot_metrics    CL    4968   1152   1024   8126   8126 -r--r----- 1667411534.165330          0.000008 "updt_hint_us"="5000000:" 
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
Total Sets: 1, Meta Data (kB): 4.97, Data (kB) 1.15, Memory (kB): 6.12
```

baallan:
@morrone Thanks; the metric timing looks great: what CPU type is this data from (if not nda)? 
The only other timing question (common to both samplers) is what kind of time does cxil_get_device_list take (I suspect it looks at local kernel info with no chance of hanging/timing out).

morrone:
cxil_get_device_list is a little longer. Around 0.0007s for 1 nic, and 0.002s for 4 nics.

baallan:
wow, that's quite a chunk of time for a high-ish frequency sampler, especially if its busy-waiting.
Any chance we can include a device discovery frequency (perhaps discover only on every nth sample event?) parameter on the sampler config?

morrone:
I don't see any evidence of busy waiting. But I suppose we can.

tom95858:
@morrone, could you please move the cxil_get_device list to after the call to base_begin_sample. That would include this overhead in the ldms_ls -v output and would allow us to realistically evaluate the overhead of this sampler.

morrone:
In the updated 5f39cd9:

cxill_get_device_list() is included inside base_begin_sample.
The device list is now cached and only updated on a selectable interval.
Except when the device list is being refreshed, the sample() times are as represented in the earlier timing results.
Other error handling improvements.

tom95858:
@morrone, this looks perfect. Does anyone see any issues with accepting this? It would be nice to have a slingshot sampler in 4.3.10 as a starting point.

baallan:
If @morrone has run the sampler with valgrind and not noticed leaks/other problems with "normal" usage, I'd rather have it than not and fix any minor bugs/missing features later.  
Minor issues easily fixable now:
- think about adding the refresh option to the info sampler as well; otherwise the maximum interval is only 35 minutes or so and running hourly or daily might be desirable.
- think about suppressing the set update (info sampler only) if in fact nothing has changed.
- run aspell or some such on the man pages for typos.

morrone:
@baallan Just FYI, I haven't done valgrind, but memory usage was holding steady in my testing.

Why is the maximum interval only 35 minutes? Is the interval stored as signed 32-bit somewhere? I guess I wasn't aware of that limit.

baallan:
@morrone sampler interval got upgraded to long (which with luck is 64 bits) in 2015, I see now; old data in cache, sorry. The updater interval still seems to be 32 bit.

PR #1053, opened 2022-11-03
Various samplers, cleanup and improved error handling

morrone:


tom95858:
@nichamon, @narategithub could you please take a look at this?

narategithub:
`sampler_baes` and `rec_def` are allocated in `config`, but `term()` releases them regardless. I think it is OK since both `base_del()` and `ldms_record_delete()` do NULL checking.

However, I think we may want to set `sampler_base = NULL` and `rec_def = NULL` in `term()` since `dlclose()` may or may not unload the library. We might have a problem in the following scenario if `dlclose()` did not unload the library:
```
load name=slingshot_info
config name=slingshot_info  # sampler_base and rec_ref are allocated
term name=slingshot_info # sampler_base and rec_def are freed

load name=slingshot_info # sampler_base and rec_def still points to the old memory
term name=slingshot_info # free the invalid sampler_base and rec_def
```

morrone:
That is true.  feae6be add the required "= NULL".

I suspect that lots of other samplers have that same issue.

ldms/src/sampler/slingshot_info/slingshot_info.c
```diff
@@ -386,6 +388,10 @@ static int sample(struct ldmsd_sampler *self)
 static void term(struct ldmsd_plugin *self)
 {
         log_fn(LDMSD_LDEBUG, SAMP" term() called\n");
+        base_del(sampler_base);
```
nichamon, line 391: `base_del()` frees `rec_def` when it deletes the schema by calling `ldms_schema_delete()`, so there is no need to call `ldms_record_delete(rec_def)`.`rec_def` is a part of the schema when `ldms_schema_record_add()` is called, so `ldms_schema_delete()` frees `rec_def` altogether. 
However, `base_del()` doesn't delete the set. You have to delete the set here.

The procedure here should be.
```
base_set_delete(sampler_base);
base_del(sampler_base);
sampler_base = NULL;
rec_def = NULL;
```

nichamon (changes requested):

ldms/src/sampler/slingshot_info/slingshot_info.c
morrone: OK, thanks.

morrone: Oh! In that case, there is no real reason for me to even have rec_def as global scope. That is even better.

morrone: Addressed in the latest push.

morrone:
Well, this cleanup patch set grew.

I added base_schema_delete(), which gives us some added symmetry and reduces laying violations.

Then I also did some similar error handling and cleanup in ibmad_records_sampler, procnetdev2, and procstat2.

nichamon (approved):

narategithub:
@morrone @nichamon @tom95858 This patch looks OK to me.

Issue #1055, opened 2022-11-03
uid reported without username in samplers is problematic 


baallan:
It has been brought to our attention by a user wanting to do analyses across multiple systems that UIDs may be different for the same username and that containerized use (with uids being mapped to in-container values that differ) may also be problematic.
They have requested that we include both username and uid, at least optionally.
This affects the /proc/$pid related samplers at minimum.

baallan:
As the fix is simple enough and lack of it constitutes a usability bug for our analysis systems, I will create a patch for it 11/3-11/4 in time for 4.3.10 inclusion.

tom95858:
Hi @baallan, I think this might be a bit of a partial fix. Could you please provide a more complete use-case. For example is this analysis being done from the store? How do we handle the converse where the uid is the same, but the username is different? For example, totucke, vs. tom for uid 1234. I certainly wouldn't want to rush a change like this into a release under test without full understanding what the goals/side-effects are.

baallan:
@tom95858 My proposed solution (add the username from getpwuid(uid) to the metric set) would be applied to a metric set that already contains the uid and the hostname; thus we have as complete context info as is available. One could argue that cluster name is also needed on some sites that recycle hostnames, but this is always available if needed. 

How one disambiguates things where there are genuine conflicts can only be solved at the analysis scope which includes multiple systems such that there is a conflict. 

Another use case is where the same user has different user names and different uids across different systems and wants to aggregate data across both. A less likely scenario in the national labs, but probably fairly common in a university setting. 

For all three cases of real person userA:
- userA>usernameA=uid1,uid2
- userA>usernameA=uid1,usernameB=uid1
- userA>username=uid1,usernameB=uid2

I claim that the most we can do (and the minimum we should do) at the sampler scope is to provide the locally correct tuple (username,uid,hostname[,optionally cluster name]).

Everything else, if aggregating across disparate identity managements, must be processed at the analysis stage-- there's nothing more possible at the sampler. If sites allow the same username to apply to two different real people, they get what they deserve and will have to define database indices/keys accordingly.
For my environment, users in many if not most cases have no idea what their uid is; they might have two usernames though this would be extremely rare and mostly violates current policies [elevated privilege accounts are not 'the same real user']. They could easily, depending on container uid mappings, have multiple uids to the same username.

One could argue on the same basis that I should also add the option or requirement that if GID is collected then the groupname collection should also be at least attempted.

[...]

PR #1056, opened 2022-11-05
add user and group names options for status_uid/gid

baallan:
This adds the ability to optionally collect names corresponding to uid and gid values, enabling analyses based on user or group name.
The library call(s) to lookup username/groupname are not repeated so long as the paired uid/gid value does not change.

### DIFF
--- a/ldms/man/Makefile.am
+++ b/ldms/man/Makefile.am
@@ -46,7 +46,6 @@ Plugin_slurm_sampler.man \
 Plugin_store_papi.man \
 Plugin_store_slurm.man \
 Plugin_syspapi_sampler.man \
-Plugin_app_sampler.man \
 Plugin_store_app.man \
 ldmsd_decomposition.man \
 Plugin_store_kafka.man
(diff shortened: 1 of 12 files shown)


tom95858:
@baallan, it seem to me that this change was made only to the linux_proc_sampler and not to the app_sampler from which it is almost entirely derived. I think we should combine these two samplers and have a config option to select either notifier or streams data.

baallan:
The combining of incompatible (but related) plugins is beyond the scope of this PR. 
We need this small revision to linux_proc_sampler to enable some of our production Q1 goals.

We can certainly have a separate discussion of the two samplers and a possible merged future. This sampler is certainly not *entirely* derived from the other. If merging/deprecating app_sampler is not in the future, I'll be happy to add the user/group name to it as well. Let's conduct our fine-grained comparison of the features and schemas of the two and then decide how to proceed.

baallan:
fixes #1055

tom95858:
Please update the appmon sampler with the equivalent functionality.

baallan:
@tom95858 user/group names added to app sampler.
Also relocated the man page to be adjacent to source.

baallan:
@tom95858 ready to merge

PR #1057, opened 2022-11-08
Introduce ldmsd_communicator module to ovis

nick-enoent:
Replaces ldmsd_request.py with ldmsd_communicator.py.
Implement API calls for all relevant ldmsd configuration requests.
Refactor ldmsd_controller to use ldmsd_communicator.
auth_add configuration requests will no longer be deffered


ldms/python/ldmsd/ldmsd_controller
```diff
@@ -61,9 +61,9 @@ import traceback
 import json
 import re
 
-from ldmsd import ldmsd_config, ldmsd_util, ldmsd_request
-from ldmsd.ldmsd_request import LDMSD_Request, LDMSD_Req_Attr
-from ldmsd.ldmsd_request import LDMSD_Req_Attr
+from ldmsd import ldmsd_config, ldmsd_util, ldmsd_communicator
```
nichamon, line 64: I think you don't have to import ldmsd_config because you use the connect facility from the ldmsd_communicator module.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -613,7 +613,6 @@ int is_req_id_priority(enum ldmsd_request req_id)
 	case LDMSD_ENV_REQ:
 	case LDMSD_VERBOSE_REQ:
 	case LDMSD_LISTEN_REQ:
-	case LDMSD_AUTH_ADD_REQ:
```
nichamon, line 616: Why did you need to remove auth_add from the priority command list? Did it give you problems with ldmsd_communicator?

nichamon (changes requested):

nichamon:
IMO, the next step is to remove ldmsd_config.py. I did 'grep -rn ldmsd_config ldms/python/*' and found that, with this patch, only the test scripts in ldms/python/ldmsd_test and ldms/python/test use the ldmsd_config module. Then, the effort to remove ldmsd_config is more toward porting the test coverages in the directories to the ldms-test repo.

ldms/src/ldmsd/ldmsd_request.c
nick-enoent: Thanks for looking at this @nichamon! I removed auth_add from the priority command list because it seemed to be having the opposite effect as intended. The event I was experiencing was the ldmsd would attempt to add a producer to an aggregator before adding the authentication domain

nichamon: What was the error message? Did you configure LDMSD using a configuration file when you saw the problem? The request priority list applies to the configuration files specified with the -c cmd-line option.

nick-enoent: I believe it was just that the aggregator was attempting to add a producer before adding the authentication domain (i.e. auth domain didn't exist). I'm going to rebase this on top of a connection timeout PR I'll be making to maestro this week, so I'll test it again with AUTH_ADD in the priorities list at that time and give you more detailed info.

tom95858:
Hi @nick-enoent, @nichamon what is the status of this pull request. It is still marked draft. In any event it needs to be rebased.

ldms/src/ldmsd/ldmsd_request.c
nichamon: @nick-enoent Any updates on the errors caused by the auth_add command in the priority list?

nick-enoent: @nichamon I'm no longer seeing an issue with keeping AUTH_ADD in the request priority list

nichamon: @nick-enoent Thanks for checking it! Could you please update the patch, so it doesn't remove the auth_add from the priority list?

nick-enoent: @nichamon I'm almost done with this patch. It has quite a few changes (along with keeping AUTH_ADD in the priority list. I'll make sure to ping you when I have pushed the new commit.

nichamon: @nick-enoent Thanks for the update.

PR #1059, opened 2022-11-09
Remove the third-plugins directory

nichamon:
The third-plugins directory is redundant with the contrib directory.


baallan:
@nichamon This directory is not redundant with src/contrib/*, as its readme indicates:
```
The preferred location of LDMS contributions is in the ldms/src/contrib where
contributions are directly integrated with the OVIS autotools build. 

That approach may not work for some, particularly developers
wanting or needing to maintain their own separate build system for legal or practical reasons.
For those developers and for other seeking a template of a project
that is dependent on but not part of LDMS, the third-plugins directory is provided.
```

If you feel, given that description, that it should be in another path such as ldms/third-party instead of ldms/src, that's fine. 
This directory provides a build-testable demonstration (if we enable it) that our "as-installed" ldms supports building third party plugins. Seems like rather than deleting this, we should be including it in pull-request testing to ensure our headers/libs remain downstream-build friendly.

nichamon:
@baallan I see. I'll leave it as is then.

PR #1060, opened 2022-11-11
Rename the component_id and job_id metrics in slurm_sampler

nichamon:
The 'component_id' metric, an uint64, is the component ID corresponding to the set. The 'job_id' metric, an uint64, is the job ID corresponding to the set at a timestamp. However, the 'component_id' metric in slurm_sampler, an uint64 array, contains the component IDs of each job in the job_id metric. Similarly, the 'job_id' metric in slurm_sampler, an uint64 array, contains the slurm job IDs on the system.

store_sos assumes that component_id and job_id are uint64 so that it can create the joined indices of component_id, job_id, and timestamps. store_sos hits an assert when it stores the slurm set because the component_id and job_id are not uint64.

The patch renames slurm_sampler's component_id and job_id metrics because they have different meanings from the other sampler plugins.

### DIFF
--- a/ldms/src/sampler/slurm/slurm_sampler.c
+++ b/ldms/src/sampler/slurm/slurm_sampler.c
@@ -273,12 +273,12 @@ static int create_metric_set(void)
 	}
 
 	/* component_id */
-	comp_id_idx = ldms_schema_metric_array_add(job_schema, "component_id",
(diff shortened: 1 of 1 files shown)


nichamon:
I forgot that the store_slurm plugin is the store we should use to store the slurm sets. I'm closing this. I note that we can use the store_sos plugin with the decomposition to store the slurm sets.

PR #1072, opened 2022-12-13
optional timeout for ldms transport

nick-enoent:
Adds connection timeout option for ldms transports

### DIFF
--- a/ldms/python/ldms.pyx
+++ b/ldms/python/ldms.pyx
@@ -1321,7 +1321,6 @@ cdef void xprt_cb(ldms_t _x, ldms_xprt_event *e, void *arg) with gil:
     if e.type == EVENT_DISCONNECTED:
         Py_DECREF(x) # taken when CONNECTED
 
-
 # This is the C callback function for passive transports (the listening
 # transport and all transports accepted by it). It calls the Python callback
 # callable if provided. Otherwise, it works internally to provide blocking
@@ -3010,8 +3009,8 @@ cdef class Xprt(object):
         if self.xprt:
             ldms_xprt_put(self.xprt)
 
-    def connect(self, host, port=411, cb=None, cb_arg=None):
-        """X.connect(host, port=411, cb=None, cb_arg=None)
+    def connect(self, host, port=411, cb=None, cb_arg=None, timeout=None):
+        """X.connect(host, port=411, cb=None, cb_arg=None, timeout=None)
 
         Connect to the remote LDMS peer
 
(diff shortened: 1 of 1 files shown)


nick-enoent:
@narategithub @nichamon Hey guys, would one of you mind taking a look at this? It creates an alternative to the ldmsd connect function, that allows a connection timeout to be specified.

ldms/python/ldms.pyx
```diff
@@ -3056,6 +3056,45 @@ cdef class Xprt(object):
...
+                                       xprt_cb, <void*>self)
+        if rc != 0:
+            # synchronously failed, self.xprt is no good. Need to "put" it down.
+            ldms_xprt_put(self.xprt)
+            self.xprt = NULL
+            raise ConnectionError(rc, "ldms_xprt_timed_connect_by_name() error: {}" \
+                                      .format(ERRNO_SYM(rc)))
+        with nogil:
+            sem_timedwait(&self._conn_sem, &tout_spec)
+        if self._conn_rc:
```
narategithub, line 3092: `self._conn_rc` was initialized as `0`. In the case of `sem_timedwait()` timeout and `xprt_cb()` has not been called yet, `self._conn_rc` will still stay `0`, leading to returning a wrong return code.

ldms/python/ldms.pyx
```diff
@@ -3056,6 +3056,45 @@ cdef class Xprt(object):
...
+        rc = ldms_xprt_connect_by_name(self.xprt, BYTES(host), BYTES(port),
+                                       xprt_cb, <void*>self)
+        if rc != 0:
+            # synchronously failed, self.xprt is no good. Need to "put" it down.
+            ldms_xprt_put(self.xprt)
+            self.xprt = NULL
+            raise ConnectionError(rc, "ldms_xprt_timed_connect_by_name() error: {}" \
+                                      .format(ERRNO_SYM(rc)))
+        with nogil:
+            sem_timedwait(&self._conn_sem, &tout_spec)
```
narategithub, line 3091: I think you also need to add a logic in `xprt_cb()` function to detect the events on the connecting endpoint that has already timed out and handle them properly (e.g. not posting a semaphore as the matching `sem_timedwait` has already timed out).

ldms/python/ldms.pyx
```diff
@@ -3056,6 +3056,45 @@ cdef class Xprt(object):
             rc = self._conn_rc
             raise ConnectionError(rc, "Connect error: {}".format(ERRNO_SYM(rc)))
 
+    def tconnect(self, host, port=411, timeout=5):
```
narategithub, line 3059: This function works very similarly to `def connect()`. I don't mind if you add `timeout` parameter to the old `connect()` function and make it timed wait instead of waiting indefinitely. I also don't mind if it is implemented separately as `tconnect()`. I'll leave the decision to combine or to separate to you :)

narategithub (changes requested):

ldms/python/ldms.pyx
narategithub: This has been resolved with 81f3355 👍

narategithub (approved):
👍

narategithub (approved):
👍

PR #1074, opened 2022-12-20
sos stores for linux_proc_sampler streams

baallan:
These stores enable sos capture of env, argv, and file data streams from the linux proc sampler.
They require #1056 be merged to provide correct data.


bschwal:
@tom95858 this is the pull request for the sos store that will eventually get overwritten by a generic json store, but we need it in TOSS for the time being. This goes with the sampler in pull request 1056. Could you add this in after our discussion today?

PR #1076, opened 2022-12-22
LDMSD configuration counter

nick-enoent:
Add configuration counter that increments when a ldmsd successfully completes a configuration request

Config counter only increments for requests that modify the current configuration of the ldmsd

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -127,6 +127,7 @@ char myname[512]; /* name to identify ldmsd */
 char myhostname[80];
 char ldmstype[20];
 int foreground;
+int cfg_cntr = 0;
 pthread_t event_thread = (pthread_t)-1;
 char *logfile;
 int log_truncate = 0;
@@ -434,6 +435,16 @@ const char *ldmsd_loglevel_to_str(enum ldmsd_loglevel level)
 	if ((level >= LDMSD_LDEBUG) && (level < LDMSD_LLASTLEVEL))
 		return ldmsd_loglevel_names[level];
 	return "LDMSD_LNONE";
+}
+
+void ldmsd_inc_cfg_cntr()
+{
+	__atomic_fetch_add(&cfg_cntr, 1, __ATOMIC_SEQ_CST);
+}
+
+int ldmsd_cfg_cntr_get()
+{
+	return cfg_cntr;
 }
 
 const char *ldmsd_myname_get()
--- a/ldms/python/ldmsd/ldmsd_request.py
+++ b/ldms/python/ldmsd/ldmsd_request.py
@@ -277,6 +277,7 @@ class LDMSD_Req_Attr(object):
 class LDMSD_Request(object):
     EXAMPLE = 1
     GREETING = 2
+    CFG_CNTR = 3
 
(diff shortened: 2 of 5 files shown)


nick-enoent:
@narategithub @nichamon Hey guys, any thoughts on this PR as implemented? It creates a "configuration counter" that increments with each successful configuration request that modifies a ldmsd's current configuration. i.e. configuration requests such as prdcr_stats will not increment the counter. 

In order to do this, I utilized a switch statement with all of the commands that do _not_ modify the ldmsd's current configuration. I could see this being mildly irritating to maintain in the future when/if we introduce new status commands, as they will have to be manually added to this switch statement.

An alternative I considered was creating an array of all "status" config requests in a centralized location.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -436,6 +437,16 @@ const char *ldmsd_loglevel_to_str(enum ldmsd_loglevel level)
 	return "LDMSD_LNONE";
 }
 
+void ldmsd_inc_cfg_cntr()
+{
+	__sync_fetch_and_add(&cfg_cntr, 1);
```
narategithub, line 442: From https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html:

> These functions are implemented in terms of the ‘__atomic’ builtins (see __atomic Builtins). They should not be used for new code which should use the ‘__atomic’ builtins instead.

So, I suggest that you use `__atomic_fetch_add(&cfg_cntr, 1, __ATOMIC_SEQ_CST)` instead.

narategithub (commented):
Just a minor suggestion about `__sync_fetch_and_add()` (please see in-line comment). Everything else looks fine to me.

ldms/src/ldmsd/ldmsd.c
nick-enoent: Thanks @narategithub, I'll update that. I actually ran into the __atomic builtins when I was implementing this, but figured I'd try to stay consistent with the existing code.

nichamon:
> @narategithub @nichamon Hey guys, any thoughts on this PR as implemented? It creates a "configuration counter" that increments with each successful configuration request that modifies a ldmsd's current configuration. i.e. configuration requests such as prdcr_stats will not increment the counter.
> 
> In order to do this, I utilized a switch statement with all of the commands that do _not_ modify the ldmsd's current configuration. I could see this being mildly irritating to maintain in the future when/if we introduce new status commands, as they will have to be manually added to this switch statement.
> 
> An alternative I considered was creating an array of all "status" config requests in a centralized location.

Two other methods are 1) to use the flag in the request_handler_entry struct https://github.com/ovis-hpc/ovis/blob/1e83ade3aa723c269507dc0d0c269951e08fa4f9/ldms/src/ldmsd/ldmsd_request.c#L186 and 2) to add another field in the request_handler_entry struct to differentiate between status and non-status request commands. For the prior, you could pick a bit and check the bit instead of the switch statement. Either of the methods will force us to specify whether the request command is a status or not in the request_handler[] array. https://github.com/ovis-hpc/ovis/blob/1e83ade3aa723c269507dc0d0c269951e08fa4f9/ldms/src/ldmsd/ldmsd_request.c#L299

PR #1079, opened 2023-01-04
Improve stream_client_dump's display in ldmsd_controller and ldmsctl

nichamon:


tom95858:
@nichamon I would like to rename this command to stream_status. What do you think?

ldms/src/ldmsd/ldmsctl.c
```diff
@@ -1763,7 +1765,19 @@ static void resp_stream_client_dump(ldmsd_req_hdr_t resp, size_t len,
 		return;
 	}
 	json_parser_free(parser);
-	__json_value_print(json, 0);
+	printf("%-15s %-15s %-60s\n", "stream", "context", "cb_fn");
+	printf("--------------- --------------- -----------------------------------------------------------\n");
+	l = json_value_find(json, "streams");
+	for (s = json_item_first(l); s; s = json_item_next(s)) {
+		n = json_value_cstr(json_value_find(s, "name"));
```
tom95858, line 1772: We should protect against getting NULL back from this function.

nichamon:
> @nichamon I would like to rename this command to stream_status. What do you think?

I agree. It's done.

ldms/src/ldmsd/ldmsctl.c
nichamon: Done.

PR #1080, opened 2023-01-05
Updates for ibmad_records_sampler

morrone:
We came across a system where there are RoCE devices on a node. Those device show up in the umad device listing, but connections to the devices for sampling always fail. Also, the device lookup seems to be especially long on that system, and my suspicion is that the RoCE device(s) may be to blame there as well.

First, we add a check to skip over device ports with a link_layer that is no InfiniBand (the RoCE ones where link_layer Ethernet).

Next we add a refresh_interval_sec configuration option so we can refresh the list of available devices on a longer interval so we can avoid the high cost of device lookup on most sampling intervals.

### DIFF
--- a/ldms/src/sampler/ibmad_records_sampler/ibmad_records_sampler.c
+++ b/ldms/src/sampler/ibmad_records_sampler/ibmad_records_sampler.c
@@ -440,33 +440,34 @@ static void interfaces_tree_refresh()
 			char name_and_port[UMAD_CA_NAME_LEN+128];
 			struct rbn *rbn;
 			struct interface_data *data;
+                        umad_port_t *port = ca.ports[j];
 
-			if (ca.ports[j] == NULL)
+			if (port == NULL)
 				continue;
 			else
 				cnt++;
 
-			if (!collect_ca_port(ca_names[i], ca.ports[j]->portnum)) {
+			if (!collect_ca_port(ca_names[i], port->portnum)) {
 				continue;
 			}
-			if (ca.ports[j]->state != PORT_STATE_ACTIVE) {
+			if (port->state != PORT_STATE_ACTIVE) {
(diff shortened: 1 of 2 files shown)


baallan:
@morrone could you use a macro or constant with some sensible name (including 'default') and defined near the top of the file instead of the magic number in:
```
                conf.refresh_interval = (time_t)600;
```
please?

ldms/src/sampler/ibmad_records_sampler/ibmad_records_sampler.c
```diff
@@ -764,6 +788,23 @@ static int config(struct ldmsd_plugin *self,
                 return rc;
         }
 
+        value = av_value(avl, "refresh_interval_sec");
+        if (value != NULL) {
+                char *end;
+                long val;
+
+                strip_whitespace(&value);
```
baallan, line 796: the strip_whitespace call (and the bit called) should be entirely unnecessary, as the av processing should be delivering values that are already stripped (lib/src/ovis_util/util.c:tokenize:strtok_r(... " \t\n"...). 
Did you find a case where something was not stripped? If so we should fix tokenize.

tom95858: @morrone, @baallan what is the status of this pull request?

baallan: If @morrone is busy but will fix the nits soon post-release, we can call it good enough for release.

morrone: Yeah, I probably won't get to it. And the same nits are elsewhere in the code already (they were literal cut-and-pastes from elsewhere).

PR #1081, opened 2023-01-06
Improve stream_status

nichamon:
Rename stream_dir to stream_status
- Add the publish statistics to stream_status


nichamon:
**stream_status result**

A sampler's stream_status
image

The samplerd republished only the app stream to the aggregator.

An aggregator's stream_status
image

nichamon:
**prdcr_stream_status result**
Users send the command to an aggregator, and the aggregator acts as a proxy and forwards the _stream_status_ command to producers that are matched the regular expressions.

`ldmsd_controller> prdcr_stream_status regex=.*`

image

PR #1082, opened 2023-01-10
Add the capability to reset the updater's counters thru updtr_status

nichamon:
Users can do 'updtr_status reset=true' in ldmsd_controller to reset the updaters' counters.

The descriptions of the updaters' counters are provided in the command help. Do 'help updtr_status' in ldmsd_controller to see the descriptions.


baallan:
So doesn't this make 'updtr_status' both a status command and a configuration command?
I wonder if we shouldn't have a reset_status command that such as "reset_counters name=updtr_status" to keep things clear.

tom95858:
@baallan, we currently define a configuration change as something that changes what is sampled, at what frequency, etc... The status commands do not do that, but they do affect what subsequent calls to status commands might see. For now, we are going to accept these changes because our users have said that they are very useful.

PR #1083, opened 2023-01-11
Make the set_stats command support push updaters

nichamon:
In the push mode, LDMSD uses the LDMS sets' update_hint to calculate the frequency of data being pushed from the upstream, instead of the update interval.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -6673,7 +6673,15 @@ static char * __set_stats_as_json(size_t *json_sz)
 						rc = regexec(&match->regex, prd_set->inst_name, 0, NULL, 0);
 						if (rc)
 							continue;
-						freq = 1000000 / (double)prd_set->updt_interval;
+						if (updtr->push_flags) {
+							/* Use the update hint to calculate the frequency */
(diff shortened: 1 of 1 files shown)


tom95858:
In push_mode, the update frequency whether from the hint or the updater isn't used. Instead the update frequency is driven by the sampler in the upstream ldmsd and how often it calls ldms_transaction_end(). Am I missing something here?

Also, the way the commit message is written makes sound like you changed it from updt_hint to prd_set->updt_intrvl not the other way around

nichamon:
> In push_mode, the update frequency whether from the hint or the updater isn't used. Instead the update frequency is driven by the sampler in the upstream ldmsd and how often it calls ldms_transaction_end(). Am I missing something here?
> 
> Also, the way the commit message is written makes sound like you changed it from updt_hint to prd_set->updt_intrvl not the other way around

I used the update_hint to calculate the set_load value. As you said, the push frequency is driven by how often the sampler plugin calls ldms_transaction_end(). When I made that patch, I considered that the sampler plugin calls ldms_transaction_end() as often as the sampling interval. I realized now that it isn't true in some sampler plugins that update their sets based on events, e.g., stream.

tom95858:
@nichamon, should we close this?

PR #1084, opened 2023-01-13
Refactor the lookup error path

tom95858:
The lookup error path will call ldms_set_delete which in turn will call ldms_xprt_set_delete which informs downstream aggregators of the deleted set. However, in the error path the downstream peer has never received a dir_add because the set was never published. This results in the downstream reporting an error that a dir_delete was received for a set that was never added.

It also fixes the case where the set is created in the local tree, but an error occurs in the RDMA of the meta data into the local set. In this case, the set was not being removed from the set tree which would cause all subsequent lookup requests to fail synchronously because the set was already present in the tree.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -818,7 +818,7 @@ static void __destroy_set(void *v)
 	pthread_mutex_unlock(&__del_tree_lock);
 }
 
-void ldms_set_delete(ldms_set_t s)
+void __ldms_set_delete(ldms_set_t s, int notify)
 {
 	ldms_t x;
 	struct ldms_set *__set;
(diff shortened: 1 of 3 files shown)


narategithub (changes requested):
@tom95858

I think the logic is correct. However, the unused variable `ldms_name_t name` variable at `ldms.c:3746` failed the build test.

narategithub (approved):
👍

nichamon:
@tom95858 I think we should call `__ldms_set_delete(set, 0)` at line 2386 in ldms_xprt.c instead of `ldms_set_delete(set)`, in `__handle_lookup()`. The set has not been published to the downstream yet, so LDMSD must not send the SET_DELETE message to the downstream daemon.

tom95858:
That's the meaning of the 0 value as the 2nd parameter. It says "don't send SET_DELETE".

tom95858:
Sorry @nichamon , misunderstood your comment. I think you are correct, but I'm going to regression test at scale to be sure.

nichamon (approved):

PR #1085, opened 2023-01-14
Remove `installcheck-local` in netlink/Makefile.am

narategithub:
The installcheck scirpt `test_slps.sh` hanged in github workflow for Ubuntu-20.04, causing the `distcheck` workflow to fail after 6 hours. This is a work around so that we can move forward while the plugin authors figuring out the cause.


tom95858:
Thanks @narategithub ...

PR #1091, opened 2023-01-18
Fix a possible deadlock caused by ldmsd_stream_publish()

nichamon:
The deadlock occurs when applications publishes data on the same stream multiple times by calling ldmsd_stream_publish().

### DIFF
--- a/ldms/src/ldmsd/ldmsd_stream.c
+++ b/ldms/src/ldmsd/ldmsd_stream.c
@@ -526,6 +526,7 @@ int ldmsd_stream_publish(ldms_t xprt,
 	s->s_pub_info.count += 1;
 	s->s_pub_info.last_ts = now;
 	s->s_pub_info.total_bytes += data_len;
+	pthread_mutex_unlock(&s->s_lock);
 
  err:
 	if (buf)


baallan:
@nichamon was this discovered by being "provoked" by the stream usage in simple_lps, or did you just find it while looking around? My intent in slps was that slps provided serialization of message publications so that this kind of bug shouldn't be hit.

nichamon:
> @nichamon was this discovered by being "provoked" by the stream usage in simple_lps, or did you just find it while looking around? My intent in slps was that slps provided serialization of message publications so that this kind of bug shouldn't be hit.

I found this by going through Github's log of one of the failed dist-check actions. The last message was from `sim_lps.c` logging that it published stream to a client. https://github.com/ovis-hpc/ovis/actions/runs/3934348367/jobs/6729008606#step:6:5294

`simple_lps.c` didn't report whether it successfully published the stream or not, so I looked into the `ldmsd_stream_publish()` API, and this is how I also found the line that you reset the variable that receives `ldmsd_stream_publish()`'s returned code in `simple_lps.c`.

narategithub:
I also made another branch in my repo ([here](https://github.com/narategithub/ovis/pull/21)) that matches this pull request minus commit f797563d so that the `installcheck` of `netlink` is called. It passed the workflow test. I think this pull request addresses the root cause that made netlink installcheck to stall.

However, since `netlink` is not a core functionality, I think it should be in `ldms-test`.

baallan:
the netlink functionality enables process monitoring that is not specific hardware dependent nor specific scheduler dependent, so I don't know why it wouldn't be included in installcheck. It's included in our jenkins runs at snl. You are welcome to test it in another repo of tests also, but please don't break/remove existing checks.

nichamon:
@tom95858 We need this pull request before the release. It is the root cause of stalled Github's dist-check actions. However, Narate and I don't understand why pull request #1081 which introduced the bug passed Github's tests.

PR #1092, opened 2023-01-20
Fix prdcr_stream_status's handler

nichamon:
- Remove a possible race between forwarding the stream_status request to the connected producer and receiving the response back. The race could cause LDMSD to free the forwarding request context before LDMSD receives all the responses.


nichamon:
The ldms-test test script accompanying the patch is here. https://github.com/ovis-hpc/ldms-test/pull/34

PR #1094, opened 2023-01-26
Make store_sos & store_csv not support multiple lists w/o decomposition

nichamon:


ldms/src/store/store_csv.c
```diff
@@ -1500,6 +1500,13 @@ static int store(ldmsd_store_handle_t _s_handle, ldms_set_t set, int *metric_arr
 		for (i = 0; i < metric_count; i++) {
 			if (LDMS_V_LIST == ldms_metric_type_get(set, metric_array[i])) {
 				s_handle->num_lists++;
+				if (s_handle->num_lists > 1) {
+					msglog(LDMSD_LERROR, PNAME
+						": set '%s' contains multiple lists. "
+						"Please store the set using a decomposition.\n",
+						ldms_set_instance_name_get(set));
+					return 0;
```
narategithub, line 1508: should this be an error code?

baallan: yes. :)

nichamon: I will change this to EINVAL.

FYI, there is only one caller, and it's in the update complete callback. LDMSD will not stop the storage policy or do anything else when the `store()` function returns an error code. The store plugin is responsible for reporting an error message and managing its resources.

PR #1098, opened 2023-01-31
subdirify simply built plugins and related man pages

baallan:
This moves many misc plugins (below) into their own build subdirectories with their associated man pages, consistent with recent development practices. No C code changes are present, except where git whitespace enforcement required it.
plugins:
clock, lnet_stats, sysclassib, opa2, fptrans, procnfs, procnetdev, perfevent msr_interlagos papi_sampler, procdiskstats, syspapi_sampler, atasmart slurm_sampler, slurm_notifier, store_slurm


ldms/src/sampler/clock/Makefile.am
```diff
@@ -0,0 +1,16 @@
...
+lib_LTLIBRARIES =
+dist_man7_MANS =
+dist_man1_MANS =
+
+AM_CPPFLAGS = @OVIS_INCLUDE_ABS@
+AM_LDFLAGS = @OVIS_LIB_ABS@
+COMMON_LIBADD = -lsampler_base -lldms -lovis_util -lcoll \
+		@LDFLAGS_GETTIME@
+
+if ENABLE_CLOCK
```
morrone, line 11: These conditionals are all redundant. They are already taken care of in the in conditional SUBDIRS in the Makefile.am on level up.

baallan:
I will remove them from the level up then, since the details controlled should be near the control statement for maintainability and for simplicity of clone-modify development.

ldms/src/sampler/clock/Makefile.am
baallan: I will remove them from the level up then, since the details controlled should be near the control statement for maintainability and for simplicity of clone-modify development.

morrone: I think staying with the conditionals around SUBDIRS is actually the way to go right now.

There is an autotools bug that would require extra handling of the man page if you move to only doing the conditional the per-sampler automake files. And we clearly are mostly using the SUBDIRS approach these days. I'm not sure that I see any different in difficult for "clone-modify development" either.

baallan: Can you cite the automake bug you would have us code around? I've been testing on rhel7 (the oldest thing we support) and haven't managed to hit any problems with distcheck. I did notice inconsistent use elsewhere of extra_dist for a man page.

baallan: @tom95858 can you clarify which subdirs approach you prefer, a or b?
A)
```
if xyz_enabled
subdirs += xyz
endif
```

or

B)
```
subdirs += xyz
# xyz/Makefile.am from the plugin author handles the conditionals
subdirs += pdq
subdir += sl
```

morrone: > Can you cite the automake bug you would have us code around? I've been testing on rhel7 (the oldest thing we support) and haven't managed to hit any problems with distcheck. I did notice inconsistent use elsewhere of extra_dist for a man page.

Yes, there needed to be an EXTRA_DIST of the manpage outside of the conditional when the man page line is inside the conditional. distcheck wouldn't necessarily find the issue. The testing probably needs to go like this:

1. "configure" ldms to explicitly disable a plugin that uses this idion
2. generate a dist tarball with "make dist"
3. Check if tarball contains the man page in question

baallan: @morrone and to be clear, the (A) approach of my question to Tom prevents this bug being triggered?

morrone: I think that is the case.

baallan:
I have refactored this update to put the conditional build checks where possible in sampler/Makefile.am rather than in subdirs
to work around potential dist rule bugs and for esthetics.

baallan:
@tom95858 I believe this pr is ready to go.

tom95858:
@morrone, are you happy with this approach?

PR #1100, opened 2023-02-01
update metric types to signed for lustre_mdc sampler

baallan:
This commit:
- changes the incorrectly typed (u64) metrics to s64 to match the lustre code producing them (schema change). 
- removes the "-mdc-$hexaddr" suffix from the "mdc" field; the hex part previously varied from node to node and made the "mdc" field useless as an index.
- adds the option (default enabled) that if rollover to negative values occurs the sampler resets the counters. The time (epoch seconds) of the most recent reset is reported in the "last_reset metric". If auto_reset=0 disables the reset process, a single log message occurs should the need for a reset occur.

### DIFF
--- a/ldms/src/sampler/lustre_mdc/Plugin_lustre_mdc.man
+++ b/ldms/src/sampler/lustre_mdc/Plugin_lustre_mdc.man
@@ -49,12 +49,21 @@ mdc_timing=0
 Optionally exclude timing data from /sys/kernel/debug/lustre/mdc/*/stats. If given,
 the sampler may be run by unprivileged users. If /sys/kernel/debug/ cannot be opened
 by the user, it is a configuration error unless mdc_timing=0 is given.
+.TP
+auto_reset=0
+.br
+Turn off the default behavior of resetting the counters when an overflow condition is detected.
+Reset is implemented by writing 0 to the corresponding /proc or /sys file.
 .RE
 
 .SH SCHEMA
 The default schema name is lustre_mdc_ops_timing with all the data described in DATA REPORTED below included. If mdc_timing=0 is given, only the operation counts from md_stats are reported and the default schema name changes to lustre_mdc_ops.
 
 .SH DATA REPORTED
+
+fs_name: The lustre file system name, e.g. xscratch.
+mdc: The mdc target that goes with the metrics, e.g. xscratch-MDT0000.
+last_reset: The time of the last reset performed by this sampler for any of its metric sets.
 
 Operation counts from /proc/fs/lustre/mdc/*/md_stats.
 See also kernel source lustre/lustre/obdclass/lprocfs_status.c and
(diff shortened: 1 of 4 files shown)


tom95858:
@morrone, @baallan it seems like we should discuss this a little bit. The patch says that rollover 'cannot be detected', but in fact it can be by comparing the current sample vs. the last sample; although checking for <0 doesn't require the last sample.

That said, it seems like we don't really have a solution here because if you're looking for true totals in long-running servers we would have to maintain these totals with either floating point, or 128b numbers (which will eventually roll over too). So can we take a look at our use cases and decide whether what we should really be saving is diffs? With a diff we can recover totals, and avoid the overflow problem altogether. We would still use S64 to make the overflow check easy and use the overflow condition to a) do special math on the negative number and b) issue the write to /proc/... to reset the counters from the sampler.

baallan:
Given Tom's concerns, I would be willing to add an optional behavior that when overflow is detected (x < 0) we reset the counters for that group. I don't think we should be in the business of synthetic counters (128 bit, or whatever, that a sysadmin cannot validate by comparing ldms_ls output to cat /proc/somefile) at the data acquisition stage; if we want any synthetic permutation, it belongs in analysis or a transform plugin elsewhere in the pipeline.

@tom95858 With 'cannot be detected' I was referring (poorly worded in retrospect) to the pre-patch state where the u64 assumption/usage hides the sign change (and worse, perhaps causes the sscan to fail on hitting the '-').  With the change of the parse to match the data actually being provided by lustre, we can then correctly detect change and do the reset.

baallan:
After more tea, it's sort of obvious that the correct condition to perform a reset (particular so the stddev computation from sumsquares can be "correct since the last reset" (the best we can do) is to check is any counter negative or has any counter decreased from the previous sample (if there is a previous sample). 
But in the presence of other sources of reset events, this leads to us 'double resetting' because their reset is also detected as a drop.

We may also optionally report the time of the last reset *we* did as metadata.
As with previous reset history with our admins (IB stats from the 32 bit days), they have the option to be resetting counters by other means, in which case they should be able to not use our reset behavior and be responsible for their choice of reset policies and the impact on collected data.

tom95858:
@morrone, @baallan - did we come to a resolution on this patch?

tom95858:
@baallan, please advise when you have made the changes you refer to. Perhaps change this to "draft" in the meantime?

morrone:
I don't have an opinion on reset behavior at the moment. The change from u64 to s64 does seem to match Lustre's code better. There is some question in my mind about when we can change schema like this between ldms release. But maybe lustre_mdc isn't used widely enough yet to worry about that at the moment?

tom95858:
@baallan, are we actually seeing this overflow behavior in deployment or is this a speculative change?

baallan:
@tom95858 lustre byte counters do roll over in practice. we have servers that have been up for 3 years.

tom95858:
@baallan, @morrone Ok, let's accept this change. But having a discussion in general about schema changes to existing samplers is something we should discuss at the UG meeting.

tom95858:
@baallan, I cannot accept this change while it is still marked as draft.

baallan:
@tom95858 correct it is still in draft and should not be accepted yet. I have not yet had time to implement and test the discussed auto-reset behavior needed when the counters drop from previous values (indicating an integer addition overflow). I will remove the draft marking when i've added and tested the rollover behavior.

baallan:
@tom95858 Are ldms_metric_get/set_timestamp on the verge of existing to go with LDMS_V_TIMESTAMP? It would be handy to have in some metric sets, including the one in work here.

baallan:
@tom95858 I've revised the patch per the revised initial description of this PR. It has been tested on one of my local systems. I believe it is ready to merge.

PR #1101, opened 2023-02-02
Normalize all ldmsd_communicator response messages to strings

nick-enoent:
Add json formatting for status request responses in ldmsd_controller


ldms/python/ldmsd/ldmsd_communicator.py
```diff
@@ -196,21 +196,29 @@
...
-        status = json.loads(resp['msg'])
-    return status
+def fmt_status(msg):
+    """
+    Format communicator status response string into json object
+    """
+    if msg is not None:
+        try:
+            msg = json.loads(msg)
+        except Exception as e:
```
tom95858, line 216: @nick-enoent  Do you want to set msg = None in this case so the caller can universally check for None to know the return value was bad? As written the msg could != None, but is otherwise text garbage in which case the caller gets a garbage string returned instead of either None or a valid a JSON object.

PR #1110, opened 2023-02-22
Fix the plugn_sets handler in ldmsd_controller and ldmsd_communicator

nichamon:


nichamon:
@nick-enoent @tom95858 Here is the patch to fix plugn_sets in ldmsd_controller and ldmsd_communicator. I'll create a pull request in ldms-test for the test cases when I complete them.

PR #1111, opened 2023-02-22
Fix prdcr_add_handler to properly return an error

nichamon:
The handler did not set the errcode when the given type is invalid. It results in a silense error. Also, the handler did not validate whether the given interval string. The patch makes the handler responds with an error if the given interval is not a positive number.


nichamon:
The issues were found by running the test coverage in https://github.com/ovis-hpc/ldms-test/pull/37.

PR #1113, opened 2023-02-24
Take a store plugin's reference when ldmsd links it to a strgp

nichamon:
Without the patch, whenever LDMSD receives a term request for a store plugin, the store plugin's resources will be freed, and the plugin will be unloaded. LDMSD will, then, segfault if the store links to a running storage policy.


nichamon:
The bug was discovered by running the test cases in https://github.com/ovis-hpc/ldms-test/pull/38.

PR #1114, opened 2023-02-24
Fix the plugn_start and plugn_term in ldmsd_controller and ldmsd

nichamon:


nichamon:
The bugs were discovered by running the test cases in https://github.com/ovis-hpc/ldms-test/pull/38.

PR #1115, opened 2023-02-27
Fix greeting communicator

nichamon:


nichamon:
I found the bugs by running the test script in https://github.com/ovis-hpc/ldms-test/pull/32. The test script uses the greeting() configuration command to create a long-string response.
@nick-enoent The patch fixes the greeting() handlers in ldmsd_communicator and ldmsd_contorller.

PR #1118, opened 2023-02-28
Add LDMS_V_TIMESTAMP accessor functions

tom95858:
The LDMS_V_TIMESTAMP was added as part of the
decomposition support, but the accessor functions
were never added.


narategithub:
Looks OK to me.

PR #1119, opened 2023-02-28
Add ldms_list_tail API

tom95858:
This change adds a function, ldms_list_last(), that returns the tail of a list.


narategithub:
👍

PR #1121, opened 2023-03-01
Fix error messages in ldmsd_controller

nichamon:
- Fix an error message in do_prdcr_start()
- Remove the error code in the message from do_plugn_load(). The error code may be -1 if LDMSD fails to load the plugin library. It is not meaningful to report -1. The message sent from LDMSD contains the reason.


nichamon:
The bug was found by the weekly regression test. `ldmsd_ctrl_test` failed in https://github.com/ldms-test/weekly-report/blob/master/summary.md. The test has been updated in https://github.com/ovis-hpc/ldms-test/pull/40.

PR #1122, opened 2023-03-02
thread_stats parameter for daemon_status

nick-enoent:
Add True/False parameter 'thread_stats' to daemon_status to optionally return the daemon's thread statistics in daemon_status response. This reduces the request load for users only seeking the daemon's status


nick-enoent:
@narategithub @nichamon I'm guessing we'll need to modify ldms-test to accommodate this change. I can take a look at ldms-test this afternoon to make sure the tests are all up to date before we merge this.

nichamon:
@nick-enoent Please go ahead to make any necessary changes. I am not working on any test scripts that relate to the daemon_status. I'm updating store_list_record_test. Thanks!

PR #1124, opened 2023-03-08
add user_debug option to control user job logging of plugin handling messages in slurm_notifier

baallan:
The output of the slurm notifier DEBUG2 macro is visible in user log files. This is not desired in normal operation. This patch adds the 'user_debug' option to the spank plugin to enable user's job log output when the slurm admin is debugging use of the plugin and to route these DEBUG2 messages to slurm_debug2() otherwise.

### DIFF
--- a/ldms/src/sampler/spank/Plugin_slurm_notifier.man
+++ b/ldms/src/sampler/spank/Plugin_slurm_notifier.man
@@ -1,4 +1,4 @@
-.\" Manpage for slurm_notifier
+.\" Man page for slurm_notifier
 .\" Contact ovis-help@ca.sandia.gov to correct errors or typos.
 .TH man 7 "30 Sep 2019" "v4" "SPANK Plugin slurm_notifier man page"
 
@@ -11,6 +11,7 @@ Within plugstack.conf:
 .IR OVIS_PREFIX / LIBDIR /ovis-ldms/libslurm_notifier.so
 .BI stream= STREAM_NAME
 .BI timeout= TIMEOUT_SEC
+.BI [user_debug]
 .BI client= XPRT : HOST : PORT : AUTH
 .RB ...
 .YS
@@ -20,9 +21,8 @@ Within plugstack.conf:
 \fBslurm_notifier\fR is a SPANK plugin that notifies \fBldmsd\fR about job
 events (e.g. job start, job termination) and related information (e.g. job_id,
 task_id, task process ID). The notification is done over \fBldmsd_stream\fR
-publish mechanism. \fBslurm_sampler\fR (see \fBPlugin_slurm_sampler\fR(7)) and
-\fBpapi_sampler\fR (see \fBPlugin_papi_sampler\fR) are known plugins that
-subscribe to the notification.
+publish mechanism. See SUBSCRIBERS below for plugins known to consume
+the spank plugin messages.
 
 .BI stream= STREAM_NAME
 specifies the name of publishing stream. The default value is \fIslurm\fR.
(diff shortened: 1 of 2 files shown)


ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -74,9 +74,11 @@
 static char *stream;
 #define SLURM_NOTIFY_TIMEOUT 5
 static time_t io_timeout = SLURM_NOTIFY_TIMEOUT;
+static int user_debug = 0;
 
 static const char *stepd_event = "";
 #define DEBUG2(FMT, ...) do { \
+	if (user_debug) \
```
tom95858, line 81: It seems like this is left-over code from when the sampler was being developed. I seem to recall that originally this code called slurm_debug2(...), not printf directly and that the slurm configuration debug level controlled what was seen. In any event it seems like that is the way that it should work, not with a global variable like this. BTW this is consistent with the spank plugin documentation here: https://slurm.schedmd.com/plugins.html

baallan: While I agree the obvious thing to do would be to revert Mon's redefinition of DEBUG2 that used printf, it turns out that from an admin-controlled feature point of view it is much easier to debug/trace what the new plugin is doing by examining the user log file with the spank messages in context. As a general rule in my shop, the slurm daemon logs get routed off to splunk and are difficult or impossible to access for the performance monitoring team. 

I'd be happy to update the patch to provide 

- 'debug=0' no logging
- 'debug=1' log to user log
- 'debug=2' log to slurm_debug2

Thoughts @nichamon ?

morrone: I'm not entirely sure that all of slurm_* log functions are available, are they? There seems to be a slurm_spank_log(), but that is for printing messages back through srun, I think. I really don't know.

tom95858: > 

The logging functions are all defined in spank.h and I have tested that they are in fact available to plugins. The macro in question DEBUG2 used to call slurm_debug2 instead of printf; the change was made (and accidentally left in) when debugging the plugin because as Ben points out the slurm_debug2 goes to the slurm log instead of the output of the job.

It is more convenient when debugging to see these messages in the job output then looking at the slurm logs. That said, IMO the default behavior should be that plugin messages go to the slurm log as governed by the slurm DebugLevel configuration, not a plugin specific configuration. As long as the default behavior is to let the slurm configuration govern where and what messages are logged, then I am good with this.

It should be noted that per Ben's change when in 'debug' mode, _all_ messages end up in the job output and the configured DebugLevel is ignored. That said, when in debug mode ... who cares...

From spank.h 
```
/*
 *  Slurm logging functions which are exported to plugins.
 */
extern void slurm_info (const char *format, ...)
  __attribute__ ((format (printf, 1, 2)));
extern void slurm_error (const char *format, ...)
  __attribute__ ((format (printf, 1, 2)));
extern void slurm_verbose (const char *format, ...)
  __attribute__ ((format (printf, 1, 2)));
extern void slurm_debug (const char *format, ...)
  __attribute__ ((format (printf, 1, 2)));
extern void slurm_debug2 (const char *format, ...)
  __attribute__ ((format (printf, 1, 2)));
extern void slurm_debug3 (const char *format, ...)
  __attribute__ ((format (printf, 1, 2)));

/*
 * Print at the same log level as error(), but without prefixing the message
 * with "error: ". Useful to report back to srun commands from SPANK plugins,
 * as info() will only go to the logs.
 */
extern void slurm_spank_log(const char *, ...)
  __attribute__ ((format (printf, 1, 2)));
```

ldms/src/sampler/spank/Plugin_slurm_notifier.man
```diff
@@ -31,6 +31,10 @@ specifies the name of publishing stream. The default value is \fIslurm\fR.
 is the number of seconds determining the time-out of the LDMS connections
 (default \fI5\fR).
 
+.BI debug=1
+enables certain plugin management debugging messages to the user's slurm log.
```
tom95858, line 35: I'm not sure this is true, if enabled, messages go to the user's job output, not the 'slurm log'. I think this is just a wording issue.

ldms/src/sampler/spank/slurm_notifier.c
```diff
@@ -427,9 +429,15 @@ void add_client(struct client_list *cl, const char *spec)
...
 	const char *timeout = NULL;
+	const char *dbg = NULL;
 	int rc;
 
 	for (rc = 0; rc < argc; rc++) {
+		if (0 == strncasecmp(argv[rc], "debug", 5)) {
+			dbg = get_arg_value(argv[rc]);
+			if (dbg && dbg[0] == '1')
+				user_debug = 1;
+		}
```
tom95858, line 440: It seems like this is a bit wonky. What if the user did 'debug=01'? Also this accepts 'debugboogie=1'. So it seems like if we see 'debug' we should just turn on this mode.

ldms/src/sampler/spank/slurm_notifier.c
tom95858, line 81: This needs an else,
```
slurm_debug2(...)
```
otherwise without _debug_ we get no messages at all regardless of the slurm message level configuration.

ldms/src/sampler/spank/slurm_notifier.c
baallan: I just followed the option handling pattern used in the rest of the plugin.

ldms/src/sampler/spank/slurm_notifier.c
morrone: Whether or not they actually go to the slurm log might depends on which context the spank plugin is called from. I'm not sure that all will result in them going there.

tom95858: @morrone, right ... sometimes these log functions are called from slurmd, sometimes from slurmstepd (srun/root) and sometimes from slurmstepd (user). But I think slurm takes care of that ... at least that is what I've seen in the past.

tom95858: @baallan, That seems reasonable to me.

ldms/src/sampler/spank/slurm_notifier.c
baallan: I have de-wonkied this in the revised patch.

baallan:
@tom95858 this is ready to merge.

Issue #1133, opened 2023-03-14
Segmentation fault in libdecomp_as_is


morrone:
With the following decomposition file:

```
{
    "type": "flex",
    "decomposition": {
        "the_default": {
            "type": "as_is"
        }
    },
    "digest": {
        "*": "the_default"
    }
}
```

I get the following segmenation fault:

```
Thread 1 "ldmsd" received signal SIGSEGV, Segmentation fault.                                                                                                                                       
0x0000155555467819 in __decomp_as_is_config () from /usr/lib64/ovis-ldms/libdecomp_as_is.so                                                                                                         
(gdb) bt                                                                                                                                                                                            
#0  0x0000155555467819 in __decomp_as_is_config () from /usr/lib64/ovis-ldms/libdecomp_as_is.so                                                                                                     
#1  0x000015555546c858 in __decomp_flex_config () from /usr/lib64/ovis-ldms/libdecomp_flex.so                                                                                                       
#2  0x0000555555582b08 in ldmsd_decomp_config ()                                                                                                                                                    
#3  0x00005555555716ff in strgp_add_handler ()                                                                                                                                                      
#4  0x000055555557358c in ldmsd_handle_request ()                                                                                                                                                   
#5  0x00005555555739c6 in ldmsd_process_config_request ()                                                                                                                                           
#6  0x0000555555567385 in __process_config_file.constprop.6 ()                                                                                                                                      
#7  0x00005555555617e4 in main ()
```

morrone:
FYI, was at commit 8838776d for this. I'm trying to get a line number, but an unrelated issue is stopping me at the moment.

morrone:
Line 256:

```
        TAILQ_FOREACH(jidx, &jidxs->item_list, item_entry) {
```

narategithub:
Thank you @morrone for the report. `jidxs` was `NULL` because the "default" `as_is` decomposition has no indices. A short-term work-around would be to define the indices, e.g.

```json
{
    "type": "flex",
    "decomposition": {
        "the_default": {
            "type": "as_is",
            "indices": [
                { "name": "time_comp", "cols": [ "timestamp", "component_id" ] }
            ]
        }
    },
    "digest": {
        "*": "the_default"
    }
}
```

I'll post a patch soon. I'm sorry for this silly bug.

PR #1137, opened 2023-03-15
Fix `as_is` decomposition SEGV when there are no indices

narategithub:
This has been tested with 1 sampler damon - 1 aggregator daemon with the following config:

- *samp.conf*
```conf
# samp.conf
env HOST=$(hostname)
listen xprt=sock port=10000

load name=meminfo
config name=meminfo instance=${HOST}/meminfo producer=${HOST}
start name=meminfo interval=2000000 offset=0

load name=vmstat
config name=vmstat instance=${HOST}/vmstat producer=${HOST}
start name=vmstat interval=2000000 offset=0
```

- *agg.conf*
```conf
# agg.conf
listen xprt=sock port=10001

load name=store_csv
config name=store_csv path=store buffer=0

prdcr_add name=local host=localhost xprt=sock port=10000 type=active interval=2000000
prdcr_start name=local

strgp_add name=any plugin=store_csv container=any regex=.* decomposition=decomp.json
strgp_start name=any

updtr_add name=all interval=2000000 offset=100000
updtr_prdcr_add name=all regex=.*
updtr_start name=all
```

- *decomp.json*
```json
{
    "type": "flex",
    "decomposition": {
        "the_default": {
            "type": "as_is"
        }
    },
    "digest": {
        "*": "the_default"
    }

}
```

The aggregator successfully produced CSV files for `meminfo` and `vmstat`.

### DIFF
--- a/ldms/src/decomp/as_is/decomp_as_is.c
+++ b/ldms/src/decomp/as_is/decomp_as_is.c
@@ -250,6 +250,8 @@ __decomp_as_is_config(ldmsd_strgp_t strgp, json_entity_t jcfg,
 	rbt_init(&dcfg->row_cfg_rbt, __row_cfg_cmp);
 	dcfg->decomp = __decomp_as_is;
 	dcfg->idx_count = n_idx;
+	if (!n_idx)
+		goto out;
 
 	/* foreach index */
 	j = 0;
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 @morrone 

This is the patch for #1133 .

Issue #1141, opened 2023-03-16
transient sets aggregated with push=onchange not leaving L1 after deletion on L0


baallan:
The pid sampler (linux-proc-sampler) creates sets for pids it's notified of and clears these sets when the pid monitored goes away.  It appears that ~1/3 of these sets (388/1200) never get cleaned up on the L1 aggregator, even after 1300 seconds.

baallan:
Will run more diagnostics to see if this looks a lot like #771 at the L1.

baallan:
The symptom is repeatable, but a bit arbitrary, thusly:
- run pid sampler on 20 nodes.
- launch 36 ranks/node of mpiGraph using mpiexec or srun.
- job completes; tracked pids all gone.
- for 2 arbitrary (varies run to run which node) nodes, the 36 local pid-related sets are deleted by the sampler, but hang in the deleting state according to set_stats.
- Also for those same nodes, the messages that the pids are gone do not get delivered to the L1, even though all such messages are published at L0 before the corresponding set_deletes are performed. This is just plain weird and in fact all subsequent message deliveries (e.g. new pid appearance) on the same 2 nodes also do not reach L1.

baallan:
It appears this issue has been resolved by #1153.

baallan:
Actually, it's still found in further testing...

nichamon:
@baallan Thanks for the info!

I assume that no L2 aggregated from L1.

baallan:
> @baallan Thanks for the info!
> 
> I assume that no L2 aggregated from L1.

Correct

nichamon:
@tom95858 

> Also for those same nodes, the messages that the pids are gone do not get delivered to the L1, even though all such messages are published at L0 before the corresponding set_deletes are performed. This is just plain weird and in fact all subsequent message deliveries (e.g. new pid appearance) on the same 2 nodes also do not reach L1.

This observation made me think that the root cause may not be related to the delete and push path. It is instead in the transport path. Did Ben and you reproduce this today? What do you think?

baallan:
@nichamon with the new logging stuff, whats the syntax to turn on logging for just the transport (and maybe the set management) code but no sampler plugin logging?
or do  we still need a recompile with extra flags for transport?

nichamon:
> @nichamon with the new logging stuff, whats the syntax to turn on logging for just the transport (and maybe the set management) code but no sampler plugin logging? or do we still need a recompile with extra flags for transport?

[...]

PR #1142, opened 2023-03-19
Netlink toss changes

baallan:
Fix bugs, man pages, and options so that linux_proc_sampler +  ( spank and/or netlink pid publisher) works with pushing metric sets.

key bugs fixed:
- uptime was not initialized in computation of process start times.
- netlink messages could not be used for dashboarding process events (lacking node name)
- corner case hit at scale lead to sampler not deleting some sets ( this is not the same as #1141) .
- push sets needed genders-generated updtr to enable auto_interval for all other sets to be correct.
- man page for ldms-notify hard to find for admin.


baallan:
@tom95858 I think this is ready to merger after much debugging. The commits are by sub-topic for commenting commits in context; will flatten if you prefer.

PR #1145, opened 2023-03-21
return test_slps to installcheck-local with time,zap io limits

### LINKED ISSUE #1095: `simple_lps.c` resets `ldmsd_stream_publish()`'s returned code variable and then checks the value
While I was investigating the stalled dist-check test, I found this. https://github.com/ovis-hpc/ovis/blob/46ab92012b8fee43dff0d2328b19c157ef49baa2/ldms/src/sampler/netlink/simple_lps.c#L1008

After the variable is set to 0, the program checks whether or not its value is zero. 
https://github.com/ovis-hpc/ovis/blob/46ab92012b8fee43dff0d2328b19c157ef49baa2/ldms/src/sampler/netlink/simple_lps.c#L1009

What is the objective of the check?

baallan:
this checks the return code and wraps test_slps in 'timeout' so it cannot hang forever, should a transport hang occur.
also
fixes #1095

### DIFF
--- a/ldms/src/sampler/netlink/input/test_slps.sh
+++ b/ldms/src/sampler/netlink/input/test_slps.sh
@@ -31,7 +31,6 @@ set_kokkos () {
 }
 
 timeout=1m
-export ZAP_MAX_IO=8
 export ZAP_IO_MAX=8
 $BIN/ldmsd -x sock:10444:localhost -a $LDMS_AUTH -l ./test_slps.daemon.1.log  -r ./test_slps.pid.1 -v DEBUG
 $BIN/ldmsd -x sock:10445:localhost -a $LDMS_AUTH -l ./test_slps.daemon.2.log  -r ./test_slps.pid.2 -v DEBUG
--- a/ldms/src/sampler/netlink/input/test_slps.sh
+++ b/ldms/src/sampler/netlink/input/test_slps.sh
@@ -30,6 +30,9 @@ set_kokkos () {
 	export KOKKOS_LDMS_RECONNECT=59
 }
 
+timeout=1m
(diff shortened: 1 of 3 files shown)


ldms/src/sampler/netlink/input/test_slps.sh
```diff
@@ -30,6 +30,9 @@ set_kokkos () {
 	export KOKKOS_LDMS_RECONNECT=59
 }
 
+timeout=1m
+export ZAP_MAX_IO=8
```
tom95858, line 34: @baallan, there is no ZAP_MAX_IO environment variable.

baallan: Did you mean to say "use ZAP_IO_MAX instead", which appears to be influential but undocumented?

tom95858:
@nichamon, did you get a chance to review this?

nichamon (approved):
I checked that the line that sets the returned code had been removed.  I saw the "no-kokkos blocking ok" messages in the build-test-ubuntu workflow. I didn't confirm that the "no-kokkos blocking test timed out" messages will be reported if the test_slps script gets timeout.

tom95858:
@baallan can you resolve the conflicts in this pull request?

baallan:
I rebased to top of tree; checking into what's causing the distcheck fails.

baallan:
test_slps is failing in the transport layer in something rails related. 
Nothing new in this patch is at fault, but if it is merged, then installcheck will be failing, possibly due to test code which has not caught up with recent rails api changes.
I've ported this patch to the 441rc2 branch as a separate pr #1300

PR #1153, opened 2023-03-28
Fix race between delete thread and xprt set delete

tom95858:
The logic in the set delete path notified subscribers of set deletion prior to adding the set to the delete tree. This could cause the set delete callback path to remove a set from the tree that had not yet been added; corrupting the tree.


baallan:
@tom95858 I will test to see if this improves either #1141 or #771.

Also, the two log messages marked critical are just errors; they don't require immediate human attention if they occur.

PR #1161, opened 2023-04-04
Refactor ldms and Zap libraries to use ovis_log

nichamon:


baallan:
This is an API breaking change for all existing stream publishing clients, is it not? I see everything in our tree is being fixed, but this change takes us to a 5.x release unless there's a compatibility mode retained.

We just got a stream-dependent plugin included in a major 3rd party c++ package (kokkos), so we're going to need a compatibility API to stay in 4.x mode and not look volatile to the 3rd parties.

ldms/src/core/ldms_xprt.h
```diff
@@ -391,8 +391,7 @@ struct ldms_xprt {
 
 	int term;
 
-	/** Transport message logging callback */
-	ldms_log_fn_t log;
+//	ldms_log_fn_t log;
```
tom95858, line 394: Please remove, let's not leave commented-out code in the tree.

nichamon: Done! Thanks for catching this.

tom95858:
> This is an API breaking change for all existing stream publishing clients, is it not? I see everything in our tree is being fixed, but this change takes us to a 5.x release unless there's a compatibility mode retained.
>

The minor version speaks to binary compatibility, not API. These changes do not affect binary compatibility.
 
> We just got a stream-dependent plugin included in a major 3rd party c++ package (kokkos), so we're going to need a compatibility API to stay in 4.x mode and not look volatile to the 3rd parties.

We can discuss the merits of leaving the API the same and simply ignoring the parameter, but I think that leads to developer confusion. One out of tree plugin that we maintain doesn't seem like much of a lift.

PR #1163, opened 2023-04-06
[WIP] Fix the delete_proc

nichamon:
The patch is to fix the bahvior that sometimes a set of sets are permanently stuck in the delete tree on the server side. It seems the server has never received the SET_DELETE reply from the client while the connection is still connected.

TODO:
  The delete_proc() skips the sets that there is a transport in the
lookup collection that is connected based on the assumption that 1) the server had sent the SET_DELETE request to the peers, and 2) the set will be freed when the server receives the SET_DELETE reply from its peers.
  We want to check whether there is a SET_DELETE context in the context
list of each transport in the lookup collection or not. If none of the transport has the SET_DELETE context, we will free the set. We believe this might be a corner case, e.g., we don't properly handle an error, or the underlying transport stuck in a stale stage.
  The problem with iterating through the transport's context list is
that we should take the transport lock. It is a nested lock inside the set lock. This may lead to a deadlock, so I have not implement the logic yet.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -3791,9 +3791,14 @@ static void *delete_proc(void *arg)
 				break;
 
 			/*
-			 * Look for a connected push/lookup peer.
+			 * Look for a connected lookup peer.
 			 * If there is a connected peer, we skip the set
 			 * and wait for the SET_DELETE reply from the connected peers.
+			 *
+			 * We don't check the push collection because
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 @baallan This is a part of the patch to fix the dangling sets on the server in #1141.

@tom95858 I haven't added the logic to consult the transport's context list because I need to take the transport lock inside the set lock.

PR #1164, opened 2023-04-06
[WIP] Fix the delete_proc

nichamon:
The patch is to fix the symptom that sometimes a set of sets are permanently stuck in the delete tree on the server side. It seems the server has never received the SET_DELETE reply from the client while the connection is still connected.

  The delete_proc() skips the sets that there is a transport in the
lookup collection that is connected based on the assumption that 1) the server had sent the SET_DELETE request to the peers, and 2) the set will be freed when the server receives the SET_DELETE reply from its peers.

  The patch consults the transports in the lookup collection if they
have the SET_DELETE context for the set or not. If none of the transports have the SET_DELETE context corresponding to the set, we will free the set. The logic is to handle when the above assumptions are not true. We believe this might be a corner case, e.g., we don't properly handle an error, or the underlying transport stuck in a stale stage.

  A problem that may arise with the patch is the deadlock between the set
lock and the transport lock. We need to take the transport lock inside the set lock to iterate through the transport's context list.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -3791,9 +3791,14 @@ static void *delete_proc(void *arg)
 				break;
 
 			/*
-			 * Look for a connected push/lookup peer.
+			 * Look for a connected lookup peer.
 			 * If there is a connected peer, we skip the set
 			 * and wait for the SET_DELETE reply from the connected peers.
+			 *
+			 * We don't check the push collection because
+			 * the push collection is always a subset of
+			 * the lookup collection. The application must lookup
+			 * the set before registering for push updates.
 			 *
 			 * This prevents the race between the SET_DELETE timeout
 			 * and the SET_DELETE reply.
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 @baallan  The patch is an attempt to fix issue #1141. My concern with the patch is the nested locks. I took the transport lock inside the set lock. I need to take the transport lock to iterate through its context list.

@baallan Could you please try with your reproducer? Thanks!

baallan:
@nichamon I still get 2 nodes out of 20 with 109 deleting_count.
:(

tom95858:
@nichamon, I think the solution here is to remove the lookup_coll altogether, but it seems likely that a configuration error is triggering this behavior. I say that because if it were a race, it seems statistically improbable that we would always end up with exactly 109 sets stuck in delete_proc.

nichamon:
> @nichamon, I think the solution here is to remove the lookup_coll altogether, but it seems likely that a configuration error is triggering this behavior. I say that because if it were a race, it seems statistically improbable that we would always end up with exactly 109 sets stuck in delete_proc.

@tom95858 Ben and I will have a call session tomorrow at 11CT. I plan to look into the configuration files, the log files, and the gdb sessions.

baallan:
> @nichamon, I think the solution here is to remove the lookup_coll altogether, but it seems likely that a configuration error is triggering this behavior. I say that because if it were a race, it seems statistically improbable that we would always end up with exactly 109 sets stuck in delete_proc.

Why would a configuration error not create the problem on all nodes instead of a sprinkling?

baallan:
Working with this patch, Mon and I found deadlocks for publishing messages (env, filename, argv) from the pid sampler plugin. Turning off the publishing stopped the lockup, but not the lingering sets.

baallan:
On a chance, @nichamon suggested switching to sock from rdma (cts1x is omnipath). On the first test, no lingering sets seen. Repeating the test.
This change would be undesirable at scale,  but perhaps points to a problem somewhere.

tom95858:
@nichamon, @baallan where are we on this? The last I heard this is only an issue with Omnipath (software IB), has this been confirmed?

baallan:
> @nichamon, @baallan where are we on this? The last I heard this is only an issue with Omnipath (software IB), has this been confirmed?

@tom95858 i will be testing this on native ib hardware (mlx4 i think) on friday to confirm/deny (using system btaco). The patch attached to this didn't change the failure, so we have continued with testing from near top of tree rather than this patch.

nichamon:
> @baallan, @nichamon any status on this?

@tom95858 I'm closing the draft pull request. It didn't fix the problem Ben found when we tested it. I will create a patch that includes more debugging messages when sending/receiving a SET_DELETE request/reply for Ben to test on Infiniband. I should have the patch by the end of this week. I'll make sure to keep you posted.

baallan:
@nichamon I'm on travel all next week.

PR #1172, opened 2023-04-14
Logical types

tom95858:
This change adds logicalTypes support for unsigned values when mapping from LDMS to Avro/Kafka.


tom95858:
@morrone, could you please take a look at this, in particular to the logical type naming and the handling of u8.

PR #1173, opened 2023-04-17
Normalize the Avro logical types

tom95858:
Any LDMS type that doesn't have an Avro equivalent is encoded in an Avro type with a logicalType annotation that indicates the type from which the value came.


ldms/src/ldmsd/ldmsd_decomp.c
```diff
@@ -748,22 +748,22 @@ int ldmsd_row_to_json_object(ldmsd_row_t row, char **str, int *len)
 static const char *col_type_str(enum ldms_value_type type)
 {
 	static char *type_str[] = {
-	    [LDMS_V_CHAR] = "\"string\"",
-	    [LDMS_V_U8] = "\"string\"",
-	    [LDMS_V_S8] = "\"int\"",
+	    [LDMS_V_CHAR] = "{\"type\":\"int\",\"logicalType\":\"signed-char\"}",
```
tom95858, line 751: @morrone, should this be int8_t, or int8?

ldms/src/ldmsd/ldmsd_decomp.c
```diff
@@ -748,22 +748,22 @@ int ldmsd_row_to_json_object(ldmsd_row_t row, char **str, int *len)
...
+	    [LDMS_V_U8] = "{\"type\":\"int\",\"logicalType\":\"unsigned-char\"}",
```
tom95858, line 752: Should this be 'uint8_t' or 'uint8'?

tom95858 (commented):
Are all of the other logicalTypes acceptable?

ldms/src/ldmsd/ldmsd_decomp.c
morrone: I don't know. How about int8.

ldms/src/ldmsd/ldmsd_decomp.c
morrone: How about uint8.

morrone:
> Are all of the other logicalTypes acceptable?

I think that LDMS_V_CHAR should probably go back to having a base avro type of "string". We could give it some kind of logicalType name that makes it clear it will always be a single character. "character", "single-character", something like that.

morrone:
I think maybe LDMS_V_U32 should have a base type of "long" so we never have issues of confusing a large value for a negative number.

ldms/src/ldmsd/ldmsd_decomp.c
tom95858: done

tom95858:
> I think that LDMS_V_CHAR should probably go back to having a base avro type of "string". We could give it some kind of logicalType name that makes it clear it will always be a single character. "character", "single-character", something like that.

It seems to me that this should not be a string. In all of the use cases I'm familiar with it is typically treated as an int, not a unity length array of char. Do you have a sample client use case that makes this easier to deal with as a string. Specifically, the avro accessor is going to be something like avro_array_get_xxx which seems odd for a single character.

morrone:
> > I think that LDMS_V_CHAR should probably go back to having a base avro type of "string". We could give it some kind of logicalType name that makes it clear it will always be a single character. "character", "single-character", something like that.
> 
> It seems to me that this should not be a string. In all of the use cases I'm familiar with it is typically treated as an int, not a unity length array of char. Do you have a sample client use case that makes this easier to deal with as a string. Specifically, the avro accessor is going to be something like avro_array_get_xxx which seems odd for a single character.

Setting aside the unfortunate situation with pointer arithmetic in C, no on should have been storing integers in chars since C99. I know old habits are hard to break, but it has been over twenty years. :) We do not need to consider the case when a user is interpreting a char as an integer. If any LDMS sampler does that, it is a bug that needs to be fixed. With that out of the way:

No avro deserializer out of the box, regardless of which language it is written in, will know to decode a 32-bit integer into a character.

Lets take the example of a sampler storing the letter "P" in an LDMS_V_CHAR. If store_avro_kafka chooses to store that in an avro int, then when any client (Python, Java, even C) automatically decodes the message what it will output for a human to read is "80". "80" is not "P". 80 is a number, P is a character. If, on the other hand store_avro_kafka chooses to store that in an avro string, again, any client can automatically output "P". That is what we want. We want the character to remain a character.

I would argue that getting "80" when you expect "P" is far more of a problem then encoding "P" in a string.

tom95858:
@morrone, please take one last look at this. I think it does what you suggest. If you are satisfied, I'll merge.

morrone:
> @morrone, please take one last look at this. I think it does what you suggest. If you are satisfied, I'll merge.

Yes, that looks good to me. Thanks!

PR #1176, opened 2023-04-18
Update updtr_match_list in ldmsd_controller and ldmsctl

nichamon:


ldms/man/ldmsd_controller.man
```diff
@@ -530,6 +530,15 @@ The updater name. If none is given, the statuses of all updaters are
...
 .RE
 
+.SS Query the updater match list
+.BR updtr_match_list
+attr=<value>
+.RS
+.TP
+.BI [name " name"]
+.br
+The Updater name. If none is given, the match list of all updaters are reported.
```
tom95858, line 540: I think this is supposed to say "the match list of all sets ..." correct. We're not matching updaters here, were matching sets that an updater will be issuing ldms_xprt_update requests for?

nick-enoent: > I think this is supposed to say "the match list of all sets ..." correct. We're not matching updaters here, were matching sets that an updater will be issuing ldms_xprt_update requests for?

That was my understanding as well. When I wrote this command, it was meant to return updater sets that matched the given regex, not just the updaters that contain them.

ldms/man/ldmsctl.man
```diff
@@ -469,6 +469,16 @@ The updater name. If none is given, the statuses of all updaters are
...
 .RE
 
+.SS Query the updater match list
+.BR updtr_match_list
+attr=<value>
+.RS
+.TP
+.BI [name " name"]
+.br
+The updater name. If none is given, the match list of all updaters are reported.
```
tom95858, line 479: Should we update this to say the "list of sets matching the updtr match regular expression" is reported. Is this correct?

tom95858, line 479: Also, maybe provide an example request with output in the man page?

ldms/man/ldmsd_controller.man
nichamon: > I think this is supposed to say "the match list of all sets ..." correct. We're not matching updaters here, were matching sets that an updater will be issuing ldms_xprt_update requests for?

Yes, the match list is the list of regular expressions to match the sets the updater will be issuing ldms_xprt_update requests for.

ldms/man/ldmsctl.man
nichamon: No. I'm reporting the regular expressions. However, it is more precise to say that the command returns the list of regular expressions to match set names or set schemas.

nichamon: Sure, I'll add a couple of examples.

PR #1180, opened 2023-04-20
Update LDMS_V*8_ARRAY logicalTypes to match LDMS_V_*8

morrone:


tom95858:
@morrone, thanks for this...

Issue #1183, opened 2023-04-21
ldmsd_controller crash in update_time_stats


baallan:
ldmsd_controller crashes on an internal json parsing error in update_time_stats. example below.
The error is from an empty list element (server swa1106 is not up, so the iteration over the producer list generates nothing but an extra comma and hence the json parse error later.) see below between the end of 1105 and the start of 1107.
```
{
  "all_swaadmin6": {
    "swa1105": {
      "swa1105/dstat": {
        "min": 166.656,
        "max": 271489.792,
        "avg": 1120.068267,
        "cnt": 1380
      },
      "swa1105/vmstat": {
        "min": 40547.328,
        "max": 275121.664,
        "avg": 45359.54662,
        "cnt": 1380
      }
    },,
    "swa1107": {
      "swa1107/dstat": {
        "min": 164.864,
        "max": 270308.352,
        "avg": 1120.709936,
        "cnt": 1380
      },
      "swa1107/vmstat": {
        "min": 39994.368,
        "max": 275653.632,
        "avg": 44401.257183,
        "cnt": 1380
      }
    }
  }
}
```

baallan:
Need to either not generate the extra comma or to generate 
```
"swa1106": {},
```

PR #1184, opened 2023-04-25
Account for unresponsive producers in update_time_stats json formatting

nick-enoent:
Resolve issue #1183 where multiple commas will be inserted into the json response object if a producer is disconnected


nick-enoent:
@nichamon @narategithub Would you guys mind taking a quick look at this? I can't seem to assign reviewers to this PR.

nichamon (approved):

narategithub (approved):

nichamon:
@nick-enoent Could you amend the commit message with "Fixes #1183"? Github will close the issue once the pull request is merged. Thanks!

baallan:
@nichamon @narategithub Should we include an empty object foo:{} rather than omit foo altogether? That would communicate that an expected server is not responding/connected rather than it simply isn't expected.

PR #1186, opened 2023-04-26
Add simple blob store to csv transformation script/manual

baallan:
This script handles simple cases only. When a fully flexible stream decomposition store policy is available, we can probably dispense with it.


baallan:
@tom95858 I'm under the impression a sos store for streams with something like the storage policy decomposition from metric sets to columns (mapping json schema elements to columns) is in the works. Is that incorrect?

baallan:
The pushed update adds the --col-sep option that Schwaller is desirable because the spank plugin produces string values with embedded commas.

PR #1191, opened 2023-05-02
Fix the strgp_add handler in ldmsd_communicator.py

nichamon:
The fix makes the handler to handle the decomposition attribute.


nichamon:
@nick-enoent Please review the patch. I need to figure out how I add you as a reviewer. 🤔

PR #1192, opened 2023-05-04
Make set_sec_mod accept value zero of UID or GID

nichamon:


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5475,7 +5475,7 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 			/* Valid username */
 			uid = pwd->pw_uid;
 			set_flags |= DEFAULT_AUTHZ_SET_UID;
-		} else if ((*endptr == '\0') && (_uid > 0)) {
+		} else if (*endptr == '\0') {
```
tom95858, line 5478: @nichamon it seems like checking for a terminating '\0' is not the right way to go. If there were trailing space in the value, this doesn't work. I think the way to do this is to check if the 1st character of value is a digit (isdigit), in which case strtol is used, otherwise getpwnam is used. In this way we can reliably validate both numbers and names. If !pwd the error is "unknown user", if uid < 0, the error is "invalid user id"

Issue #447, opened 2020-10-02
ldmsd_set_[register|deregister] need code documentation


morrone:
Can someone please share the semantics surrounding ldmsd_set_register and ldmsd_set_deregister? The prototypes lack any code documentation.

What do the functions actually do?

At what points in set creation and destruction should a plugin be calling those functions?

And maybe we should think of hiding these functions from the plugins entirely. They seem like a strange break in the metric set API by coming from a completely different header than almost everything else in metric set API that is use by sampler plugins. At a first read through the interface, it looks like we would be better off hiding these calls under ldms_set_new() and ldms_set_delete(). It looks like it might only require adding the plugin name as an option to ldms_set_new() and storing the name in the set structure to make that possible.

tom95858:
The ldmsd_set_register function is not part of LDMS, but rather LDMSD, i.e. the application. That's why it's in a different header and has a different name.

At some point there was a desire to know what plugins provided which sets. The **ldmsd** 'plugn_sets' configuration command reports what sets each plugin provides. The ldms**d**_set_register() function informs the **ldmsd** daemon that it provides a particular set name. IOW, this functionality is implemented in the daemon, not in the underlying LDMS library.

plugins, etc... are not part of the LDMS API, they are part of the ldmsd daemon application.

tom95858:
@morrone, to your point, I'm not disinclined to remove that function and roll it into some other API, for example, **ldmsd_set_new**, I'm just explaining how it got there.

morrone:
I don't remember these being properly documented.

tom95858:
@morrone please reopen if necessary.

PR #1197, opened 2023-05-08
Add documentation for ldmsd set registration API

tom95858:
Fixes issue #447


tom95858:
@morrone, does this fix #447 ?

tom95858:
@morrone, does this fix #447?

morrone:
Yes, that works for me!

PR #1199, opened 2023-05-08
Command to dump the current running configuration

nick-enoent:
Command dump_cfg added to ldmsd_controller and ldmsd_communicator. Writes the current running configuration to a file named after the daemon's host and communication port, <hostname>-<port>.conf, in path specified in argument. If no path is specified it writes the file to the user's home path.

Parameters:
path - [optional] The path to the configuration file


baallan:
I suggest if no path is specified that the output go to stdout, enabling echo "dump_cfg | ldms...".

PR #1201, opened 2023-05-09
Fix ovis_log_list in the ovis_log library

nichamon:
The ovis_log_list() function reused a va_arg without copying it first. Resulting in, a bogus string in the returned string containing the list of all loggers.

### DIFF
--- a/lib/src/ovis_log/ovis_log.c
+++ b/lib/src/ovis_log/ovis_log.c
@@ -448,17 +448,23 @@ struct ovis_log_buf {
 
 static int __buf_append(struct ovis_log_buf *buf, const char *fmt, ...)
 {
-	va_list ap;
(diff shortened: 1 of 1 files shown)


lib/src/ovis_log/ovis_log.c
```diff
@@ -527,7 +533,7 @@ char *ovis_log_list(const char *subsys)
 			goto err;
 		}
 	} else {
-		rc = __buf_append(buf, "{\"name\":\"%s\","
+		rc = __buf_append(buf, "{\"name\":\"%s (default)\","
```
tom95858, line 536: @nichamon what is the (default) for?

nichamon: It is to specify that this is the default log.

PR #1207, opened 2023-05-12
Refactor all LDMSD plugins to use the ovis_log lib and remove the logic that passes log function handle around

nichamon:


nichamon:
@tom95858 The draft changes the `get_plugin` signature so that it won't receive a log function handle. The patches also refactored all LDMSD plugins in the tree to use the new `get_plugin` signature and the logging infrastructure from the ovis_log library.

The draft depends on PR #1202.

nichamon:
The following plugins have been build-tested according to config.log.
* app_sampler
* array_example
* blob_stream
* clock
* coretemp
* store_csv
* dstat
* store_flatfile
* generic_sampler
* hello_stream
* ibnet
* jobid
* jobinfo
* loadavg
* lustre
* meminfo
* procdiskstats
* procinterrupts
* procnetdev
* procnet
* procnfs
* procstat
* tsampler
* vmstat
* store_sos
* rabbit storage plugins

None of the cray-related samplers, lustre samplers, and the plugins in ldms/src/contric are tested.

nichamon:
@tom95858 I have squashed the commits. Each commit is buildable.

PR #1209, opened 2023-05-19
Make ldmsd_controller exit on connection error

nichamon:
The patch also updates the connection error message to include the transport and the authentication method.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_communicator.py
+++ b/ldms/python/ldmsd/ldmsd_communicator.py
@@ -933,7 +933,15 @@ class Communicator(object):
         try:
             self.ldms.connect(self.host, self.port, timeout=timeout)
         except Exception as e:
-            print(f'Error {e}: connecting to {self.host} on port {self.port}')
+            if self.auth is not None:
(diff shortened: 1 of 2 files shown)


nichamon:
@nick-enoent Please review.

nichamon:
`ldmsd_controller` takes a second to exit after it prints the connection error message. I don't know why. @tom95858 @nick-enoent @narategithub Do you have any ideas?

tom95858:
> `ldmsd_controller` takes a second to exit after it prints the connection error message. I don't know why. @tom95858 @nick-enoent @narategithub Do you have any ideas?

My guess is that there is a pthread_join on the I/O threads and it doesn't exit until the join completes. Most of those threads are blocked in epoll_wait I think.

baallan:
Please note that ldmsd_controller should exit on a connection error *only* for connections being provoked by a command line argument. Once i'm interactive, i should be able to try connecting to things i know are not there (or have different connection parameters) without being forced out of the command line.

It's not entirely clear to me from the patch that the 'exit on connection error' is only to happen on a failed argv processing.

PR #1210, opened 2023-05-22
Fix a typo in the code path that loads a decomp lib

nichamon:


ldms/src/ldmsd/ldmsd_decomp.c
```diff
@@ -156,7 +156,7 @@ static ldmsd_decomp_t __decomp_get(const char *decomp, ldmsd_req_ctxt_t reqc)
 		d = dlopen(library_name, RTLD_NOW);
 		if (d)
 			break;
-		if (!stat(library_name, &st))
+		if (stat(library_name, &st))
```
narategithub, line 159: woops. sorry for the bug & thanks for catching it.

tom95858: It seems like this logic could be more helpful to the user. The failing stat() doesn't mean the library does not exist, it could be a permission error; regardless printing the error would be helpful to the user.

If stat() succeeds, we should probably continue after printing the dlerror() and try the next pathdir element, not give up and return NULL.

ldms/src/ldmsd/ldmsd_decomp.c
```diff
@@ -156,7 +156,7 @@ static ldmsd_decomp_t __decomp_get(const char *decomp, ldmsd_req_ctxt_t reqc)
...
 			continue;
 		dlerr = dlerror();
 		errno = ELIBBAD;
```
tom95858, line 162: If we run through all the pathdir and can't find the decomposer, it doesn't mean the decomposer is bad, it more likely means the specified decomposer does not exist; again the stat() tells you which.

tom95858: This comment is supposed to be at line 169...

tom95858: At line 179, the error "get() not found" is not useful to the user unless they develop decomposers. It would be more useful to say "The library %s is not a valid decomposer plugin."

tom95858:
@nichamon?

nichamon:
> @nichamon?

Hi Tom, I just pushed the updated code that addresses your comments.

PR #1212, opened 2023-06-05
Make strgp_status report the regex and decomposition

nichamon:


nick-enoent:
Looks good to me @nichamon

PR #1234, opened 2023-07-18
Fix 'credits' and 'rail' optional attributes in ldmsd_controller prdcr_add

narategithub:
'credits' and 'rail' are optional attributes in prdcr_add command. They were accidentally listed under the required attribute list.


narategithub:
@tom95858 the rebase is done. I talked with @nichamon, her PR #1233 moved 'reconnect' and 'interval' under the optional list because the user can specify either 'reconnect' or 'interval', until the 'interval' attribute is completely removed from the command. She also put logic in `ldmsd_controller` to sanity check these options (i.e. either 'reconnect' or 'interval' is specified, but not both or none).

PR #1237, opened 2023-07-21
Add formatted print attribute to `ovis_log()`

narategithub:
This is to help compiler warns about print format error.


tom95858 (commented):
@narategithub this is a great catch. This should be cherry picked into 4.4.1 I think. Some of these errors will cause crashes if the message were ever logged.

PR #1238, opened 2023-07-21
Fix accepting endpoint race in zap_sock

narategithub:
zap_sock uses the thread of the listening to initially process new passive endpoints (from connection requests). It correctly removed the new endpoints from the listening thread epoll as the threads assigned to the new endpoints will further process the events. Hoever, zap_sock forgot to stop the processing loop in the listening thread, resulting in wrong endpoint processing. This patch fixes the bug.


baallan:
@narategithub does this apply to the rails work or generally to earlier code also?

narategithub:
@baallan Just the rails. The earlier code does not have this issue.

PR #1239, opened 2023-07-21
zap_sock: fix print format warnings and add tid to debug log

narategithub:


narategithub:
This is now a part of #1237 .

PR #1240, opened 2023-07-21
Add 'reconnect' to prdcr_start* to deprecate 'interval'

nichamon:


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -1825,16 +1831,22 @@ static int prdcr_start_regex_handler(ldmsd_req_ctxt_t reqc)
 		goto send_reply;
 	}
 
-	interval_str = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_INTERVAL);
-
+	interval_str = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_RECONNECT);
+	if (!interval_str) {
+		interval_str = ldmsd_req_attr_str_value_get_by_id(reqc,
+							LDMSD_ATTR_INTERVAL);
+		ovis_log(config_log, OVIS_LWARN, "The 'interval' in "
```
tom95858, line 1838: I don't think we need these messages in the server.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -1729,11 +1729,17 @@ static int prdcr_start_handler(ldmsd_req_ctxt_t reqc)
 
 	ldmsd_req_ctxt_sec_get(reqc, &sctxt);
 	interval_str = ldmsd_req_attr_str_value_get_by_id(reqc,
-							LDMSD_ATTR_INTERVAL);
+							LDMSD_ATTR_RECONNECT);
```
tom95858, line 1732: Why do we need a new attribute?

tom95858 (changes requested):

ldms/man/ldmsd_controller.man
```diff
@@ -334,10 +339,15 @@ attr=<value>
 .br
 A regular expression
 .TP
-.BI [interval " interval"]
+.BI [reconnect " interval"]
```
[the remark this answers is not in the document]
nichamon: I'll submit the change to support strings in another patch.

ldms/src/ldmsd/ldmsd_request.c
nichamon: The log message in the server is for the case that the 'interval' attribute is used in configuration files.

ldms/src/ldmsd/ldmsd_request.c
nichamon: The only reason I added `LDMSD_ATTR_RECONNECT` is to differentiate between the uses of 'interval' and 'reconnect' in configuration files, so LDMSD could give a warning message that the 'interval' attribute is being deprecated.

tom95858: Hi @nichamon, I understand, but my suggestion is that the server shouldn't do this, ldmsd_controller, ldmsdctl and the config parsing should.

ldms/man/ldmsd_controller.man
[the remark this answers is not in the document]
tom95858: Please put it in this patch. We are logically introducing a 'new' attribute or at least an alias with expanded syntax; this would be the place to get the documentation and the implementation all sync'd up. This patch is already pretty concise.

ldms/src/ldmsd/ldmsd_request.c
nichamon: I pushed a patch that gets rid of the new attribute `LDMSD_ATTR_RECONNECT` and let the parsers for ldmsd_controller, ldmsctl, and config files give the warning about 'interval' being deprecated.

tom95858:
@nichamon, is this pull request obsolete?

nichamon:
> @nichamon, is this pull request obsolete?

No, it is not. I just rebased it on top of OVIS-4, and it is ready to merge.

PR #1242, opened 2023-07-24
Make LDMSD support interval strings

nichamon:
The patch makes LDMSD support time-interval strings. A time-interval string is an integer followed by a unit string. A unit string is one of the followings:
  us -- microseconds
  ms -- milliseconds
  s  -- seconds
  min -- minutes
  hr  -- hours
  day -- days

For example, with this patch, to specify a sampling interval of 1 second, '1s' can be used.

### DIFF
--- a/ldms/man/ldmsd_controller.man
+++ b/ldms/man/ldmsd_controller.man
@@ -166,15 +166,22 @@ The plugin name.
 .TP
 .BI interval " interval"
 .br
-The sample interval in microseconds.
+The sample interval, which is a float followed by a unit string.
+If no unit string is given, the default unit is microseconds.
+A unit string is one of the followings:
+  us -- microseconds
+  ms -- milliseconds
+  s  -- seconds
+  m  -- minutes
+  h  -- hours
+  d  -- days
 .TP
 .BI [offset " offset"]
 .br
-Offset (shift) from the sample mark in microseconds.
+Offset (shift) from the sample mark in the same format as intervals.
 Offset can be positive or negative with magnitude up to 1/2
-the sample interval. If this offset is specified, including 0,
-collection will be synchronous; if the offset is not specified,
-collection will be asynchronous. Optional.
+the sample interval. The default offset is 0. Collection is always synchronous.
+
 .RE
 
 
@@ -269,7 +276,15 @@ The connection type [active, passive]
 .TP
 .BI reconnect " interval"
 .br
-The connection retry interval
+The connection retry interval, which is a float followed by a unit string.
+If no unit string is given, the default unit is microseconds.
+A unit string is one of the followings:
+  us -- microseconds
+  ms -- milliseconds
+  s  -- seconds
+  m  -- minutes
+  h  -- hours
+  d  -- days
 .TP
 .BI interval " interval"
 .br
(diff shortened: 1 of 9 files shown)


nichamon:
@nick-enoent Here is the patch that removes `check_offset()` from `ldmsd_communicator.py`. Please review the patch. Thanks!

ldms/man/ldmsd_controller.man
```diff
@@ -514,8 +553,16 @@ The update policy name
...
 .br
-The update interval in micro-seconds. If this is not
-specified, the previously configured value will be used. Optional.
+The update interval, which is an integer followed by a unit string.
+If no unit string is given, the default unit is microseconds.
+A unit string is one of the followings:
+  us -- microseconds
+  ms -- milliseconds
+  s  -- seconds
+  min -- minutes
```
tom95858, line 562: 'm'
'h'
'd'
seems more concise.

tom95858:
@nichamon, @nick-enoent can we make sure whatever these strings are they need to be match what maestro uses.

nichamon:
> @nichamon, @nick-enoent can we make sure whatever these strings are they need to be match what maestro uses.

@tom95858 After I shorten 'min' to 'm', the strings in this patch will be the same as in Maestro. However, there are 2 main differences: 
1) Maestro supports floats as an interval but my patch does not. You can give '1.5s' in Maestro but you need to give '1500ms' in ldmsd_controller. I'll update my patch to support floats.
2. My patch supports more time units, i.e., nanosecond (n), h (hour), and d (day).

I will update the patch during the weekend.

tom95858:
@nichamon The only one I think is unreasonable is nano-second.

ldms/man/ldmsd_controller.man
nick-enoent: @nichamon Maestro currently supports "m" for minutes as well. It doesn't currently support hours or days, but I can add a quick patch to address this once we land on a final format for minutes/hours/days. I'm in favor of shortening it like @tom95858 suggested, as it lends itself less to typos.

nick-enoent (commented):
@nichamon This patch works just fine with maestro currently, aside from the "m" vs "min" format for minutes.

ldms/man/ldmsd_controller.man
nichamon: @tom95858 @nick-enoent I changed the minute, hour, and day strings to 'm', 'h', and 'd', respectively. Also, the patch supports floats, e.g., '1.5s' as in Maestro.

nichamon:
@tom95858 Per your request in https://github.com/ovis-hpc/ovis/pull/1240#discussion_r1271371262, I'll add this patch on top of #1240 and close this pull request when we think this patch is good .

tom95858:
@nichamon,  @nick-enoent is this patch ready to go?

ldms/python/ldmsd/ldmsd_communicator.py
```diff
@@ -204,17 +204,6 @@
                       'auth_add': {'req_attr': ['name', 'plugin'], 'opt_attr': []},
                       }
 
-def check_offset(interval_us, offset_us=None):
```
tom95858, line 207: Why did we decide that this check was no longer necessary?

nichamon: - The main reason is that LDMSD also checks the offset value.
- With the patch, LDMSD converts interval strings to microseconds, so we don't have to duplicate the code in ldmsctl, ldmsd_controller, and the config file parser.
- If we want to keep the check_offset(), I need to convert interval strings to microseconds before passing the values to check_offset(). I will probably move the conversion and the check logic out from LDMSD and put them in ldmsctl and config file parser, similar to ldmsd_controller.

tom95858:
@nichamon, ready to merge?

tom95858 (approved):

nichamon:
> @nichamon, ready to merge?

Yes, it is ready.

PR #1250, opened 2023-08-08
Make ldmsd_plugattr report log messages to the application's default log

nichamon:
The patch is to fix build problems of stand-alone application in the project that uses the library.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_plugattr.c
+++ b/ldms/src/ldmsd/ldmsd_plugattr.c
@@ -54,8 +54,10 @@
 #include <stdarg.h>
 #include "ovis_util/dstring.h"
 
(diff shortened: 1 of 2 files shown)


nichamon:
@morrone Could you give this a try?

morrone:
Yes, it does compile now. Thanks!

nichamon:
> Yes, it does compile now. Thanks!

@morrone  Thanks for trying it out!

baallan:
I will look into this. 

The plugattr library is utilities for processing configuration values (all during the config method). As such, the messages should really be controlled by the debug level at which config-statement processing is being handled, not by the sampler-specific logging level which is appropriate during sample() and the other methods.

I have a patch already for the standalone program references to logging.

PR #1251, opened 2023-08-09
Fix print formats in ibmad_sampler & ibmad_record_sampler plugins

nichamon:


nichamon:
The flag `-Werror -Wall` caught this after I applied PR #1250 .

tom95858:
@nichamon, these changes need to be made in top of tree too, correct? Or did Narate already fix those?

tom95858:
@nichamon, I'll make the change in OVIS-4.4.1 then.

nichamon:
> @nichamon, I'll make the change in OVIS-4.4.1 then.

I just created a pull request #1253 for 4.4.1-rc.

nichamon:
> @nichamon, I have already done this in a staging branch for 4.4.1-RC2. There can be no pull requests on 4.4.1-RC because it contains a couple of changes that resulted in extensive build breakage with -Wall -Werror. 4.4.1-RC2 is a rebase on 4.3.11 minus these changes plus some other fixes. I will push this shortly.

Got it! I'll run the ldms-test suite against 4.4.1-rc2 after you push the changes.

morrone:
Seems fine. But if we are comfortable with C99 stuff, maybe we should include <inttypes.h> and use the PRIx64 macro for this?

nichamon:
A similar fix in #1256 was merged.

PR #1253, opened 2023-08-09
Fix print formats in ibmad_sampler and ibmad_record sampler plugins

nichamon:


nichamon:
@tom95858 Here is the patch to go to V4.4.1-rc.

By the way, should we port Narate's print format fixes to v4.4.1-rc? I don't see them in the branch. I'll compare cherry-picking Narate's commits and manually fixing the formats, I'll do the one that is the easiest.

nichamon:
I am closing this because Tom already made the change in a staging branch for the release.

PR #1254, opened 2023-08-09
Fix print race in ldmsd_stream_subscribe

narategithub:
The msglog printing in ldmsd_stream_subscribe does not lock the print mutex for the entire EVENT printing, leading to interleaving EVENT records in the output. This patch fixed the bug.

### DIFF
--- a/ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
+++ b/ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
@@ -27,18 +27,20 @@ static int daemon_io;
 static int daemon_noroot;
 static int events_raw;
 
+static pthread_mutex_t msglog_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+#define MSGLOG_LOCK() pthread_mutex_lock(&msglog_mutex)
+#define MSGLOG_UNLOCK() pthread_mutex_unlock(&msglog_mutex)
+
 void msglog(const char *fmt, ...)
 {
 	if (quiet)
 		return;
 	va_list ap;
-	static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 
-	pthread_mutex_lock(&mutex);
 	va_start(ap, fmt);
 	vfprintf(file, fmt, ap);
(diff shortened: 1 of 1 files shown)


tom95858:
@narategithub I don't understand this fix. The static variable should be identical to declaring it global. There should only be a single instance of it. There mutex is taken inside the function instead of outside. What am I missing?

morrone:
I think the purpose of moving the lock to global file scope is so that the lock can be held across multiple calls to msglog(). If the lock is static function scope, the lock can only be held within a single msglog() call.

narategithub:
@tom95858 It's what @morrone said. The problem was the multiple calls of `msglog()` here

https://github.com/ovis-hpc/ovis/pull/1254/files#diff-6b46340af13fc098b2850942cefccb52c53db44579e025f2c230f492fd00af88L168-R180

However, I think I'll close this PR in favor of a simpler patch #1255 .

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
```diff
@@ -175,6 +178,7 @@ static int stream_recv_ev(ldms_stream_event_t ev, void *arg)
 	if (!events_raw)
 		msglog("}");
 	msglog("\n");
+	MSGLOG_UNLOCK();
```
tom95858, line 181: I don't believe it has anything to do with the "scope of the lock" it has to do with the need to hold the lock across multiple calls to msglog(). This is probably what @morrone meant; so it's the description of the problem and the solution that was giving me pause. I was interpreting "scope" to mean "symbol scope."

ldms/src/ldmsd/test/ldmsd_stream_subscribe.c
```diff
@@ -27,18 +27,20 @@ static int daemon_io;
 static int daemon_noroot;
```
tom95858, line 27: @narategithub why did you close this?

narategithub: #1255 is a simpler fix to this problem.

PR #1255, opened 2023-08-10
Fix ldmsd_stream_subscribe intermingled events output

narategithub:
Multiple threads in `ldmsd_stream_subscribe` can race to handle stream recv events. The recv handler called `msglog()` multiple times to print out the EVENT. This lead to intermingled events in the output. This patch modifies the handler to print the events in single msglog call to avoid the intermingled output.


narategithub:
@tom95858 why do you close this one?

tom95858:
We had two versions of this...yes? Or is it a different pull request. I thought you had a different fix.

narategithub:
We have 2 version of this fix. This version is a lot easier to read than the other one.

PR #1258, opened 2023-08-11
Enable list_sampler & record_sampler when --enable-ldms-test is given

nichamon:


narategithub (commented):
Looks good to me.

nichamon:
I made this patch so we don't need to enable the list_sampler and record_sampler plugins when testing explicitly. The ldms-test suite relies on both of them.

PR #1264, opened 2023-08-16
more samplers to subdirs and per-dir man pages

baallan:
This makes subdirs out of more top-level samplers and moves sampler-specific man pages into the directories with their source.

git mv provokes a rescan for space line-endings, so a bunch of whitespace cleanup was required.

Missing man page for hweventpapi is now also stubbed.


baallan:
@valleydlr perhaps you or Tom can merge this #1264?

PR #1266, opened 2023-08-16
Fix a race between ovis_log_destroy() and ovis_vlog()

nichamon:
The log handle may be destroyed while ovis_vlog() is accessing it.


tom95858:
@nichamon which of these two pull requests is the one you would like to merge?

nichamon:
> @nichamon which of these two pull requests is the one you would like to merge?

I want both. The difference is that this one goes to OVIS-4.4.1-RC2, and #1265 goes to OVIS-4.

PR #1269, opened 2023-08-22
Memory leak

morrone:
We are seeing huge memory leakage when store_avro_kafka is used, but kafka is down. During that time we see these errors:

```
ERROR: store_avro_kafka: {"name":"meminfo","type":"record","fields":[{"name":"timestamp","type":{"type":"long","logicalType":"time>
ERROR: store_avro_kafka: Error 'REST request failed (code -1): HTTP request failed: Couldn't connect to server' creating schema 'm>
ERROR: store_avro_kafka: A serdes schema for 'meminfo' could not be constructed.
```

The attached memory leak fixes compile, but haven't actually been tested. They need close review.


tom95858:
@morrone, can you check out #1274?

PR #1270, opened 2023-08-24
Including the ldms_xprt.h library

Snell1224:
The ``ldms_xprt.h`` library was removed in ``ldms_stream.h`` in a previous commit and caused compile failures with the darshanConnector code. 

I would prefer to add this back in but if there are issues with this, I can always explicitly include it in the darshanConnector code.

### DIFF
--- a/ldms/src/core/ldms_stream.h
+++ b/ldms/src/core/ldms_stream.h
@@ -58,6 +58,7 @@
 #include <sys/queue.h>
 #include <pthread.h>
 
+#include "ldms_xprt.h"
 #include "ovis_ref/ref.h"
 #include "coll/rbt.h"
 #include "ldms.h"


tom95858:
@narategithub can you please take a look at this?

narategithub:
@tom95858 I think this is not required. The application only include "ldms.h" and got everything needed to use LDMS Rail + Stream. 'ldms_stream.h' is private to LDMS internal.

Snell1224:
@narategithub @tom95858 I am checking if the ldms_t struct is connected to the daemon with:
 ``if (<struct_name>->disconnected)``

However, I receive the error:

 ``invalid use of undefined type ‘struct ldms_xprt’`` 

if ``ldms_xprt.h`` is not included. 

I'm sure most developers won't be trying to access/use the variables in the ldms_t struct and that this is just an unusual case so I'll go ahead and include ``ldms_xprt.h`` in the darshanConnector code and close this PR.

I figured I'd bring it up in case someone else has the same/similar issue later on when implementing an LDMS streams connector into their application.

PR #1271, opened 2023-08-25
Improve dump_cfg

nichamon:
- dump_cfg exports the command-line options, e.g., the log path and verbosity
- Make dump_cfg support plugins that allow multiple config command lines
- Make dump_cfg export the regex instead of schema at strgp_add if regex was originally given.


nichamon:
@nick-enoent Please review.

PR #1272, opened 2023-08-25
Improve dump_cfg

nichamon:
- dump_cfg exports the command-line options, e.g., the log path and verbosity
- Make dump_cfg support plugins that allow multiple config command lines
- Make dump_cfg export the regex instead of schema at strgp_add if regex was originally given.


nichamon:
The pull request is similar to #1271. The difference is that this one goes to 4.4.1-RC2, and #1271 goes to OVIS-4.

@nick-enoent Please review.

PR #1274, opened 2023-08-28
Fix error path memory leaks in avro_kafka

tom95858:
@morrone could you please test this out on your system to see if it resolves the memory leak issue?


morrone:
Will do.

morrone:
I _think_ this is OK. I think we may have a general aggregator (store_avro_kafka?) memory leak beyond the error paths. But I'm not sure where it is yet. I tried valgrind, but it is giving me crazy numbers, saying that "definitely" memory was lost that are just absurdly large, far larger than ldmsd every got (reported by ps). I'll keep digging.

PR #1276, opened 2023-08-30
Fix the set_sec_mod handler

nichamon:
Without the patch, the handler does not send the error code back to the interface when the given UID or GID is invalid.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -5539,6 +5539,7 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 		if (isdigit(value[0])) {
 			uid = strtol(value, &endptr, 0);
 			if (uid < 0) {
+				reqc->errcode = EINVAL;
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -5539,6 +5539,7 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 		if (isdigit(value[0])) {
```
tom95858, line 5539: @nichamon the value variable here is leaked throughout this function. Although it is freed at the end, because it is reused throughout, only the last value is actually freed. This fix is part of the "memory-leak" branch I was going to send you, but I think my change and yours will conflict. Maybe I just send you the branch and you combine them all into one pull request that you submit?

nichamon: I'll do that.

tom95858:
@nichamon is this patch subsumed by the other one? Can we just close it?

nichamon:
Yes, it is a part of #1278. I'm closing this.

PR #1279, opened 2023-09-05
[WIP] Fix gpumetrics sampler

nichamon:


tom95858:
Hi @nichamon, what is the status of this change.

tom95858:
@nichamon what is that status of this pull request?

nichamon:
> @nichamon what is that status of this pull request?

Intel had problems building it. I asked them to check their Makefile.am if it contains a typo or not. I'm waiting for their response.

However, the patch does not fix the GPU sampler's inconsistency state problem reported [here ](https://github.com/ovis-hpc/ovis/discussions/1273). I made a patch to build the plugin on Ubuntu and fix the missing -lovis_log in the Makefile.am.

We first suspected that the ovis_log patch caused the GPU set to get stuck in an inconsistent state. I don't have the GPU cards to create and sample the metrics properly. However, with the patch and some debugging code, I can load and run the plugin on my system. The debugging code was to bypass the check if the devices/cards were present on my machine or not.

According to the issue, they can load and run the plugin, so it is unlikely that the inconsistency was caused by the patch refactoring the plugin to use the ovis_log library.

nichamon:
I'm closing this because it doesn't fix the problem reported in https://github.com/ovis-hpc/ovis/discussions/1273.

PR #1281, opened 2023-09-07
Make json_stream_sampler handle UID, GID, and perm

nichamon:
Without the patch, the json_stream_sampler receives the UID, GID, and perm values at the config line, but it does not apply the values to the metric sets.

The patch makes the plugin use the specified UID, GID, and permission; otherwise, it will use the security data from the received stream data. The security data is applied at the set creation time. The configuration command set_sec_mod must be used to modify the sets' security. The plugin does not automatically revise the set security according to the stream data's security changes.

### DIFF
--- a/ldms/python/ldms.pxd
+++ b/ldms/python/ldms.pxd
@@ -298,6 +298,8 @@ cdef extern from "ldms.h" nogil:
 		           sockaddr *remote_sa,
 		           socklen_t *sa_len)
 
+    const char *ldms_metric_type_to_str(ldms_value_type t)
+
     # --- dir operation related --- #
     struct ldms_key_value_s:
         char *key
--- a/ldms/python/ldms.pyx
+++ b/ldms/python/ldms.pyx
@@ -3009,6 +3009,7 @@ cdef class Set(object):
 
     def json_obj(self):
         """Return dict/list values appropriate for json.dumps()"""
+        cdef const char *tmp
         ret = dict()
         meta_lst = [ 'name', 'schema_name', 'transaction_timestamp',
                      'transaction_duration', 'card', 'data_gn', 'data_sz',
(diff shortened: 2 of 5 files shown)


nichamon:
@tom95858 Do you think the plugin should revise the sets' UID, GID, or perm according to the stream data's security changes? The patch does not do that. It only makes sense when the UID, GID, or perm are from the received stream data at the set creation time. It could be confusing when the UID, GID, or perm values are specified at the config line.
We can make it happen in the way that the plugin will automatically modify the UID, GID, or perm only when its value is not given at the config line.

nichamon:
I'll mark this as ready when I have the test coverage done, and the patch passes the test cases.

nichamon:
@narategithub I use the UID, GID, and perm in the `ldms_stream_recv_data_s` structure inside the `ldms_stream_event_s` struct. What is the 'perm' in the `ldmsd_stream_recv_data_s`? Does it make sense to use it as the permission of the metrics sets encoded the stream data?

narategithub:
@nichamon I think it makes sense for most cases. The corner case I can think of right now is stream data sending via root's channel (uid0, gid0), but somehow we want to create a set owned by a user (e.g. `bob`) from it. In that case, the stream data must carry UID/GID/Perm.

nichamon:
> @nichamon I think it makes sense for most cases. The corner case I can think of right now is stream data sending via root's channel (uid0, gid0), but somehow we want to create a set owned by a user (e.g. `bob`) from it. In that case, the stream data must carry UID/GID/Perm.

In this case, admins can specify the UID, GID, and perm at the config line of the plugin. The given values take precedent over the values in the stream data.

baallan:
> @nichamon I think it makes sense for most cases. The corner case I can think of right now is stream data sending via root's channel (uid0, gid0), but somehow we want to create a set owned by a user (e.g. `bob`) from it. In that case, the stream data must carry UID/GID/Perm.

We're running root-owned app data collection about linux processes, and part of that data is the uid/gid of each process. At the store, admins may use storage policies or set mappings with these uid/gid in setting up databases,  but these uid/gid should not be confused with the ownership/group of the stream (who originally published it).

nichamon:
The patch passes against the test coverage [here](https://github.com/ovis-hpc/ldms-test/pull/59).

PR #1285, opened 2023-09-26
Patches for the json stream sampler plugin

nichamon:


nichamon:
The patch has passed the test cases in https://github.com/ovis-hpc/ldms-test/pull/59.

nichamon:
The changes in `ldmsd_stream_publish.c` and `ldms.pyx` are needed for the test script in https://github.com/ovis-hpc/ldms-test/pull/59.

PR #1286, opened 2023-09-29
Add the bridge connection mode

nichamon:


nichamon:
The patch adds the bridge connection mode to LDMSD. The bridge mode used on a sampler daemon works with a passive producer on an aggregator.

nichamon:
There is a newer version in another branch, which I'm debugging, so I'm closing this pull request.

PR #1289, opened 2023-10-06
Fix zap_sock race in ep accepting path

narategithub:


narategithub:
@nichamon could you please give this a try?

nichamon:
> @nichamon could you please give this a try?

I'll include the patch in my passive/bridge branch and test.

PR #1290, opened 2023-10-06
Fix zap_ugni buffer race in accept path

narategithub:


tom95858:
@narategithub so what was exactly happening here? The connect data was getting corrupted by subsequent receive data on the same connection?

narategithub:
@tom95858 In short, yes. This is because the thread that processed the CONNECT_REQUEST and notify the application is the listening thread. When the application 'zap_accept()' the new endpoint, the endpoint is assigned to another thread (the least busy one from the thread pool). The new thread can then start manipulating the buffer right away (e.g. by subsequent message after the peer got the accept message) while the old thread still has a pointer pointing to the same buffer. Realistically, I think it is rare to hit, but thanks to @nichamon we found out that this can happen :sweat_smile:.

narategithub:
@tom95858 BTW, we found this on 'zap_sock' first. Since 'zap_ugni' uses pretty much the same connection logic, it applied here too.

tom95858:
@nichamon is the boss...

PR #1292, opened 2023-10-10
Fix configure script GIT SHA issue

narategithub:
Since git v2.30.3, git added a top-level directory ownership check
(please see https://github.com/git/git/commit/8959555ce). If the owner
of the top-level directory does not match the current user, all git
commands failed with message "fatal: detected dubious ownership in
repository at _PATH_". However, our configure script does not properly
report the error nor give a recommended action. The configure script
silently moving on with 'NO_GIT_SHA' value.

This patch modify the configure script to check the git command if we
are in a git repository. If the git command resulted in an error, the
configure script report the error messages from the git command and
exited with a failure status. Please note that if the source tree is not
in a git repository (no .git directory), the configure scirpt will skip
git command checking.

Example error output with recommended action from the git command:
```sh
configure: '.git' directory presented. Checking git ...
fatal: detected dubious ownership in repository at '/home/narate/projects/ovis'
To add an exception for this directory, call:

        git config --global --add safe.directory /home/narate/projects/ovis
configure: error: git command error
make: *** [Makefile:636: config.status] Error 1
```

Example of a good output:
```sh
configure: '.git' directory presented. Checking git ...
__GIT_DIR_PATH__
OK
```

### DIFF
--- a/.github/workflows/ldms-test-build.yaml
+++ b/.github/workflows/ldms-test-build.yaml
@@ -30,6 +30,7 @@ jobs:
     - name: build-ovis
       shell: bash
       run: |
+        git config --global --add safe.directory ${PWD}
         mkdir -p build
         PREFIX=/opt/ovis
         OPTIONS=(
--- a/m4/options.m4
+++ b/m4/options.m4
@@ -623,16 +623,33 @@ dnl dnl queries git for version hash and branch info.
 AC_DEFUN([OPTION_GITINFO], [
 
 	export srcdir
+
+	dnl git test
+	if test -d "${srcdir}/.git"; then
+		AC_MSG_NOTICE(['.git' directory presented. Checking git ...])
+		dnl For configure output
+		git rev-parse --git-dir >&AS_MESSAGE_FD 2>&1
+		dnl For configure log
+		git rev-parse --git-dir >&AS_MESSAGE_LOG_FD 2>&1
+		if test 0 -ne $?; then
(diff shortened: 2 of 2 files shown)


narategithub:
@nichamon could you please see if this solve your build issue?

tom95858:
Hi @narategithub. Why not use the safe.directory attribute as recommended by https://github.com/git/git/commit/8959555ce? Does it not address this issue?

narategithub:
> Hi @narategithub. Why not use the safe.directory attribute as recommended by [git/git@8959555ce](https://github.com/git/git/commit/8959555ce)? Does it not address this issue?

`safe.directory` is a user's git option (in ~/.gitconfig, or /etc/gitconfig). I don't think ovis build script should change user's git option.

Do you think we should give an error and let users add ovis to `safe.directory` manually instead of this 'SUDO_UID' work around?

tom95858:
I think we are contravening the very security hole that github is trying to close. Why not just let it error?

narategithub:
@tom95858 Got it. I'll make it a build error. It was just a build log message previously (among a lot of other build logs), with a 'NO_GIT_SHA' as git SHA value. I'll make this a draft & update it. Will let you know when it is ready to review.

nichamon (approved):
This fixes the problem I found when I build as root while a user owns the source. The problem was that LDMS was built, but Git SHA wasn't available without an error message.

PR #1293, opened 2023-10-13
Fix LDMS/LDMSD passive/bridge support

nichamon:


nichamon:
The patch has been tested against https://github.com/ovis-hpc/ldms-test/pull/60. I also manually tested the disconnection situation.

PR #1298, opened 2023-11-17
fixes to tx2mon due to log change and other errors

oceandlr:


oceandlr:
Needed because tx2mon does not build due to errors. Some of these items need to be fixed in RC as well.

PR #1299, opened 2023-11-17
fixes to t2mon (similar to update in OVIS-4)

oceandlr:


oceandlr:
I don't know how this built. These and log conversion errors failed in top of tree. 
Is it appropriate to push to the RC? Is there a non-RC branch where this should be pushed?

PR #1301, opened 2023-11-30
Fix CYTHON_CHECK m4 macro

narategithub:
This patch addresses 2 things:

1) Add `cython{$python_major_version}` (e.g. 'cython3') into the
   cython program check list. Let's suppose the Python version is 3.6,
   the current check list only contains 'cython-3.6', 'cython3.6', and
   'cython'. This caused an issue on Ubuntu 22.04 since the cython
   package installed the program as 'cython3'.

2) Fix the version checking logic when CYTHON configure variable is
   specified (e.g.  './configure CYTHON=/opt/cython/bin/cython3').
   The existing CYTHON_CHECK macro uses 'AC_PATH_PROGS_FEATURE_CHECK'
   macro on CYTHON variable, which will skip the 'feature-test' part of
   the macro if the variable (CYTHON) is defined. Since the version
   checking logic is in the 'feature-test' part, the cython version has
   never been checked and 'cython not found' error is reported. This
   patch adds a logic to cover this case.

### DIFF
--- a/m4/cython_check.m4
+++ b/m4/cython_check.m4
@@ -6,7 +6,7 @@
 #
 #   Look for variations on the cython binary name, including ones
 #   that contain the PYTHON_VESRION parameter. For instance,
-#   if PYTHON_VERSION is 3.6, look for cython-3.6, cython3.6, and cython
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 This fixes cython check issue on Ubuntu 22.04 found by John Stile.

PR #1303, opened 2023-12-04
Modify the releasing Zap endpoints from IO thread process

nichamon:
The patch separates the process of reducing the number of Zap endpoints associated with an IO thread from the process of releasing the endpoint from the IO thread. Additionally, it defers the nullification of `zap->thread` until the endpoint reference reaches zero.

`zap_sock` and `zap_ugni` need to remove their events from epoll before delivering the disconnected event to applications, preventing the reception of additional epoll events. However, without this patch, this results in resetting `zap->thread` to NULL. Thus, the applications cannot retrive Zap thread information in the disconnected path. The patch addresses this limitation.

### DIFF
--- a/lib/src/zap/zap_priv.h
+++ b/lib/src/zap/zap_priv.h
@@ -483,8 +483,33 @@ zap_err_t zap_io_thread_ep_assign(zap_ep_t ep, int tpi);
  * The transport shall call this function to release an endpoint from the
  * associated io thread. \c zap.io_thread_ep_release() will also be called as a
  * subsequence.
+ *
+ * Consequently, the endpoint will not be processed by the thread any further.
+ * However, the endpoint still keeps a reference to the thread for further statistics data access.
+ * The endpoint references to the thread is removed when \c zap_io_thread_ep_remove() is called.
+ *
+ * \param ep   A Zap endpoint
+ *
+ * \return ZAP_ERR_OK on success. Otherwise, a Zap error code is returned.
+ *
+ * \see zap_io_thread_ep_remove
  */
 zap_err_t zap_io_thread_ep_release(zap_ep_t ep);
+
+/**
+ * Remove \c ep reference to the zap io thread.
+ *
+ * The function nullifies the endpoint reference to the thread and decrements
+ * the number of endpoints associated to the thread. This results in reducing
+ * the thread's load counter.
+ *
+ * \param ep    A Zap endpoint
+ *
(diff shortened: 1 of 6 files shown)


nichamon:
The patch is a part of the bigger patch to improve the thread_stats command.

lib/src/zap/zap_priv.h
```diff
@@ -478,14 +478,23 @@ double zap_env_dbl(char *name, double default_value);
 zap_err_t zap_io_thread_ep_assign(zap_ep_t ep, int tpi);
 
 /**
- * Release \c ep from the zap io thread.
+ * Disable \c ep assigned to the zap io thread.
```
narategithub, line 481: The explanation on `zap_io_thread_ep_release()` and `zap_io_thread_ep_remove()` are confusing to me. Not sure how to make it better though ... maybe just saying `release` means the endpoint won't be processed by the thread any more (i.e. no more event callbacks to the application) but the endpoint still keep a reference to the thread for further stat data access, while the `remove` means disassociate the endpoint from the thread and reducing thead's ep load? I'm not sure ...

tom95858:
I think this creates a race between disconnect and ldms_xprt_update. With this change isn't it possible to have an ldms_xprt_update succeed after a disconnect?

nichamon:
> I think this creates a race between disconnect and ldms_xprt_update. With this change isn't it possible to have an ldms_xprt_update succeed after a disconnect?

The patch keeps the existing logic for event removal from the epoll by `zap_sock`, `zap_ugni`, and `zap_fabric`, so it should not cause a race between the disconnected and update complete events. I also went through (my code inspection) every if-condition that checks for `ep->thread` whether it would cause any race condition by the patch but I didn't see any problems.

I made this patch so that Zap's applications can obtain the IO thread statistics of a Zap endpoint in the disconnected path.

@narategithub has suggested another change to facilitate this. The suggestion was to add a pointer to the `zap_thrstat` structure to the `zap_ep` structure so that Zap could retrieve the thread stats when `ep->thread` is NULL. However, this could mess up the reference count when an endpoint has been released from the IO thread and `t->_n_eps` and `t->stats->n_ep` have been decremented; although, the endpoint is still able to access the thread statistics, which is a part of the thread structure. Note that `t->_n_eps` and `t->stats->n_ep` are always in sync.

tom95858:
@nichamon thank you for looking into this...

tom95858:
Is everyone happy with this??

narategithub:
@tom95858 I think this patch looks OK to me.

Issue #1262, opened 2023-08-15
store_avro_kafka: avro schema evolution not working


morrone:
I don't know if this is an issue in decomposition or store_avro_kafka. But I have a decomp column that looks like this:

```
{"src":"cluster", "dst":"cluster", "type":"char_array", "array_len":12, "fill":"kern"},
```

I was hoping that this would add a "cluster" column with the name "kern" if it didn't appear in the schema. But when used with store_avro_kafka I see the error:

```
ERROR: store_avro_kafka: Error 22 retrieving 'cluster' from 'zfs_zpool' schema
```

tom95858:
@morrone, this looks like a bug, I'll look into it.

morrone:
OK, I think that I know what happened here. I had a configuration of ldms running _without_ the "cluster" column, and then I restarted it with a configuration that added the "cluster" column.

The issue is in serdes_schema_find(). The first time that store_avro_kafka sees an ldms decomposed row, it looks up the ldms schema name in the kafka schema registrty, looking for an avro schema of the same name. If it does find a matching schema name, it unfortunately just accepts that schema without looking at the schema contents. It doesn't bother to generate an avro schema from the decomposed data, and therefore it definitely doesn't compare the two to see if they are different.

If they are different,  like the are in my example (the avro schema on in the registry is missing the "cluster" field) it will error out later when it cannot find the field in the schema.

So what do we do about that?

Ideally, we _do_ want to support schema evolution. schema evolution would require store_avro_kafka to upload a new, compatible, version of a schema.

My recommendation it that we:

1) Always craft the avro schema the first time we see a row from a particular ldms schema name
2) Always include a "default" for all fields in the avro schema.

We need to always craft the avro schema ourselves, not just rely on the version in the registry to be correct. There are probably two main ways to hand things after that: do intelligent schema comparison ourselves, or just push it to the registry and let that figure things out.

I don't know if any of the libraries provide conventient avro schema comparison functions. If they don't, then it could be fairly complicated to write the code to do that correctly. I would suggest as a first pass that we just push it to the registry and let it figure things out for us. I suspect that it won't have a problem with pushing the same schema multiple times, and when the schema is different, the registry will perform schema compatibility tests for us.

[...]

PR #1305, opened 2023-12-06
Support avro schema evolution

morrone:
This pair of patches allows avro schema evolution in a kafka registry. It required two changes:

1. Add avro "default" values to each field in a schema
2. Always generate the avro schema and attempt to push it to the kafka registry (because it may be newer than what is in the registry).


morrone:
Fixes #1262

morrone:
@tom95858 ping. I've been able to evolve schemas multiple times with these patches, and I'm carrying these in the toss release of ldms. We couldn't be able to function without them.

PR #1310, opened 2023-12-11
Fix xprt_stats how it handle the reset attribute

nichamon:


nichamon:
@tom95858 Here is the fix to make `xprt_stats reset=true` work in V4.4.1. I cherry-picked the commit from OVIS-4.

PR #1316, opened 2023-12-20
Improve the thread_stats result

nichamon:


nichamon:
The pull request contains the commits that make the result of the thread_stats command more informative. The patches add the usage of IO threads of LDMS operations, e.g., update, lookup, and connection setup. They also add the statistics of LDMSD's worker threads.

Here is an example of thread_stat's new output.
image

PR #1320, opened 2024-01-12
Ovis 4.4.1 slingshot

oceandlr:
Two slingshot switch samplers. configure.ac whitespace and test fixes.
Note: while these are combined in the same pull request, they should both go in and putting them in separately would still result in a merge conflict.


oceandlr:
Closing this since it will go on 4.4.2

oceandlr:
Closing this so it will go on 4.4.2

PR #1322, opened 2024-01-12
Ovis 4.4.2 slingshot 

oceandlr:
two slingshot samplers. configure.ac whitespace and test fixes
(A future request will be made on master)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -246,6 +246,7 @@ OPTION_DEFAULT_DISABLE([yaml], [ENABLE_YAML])
 
 dnl Options for store
 OPTION_DEFAULT_ENABLE([store], [ENABLE_STORE])
+OPTION_DEFAULT_ENABLE([decomp], [ENABLE_DECOMP])
 OPTION_DEFAULT_ENABLE([flatfile], [ENABLE_FLATFILE])
 OPTION_DEFAULT_ENABLE([csv], [ENABLE_CSV])
 OPTION_DEFAULT_DISABLE([rabbitkw], [ENABLE_RABBITKW])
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -371,6 +372,7 @@ OPTION_DEFAULT_DISABLE([test_sampler], [ENABLE_TEST_SAMPLER])
 OPTION_DEFAULT_DISABLE([list_sampler], [ENABLE_LIST_SAMPLER])
 OPTION_DEFAULT_DISABLE([record_sampler], [ENABLE_RECORD_SAMPLER])
 dnl test_sampler will also build with --enable-ldms-test
+OPTION_DEFAULT_ENABLE([json_stream_sampler], [ENABLE_JSON_STREAM_SAMPLER])
```
narategithub, line 375: I don't think we have json_stream_sampler in 4.4.x

configure.ac
```diff
@@ -662,6 +665,8 @@ AS_IF([test "x$with_slurm" != no],[
 ])
 AM_CONDITIONAL([HAVE_SLURM], [test "x$have_slurm" = xyes])
 
+OPTION_DEFAULT_DISABLE([spank_subscriber], [ENABLE_SPANK_SUBSCRIBER], [Builds sampler that subscribes to spank plugin])
+
```
narategithub, line 669: Why do we disable slurm/spank plugin by default? I think it is already automatically determined by the `with slurm` option above. This will NOT build slurm by default unless you explicitly enable it.

narategithub (changes requested):
Aside from the comments I have in the patch, I think the configure.ac change should be in its own PR. Otherwise, it is hindering slingshot sampler changes from merging.

configure.ac
oceandlr: we want the option to have slurm but not build the plugin.

configure.ac
```diff
@@ -246,6 +246,7 @@ OPTION_DEFAULT_DISABLE([yaml], [ENABLE_YAML])
 
 dnl Options for store
 OPTION_DEFAULT_ENABLE([store], [ENABLE_STORE])
+OPTION_DEFAULT_ENABLE([decomp], [ENABLE_DECOMP])
```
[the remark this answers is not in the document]
oceandlr: we want the option to not have to build decomp for scenarios where we will not have a store.

configure.ac
oceandlr: Looks like residual ... we will remove this.

configure.ac
tom95858: That is fine, but with this solution, you should have to explicitly disable it, otherwise the vast majority of users will have to explicitly enable it.

oceandlr:
Putting in new one with configure.ac and slingshot_switch as two different commits.
Have to put in new pull request, since issues with force push

PR #1323, opened 2024-01-15
Enable github workflow tests on 'OVIS-4.x.x'

narategithub:
The github workflow tests we have only target pull requests for 'OVIS-4' branch. This patch make them target pull requests for any branch matching 'OVIS-4**' (any branch starting with 'OVIS-4').

### DIFF
--- a/.github/workflows/4.3.3-compat-test.yaml
+++ b/.github/workflows/4.3.3-compat-test.yaml
@@ -23,7 +23,7 @@ on:
   push:
     branches: [ OVIS-4 ]
   pull_request:
-    branches: [ OVIS-4 ]
+    branches: [ 'OVIS-4**' ]
 
 defaults:
   run:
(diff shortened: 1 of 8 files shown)


narategithub:
@tom95858 I think we need this to be merged into OVIS-4.4.2 and all pull requests targeting 4.4.2 shall include this so that at least the regular build and compatibility tests in github workflows will run against the PR for 4.4.2.

I'll notify other Pull Requests after this is merged.

morrone:
I would suggest instead that we just make a branch called "OVIS-4.4" for _all_ 4.4.X releases, rather than making a new branch for every minor tag.  Or maybe better yet, call is "b4.4" to match what I wrote here: https://github.com/ovis-hpc/ovis-wiki/wiki/OVIS-LDMS-Versioning.

Do we plan for the 4.4.X series to have many more releases?

tom95858:
@narategithub is this good to go?

narategithub:
@tom95858 I think it is good to go. Even though we expect this branch to be short-lived, we still at the least want to run the basic github workflow tests on pull requests targeting it.

PR #1325, opened 2024-01-16
Ovis 4.4.2

oceandlr:
Updated pull request, separating out fixes to configure.ac and slingshot_switch sampler.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -911,6 +911,7 @@ AM_CONDITIONAL([SYSCONFDIR_NOT_ETC], [test "${sysconfdir}" != "/etc"])
 AC_LIB_HAVE_LINKFLAGS([cxi], [], [
 #include <stddef.h> /* libcxi.h fails to include this */
 #include <libcxi/libcxi.h>
+#include <cassini_cntr_desc.h> /* needed at least starting with shs-2.1.0 */
 ])
 AM_CONDITIONAL([HAVE_LIBCXI], [test "x$HAVE_LIBCXI" = xyes])
 
@@ -918,7 +919,7 @@ AC_ARG_ENABLE([slingshot],
     [AS_HELP_STRING([--enable-slingshot], [require the slinghost related plugins @<:@default=check@:>@])],
     [],
     [enable_slingshot="check"])
-AM_CONDITIONAL([ENABLE_SLINGSHOT], [test "x$enable_slingshot" != xno])
+AM_CONDITIONAL([ENABLE_SLINGSHOT], [test "x$enable_slingshot" != xno -a "x$HAVE_LIBCXI" = xyes])
 AS_IF([test "x$enable_slingshot" = xyes],[
     AS_IF([test "x$HAVE_LIBCXI" = xno],
         [AC_MSG_ERROR([libcxi or its headers not found])])
--- a/ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
+++ b/ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
@@ -16,6 +16,7 @@
 
 #include <stddef.h> /* libcxi.h neglects to include this */
 #include <libcxi/libcxi.h>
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -927,6 +927,12 @@ AS_IF([test "x$enable_slingshot" = xyes],[
         [AC_MSG_ERROR([libcxi or its headers not found])])
 ])
 
+AC_ARG_ENABLE([slingshot_switch],
```
morrone, line 930: As far as I can tell, there are no external software dependencies in this sampler. So maybe we don't need to make compilation conditional?

oceandlr: @valleydlr We want the ability to optionally not build for those wanting to explicitly limit the build

configure.ac
```diff
@@ -927,6 +927,12 @@ AS_IF([test "x$enable_slingshot" = xyes],[
...
+    [AS_HELP_STRING([--enable-slingshot_switch], [require the slinghost on-switch plugins @<:@default=check@:>@])],
+    [],
+    [enable_slingshot_switch="check"])
```
morrone, line 933: While it won't actually change the behavior, it would be better to have this say enable_slingshot_switch="yes", because there is no actually checking of dependencies. If build is not explicitly disabled it _will_ attempt to build it.

configure.ac
```diff
@@ -246,6 +246,7 @@ OPTION_DEFAULT_DISABLE([yaml], [ENABLE_YAML])
 
 dnl Options for store
 OPTION_DEFAULT_ENABLE([store], [ENABLE_STORE])
+OPTION_DEFAULT_ENABLE([decomp], [ENABLE_DECOMP])
```
morrone, line 249: In a previous review, you said that we want to be able to disable decomp build when there are no stores. So could this just use the ENABLE_STORE option? Does it really need its own?

morrone:
@oceandlr Could you please update this PR with a title that reflects the content of the changes?

configure.ac
valleydlr: Added this to support a minimum build that could be installed on a network switch (slingshot switch in this case).

morrone: I am sorry; I am not following. Are you saying that on network switches, you need to enable stores, but disable decomp?

If you want to disable both stores and decomp at the same time, then decomp could be disabled with the "ENABLE_STORE" variable, and wouldn't need its own.

valleydlr: I see. You are suggesting changing the ENABLE store option for both. Will change and resubmit:)

tom95858:
@oceandlr is everyone ready to merge this? It does need to be rebased. I think @morrone would like the description to be changed to "slingshot switch monitoring software" or something like that.

tom95858:
@oceandlr, I think all we need to do here is rebase (to get rid of conflicts) and address @morrone's request to change the pull request description.

configure.ac
```diff
@@ -468,7 +468,8 @@ AC_ARG_ENABLE([ibnet],
 	[],
 	[enable_ibnet="check"])
 AM_CONDITIONAL([ENABLE_IBNET], [test "x$enable_ibnet" != xno])
-AS_IF([test "x$enable_ibnet" = xyes],[
+AS_IF([test "$enable_ibnet" = xyes],[
```
morrone, line 471: That change is not correct.

configure.ac
```diff
@@ -468,7 +468,8 @@ AC_ARG_ENABLE([ibnet],
...
+	AC_MSG_NOTICE([Disable ibnet module NOT requested])
```
morrone, line 472: I'm not sure that his notice makes sense to add.

tom95858: @morrone, no it's broken. I was playing with the github editor and accidentally broke it. the 'x' is missing, etc... @valleydlr has changes staged that he will test today and push tomorrow.

morrone:
I rebased this and cleaned it up over in #1361.

PR #1326, opened 2024-01-17
backport pr 1301 to fix cython in configure

baallan:
need narate's fix in #1301 on stable branch

### DIFF
--- a/m4/cython_check.m4
+++ b/m4/cython_check.m4
@@ -6,7 +6,7 @@
 #
 #   Look for variations on the cython binary name, including ones
 #   that contain the PYTHON_VESRION parameter. For instance,
-#   if PYTHON_VERSION is 3.6, look for cython-3.6, cython3.6, and cython
+#   if PYTHON_VERSION is 3.6, look for cython-3.6, cython3.6, cython3, and cython
 #   in that order. Each found binary is feature checked to determine
 #   if its Cython version is greater than or equal to the supplied
 #   CYTHON_MIN_VESRION parameter. The serach stops on the first binary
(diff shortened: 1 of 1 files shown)


tom95858:
@narategithub is this good with you?

baallan:
@tom95858 closed? 4.4.1/2 do not build on rhel 8 using python3.8/cython-3.8 (cython 0.29.x) without this patch picked directly from OVIS-4.

tom95858:
Sorry @baallan, didn't mean to close this just meant to comment...

narategithub:
@tom95858 I think this is exactly commit 05de014 in OVIS-4. I wish that this is a cherry-pick commit instead of a new commit (not sure why it couldn't be a cherry-pick though, should be no conflict). But, I'm OK to merge this to OVIS-4.4.2.

baallan:
merge or cherrypick makes no difference to me. just want to be able to build using the updated macro Narate provided.

morrone:
> merge or cherrypick makes no difference to me. just want to be able to build using the updated macro Narate provided.

@baallan, I think the take away is that this PR should have been from a cherry-pick of commit 05de0143, and really shouldn't be landed in its current form. Cherry-picking to keep changes in sync between branches of a project is pretty important for the sanity of the maintainers. It has a couple of advantages:

- It retains the original Author listed in the patch
- It retains the original commit message in the patch (you can feel free to add _additional_ comments at the end of the original one)

Maintaining the same commit message in the patch is especially important. When maintainers and developers look at different git branches of the same code base with "git log --oneline", we really need to see the exact same first commit line in the listing on each branch. In this case that should be:

```Fix CYTHON_CHECK m4 macro```

But if we landed your suggested commit on the 4.4 branch series, we would see a one line commit message of:

```backport pr 1301 to fix cython in configure```

There is no way we are going to know, on sight, that those were supposed to be the same fix. It creates unnecessary extra work.

So I think this PR needs to be reworked before it can be landed.

baallan:
@morrone @tom95858 As Tom is the one with commit access I suggest he do the cherrypick of https://github.com/ovis-hpc/ovis/commit/05de014390be2553f12723462708b58a83fb5f4c to OVIS-4.4.2 and we move on. It's unfortunate Narate didn't ask him to do the equivalent at the time of the original PR.

morrone:
An example of how to do a PR of a cherry-pick back-ported commit: PR #1336

PR #1328, opened 2024-01-18
Fix `zap_ugni` bug in endpoint rejection path

narategithub:
This is not tested yet.


tom95858:
@morrone would it  be possible for you to test this. We have lost access to a uGNI machine.

morrone:
> @morrone would it be possible for you to test this. We have lost access to a uGNI machine.

I don't think that we have uGNI either.

tom95858:
@narategithub please add @bschwal or someone else who has access to uGNI as a reviewer to test this.

bschwal:
Brandt is the only person who really has a knowledge of building and running on cray.

tom95858:
@narategithub Let's confirm that this will never see the light of day and then just close it.

PR #1330, opened 2024-01-19
protect ovis_log_flush from invalid arguments.

baallan:
ovis_log_flush can be called in any order wrt initialization, so it must be guarded against null and syslog values. This prevents segfault in fflush.

### DIFF
--- a/lib/src/ovis_log/ovis_log.c
+++ b/lib/src/ovis_log/ovis_log.c
@@ -890,7 +890,9 @@ int ovis_log_open(const char *path)
 
 int ovis_log_flush()
 {
-	return fflush(log_fp);
+	if (log_fp && log_fp != OVIS_LOG_SYSLOG)
+		return fflush(log_fp);
+	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


tom95858:
@nichamon I think having OVIS_LOG_SYSLOG and overloading the value 0x7 as a FILE* is bad practice. I think the initialization of log_fp should be to stdout. This will cause anything started via systemd to go to syslog as expected. It will also avoid the checking around the code for this value and disabuse accidentally calling a function that is expecting a FILE* with an invalid address. What do you think?

baallan:
The use of a sentinel value instead of a separate boolean is a memory optimization we can afford to live without if anyone finds it bothersome; we will either way need some flag that differentiates syslog from logging to files.
Given that afaik we still have the goal of running and controlling "syslog or not" choices in non-systemd environments, I suggest we should not get rid of log_fp altogether.
Unless a change was introduced with the multi-log-object library, we have always started with log_fp defaulted to stdout and syslog being something the user must opt into.

PR #1331, opened 2024-01-19
fix segfault caused by attempting to flush syslog

baallan:
when logging is via syslog, flush must not be called. Calling syslog is achieved by assigning log_fp a sentinel value which can never be a file pointer and checking for it before calls that expect a file pointer. This patch fixes a missing check before an fflush that was introduced with the new logging infrastructure.


tom95858:
why.

baallan:
> why.

expanded original comment.

tom95858:
I would really remove this and the associated check. If we start things with systemctl and default log_fp to stdout doesn't everything go to syslog anyway? But let's get it fixed in the tree and then we can discuss how to move forward.

Issue #1311, opened 2023-12-12
LDMSD Bug When Using IPv6 address


Snell1224:
I have a github actions .yaml file that tests the Darshan-LDMS integrated coded (i.e. darshanConnector) and noticed that it was failing to send the stream data to the LDMS streams daemon. The ``ldms_ls`` was also returning ``Connection failed/rejected``. Did some debugging and it turns out this error appeared on commit https://github.com/ovis-hpc/ovis/commit/39468f0bf3d60311b0d269a2906745064ec61791. From what I understand this commit enabled support for ipv6 by checking which ipv4/6 is enabled on the system.

My current github actions test script ran on ``ubuntu-latest`` which has ipv6 enabled so, I'm assuming, LDMS detected this and initialized a daemon using the ipv6 address (i.e. for ``localhost``) over ipv4. 

I don't know if there is an option to tell LDMS to use ipv4 over ipv6 so I ended up disabling ipv6 altogether with
``sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1``. Once disabled, the LDMSD was able to collect the streams data and ``ldms_ls`` did not return any errors. Unless there's something I'm misunderstanding, I think there is a bug somewhere in this commit. 

RECREATE PROBLEM:
Using ``build-test-ubuntu-22.04.yaml``, add the following options to enable ipv6 and re-run the script:
```
container:
        image: ovishpc/ovis-ubuntu-build
        options: --privileged --sysctl net.ipv6.conf.all.disable_ipv6=0
```

When ipv6 is enabled:

![Screenshot 2023-12-11 152226](https://github.com/ovis-hpc/ovis/assets/60634446/e1d6cc62-fd9a-4ded-bcc6-5a35dc2ae7f4)
![Screenshot 2023-12-12 121312](https://github.com/ovis-hpc/ovis/assets/60634446/ef07b0eb-b541-44f1-b95a-68854a28ac37)

When ipv6 is disabled:

![Screenshot 2023-12-11 152344](https://github.com/ovis-hpc/ovis/assets/60634446/a6af7334-ed54-486d-ace0-b75406093993)
![Screenshot 2023-12-12 121354](https://github.com/ovis-hpc/ovis/assets/60634446/e5008dd4-ef1e-4d95-b21e-6c6ac3bb1f60)

tom95858:
@narategithub is this still an issue or can we close this?

narategithub:
[...]

PR #1333, opened 2024-01-19
Make `ldms_xprt_listen_by_name()` prefers IPv6.

narategithub:
`ldms_xprt_listen_by_name()` listened to 0.0.0.0 IPv4 address by default if `host` is `NULL`. This patch made
`ldms_xprt_listen_by_name()` to fully rely on `getaddrinfo()` in resolving the addresses, and then we pick the first IPv6 address. If no IPv6 addresses available, the first address returned by `getaddrinfo()` is used.

This has been tested with `ldmsd` on IPv4-only machine, and IPv4+6 machine.

On IPv4+6 machine, if the users only to listen on IPv4 addresses, they have to either provide listening address on the CLI (e.g. `-x sock:411:0.0.0.0`), or specify the host in config file (e.g. `listen port=411 xprt=sock host=0.0.0.0`).

### DIFF
--- a/ldms/src/auth/ldms_auth_munge.c
+++ b/ldms/src/auth/ldms_auth_munge.c
@@ -204,9 +204,46 @@ int __auth_munge_xprt_bind(ldms_auth_t auth, ldms_t xprt)
 }
 
 static
+void __ipv4_ipv6_xlat(struct sockaddr *sa, socklen_t *len)
+{
+	/*
+	 * NOTE
+	 * ----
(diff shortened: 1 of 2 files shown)


narategithub:
This is a patch to address #1311 .

narategithub:
NOTE: Compatibility test failed due to ldms_ls-4.3.3 to ldmsd-4 (top-of-tree) having an issue with munge authentication. This is because ldmsd-4 sent back IPv6 munge payload, but ldms_ls-4.3.3 expects IPv4 in the payload. I'm looking into it.

PR #1338, opened 2024-01-26
ldms-notify enhancements needed for production convenience.

baallan:
While not always calculable, the duration of tracked pids is of great interest and rather hard to compute downstream. This adds the duration field to process end messages and a format option to suppress it for sites not ready to migrate to the expanded schema.

Unlike ldmsd, the status of ldms-notify cannot be determined remotely, so this also adds a heartbeat message, which by default is not generated.


baallan:
There is not yet a OVIS-4 branch version of this, as a fixed version of test_slps for OVIS-4 is still in work.

PR #1345, opened 2024-02-01
Adjust libcxi test to be correct on toss4 and some libcxi versions it supplies

baallan:
Toss4 libcxi is incompatible with slingshot sampler source which assumes libcxi.h defines c1_cntr_descs.
Specifically toss package cray-libcxi-2.1.0-4.t4.x86_64 doesn't have a header defining c1_cntr_descs, but it passes the old configure check and then fails at build time.
This includes checking for c1_cntr_descs in the config check.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -910,6 +910,7 @@ AM_CONDITIONAL([SYSCONFDIR_NOT_ETC], [test "${sysconfdir}" != "/etc"])
 AC_LIB_HAVE_LINKFLAGS([cxi], [], [
 #include <stddef.h> /* libcxi.h fails to include this */
 #include <libcxi/libcxi.h>
+const char *c = c1_cntr_descs[0].name;
 ])
 AM_CONDITIONAL([HAVE_LIBCXI], [test "x$HAVE_LIBCXI" = xyes])
 


baallan:
The default workaround, which would be to add --disable-slingshot to the configure line does not actually disable slingshot, at least if libcxi is detected as present.

morrone:
> The default workaround, which would be to add --disable-slingshot to the configure line does not actually disable slingshot, at least if libcxi is detected as present.

I do not think this is correct. "--disable-slinghost" disables the build of the slingshot samplers even when libcxi is present.

morrone:
I have the fix for this actually, I just need to push it upstream. HPE changed their header layout (and, frankly, broke it).

morrone:
> > The default workaround, which would be to add --disable-slingshot to the configure line does not actually disable slingshot, at least if libcxi is detected as present.
> 
> I do not think this is correct. "--disable-slinghost" disables the build of the slingshot samplers even when libcxi is present.

OH! I see, 4.4.X still hasn't gotten the fix that add the missing $ that landed on OVIS-4. So it works on OVIS-4 but not on OVIS-4.4.X.

oceandlr:
BTW, We do want the option to disable slingshot even if the libraries are present (and to disable anything, really).

morrone:
> BTW, We do want the option to disable slingshot even if the libraries are present (and to disable anything, really).

@oceandlr Yes, but for a typo, that functionality has always been there.

PR #1346, opened 2024-02-01
Adjust libcxi test to be correct on toss4 and some libcxi versions  supplies

baallan:
Toss4 libcxi is incompatible with slingshot sampler source which assumes libcxi.h defines c1_cntr_descs.
cray-libcxi-2.1.0-4.t4.x86_64 doesn't have a header defining c1_cntr_descs.

cherry picked from 4.4.2 patch https://github.com/ovis-hpc/ovis/pull/1345/commits/fac259c837f1b2f49f3195234eaedcb2dbd801ed


morrone:
See fixes in #1348 and #1349.

PR #1348, opened 2024-02-01
slingshot_metrics: new header include needed by shs-2.1.0

morrone:
libcxi.h fails to include cassini_cntr_desc.h, which is almost certainly a bug in their header include tree. We include it explicitly until they fix things.


baallan:
@morrone thanks for providing a more complete fix.  I built your branch slingshot_header on my local toss-release-4.6-6 machine with no configure arguments and it handled slingshot plugins correctly afaict.

baallan (approved):

baallan:
@tom95858 Is there something else this patch from @morrone needs before merging to 4.4.2? It solves the out-of-the-box build failures for toss3/toss4.

PR #1351, opened 2024-02-01
Check for the openssl library and always assign OPENSSL_LIBS

nichamon:
Without the patch, the build variable OPENSSL_LIBS is only assigned when `ovis_auth` is enabled. However, LDMS uses OpenSSL to generate digests for LDMS Sets. The patch ensures that OPENSSL_LIBS is always populated.


nichamon:
I found this nit because I'm building LDMS without ovis_auth, and the build failed.

PR #1359, opened 2024-02-12
cleanup dcgm_sampler dev cruft

baallan:
This brings dcgm_sampler in top of tree in line with ovis-4.4.2 version.
changes:
eliminate code generator
eliminate file-scope variables use_base, termed, and pthread usage.


morrone (approved):
Looks good.

PR #1363, opened 2024-02-16
Slingshot Switch Samplers Plus Config fixes

oceandlr:
This includes
1) Slingshot switch samplers
2) Fixes to the top configuration for build checks

This is being submitted as one request with two commits since a merge of the two would necessary anyway

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -429,6 +429,7 @@ AS_IF([test "x$enable_papi" != xno],[
 ])
 AM_CONDITIONAL([HAVE_LIBPAPI], [test "x$HAVE_LIBPAPI" = xyes])
 AM_CONDITIONAL([HAVE_LIBPFM], [test "x$HAVE_LIBPFM" = xyes])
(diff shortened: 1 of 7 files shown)


tom95858:
@morrone, please take a look? I really want to get 4.4.2 out the door.

configure.ac
```diff
@@ -429,6 +429,7 @@ AS_IF([test "x$enable_papi" != xno],[
 ])
 AM_CONDITIONAL([HAVE_LIBPAPI], [test "x$HAVE_LIBPAPI" = xyes])
 AM_CONDITIONAL([HAVE_LIBPFM], [test "x$HAVE_LIBPFM" = xyes])
+AM_CONDITIONAL([ENABLE_PAPI], [test "x$enable_papi" != no -a "x$HAVE_LIBPAPI" = xyes -a "x$HAVE_LIBPFM" = xyes])
```
morrone, line 432: The first "no" here should be "xno"

morrone:
Just the one minor issue, other than that it looks good.

configure.ac
oceandlr: In line 432 you mean?

PR #1366, opened 2024-02-27
Fix buffer size declaration in syspapi

baallan:
pfm_name copy may be up to size of .info field being copied.
this uses the papi header macro to match.

### DIFF
--- a/ldms/src/sampler/syspapi/syspapi_sampler.c
+++ b/ldms/src/sampler/syspapi/syspapi_sampler.c
@@ -89,7 +89,7 @@ typedef struct syspapi_metric_s {
 	int init_rc; /* if 0, attr is good for perf_event_open() */
 	struct perf_event_attr attr; /* perf attribute */
 	char papi_name[256]; /* metric name in PAPI */
-	char pfm_name[256]; /* metric name in perfmon (PAPI native) */
+	char pfm_name[PAPI_HUGE_STR_LEN]; /* metric name in perfmon (PAPI native) */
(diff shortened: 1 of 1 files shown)


baallan:
@baallan need to cherry pick this to ovis-4

ldms/src/sampler/syspapi/syspapi_sampler.c
```diff
@@ -89,7 +89,7 @@ typedef struct syspapi_metric_s {
 	int init_rc; /* if 0, attr is good for perf_event_open() */
 	struct perf_event_attr attr; /* perf attribute */
 	char papi_name[256]; /* metric name in PAPI */
-	char pfm_name[256]; /* metric name in perfmon (PAPI native) */
+	char pfm_name[PAPI_HUGE_STR_LEN]; /* metric name in perfmon (PAPI native) */
```
tom95858, line 92: @baallan, I don't believe that this field is used anywhere. We should just remove it instead of consuming 1k for every papi metric.

baallan: @tom95858 it is copied at line 251 into a buffer which is just 256 bytes while the papi source string is up to 1024. The alternative is to make the structure's pfm_name field separately allocated and freed, which would be  a much bigger change.

tom95858: @baallan, after it is copied in there it is never used again. My suggestion is to just remove the field instead of wasting 1k for every event.

PR #1368, opened 2024-02-27
fix free of wild pointer if user omits an arg

baallan:
strgp_decomp_init 'goto einval' jumps free an
uninitialized pointer (decomp) if the user supplies the wrong arguments.


baallan:
@baallan may need to cherry pick this to ovis-4

baallan:
fixed already on main branch in commit 4209fbc26

PR #1369, opened 2024-03-04
Cleanup LDMSD's command-line options

nichamon:


ldms/src/ldmsd/ldmsd.c
```diff
@@ -2097,13 +2126,6 @@ int main(int argc, char *argv[])
...
 
-	if (!foreground) {
-		if (daemon(1, 1)) {
-			perror("ldmsd: ");
-			cleanup(8, "daemon failed to start");
-		}
-	}
-
 	/*
 	 * TODO: It should be a better way to get this information.
```
tom95858, line 2130: @nichamon can you elaborate on this comment?

ldms/src/ldmsd/ldmsd.c
```diff
@@ -2173,25 +2194,7 @@ int main(int argc, char *argv[])
...
-			}
-			if (!pidfile) {
-				ovis_log(NULL, OVIS_LERROR, "Out of memory\n");
-				av_free(auth_opt);
-				exit(1);
-			}
-		}
+	if (pidfile) {
 		if( !access( pidfile, F_OK ) ) {
 			ovis_log(NULL, OVIS_LERROR, "Existing pid file named '%s': %s\n",
```
tom95858, line 2199: @nichamon, @valleydlr do we need this message?

ldms/src/ldmsd/ldmsd.c
```diff
@@ -2241,13 +2245,12 @@ int main(int argc, char *argv[])
 	ldms_version.flags, OVIS_GIT_LONG
 
 #if OVIS_LDMS_HAVE_AUTH
-				fprintf(bfile, BANNER_PART1_A
+			fprintf(bfile, BANNER_PART1_A
```
tom95858, line 2248: @nichamon, @valleydlr did we talk about getting rid of this banner file?

ldms/src/ldmsd/ldmsd.c
nichamon: I added that when I refactored LDMSD to use the `ovis_log` library. The comment was to remind myself to find a better way for admins to change the time format of log messages from date-time to timestamp. Currently, admins could configure LDMSD to use timestamp from the epoch (seconds) in log messages by setting the environment variable `LDMSD_LOG_TIME_SEC`, which I didn't like this method. However, I didn't introduce an additional command-line option at the time because there were so many options already.

I don't remember the motivation of different time-formats of log messages. If I think of two path forwards:

1. We think there is no practical use of supporting multiple time formats in log messages.
In this case, I will remove the code block.

2. We decide to keep the support.
I would change the new `log_file` config command from
`log_file path=<log file path>`
to
`log dest=<log file path> [time_format=<date-time|timestamp|none>]`.
The `ovis_log` library supports the three modes: date-time (LDMSD default), timestamp in second from epoch, and no time information.
Beside modifying the config command, I would remove the logic to get the time format from the environment.

What do you think?

ldms/src/ldmsd/ldmsd.c
nichamon: According to the spreadsheet we went over together in Austin (I left early), it was decided to remove the command-line option `-B` and add a new config command to replace the option.

ldms/src/ldmsd/ldmsd.c
nichamon: It seems this message could be helpful in identifying misconfiguration of the `-r` in LDMSD. For example, if two LDMSDs are configured to create a PID file at `/home/my/var/pid/ldmsd.pid` on different nodes that share the same home directory, they will overwrite each other's PID files.This scenario is more likely to occur in a test environment. Ultimately, I am neutral on whether to keep or remove the message.

tom95858: @nichamon, I think this is reasonable.

ldms/src/ldmsd/ldmsd.c
tom95858: @valleydlr, do you want to retain the Banner file?

baallan:
@nichamon @tom95858 This merge makes subsequent subsequent pull requests fail in for example:
https://github.com/ovis-hpc/ovis/actions/runs/8175122029/job/22351519845?pr=1370
which hangs forever in 'starting agg4'
repeatably.

tom95858:
@narategithub I think this causes the compatibility test to hang trying to detect the successful start of a daemon. Could you please take a look a this and fix it?

narategithub:
@tom95858 I'm looking into it.

nichamon:
@narategithub created a patch to fix the problem here. https://github.com/ovis-hpc/ovis/pull/1371

PR #1370, opened 2024-03-06
fix task end message env variables copy code in netlink notifier

baallan:
cloned code from init message env variable copy code needed conversion to the _end variable.


tom95858:
@baallan, can you please rebase this change and force push the change?

baallan:
Will do. does someone have enough privs to kill the current check, which will be hanging for 3 more hours?

tom95858:
It will automatically kill the previous check and restart when you force push the branch.

baallan:
@tom95858 looks like the pr test is working again.

PR #1371, opened 2024-03-06
Run ldmsd in compat-test in background

narategithub:


narategithub:
@tom95858 this should fix it. The other pull requests still need to rebase after this is merged though.

PR #1374, opened 2024-03-20
add the key piece of the json parser error output: the bad input

baallan:
When users send malformed json, we get urps such as
```
first_line   : 0
last line    : 0
first column : 0
last column  : 5
str: a cryptic message about colons, commas or curlies.
```
This patch includes 'input' in the message, which tends to be truncated from the complete input very near the error.
With this, the admin has a modest chance of figuring out what message went wrong.


tom95858:
@baallan, there is ongoing work to replace ovis_json with libjansson which provides much better error reporting.

baallan:
@tom95858 any schedule yet on jansson landing?

PR #1378, opened 2024-03-25
Update workflows to support b4.4

morrone:
Also added support for using main and v* style tags in the future.


narategithub (commented):
:+1:

morrone:
I update the PR to remove the use of parenthesis in the patterns. I think those might not be supported by the github actions pattern matching. Probably good now.

tom95858:
@narategithub did you take a look at this?

PR #1379, opened 2024-03-25
Update workflows to support b4.4 (branch OVIS-4 edition)

morrone:
Also added support for using main and v* style tags in the future.


narategithub (commented):
:+1:

PR #1385, opened 2024-04-04
Netlink add jobid 4.4.x

baallan:
This takes care of getting the jobid to the netlink-notifier messages for the many ways in which a process can be created in slurm/lsf without receiving a resource manager environment variable list. The intent is to use the same data source as the job_info/jobid plugins now use.

Processes with uid's below the user range do not get a jobid applied to them.
If enabled, a list of default locations and variable names is checked for a jobid string.
The file location can be overridden. It is reread on inotify events.
This feature is not enabled by default.
The integration test is supplied:ldms-static-test.sh linux_proc_sampler.job.


bschwal:
This change provides a huge improvement in getting job ids for jobs that drop into an environment without slurm environment variables, which is quite often. It will also be a good template for other streams services to get job_id.

PR #1381, opened 2024-04-01
Update LDMSD's usage and man page to reflect the current cmd-line options

nichamon:

PR #1386, opened 2024-04-04
Fix typos in ldmsd.man

nichamon:


nichamon:
@morrone Thanks for catching the typos. The change is only available on OVIS-4.

nichamon:
The patch addressed comments in #1381 .

PR #1387, opened 2024-04-17
Clarify the dependencies in README.md

nichamon:
The patch clarifies that Python3 and munge development libraries are dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ one the "Code" button. Or use the following command:
 * flex
 * libreadline
 * openssl development library (for OVIS, LDMS Authentication)
-* libmunge (for Munge LDMS Authentication plugin)
-* Python >= 3.6 and Cython >= 0.25 (for the LDMS Python API and ldmsd_controller)
+* libmunge development library (for Munge LDMS Authentication plugin)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,8 +72,8 @@ one the "Code" button. Or use the following command:
 * flex
 * libreadline
 * openssl development library (for OVIS, LDMS Authentication)
-* libmunge (for Munge LDMS Authentication plugin)
-* Python >= 3.6 and Cython >= 0.25 (for the LDMS Python API and ldmsd_controller)
+* libmunge development library (for Munge LDMS Authentication plugin)
+* Python >= 3.6 development library and Cython >= 0.25 (for the LDMS Python API and the LDMSD Interface, ldmsd_controller)
```
tom95858, line 76: I think we rely on 0.29 for Cython now., but the issue that prompted this was that the user in question didn't read any of this, they just went directly to the Building section which doesn't reference these dependencies. So I think we need a pointer in the "Building" section that refers back to the dependencies section.

tom95858: @narategithub could you please collect the versions from the build container that we have? I know it's a pain, but we know that works. Also this readme should refer to that container and how one builds using it.

tom95858 (changes requested):

narategithub:
Closing this one and use #1388 instead.

PR #1388, opened 2024-04-19
Clarify the dependencies in README.md

narategithub:
The patch clarifies that Python3 and munge development libraries are dependencies. This patch also adds a section that points to pre-built containers.

PR #1391, opened 2024-04-22
Clarify the dependencies in README.md

nichamon:
The patch clarifies that Python3 and munge development libraries are dependencies. This patch also adds a section that points to pre-built containers.


nichamon:
This change is identical to #1388 for the b4.4 branch.

PR #1392, opened 2024-04-22
Netlink stream format harmonization for 4.4

baallan:
Adds format 3 option to the netlink publisher. This makes linux/slurm/lsf json messages and init/exit messages similar in field ordering to the maximum extent possible so that creating policies and tools transforming stream data across resource managers is simpler.


bschwal (approved):
Looks good and will be extremely helpful for better pid collection. Testing on hpc cluster over several days showed we captured all jobs ran on the system .

PR #1393, opened 2024-04-24
add rollover to blob_stream_writer in b4.4

baallan:
This eliminates the need for cron-driven scripting to make interactive plugin reconfig calls just to achieve file rollover.
This will help reduce incomplete blob file writes as well.


bschwal (approved):
Very useful for better usage of the blob store. Current manual reconfig to rollover is ineffective.

PR #1397, opened 2024-05-13
Change prefix used by git-version-gen to "v" from "OVIS-".

morrone:


tom95858:
So should I delete the release, update the tag, and then regenerate the release?

morrone:
> So should I delete the release, update the tag, and then regenerate the release?

Sure, that or just make a new tag.

PR #1400, opened 2024-05-31
Decomp func op

tom95858:
This set of changes implements functional operators in the decomposition path. These changes also change the JSON parser to use the libjansson library which provides better error reporting and an easier to use API.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -65,6 +65,8 @@ AC_C_FLEXIBLE_ARRAY_MEMBER
 OVIS_PKGLIBDIR
 
 AC_LIB_HAVE_LINKFLAGS([jansson], [], [#include <jansson.h>])
+AS_IF([test "x$HAVE_LIBJANSSON" = xno],
+	[AC_MSG_ERROR([libjansson or jansson.h not found])])
 AM_CONDITIONAL([HAVE_LIBJANSSON], [test "x$HAVE_LIBJANSSON" = xyes])
 
 dnl change sharedstatedir default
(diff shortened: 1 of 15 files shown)


morrone:
I think the containers need libjansson >=2.11 to support json_error_code().

narategithub:
Yes. libjansson 2.13.1 was installed from source in the CentOS-based 
images (ovishpc/ovis-centos-build and ovishpc/ovis-compat-centos7).

narategithub:
I believe this should pass all the workflow once @tom95858 trigger the re-run of the workflows.

tom95858:
Should we get rid of the code that relies on the later version? What can we
reasonably expect to be installed?

narategithub:
@tom95858 I noticed that you have `json_error*` stuff conditioned on JANSSON version. I want to confirm that I should downgrade libjansson in CentOS7 images back to 2.10 (shipped with the distro), right?

tom95858:
@narategithub I think the conditional you added is fine. For the CentOS7 images, then yeah, you should install the libjansson-devel version to whatever is shipped with the distro.

narategithub:
@tom95858 the images were updated accordingly with libjansson that shipped with the distro. 

FYI, I tried the commit without the JANSSON_VERSION_HEX conditional here:  https://github.com/narategithub/ovis/pull/27 (in my repo) and it failed as expected (undefined `json_error_code`).

The commit with JANSSON_VERSION_HEX condition passed the workflows: https://github.com/narategithub/ovis/pull/28 (with libjansson that shipped with the distro).

PR #1402, opened 2024-06-03
Add LDMS_STREAM_EVENT_CLOSE

narategithub:
Due to the asynchronous nature of `ldms_stream`, a stream event may already in the process of being delivered to the application callback function at the same time that the application closes the stream client and releases application's resources associated with it. This results in a use-after-free.

This patch adds `LDMS_STREAM_EVENT_CLOSE` callback event that is delivered to the application callback function when the stream client is closed. The `LDMS_STREAM_EVENT_CLOSE` is the last event to be delivered to the client call back function, so that the application can then safely release resources asscoiated with the stream client after receiving it.


baallan:
does this issue not occur in b4.4 also? might explain some mysterious things. @narategithub

PR #1405, opened 2024-06-17
Add sampler plugin 'ldmscon2024_sampler'

nichamon:
The sampler plugin is an example and a reference point from the Sampler Plugin Implement focus group at LDMSCON2024.

The commit infomation also shows the necessary files to be modified and created in order to adding a new sampler plugin to the git tree.

The sampler plugin is disabled by default as it is not for production environment. To build the plugin, --enable-ldmscon2024_sampler must be given when ./configure.


tom95858:
@nichamon did we decide to move this to the UG page?

nichamon:
> @nichamon did we decide to move this to the UG page?

We decided to host this on the LDMSCON2024 website. I will upload the files to the Google folder we share (or about to share) with the attendees. I would like to include the `git show` of the commit too because it's obvious which .ac and .am files they need to add and modify when they add a new sampler plugin. Let me think a bit about how to do this. I'll close this pull request now.

PR #1407, opened 2024-06-20
Do not automatically create a github release

narategithub:
~~This will allow `config/git-version-gen` to get version from unannotated
tags. This also enable marking a version without automatically create a
release~~. (Keeping this as a history).


narategithub:
@tom95858 the compatibility test failed is fixed in #1408. Once you merge #1408, I'll rebase this and it should pass all tests.

tom95858:
@narategithub I think we decided to just remove the auto-release generation. Could you refactor and re-submit?

morrone:
I don't think we should do this! We _should_ always be using annotated tags for releases. Period.

morrone:
> @narategithub I think we decided to just remove the auto-release generation. Could you refactor and re-submit?

What auto-release generation would we remove? What is the current issue?

tom95858:
> I don't think we should do this! We _should_ always be using annotated tags for releases. Period.

We're not suggesting anything other than that we can use annotated tasks  even when we don't want a release, e.g. v4.5.0

morrone:
A "v4.5.0" tag looks like a release to me. If it is a pre-release, shouldn't is be something like "v4.5.0-rc1"?

In any event, yes, using an annotated tag is the way to go.

narategithub:
@morrone

The actual issue I have is that the top-of-tree cannot resolve the package version (in autoconf terms). The version string becomes "UNKNOWN" and broke test scripts that use version numbers to determine which code to execute. This "UNKNOWN" issue was introduced by the change of the versioning tag prefix from "OVIS-" to "v". 

```sh
$ ldmsd -V
LDMSD Version: UNKNOWN
LDMS Protocol Version: 4.2.0.0
LDMSD Plugin Interface Version: 3.2.2.0
git-SHA: 277d3f09ecc75dc27b66658d846a15a755cac2c9

```

At first I think we can easily avoid this patch by just annotated tagging the current "OVIS-4.3.11" with "v4.3.11" and the package version of the top-of-tree will be "4.3.11.{NUM}-{SHA}" and we can move on. However, the leading "4.3.11" looks misleading. It makes the top-of-tree look "older" than the 4.4.x releases.

So, we (Tom and I) wanted to make it so that the package version of the top-of-tree would report "4.5" something without triggering the automatic GitHub release script, and that was the original pull request was for: allowing the package version to be obtained from "v4.5.0-{SOMETHING}" without an annotation so that the package version is not "UNKNOWN" and the GitHub release is not triggered.

However, with further discussion with Tom, we reached a decision to not automatically make GitHub release and Tom will be the one who pulls the trigger through GitHub GUI as he usually does. So, the pull request has been updated to remove the GitHub workflow that automatically creates a release tar ball on GitHub.

As for the naming, I'm not sure if we should call it "4.5.0-rc1" since this is not a release candidate just yet. But I get that it is something along this "4.5.0-{SOMETHING}". I'll leave this to @tom95858 :)

morrone:
I do not think this is the right way to go. I don't think you want to start to have to do the things that the automatic release process does by hand.

This is a fairly one-off situation. So just make a tag like "v4.5.0-rc1" and if you _really_ don't want the tag to appear under the github "releases" section, just go delete that release. Its a fairly one-off situation, so a one-off delete of the release seems pretty simple. Whereas release tags creating releases would tend to be the _normal_ behavior, so let's not change the automation about that.

PR #1409, opened 2024-06-25
Resolve segfault when using deprecated producer type "passive"

nick-enoent:
Check that transport exists before attempting to assign event callback
when using deprecated producer type passive


tom95858:
@nick-enoent please re-base this request.

nick-enoent:
> @nick-enoent please re-base this request.

Rebased.

PR #1411, opened 2024-07-01
Stream-Rail distribution with addr + port

narategithub:
When the stream forward the stream data, it used the peer address to determine the endpoint in the rail to forward the data. This patch adds `port` into the calculation as well to increase the stream data distribution among the endpoints in the rail.

### DIFF
--- a/ldms/src/core/ldms_stream.c
+++ b/ldms/src/core/ldms_stream.c
@@ -70,6 +70,7 @@
 
 #include "ovis_json/ovis_json.h"
 #include "coll/rbt.h"
+#include "coll/fnv_hash.h"
 #include "ovis_ref/ref.h"
 #include "ovis_log/ovis_log.h"
 
@@ -221,6 +222,7 @@ static int __part_send(struct ldms_rail_ep_s *rep,
 }
 
 static int __rep_publish(struct ldms_rail_ep_s *rep, const char *stream_name,
+			uint32_t hash,
                         ldms_stream_type_t stream_type,
 			struct ldms_addr *src, uint64_t msg_gn,
 			ldms_cred_t cred, int perm,
@@ -254,6 +256,7 @@ static int __rep_publish(struct ldms_rail_ep_s *rep, const char *stream_name,
 	msg.cred.uid = htobe32(cred->uid);
 	msg.cred.gid = htobe32(cred->gid);
 	msg.perm = htobe32(perm);
+	msg.name_hash = hash;
 	rc = __part_send(rep, &msg.src, msg_gn,
 			 &msg, sizeof(msg), /* msg hdr */
 			 stream_name, name_len, /* name */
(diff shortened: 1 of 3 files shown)


narategithub:
@tom95858 it failed the compatibility test due to IPv6 issue in the test script that was addressed in #1408 . Once that pull request is merged, I'll rerun the test in this PR.

narategithub:
@tom95858 Shall I also include the stream name into the calculation to increase the distribution?

narategithub:
I'll add `name` into the calculation too. I'll mark this as `draft` for now.

tom95858:
@narategithub what's the status of this?

narategithub:
@tom95858 the code is done but the `ldms_stream_test` script failed. I'm debugging it.

narategithub:
@tom This is ready for your review.

ldms/src/core/ldms_stream.c
```diff
@@ -254,6 +256,7 @@ static int __rep_publish(struct ldms_rail_ep_s *rep, const char *stream_name,
 	msg.cred.uid = htobe32(cred->uid);
 	msg.cred.gid = htobe32(cred->gid);
 	msg.perm = htobe32(perm);
+	msg.name_hash = htobe32(hash);
```
tom95858, line 259: Nit: Hashes's don't need to be byte-swapped. The are not interpreted by the CPU like integers. This applies to the logic below as well, e.g. even when using the port_no as part of the hash-key, it's just a string of bits at that point.

ldms/src/core/ldms_stream.c
```diff
@@ -478,6 +491,7 @@ __stream_deliver(struct ldms_addr *src, uint64_t msg_gn,
 			.name_len = name_len,
 			.data_len = data_len,
 			.name = stream_name,
+			.name_hash = hash,
```
tom95858, line 494: If you're going to put the hash in the message, you could put the entire thing in there and avoid the computation on every message. The port and addr aren't changing. This would also mean that the same hash is used all the way up the chain from sender to consumer.

narategithub: The address got changed in the following case.

There is a logic to exclude the use of `local` address (e.g. `127.0.0.1`) as `src` address here:

https://github.com/ovis-hpc/ovis/blob/f71e7adcff4a852f3ccc05b143a7794a6bc71986/ldms/src/core/ldms_stream.c#L1131-L1153

This is to prevent the case that the application running on the same host connects to sampler ldmsd using `localhost` address and make all `src` in the messages from all nodes being `127.0.0.1`. Basically, if the `src` address is local, do not resolve it and let the next guy resolve it instead.

tom95858: So there must be code somewhere else that checks the address and family and if 0 sets it to the address and port of the forwarder? Where is that logic?

tom95858: @narategithub In any event, all we're doing here is randomizing the seed to what we hope will be a random hash. So as-usual, we've found ourselves in a hole so we keep digging. Let's just back up and decide how we're going to generate a random hash.

narategithub:
Closing this PR. This will be a part of adaptive flow control PR.

PR #1412, opened 2024-07-03
Make ldmsd_controller exit on connection error

nichamon:
The patch also updates the connection error message to include the transport and the authentication method.


nichamon:
@tom95858 I cherry-picked this from OVIS-4.

PR #1416, opened 2024-07-10
Workaround for GitHub workflow dropping CentOS7 support

narategithub:
CentOS7 is old, and GitHub Runner recently uses 'node20' which requires newer GLIBC that does not exist in CentOS7. As a result, the three workflows in our repository that use 'ovis-centos-build' image fails to even checkout the code since `actions/checkout@v3` was implemented in TypeScript that run using 'node.js'.

According to this:
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/ CentOS7 is also removed from the supported OS list due to node20 not supported on CentOS7.

This patch does the following on the CentOS7-based workflows:

- Use `actions/checkout@v1` (instead of v3) because it is implemented as a plugin that comes with the GitHub Runner (C++). We do not use any advanced features that does not exist in v1.

- Remove the use of `actions/upload-artifact`. v3 is implemented in TypeScript, and v1 (C++ Runner Plugin) does not work. The artifact's purpose was for debugging.

### DIFF
--- a/.github/workflows/4.3.3-compat-test.yaml
+++ b/.github/workflows/4.3.3-compat-test.yaml
@@ -41,7 +41,7 @@ jobs:
     container:
         image: ovishpc/ovis-compat-centos7
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v1
     - run: sh autogen.sh
     - name: build and install
       run: |
@@ -332,7 +332,3 @@ jobs:
         for V in 4.3.{3..6}; do
           diff -u /test/${V}.enum /test/HEAD-${V}.enum > /test/enum-${V}.diff
         done
-    - uses: actions/upload-artifact@v3
-      if: ${{ always() }}
-      with:
-        path: /test/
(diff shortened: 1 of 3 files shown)


baallan:
We anticipate retiring our last rhel7 systems this calendar year, so we're probably good to drop centos7 if there is an equivalent ancient ubuntu using gcc 4.8 still available in the interim.

tom95858:
@narategithub not sure if I'm doing this correctly, but restarting the workflows on the pull requests still results in failure.

narategithub:
@tom95858 The existing pull requests have to rebase to include this commit. I'll rebase mine later today.

narategithub:
> @narategithub why don't we just remove the Centos7 workflows?

@tom95858 I think we can remove the build tests on CentOS7. We still need 4.3.11 compatibility test though. I'll check with Jim on what distro we should test this on.

@valleydlr Do you know what distro we should use for v4.3.11 compatibility test?

PR #1418, opened 2024-07-11
Fix a leak in store_sos.c

nichamon:


nichamon:
@tom95858 Here is the patch for the top of tree. Thank you for finding the leak cause.

PR #1422, opened 2024-07-16
Correct rails, credits, rx_rate values in advertised producers

nichamon:


ldms/man/ldmsd_sampler_advertisement.man
```diff
@@ -32,7 +32,7 @@ ldmsd_sampler_advertisement - Manual for LDMSD Sampler Advertisement
 
 .IP \fBprdcr_listen_add
 .RI "name=" NAME "
-.RI "[disabled_start=" TURE|FALSE "] [regex=" REGEX "] [ip=" CIDR "] [rail=" SIZE "] [credits=" BYTES "] [rx_rate=" RATE_LIMIT "]"
+.RI "[disable_start=" TURE|FALSE "] [regex=" REGEX "] [ip=" CIDR "]"
```
tom95858, line 35: TURE --> TRUE
Do we want the spaces around the values? A a user that copied that would get an error; and likely a very confusing one.

ldms/man/ldmsd_sampler_advertisement.man
```diff
@@ -173,13 +187,16 @@ aggregator (\fBagg\fR) runs on \fBnode-3\fR. All LDMSD listen on port 411.
...
+\fBnode-3\fR at port 411. They will try to reconnect to the aggregator every 10
+seconds until the connection is established. The following are the configuration
+files of the \fBsamplerd-1\fR and \fBsamplerd-2\fR.
 
 .EX
 .B
 > cat samplerd-1.conf
 .RS 4
+# Create a listening endpoint
+listen xprt=sock port=411
```
tom95858, line 199: This isn't strictly needed for the advertiser function, correct? What is the reasoning for including it in the example?

tom95858 (approved):
We need to discuss how the ADVERTISED producer name is determined. We will have to make this configurable in the future.

PR #1432, opened 2024-08-15
Fix decomp operations (b4.4. branch)

morrone:
There are multiple fixes and improvements in this PR.

The bug that instigated this work is fixed in the commit titled "ldmsd_row_cache: fix tree_comparator() function".

Also notable is the fix for the timestamp calculations in commit "decomp_static: correct timestamp calculations".


morrone:
@tom95858 All of the changes should be reviewed, of course, but it would be especially good to double check my change in the commit "decomp_static: call ldmsd_row_cache_make_list() just once". So far testing bears out that my understanding was correct, but it would be good for you to verify that I didn't miss anything there.

PR #1433, opened 2024-08-15
Fix decomp operations (OVIS-4 branch)

morrone:
This is the OVIS-4 branch version of PR #1432 (which is for b4.4).

### DIFF
--- a/ldms/src/decomp/static/decomp_static.c
+++ b/ldms/src/decomp/static/decomp_static.c
@@ -294,14 +294,6 @@ decomp_static_release_decomp(ldmsd_strgp_t strgp)
 	}
 }
 
-static int init_row_cache(ldmsd_strgp_t strgp, int row_limit)
-{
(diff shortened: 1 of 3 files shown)


ldms/src/decomp/static/decomp_static.c
```diff
@@ -1785,22 +1805,24 @@ static int decomp_static_decompose(ldmsd_strgp_t strgp, ldms_set_t set,
 			dup_row = row_cache_dup(cfg_row, mid_rbn, row);
 
 			/* Apply functional operators to columns */
+                        struct ldmsd_row_list_s tmp_row_list;
+                        int count;
+                        count = ldmsd_row_cache_make_list(&tmp_row_list,
```
tom95858, line 1810: @morrone, the reason it was inside the col loop is to support operators that work with multiple columns, e.g. +, -, /. But we don't have any of those yet, so this change is fine.

tom95858:
@morrone, thanks for fixing this.

PR #1435, opened 2024-08-20
Fix null reference in ldms_xprt.c

nick-enoent:
Move ref_put before assigning local set to NULL to avoid referencing a NULL address. Resolves segfault.


nick-enoent:
@tom95858 Will you check this?

nichamon (approved):

PR #1437, opened 2024-08-26
Fix `ldmsd_controller` completion without ldmsd connection

narategithub:
When there is no `ldmsd` connection, `ldmsd_controller` command parameter completion does not work. This patch fixes the issue.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_communicator.py
+++ b/ldms/python/ldmsd/ldmsd_communicator.py
@@ -226,6 +226,24 @@ LDMSD_CTRL_CMD_MAP = {'usage': {'req_attr': [], 'opt_attr': ['name']},
                       'prdcr_listen_status': {'req_attr': [], 'opt_attr': []},
                       }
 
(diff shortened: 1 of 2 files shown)


narategithub:
@nick-enoent I cannot assign you as a reviewer .. So here's a review invitation :)

ldms/python/ldmsd/ldmsd_communicator.py
```diff
@@ -226,6 +226,24 @@
                       'prdcr_listen_status': {'req_attr': [], 'opt_attr': []},
                       }
 
+def get_cmd_attr_list(cmd_verb):
+    """Return the dictionary of command attributes
+
+    If there are no required/optional attributes, the value of the
+    'req'/'opt' key is None. Otherweise, the value is a list of attribute
```
tom95858, line 233: @narategithub nit: otherwise

narategithub: Whoops. Thank you, @tom95858, for catching this.

PR #1438, opened 2024-08-26
[b4.4] Fix `ldmsd_controller` completion without ldmsd connection

narategithub:
When there is no `ldmsd` connection, `ldmsd_controller` command parameter completion does not work. This patch fixes the issue.


narategithub:
Same as #1437, but for `b4.4`.

PR #1439, opened 2024-08-27
Fix incorrect remote hostname and port produced by ldms_xprt_names()

nichamon:
Previosly, ldms_xprt_names() incorrectly passed the local address to getnameinfo() to obtain the remote hostname and port. This resulted in incorrect output.


nichamon:
Symptoms:

- Same as in b4.4 (#1440)
- `xprt_stats` could report an incorrect total send-queue dept.

PR #1442, opened 2024-09-05
Interpose rail update callback to inject rail handle during update

nichamon:
Ensure that the update operation is handled by `__rail_update()`, allowing the rail handle to be delivered upon update completion


narategithub (approved):
Looks good to me. Thank you, @nichamon !.

PR #1444, opened 2024-09-13
Fix zap_sock send/send2 data length check

narategithub:
`zap_sock` did not check the send data length. It is added by this patch.


tom95858:
@narategithub Did this break something or is this something found by inspection?

PR #1447, opened 2024-09-18
Update LDMS Variorum sampler to support Variorum v0.8

### LINKED ISSUE #1431: Update Variorum sampler to support v0.8
Variorum v0.8 has updated the JSON API for obtaining vendor-neutral power telemetry.
We will update the Variorum sampler in LDMS to use the new API.

tpatki:
Work in Progress, not ready for review.

Fixes #1431. 

This PR updates the JSON API and the parsing of power data based on Variorum 0.8.

ToDo: 
- [x] Code Formatting
- [x] Documentation updates
- [ ] Squash commits 
- [ ] Autoconf checks
- [ ] Testing (see below)

Testing:
- [x] IBM Alehouse node, CPU+GPU build
- [ ] IBM Alehouse node, GPU-build only
- [ ] IBM Alehouse node, CPU-build only
- [ ] Other LLNL systems with @kshoga1 and @slabasan.

### DIFF
--- a/ldms/src/contrib/sampler/variorum_sampler/Plugin_variorum_sampler.man
+++ b/ldms/src/contrib/sampler/variorum_sampler/Plugin_variorum_sampler.man
@@ -19,7 +19,7 @@ GPU power consumption in Watts (aggregated across all GPUs on the socket, and
 reported as -1 on unsupported platforms); and memory power consumption in Watts.
 
 .PP
-The variorum sampler depends on Variorum 0.6.0 or higher and Jansson. The sampler cannot be built without these libraries. If either library is installed in a non-standard location, paths to the respective install directories should be provided to Autoconf using
+The variorum sampler depends on Variorum 0.8.0 or higher and Jansson. The sampler cannot be built without these libraries. If either library is installed in a non-standard location, paths to the respective install directories should be provided to Autoconf using
 the --with-libjansson-prefix and/or --with-libvariorum-prefix flag.
 
 .SH CONFIGURATION ATTRIBUTE SYNTAX
--- a/ldms/src/contrib/sampler/variorum_sampler/README.md
+++ b/ldms/src/contrib/sampler/variorum_sampler/README.md
@@ -9,7 +9,7 @@ architecture and implementation.
 Build Requirements
 ------------------
 
-The Variorum LDMS sampler currently requires version 0.6.0 or higher
+The Variorum LDMS sampler currently requires version 0.8.0 or higher
 of the Variorum library (``libvariorum.so``). This library must be built
 from source. The sampler also requires jansson, which is a Variorum
 dependency. If both libraries are installed in standard locations,
@@ -52,7 +52,7 @@ Using the Variorum LDMS Sampler
 
 The sampler, when configured, automatically detects the number of sockets
 on the host machine and then provides, for each socket, an LDMS record
-containing power data. The sampler calls ``variorum_get_node_power_json``
+containing power data. The sampler calls ``variorum_get_power_json``
 internally, for which documentation can be found here:
 [Variorum JSON-Support Functions](https://variorum.readthedocs.io/en/latest/api/json_support_functions.html)
 
--- a/ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
+++ b/ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
@@ -218,7 +218,7 @@ static int sample(struct ldmsd_sampler *self)
                 ldms_record_set_double(rec_idxs[socket], i_node, power_node);
 
                 // Obtain Socket Object
-                snprintf(socketID, 20, "socket_%d", i);
+                snprintf(socketID, 20, "socket_%d", socket);
                 json_t *socket_obj = json_object_get(node_obj, socketID);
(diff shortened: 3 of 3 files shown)


morrone:
@tpatki Could you also please this squash this PR together into just a small set of clean commits? 

You will also need to include appropriate autoconf checks since this will now only work with variorum 0.8 or newer.

tpatki:
@morrone 
Yes, will squash into clean commits when we're done testing and mark this as ready for review (will be a few weeks until @kshoga1 can test on Mutt.)

Will update the autoconf checks as well, thanks for catching that!

morrone:
> Could you please clarify this? We were not checking for a specific version in `configure.ac` [here](https://github.com/ovis-hpc/ldms/blob/8beef09a1100b35965976f77ba5ed24cfc039e17/configure.ac#L374-L391) earlier. Where/how should we include the autoconf checks (is there an example we can follow)?

Better than a specific version check is to test for _features_ in the library that the code requires. For instance, are there some unique new function in the new version of the library? You could add a test for that function in the library.

LDMS's configure.ac has an AC_LIB_HAVE_LINKFLAGS test for variorum. There is an additional "testcode" field for that macro where you could put a tiny bit of C code that calls a function that is unique to the newer version of the code.

tom95858:
@tpatki, @morrone what is the status of this PR?

tpatki:
Hi @tom95858 : This is draft PR at the moment. Kathleen (@kshoga1) will need to test this at scale internally. We plan to do that in this month. Once we have tested at scale, we'll address Chris' comments, update it as ready for review, and it can be merged.

tom95858:
@tpatki Ok great, thanks for the update.

tom95858:
@morrone, please confirm that this is not a target for 4.4.5.

morrone:
> @morrone, please confirm that this is not a target for 4.4.5.

Except for reviewing here, I'm not really involved in that work. But my understanding is that it'll happen when it happens, and we should not hold up the release for this work.

tom95858:
@tpatki please reopen this pull request when it is ready for review.

tpatki:
Hi @morrone :

@kshoga1 tested this on an LC system this afternoon, and we can confirm that the plugin builds correctly on LC systems and the power data reported is correct. 

We also discussed your suggestion of adding a small testcode with `AC_LIB_HAVE_LINKFLAGS` -- but we realized that this doesn't make sense for Variorum as we don't have a small-enough check that can specifically test for version 0.8. What has changed is not the Variorum API, but how to data is reported (nested JSON as opposed to a non-nested format; the older format is not supported any more). This changes how the data is parsed in any tools that Variorum integrates with, e.g. the LDMS Variorum sampler. 

If you have other suggestions on how we can do a version check, let us know. If not, this PR is ready for your review and ready to go.

morrone:
@tpatki This PR was opened against OVIS-4, which we've done away with. We now use the more standard "main" branch name for the main development branch. github won't let me reopen either becasue OVIS-4 doesn't exist.

I would recommend just opening a new PR.

On the topic of version checking:

It looks like variorum supplies a variorum.pc file, which is great! But at least as of 0.6.0 it looks like that file is installed in the wrong path on the system. It looks like currrently it is going in /usr/share/, and I think it needs to be in ${libpath}/pkgconfig/ or /usr/share/pkgconfig/

If that can be fixed in variorum 0.8's rpm, then the following patch might meet our needs.

```
diff --git a/configure.ac b/configure.ac
index 4eb068da0..9b4a3ea14 100644
--- a/configure.ac
+++ b/configure.ac
@@ -388,6 +388,11 @@ AC_ARG_ENABLE([variorum],
     [],
     [enable_variorum="check"])
 AS_IF([test "x$enable_variorum" != xno],[
+    PKG_CHECK_MODULES([variorum], [variorum >= 0.8], [variorum_version_check=yes], [variorum_version_check=no])
+    AS_IF([test "x$enable_variorum" != xcheck],[
+        AS_IF([test "x$variorum_version_check" = xno],
+            [AC_MSG_ERROR([variorum pkgconfig file not found, or variorum version is not >= 0.8])])
+    ])
     AC_LIB_HAVE_LINKFLAGS([variorum], [], [#include <variorum.h>
                            #include <variorum_topology.h>])
     AS_IF([test "x$enable_variorum" != xcheck],[
@@ -399,7 +404,7 @@ AS_IF([test "x$enable_variorum" != xno],[
             [AC_MSG_ERROR([libjansson or headers not found])])
     ])
 ])
-AM_CONDITIONAL([ENABLE_VARIORUM_SAMPLER], [test "x$enable_variorum" != no -a "x$HAVE_LIBVARIORUM" = xyes -a "x$HAVE_LIBJANSSON" = xyes])
+AM_CONDITIONAL([ENABLE_VARIORUM_SAMPLER], [test "x$enable_variorum" != no -a "x$HAVE_LIBVARIORUM" = xyes -a "x$variorum_version_check" = xyes -a "x$HAVE_LIBJANSSON" = xyes])

 dnl check for libcurl if influx is configured
 OPTION_DEFAULT_DISABLE([influx], [ENABLE_INFLUX])

```

tpatki:
Hi @morrone Noted the above, we will take a look in early March once @slabasan and I are back in office, update and then and re-open the PR like you suggested.

PR #1456, opened 2024-09-27
Fix static decomposition

narategithub:
`ldms-test/ldmsd_decomp_test` was failing and the following bugs were discovered:
- "type" from the configuration was ignored,
  - "fill" was not working because "type" was ignored,
- the ARRAY values were ignored.

This pull request addressed the discovered issued as well as keeping the destination rows consistent within `ldmsd` process life span as follows:
- The type of a destination column is stored in `cfg_col->type`.
  - If "type" of a column is specified in the config file, `cfg_col->type` is set
    to the specified "type".
  - Otherwise, `cfg_col->type` is discovered from the `metric` in the `set` being
    processed. If it cannot be resolved, the row is skipped.
- If the type of the metric from `src` (e.g. from `set`) is not the same as
  the same type of the `dst`, the value will be coerced. The array values are
  coerced element-by-element.

(old texts)
~~In addition, four `ldms_mval` interfaces were also introduced:~~
- ~~`ldms_mval_set_int()`~~
- ~~`ldms_mval_set_real()`~~
- ~~`ldms_mval_array_set_int()`~~
- ~~`ldms_mval_array_set_real()`~~
~~The APIs received the biggest size of integer or real number. The conversion of values are handled by the compiler.~~

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -5248,3 +5248,734 @@ int ldms_getsockaddr(const char *host, const char *port,
 	struct addrinfo hints = { .ai_socktype = SOCK_STREAM };
 	return __getsockaddr(host, port, &hints, sa, sa_len);
 }
+
+#define _FROM_TYPE_BODY(v, t, i) \
+	switch (t) { \
+	case LDMS_V_CHAR: \
+		return (v)->v_char; \
+	case LDMS_V_U8: \
+		return (v)->v_u8; \
+	case LDMS_V_S8: \
+		return (v)->v_s8; \
+	case LDMS_V_U16: \
+		return (v)->v_u16; \
+	case LDMS_V_S16: \
+		return (v)->v_s16; \
+	case LDMS_V_U32: \
+		return (v)->v_u32; \
+	case LDMS_V_S32: \
+		return (v)->v_s32; \
+	case LDMS_V_U64: \
+		return (v)->v_u64; \
+	case LDMS_V_S64: \
+		return (v)->v_s64; \
+	case LDMS_V_F32: \
+		return (v)->v_f; \
+	case LDMS_V_D64: \
+		return (v)->v_d; \
+	case LDMS_V_CHAR_ARRAY: \
+		return (v)->a_char[i]; \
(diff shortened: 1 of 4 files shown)


tom95858:
I think we need to rethink this. We don't need the type because it can be inferred from the destination mval. That's why the decomp parsing logic was moved until after we had the 1st set handle.

tom95858:
@narategithub Can you point me at the test failures and I will take a look?

tom95858:
@narategithub sorry nevermind the test case is in the pull request

tom95858:
@narategithub can you remind me how to set up the containers

narategithub:
@tom95858 FYI, this change won't require "type". If the config does not specify "type", it will be inferred. However, if "type" is specified in the config, the top-of-tree code ignored it. This patch just changed it to take the "type" if it is in the config.

I'll email you when I finished setting up the test on cygnus (it is building ...).

narategithub:
The array value assignment was posted in #1457 . I convert this PR to Draft to keep it open for discussion.

narategithub:
@tom95858 This is ready for your review.

narategithub:
This PR has been updated to includes "fill" patches #1500 and #1499 (since it depends on a new interface `ldms_mval_as_u64()`). I've tested this with the following test cases from `ldms-test` (branch `master`):
- [ldmsd_decomp_test](https://github.com/ovis-hpc/ldms-test/blob/master/ldmsd_decomp_test) (updated to include all "fill" values)
- [ldmsd_decomp_no_fill_test](https://github.com/ovis-hpc/ldms-test/blob/master/ldmsd_decomp_no_fill_test)
- [ldmsd_decomp_static_omit_test](https://github.com/ovis-hpc/ldms-test/blob/master/ldmsd_decomp_static_omit_test)
- [ldmsd_flex_decomp_test](https://github.com/ovis-hpc/ldms-test/blob/master/ldmsd_flex_decomp_test)

tom95858:
@narategithub is this pull request obsolete?

narategithub:
@tom95858 this is up-to-date for `main`. The ones you merged previously were for b4.4

PR #1457, opened 2024-09-30
Fix static decomposition array value assignment

narategithub:
Array values were not handled in static decomposition value assignment.

### DIFF
--- a/ldms/src/decomp/static/decomp_static.c
+++ b/ldms/src/decomp/static/decomp_static.c
@@ -1030,6 +1030,18 @@ static void assign_value(ldms_mval_t dst, ldms_mval_t src,
 	case LDMS_V_CHAR_ARRAY:
 		memcpy(dst->a_char, src->a_char, count);
 		break;
+	case LDMS_V_U8_ARRAY:
+	case LDMS_V_S8_ARRAY:
+	case LDMS_V_U16_ARRAY:
+	case LDMS_V_S16_ARRAY:
+	case LDMS_V_U32_ARRAY:
+	case LDMS_V_S32_ARRAY:
+	case LDMS_V_U64_ARRAY:
+	case LDMS_V_S64_ARRAY:
+	case LDMS_V_F32_ARRAY:
+	case LDMS_V_D64_ARRAY:
+		memcpy(dst, src, ldms_metric_value_size_get(type, count));
+		break;
 	default:
 		break;
 	}


tom95858:
Are we saying that array assignment for anything other than strings never worked?

ldms/src/decomp/static/decomp_static.c
```diff
@@ -1030,6 +1030,18 @@ static void assign_value(ldms_mval_t dst, ldms_mval_t src,
...
+	case LDMS_V_S8_ARRAY:
+	case LDMS_V_U16_ARRAY:
+	case LDMS_V_S16_ARRAY:
+	case LDMS_V_U32_ARRAY:
+	case LDMS_V_S32_ARRAY:
+	case LDMS_V_U64_ARRAY:
+	case LDMS_V_S64_ARRAY:
+	case LDMS_V_F32_ARRAY:
+	case LDMS_V_D64_ARRAY:
+		memcpy(dst, src, ldms_metric_value_size_get(type, count));
```
tom95858, line 1043: @narategithub it seems like we make no assurances that dst, and src are the same type. Which means that we could be copying garbage. The functions that call this, however, have this information.

nichamon:
@narategithub Is the problem also in b4.4 branch?

ldms/src/decomp/static/decomp_static.c
narategithub: I'm confused. It looks to me like all of the call sites use `assign_value()` to copy `src` to `dst` of the same type. Are you suggesting that `assign_value()` should have `src_type` and `dst_type` parameters instead of just `type`?

narategithub:
If I remember correctly, "op" does not support array. This patch does not make "op" support array. It just fix the bug that the array values are ignored even without "op".

narategithub:
> @narategithub Is the problem also in b4.4 branch?

yes. Let's have a conclusive solution for OVIS-4, then I'll port it back to b4.4.

nichamon:
> > @narategithub Is the problem also in b4.4 branch?
> 
> yes. Let's have a conclusive solution for OVIS-4, then I'll port it back to b4.4.

Thanks for the confirmation. Yes, I agree with your plan.

ldms/src/decomp/static/decomp_static.c
tom95858: @narategithub, this is __in__ assign_value(). I'm suggesting that assign_value() may be broken.

narategithub:
This pull request is deprecated. Please go to #1456.

tom95858:
@narategithub is this pull request replaced by another one? If so, please close this one.

PR #1458, opened 2024-10-03
Fix `dump_cfg` producing deprecated CLI options

narategithub:
`dump_cfg` ldmsd command produced deprecated CLI options (e.g. -C, -P). This patch removed the deprecated CLI options and produced commands in config file instead.


nichamon:
@narategithub Just out of curiosity, does the `dump_cfg` dump a CLI with only a `-c` option?

narategithub:
> @narategithub Just out of curiosity, does the `dump_cfg` dump a CLI with only a `-c` option?

No. It still produces '-l {LOG_FILE}' or '-r {PID_FILE}' if they were given.

nichamon:
> > @narategithub Just out of curiosity, does the `dump_cfg` dump a CLI with only a `-c` option?
> 
> No. It still produces '-l {LOG_FILE}' or '-r {PID_FILE}' if they were given.

Thanks for confirming this.

nichamon (approved):

PR #1460, opened 2024-10-08
Introduce LDMS Profiling Capability

nichamon:


nichamon:
The PR includes #1463 because LDMS could crash without it.

ldms/src/core/ldms.c
```diff
@@ -100,9 +100,83 @@ const char *ldms_xprt_op_names[] = {
...
+	0,          /* update */
+	-ENOSYS,    /* Publish */
+	0,          /* set_delete */
+	-ENOSYS,    /* dir_req */
+	-ENOSYS,    /* dir_rep */
+	0,          /* send */
+	-ENOSYS,    /* receive */
+	0,          /* stream_publish */
+	-ENOSYS,    /* stream_subscribe */
+	-ENOSYS	    /* stream_unsubscribe */
```
tom95858, line 121: Why are these error codes negative?

nichamon: I made the error code negative because I used 1 as enabled and 0 as disabled. I'll add a comment that includes the meanings of 0 and 1.

tom95858 (changes requested):

tom95858:
@nichamon, this is a convention that is used throughout the kernel for interfaces that return pointers to determine whether it's an error value or a valid pointer. In the kernel everyone uses the same convention. In our logic, anyone interpreting the return value would have to do so by interface. In other words, it is not consistent. In user mode, the convention for interfaces returning integers: -1 means error, and then the application consults errno for more information.

ldms/src/core/ldms_rail.c
```diff
@@ -86,13 +89,14 @@ static int __rail_sockaddr(ldms_t _r, struct sockaddr *local_sa,
 	       struct sockaddr *remote_sa,
 	       socklen_t *sa_len);
 static void __rail_close(ldms_t _r);
-static int __rail_send(ldms_t _r, char *msg_buf, size_t msg_len);
+static int __rail_send(ldms_t _r, char *msg_buf, size_t msg_len,
```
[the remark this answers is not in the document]
tom95858: @nichamon, by that logic every non-static symbol inside a .c file would be proceeded by an __, and that is not descriptive at all in my opinion. BTW, I am guilty of this too ... and recently.

ldms/python/ldmsd/ldmsd_communicator.py
```diff
@@ -616,6 +617,7 @@ class LDMSD_Request(object):
     SET_SEC_MOD = 0x600 + 19
     LOG_STATUS = 0x600 + 20
     STATS_RESET = 0x600 + 21
+    PROFILING = 0x600 + 31
```
tom95858, line 620: @nichamon, is there some reasons we are bumping by 31 instead of 22?

nichamon: > @nichamon, is there some reasons we are bumping by 31 instead of 22?

It's the same number defined in [ldmsd_request.h:168](https://github.com/nichamon/ovis/blob/a77173a37fbd456183f6095f4dd0121532986ea7/ldms/src/ldmsd/ldmsd_request.h#L168) in this change.

nichamon: @tom95858 I added the comment to explain why the ID skips the range.

PR #1461, opened 2024-10-08
Yaml support

nick-enoent:
Support for YAML configuration files for LDMSD.
YAML configuration file allows configuration of an entire cluster in a single YAML file.
Add -y option to LDMSD CLI that loads a YAML configuration file in the same manner as "-c" for LDMS v4 configuration files.
The -y option parses a YAML cluster file and applies relevant configuration to a specific daemon.
New command ldmsd_yaml_parser command that can generate v4 configuration files from a YAML file.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_yaml_parser
+++ b/ldms/python/ldmsd/ldmsd_yaml_parser
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+import os, sys
+import yaml
+import errno
+import argparse
+from ldmsd.parser_util import *
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="LDMS Monitoring YAML Cluster Configuration Parser")
+    required = parser.add_argument_group("required arguments")
+    required.add_argument("--ldms_config", metavar="FILE", required=True,
+                        help="Path to LDMSD YAML configuration file. ")
+    parser.add_argument("--generate-config-path", metavar="STRING", required=False,
+                        help="Path to directory to store generated v4 configuration files for an entire LDMS cluster. "
+                             "Samplers with similar configurations share a single configuration file. ",
+                        default=False)
+    parser.add_argument("--daemon_name", metavar="STRING", required=False,
+                        default=False,
+                        help="Daemon name to generate configuration from YAML file")
+    parser.add_argument("--debug", action="store_true",
+                        help="Enable debug information")
+    args = parser.parse_args()
+    if not args.debug:
+        import sys
+        sys.tracebacklimit=0
+    config_fp = open(args.ldms_config)
+    conf_spec = yaml.safe_load(config_fp)
+
+    cluster = YamlCfg(None, None, conf_spec, args)
+
+    if args.daemon_name:
+        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
+        print(f'{ldmsd_cfg_str}')
+        sys.exit(0)
+
+    if args.generate_config_path:
+        cluster.config_v4(args.generate_config_path)
+        print("LDMSD v4 config files generated")
+        sys.exit(0)
+
+    if not args.generate_config_path and not args.ldms_config and not args.daemon_name:
+        print(f'No action detected. Exiting...')
+
+    sys.exit(0)
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -1,0 +1,1088 @@
+import os, sys
+import errno
+import json
+import subprocess
+import socket
+import time
+import itertools as it
+import collections
+import ldmsd.hostlist as hostlist
+
+AUTH_ATTRS = [
+    'auth',
+    'conf'
+]
+
+CORE_ATTRS = [
+    'daemons',
+    'aggregators',
+    'samplers',
+    'plugins',
+    'stores'
+]
+
+DEFAULT_ATTR_VAL = {
+    'maestro_comm' : False,
+    'xprt'         : 'sock',
+    'interval'     : 1000000,
+    'auth'         : 'none',
+    'mode'         : 'static'
+}
+
+INT_ATTRS = [
+    'interval',
+    'offset',
+    'reconnect',
+    'flush'
+]
+
+unit_strs = {
+    'ms' : 1000,
+    'us' : 1,
+    'm' : 60000000,
+    's' : 1000000,
+    'h' : 3600000000,
+    'd' : 86400000000
+}
+LDMS_YAML_ERR = 'Error parsing ldms_config yaml file'
+LIST_ERR = 'spec must be a list of dictionaries, specified with "-" in the ldms_config yaml file'
+DICT_ERR = 'spec must be a dictionary, with keys referencing a plugin instance name'
+
+def check_intrvl_str(interval_s):
+    """Check the format of the interval string is correct
+
+    A time-interval string is an integer or float followed by a
+    unit-string. A unit-string is any of the following:
+
+    'us' - microseconds
+    'ms' - milliseconds
+    's'  - seconds
+    'm'  - minutes
+    'h'  - hours
+    'd'  - days
+
+    Unit strings are not case-sensitive.
+
+    Examples:
+    '1.5s' - 1.5 seconds
+    '1.5S' - 1.5 seconds
+    '2s'   - 2 seconds
+    """
+    error_str = f"{interval_s} is not a valid time-interval string\n"\
+                f"'Only a single unit-string is allowed. e.g. '50s40us' is not a valid entry."\
+                f"Examples of acceptable format:\n"\
+                f"'1.5s' - 1.5 seconds\n"\
+                f"'1.5S' - 1.5 seconds\n"\
+                f"'2us'  - 2 microseconds\n"\
+                f"'3m'   - 3 minutes\n"\
+                f"'1h'   - 1 hour\n"\
+                f"'1d'   - 1 day\n"\
+                f"\n"
+    if type(interval_s) == int or type(interval_s) == float:
+        return interval_s
+    if type(interval_s) != str:
+        raise ValueError(f'{error_str}')
+    interval_s = interval_s.lower()
+    unit = next((unit for unit in unit_strs if unit in interval_s), None)
+    if unit:
+        if interval_s.split(unit)[1] != '':
+            raise ValueError(f'{error_str}')
+        ival_s = interval_s.split(unit)[0]
+        try:
+            ival_s = float(ival_s) * unit_strs[unit]
+        except Exception as e:
+            raise ValueError(f'"{interval_s}" is not a valid time-interval string\n')
+    else:
+        ival_s = interval_s
+    return int(ival_s)
+
+def check_opt(attr, spec):
+    # Check for optional argument and return None if not present
+    if attr in AUTH_ATTRS:
+        if attr == 'auth':
+            attr = 'name'
+        if 'auth' in spec:
+            spec = spec['auth']
+    if attr in spec:
+        if attr in INT_ATTRS:
+            return check_intrvl_str(spec[attr])
+        return spec[attr]
+    else:
+        if attr in DEFAULT_ATTR_VAL:
+            return DEFAULT_ATTR_VAL[attr]
+        else:
+            return None
+
+def check_required(attr_list, container, container_name):
+    """Verify that each name in attr_list is in the container"""
+    for name in attr_list:
+        if name not in container:
+            raise ValueError(f'The "{name}" attribute is required in {container_name}\n')
+
+def NUM_STR(obj):
+    return str(obj) if type(obj) in [ int, float ] else obj
+
+def expand_names(name_spec):
+    if type(name_spec) != str and isinstance(name_spec, collections.abc.Sequence):
+        names = []
+        for name in name_spec:
+            names += hostlist.expand_hostlist(NUM_STR(name))
+    else:
+        names = hostlist.expand_hostlist(NUM_STR(name_spec))
+    return names
+
+def check_auth(auth_spec):
+    name = check_opt('auth', auth_spec)
+    if not name:
+        return None, None, None
+    plugin = check_opt('plugin', auth_spec['auth'])
+    auth_opt = check_opt('conf', auth_spec)
+    return name, plugin, auth_opt
+
+def check_plugin_config(plugn, plugin_spec):
+    if plugn not in plugin_spec:
+        raise ValueError(f'Configuration for plugin instance "{plugn}" '
+                         f'is not defined in the top level "plugins" dictionary\n')
+    plugin = plugin_spec[plugn]
+    check_required([ 'name' ], plugin, f'"plugin" entry. Error in "'+ plugn +'" configuration')
+    check_required(['config'], plugin, '"plugin" entry')
+    if type(plugin['config']) is not list:
+        raise TypeError('"config" must be a list of configuration commands\n')
+    for cfg in plugin['config']:
+        if type(cfg) is not dict and type(cfg) is not str:
+            raise TypeError('"config" list members must be a dictionary or a string\n')
+    return plugin
+
+def parse_to_cfg_str(cfg_obj):
+    cfg_str = ''
+    for key in cfg_obj:
+        if key not in INT_ATTRS:
+            if len(cfg_str) > 1:
+                cfg_str += ' '
+            cfg_str += key + '=' + str(cfg_obj[key])
+    return cfg_str
+
+def parse_yaml_bool(bool_):
+    if bool_ is True or bool_ == 'true' or bool_ == 'True':
+        return True
+    else:
+        return False
+
+def perm_handler(perm_str):
+    if perm_str is None or type(perm_str) is int:
+        return perm_str
+    nperm = "0"
+    perm_str = perm_str.split('-')
+    if len(perm_str) == 1:
+        try:
+            z = int(perm_str[0])
+            return perm_str[0]
+        except:
+            raise ValueError(f'Error: permission {perm_str[0]} is not a valid value')
+    if len(perm_str) > 3:
+        raise ValueError(f'There are only 3 definable characteristics of linux permissions; users, group, and global. Please modify {perm_str} to reflect this\n')
+    for uog in perm_str:
+        x = 0
+        if 'r' in uog:
+            x += 4
+        if 'w' in uog or '+' in uog:
+            x += 2
+        if 'x' in uog:
+            x += 1
+        nperm += str(x)
+    return nperm
+
+class YamlCfg(object):
+    def build_daemons(self, config):
+        """Generate a daemon spec list from YAML config
+
+        Builds a dictionary of endpoint definitions. The 'config' is a
+        list of daemon specifications. Each daemon specification contains
+        'names', 'host', and 'endpoint' attributes. All attributes are
+        expanded per the slurm hostlist rules. The length of the
+        expanded name-list, must equal the length of the expanded
+        host-list.
+
+        Example:
+
+        daemons:
+          - names : "agg-[1-10]"
+            hosts : "node[1-10]"
+            endpoints :
+              - names : "node-[1-10]-[10002]
+                ports : "[10002]"
+                maestro_comm : True
+                xprt : sock
+                auth :
+                  name : munge
+                  plugin : munge
+
+        results in the following host-spec dictionary:
+
+        {
+        "agg-[[1-3]-[10002]" : {
+          "agg-1" : {
+            "endpoints": {
+              "node-1-10002" : { "host" : "node-1", "port" : 10002 },
+              "node-2-10002" : { "host" : "node-2", "port" : 10002 },
+              "node-3-10002" : { "host" : "node-3", "port" : 10002 },
+              ...
+            }
+          }
+        }
+
+        """
+        ep_dict = {}
+        node_config = config['daemons']
+        if type(node_config) is not list:
+            raise TypeError(f'{LDMS_YAML_ERR}\n'
+                            f'daemons {LIST_ERR}\n'
+                            f'e.g. daemons:\n'
+                            f'       - names : &l1-agg "l1-aggs-[1-8]"\n'
+                            f'         hosts : &l1-agg-hosts "node-[1-8]"\n')
+        for spec in node_config:
+            check_required([ 'names', 'endpoints', 'hosts' ],
+                           spec, '"daemons" entry')
+            dnames = expand_names(spec['names'])
+            hosts = expand_names(spec['hosts'])
+            hostnames = hosts
+            if len(dnames) != len(hostnames):
+                hosts = [ [host]*(len(dnames)//len(hostnames)) for host in hostnames ]
+                hosts = list(it.chain.from_iterable(hosts))
+            ep_names = []
+            ep_ports = []
+            ep_hosts = []
+            if type(spec['endpoints']) is not list:
+                raise TypeError(f'{LDMS_YAML_ERR}\n'
+                                f'endpoints {LIST_ERR}\n'
+                                f'e.g endpoints :\n'
+                                f'      - names : &l1-agg-endpoints "node-[1-8]-[10101]"\n'
+                                f'        ports : &agg-ports "[10101]"\n'
+                                f'        maestro_comm : True\n'
+                                f'        xprt  : sock\n'
+                                f'        auth  :\n'
+                                f'          name : munge1\n'
+                                f'          plugin : munge\n')
+            for endpoints in spec['endpoints']:
+                check_required(['names','ports'],
+                               endpoints, '"endpoints" entry')
+                cur_epnames = expand_names(endpoints['names'])
+                ep_names.append(cur_epnames)
+                cur_ports = expand_names(endpoints['ports'])
+                _ports = cur_ports
+                if len(cur_ports) != len(cur_epnames):
+                    cur_ports = [ _ports for i in range(0, len(cur_epnames)//len(_ports)) ]
+                    cur_ports = list(it.chain.from_iterable(cur_ports))
+                ep_ports.append(cur_ports)
+                if check_opt('hosts', endpoints):
+                    cur_ephosts = expand_names(endpoints['hosts'])
+                    _ephosts = cur_ephosts
+                    if len(cur_ephosts) != len(cur_epnames):
+                        cur_ephosts = [ [host]*(len(cur_epnames)//len(_ephosts)) for host in _ephosts ]
+                        cur_ephosts = list(it.chain.from_iterable(cur_ephosts))
+                    ep_hosts.append(cur_ephosts)
+                else:
+                    ep_hosts.append([])
+            ep_dict[spec['names']] = {}
+            env = check_opt('environment', spec)
+            cli_opt = {}
+            for key in spec:
+                if key not in ['hosts','names','endpoints','environment']:
+                    cli_opt[key] = spec[key]
+            for dname, host in zip(dnames, hosts):
+                ep_dict[spec['names']][dname] = {}
+                ep_dict[spec['names']][dname]['addr'] = host
+                ep_dict[spec['names']][dname]['environment'] = env
+                ep_dict[spec['names']][dname]['endpoints'] = {}
+                if len(cli_opt):
+                    ep_dict[spec['names']][dname]['cli_opt'] = cli_opt
+                dcount = 0
+                for ep_, ep_port, ep in zip(ep_names, ep_ports, spec['endpoints']):
+                    port = ep_port.pop(0)
+                    ep_name = ep_.pop(0)
+                    xprt = check_opt('xprt', ep)
+                    auth_name = check_opt('auth', ep)
+                    auth_conf = check_opt('conf', ep)
+                    plugin = check_opt('plugin', ep['auth'])
+                    maestro_comm = parse_yaml_bool(check_opt('maestro_comm', ep))
+                    h = {
+                        'name' : ep_name,
+                        'port' : port,
+                        'xprt' : xprt,
+                        'maestro_comm' : maestro_comm,
+                        'auth' : { 'name' : auth_name, 'conf' : auth_conf, 'plugin' : plugin }
+                    }
+                    if check_opt('bind_all', ep):
+                        h['bind_all'] = ep['bind_all']
+                    _ephost = check_opt('hosts', ep)
+                    if _ephost:
+                        h['host'] = ep_hosts[dcount].pop(0)
+                    ep_dict[spec['names']][dname]['endpoints'][ep_name] = h
+                    dcount += 1
+            if len(ep_dict[spec['names']]) == 0:
+                raise ValueError(f'Error processing regex of hostnames {spec["hosts"]} and daemons {spec["names"]}.\n'
+                                 f'Number of hosts must be a multiple of daemons with appropriate ports or equivalent to length of daemons.\n'
+                                 f'Regex {spec["hosts"]} translates to {len(hostnames)} hosts\n'
+                                 f'Regex {spec["names"]} translates to {len(dnames)} daemons\n')
+        return ep_dict
+
+    def build_advertisers(self, spec):
+        if 'advertise' not in spec:
+            return
+        ad_grp = spec['advertise']
+        check_required(['names', 'hosts', 'xprt', 'port', 'reconnect'],
+                       ad_grp, '"advertise" entry')
+        names = expand_names(ad_grp['names'])
+        dmns = expand_names(spec['daemons'])
+        if len(names) != len(dmns):
+            raise ValueError(f'Please provide a regex for "names" that is equal to the number of daemons '
+                             f'to advertise from.\n')
+            sys.exit()
+        auth_name, plugin, auth_opt = check_auth(ad_grp)
+        perm = check_opt('perm', ad_grp)
+        perm = perm_handler(perm)
+        rail = check_opt('rail', ad_grp)
+        credits = check_opt('credits', ad_grp)
+        rx_rate = check_opt('rx_rate', ad_grp)
+        ad_list = expand_names(spec['daemons'])
+        self.advertisers[spec['daemons']] = {'names'     : ad_grp['names'],
+                                             'hosts'     : ad_grp['hosts'],
+                                             'xprt'      : ad_grp['xprt'],
+                                             'port'      : ad_grp['port'],
+                                             'reconnect' : ad_grp['reconnect'],
+                                             'auth'      : { 'name' : auth_name,
+                                                             'conf' : auth_opt,
+                                                             'plugin' : plugin },
+                                             'perm'      : perm,
+                                             'rail'      : rail,
+                                             'credits'   : credits,
+                                             'rx_rate'   : rx_rate,
+                                             'ad_list'   : ad_list
+        }
+
+    def build_prdcr_listeners(self, spec):
+        if 'prdcr_listen' not in spec:
+            return
+        for pl in spec['prdcr_listen']:
+            check_required(['name'], pl,
+                           '"prdcr_listen" entry')
+            node_listen = {}
+            ip = check_opt('ip', pl)
+            regex = check_opt('regex', pl)
+            dstart = check_opt('disable_start', pl)
+            node_listen[pl['name']] = { 'disable_start' : dstart,
+                                        'regex'   : regex,
+                                        'ip'      : ip
+            }
+            self.prdcr_listeners[spec['daemons']] = node_listen
+
+    def build_aggregators(self, config):
+        aggregators = {}
+        if 'aggregators' not in config:
+            return aggregators
+        agg_conf = config['aggregators']
+        if type(agg_conf) is not list:
+            raise TypeError(f'{LDMS_YAML_ERR}\n'
+                            f' aggregators {LIST_ERR}\n'
+                            f'e.g. aggregators:\n'
+                            f'       - daemons: "l1-aggregators"\n'
+                            f'         peers :\n'
+                            f'           - daemons : "samplers"\n'
+                            f'             ...     :  ...\n')
+        for agg_spec in agg_conf:
+            check_required([ 'daemons' ],
+                           agg_spec, '"aggregators" entry')
+            names = expand_names(agg_spec['daemons'])
+            group = agg_spec['daemons']
+            plugins = check_opt('plugins', agg_spec)
+            if plugins:
+                if plugins is not list:
+                    raise ValueError(f'"plugins" must be a list of plugin instance names"\n')
+                for plugin in plugins:
+                    check_plugin_config(plugin, self.plugins)
+            daemons_ = None
+            for daemons in config['daemons']:
+                if group == daemons['names']:
+                    daemons_ = daemons
+            if daemons_ is None:
+                raise ValueError(f'Daemons regex {daemons["names"]} does not match daemon regex "{group}"\n')
+            if group not in aggregators:
+                aggregators[group] = {}
+            subscribe = check_opt('subscribe', agg_spec)
+            if subscribe:
+                for stream in subscribe:
+                    check_required([ 'stream', 'regex' ], stream, "stream specification")
+            for name in names:
+                aggregators[group][name] = { 'state' : 'stopped' } # 'running', 'error'
+                self.build_advertisers(agg_spec)
+                self.build_prdcr_listeners(agg_spec)
+                if subscribe:
+                    aggregators[group][name]['subscribe'] = subscribe
+                if plugins:
+                    aggregators[group][name]['plugins'] = plugins
+        return aggregators
+
+    def build_producers(self, config):
+        """
+        Return a dictionary keyed by the group name. Each dictionary
+        entry is a list of producers in that group.
+        """
+        producers = {}
+        for agg in config.get('aggregators', []):
+            if 'peers' not in agg:
+                continue
+            if type(agg['peers']) is not list:
+                raise TypeError(f'{LDMS_YAML_ERR}\n'
+                                f'peers {LIST_ERR}\n'
+                                f'e.g. peers:\n'
+                                f'       - daemons: "samplers"\n'
+                                f'         endpoints : "sampler-endpoints"\n'
+                                f'         ...       : ...\n')
+            for prod in agg['peers']:
+                check_required([ 'endpoints', 'reconnect', 'type', ],
+                               prod, '"peers" entry')
+                # Use endpoints for producer names and remove names attribute?
+                if prod['daemons'] not in self.daemons:
+                    dmn_grps = prod['daemons'].split(',')
+                    eps = prod['endpoints'].split(',')
+                else:
+                    dmn_grps = [ prod['daemons'] ]
+                    eps = [ prod['endpoints'] ]
+                for daemons, endpoints in zip(dmn_grps, eps):
+                    names = expand_names(endpoints)
+                    endpoints = expand_names(endpoints)
+                    group = agg['daemons']
+                    smplr_dmns = expand_names(daemons)
+                    if group not in producers:
+                        producers[group] = {}
+
+                    upd_spec = check_opt('updaters', prod)
+                    # Expand and generate all the producers
+                    typ = prod['type']
+                    reconnect = check_intrvl_str(prod['reconnect'])
+                    perm = check_opt('perm', prod)
+                    perm = perm_handler(perm)
+                    ports_per_dmn = len(endpoints) / len(smplr_dmns)
+                    ppd = ports_per_dmn
+                    try:
+                        for name in names:
+                            if ppd > 1:
+                                smplr_dmn = smplr_dmns[0]
+                                ppd -= 1
+                            else:
+                                smplr_dmn = smplr_dmns.pop(0)
+                                ppd = ports_per_dmn
+                            endpoint = endpoints.pop(0)
+                            prod = {
+                                'daemon'    : smplr_dmn,
+                                'dmn_grp'   : daemons,
+                                'name'      : name,
+                                'endpoint'  : endpoint,
+                                'type'      : typ,
+                                'group'     : group,
+                                'reconnect' : reconnect,
+                                'perm'      : perm,
+                                'updaters'  : upd_spec
+                            }
+                            producers[group][endpoint] = prod
+                    except:
+                        raise ValueError(f'Mismatch in producer config:\n'
+                                         f'Please ensure "endpoints" is configured to the correct number of ports specified.\n')
+        return producers
+
+    def build_updaters(self, config):
+        """
+        Return a dictionary based on the aggregator. Each dictionary
+        entry is a list of updaters in that group.
+        """
+        updaters = {}
+        updtr_cnt = 0
+        for agg in config.get('aggregators', []):
+            if 'peers' not in agg and 'prdcr_listen' not in agg:
+                continue
+            peer_list = []
+            if 'peers' in agg:
+                peer_list += agg['peers']
+            if 'prdcr_listen' in agg:
+                peer_list += agg['prdcr_listen']
+            for prod in peer_list:
+                if prod['updaters'] is None:
+                    continue
+                if type(prod['updaters']) is not list:
+                    raise TypeError(f'{LDMS_YAML_ERR}\n'
+                                    f'"updaters" {LIST_ERR}\n'
+                                    f'e.g. updaters:\n'
+                                    f'       - mode     : pull\n'
+                                    f'         interval : "1.0s"\n'
+                                    f'         sets     :\n'
+                                    f'           - regex : ".*"\n'
+                                    f'             field : inst\n')
+                for updtr_spec in prod['updaters']:
+                    check_required([ 'interval' ],
+                                   updtr_spec, '"updaters" entry')
+                    updtr_sets = check_opt('sets', updtr_spec)
+                    if updtr_sets and type(updtr_sets) is not list:
+                        raise ValueError(f'Error parsing YAML configuration in "updaters". "sets" must be a list of dictionaries')
+                    group = agg['daemons']
+                    if group not in updaters:
+                        updaters[group] = {}
+                    grp_updaters = updaters[group]
+                    updtr_name = f'updtr_{updtr_cnt}'
+                    if updtr_name in grp_updaters:
+                        raise ValueError(f'Duplicate updater name "{updtr_name}". '
+                                         f'An updater name must be unique within the group\n')
+                    perm = check_opt('perm', updtr_spec)
+                    perm = perm_handler(perm)
+                    prod_regex = check_opt('producers', prod)
+                    prod_list = []
+                    if prod_regex:
+                        prod_list.append({'regex' : prod_regex})
+                    else:
+                        prod_regex = '.*'
+                        prod_list.append({'regex' : '.*'})
+                    updtr = {
+                        'name'      : updtr_name,
+                        'interval'  : check_intrvl_str(updtr_spec['interval']),
+                        'perm'      : perm,
+                        'group'     : agg['daemons'],
+                        'sets'      : updtr_sets,
+                        'producers' : prod_list
+                    }
+                    if 'offset' in updtr_spec:
+                        updtr['offset'] = check_intrvl_str(updtr_spec['offset'])
+                    if 'mode' in updtr_spec:
+                        updtr['mode'] = updtr_spec['mode']
+                    else:
+                        updtr['mode'] = 'pull'
+                    grp_updaters[updtr_name] = updtr
+                    updtr_cnt += 1
+        return updaters
+
+    def build_stores(self, config):
+        """
+        Return a dictionary keyed by the group name. Each dictionary
+        entry is a list of stores in that group.
+        """
+        if 'stores' not in config:
+            return None
+        stores = {}
+        if type(config['stores']) is not dict:
+            raise ValueError(f'{LDMS_YAML_ERR}\n'
+                             f'store {DICT_ERR}\n'
+                             f'e.g. stores:\n'
+                             f'       sos-meminfo :\n'
+                             f'         daemons   : "l1-aggregators"\n'
+                             f'         container : ldms_data\n'
+                             f'         ...       : ...\n')
+        for store in config['stores']:
+            store_spec = config['stores'][store]
+            store_spec['name'] = store
+            check_required([ 'plugin', ],
+                           store_spec, '"store" entry')
+            container = check_opt('container', store_spec)
+            if not container:
+                container = ""
+            decomp = check_opt('decomposition', store_spec)
+            if not decomp:
+                check_required(['schema'], store_spec, '"store" entry')
+            schema = check_opt('schema', store_spec)
+            regex = check_opt('regex', store_spec)
+            if decomp and not schema and not regex:
+                raise ValueError(f'Decomposition plugin configuration requires either'
+                                 f' "schema" or "regex" attribute"\n')
+            if decomp and schema and regex:
+                raise ValueError(f'Please specify either "schema" or "regex" when using decomposition - these parameters are mutually exclusive.\n')
+            group = store_spec['daemons']
+            if group not in stores:
+                stores[group] = {}
+            grp_stores = stores[group]
+            if store in grp_stores:
+                raise ValueError(f'Duplicate store name "{store}". '
+                                 f'A store name must be unique within the group\n')
+            check_opt('flush', store_spec)
+            perm = check_opt('perm', store_spec)
+            store_spec['perm'] = perm_handler(perm)
+            check_plugin_config(store_spec['plugin'], self.plugins)
+            grp_stores[store] = store_spec
+        return stores
+
+    def build_samplers(self, config):
+        """
+        Generate samplers from YAML config.
+        Return a dictionary keyed by the samplers group name. Each dictionary
+        entry is a single ldms daemon's sampler configuration.
+        """
+        if 'samplers' not in config:
+            return None
+        smplrs = {}
+        if type(config['samplers']) is not list:
+            raise TypeError(f'{LDMS_YAML_ERR}\n'
+                            f'samplers {LIST_ERR}\n'
+                            f'e.g. samplers:\n'
+                            f'       - daemons : "samplers"\n'
+                            f'         plugins :\n'
+                            f'           - name        : meminfo\n'
+                            f'             interval    : "1.0s"\n'
+                            f'             offset      : "0s"\n'
+                            f'             config :\n'
+                            f'               - schema : meminfo\n'
+                            f'                 component_id : "10001"\n'
+                            f'                 producer : "node-1"\n'
+                            f'                 perm : "rwx-rwx-rwx"\n')
+        for smplr_spec in config['samplers']:
+            check_required([ 'daemons', 'plugins' ],
+                           smplr_spec, '"sampler" entry')
+            self.build_advertisers(smplr_spec)
+            for plugin in smplr_spec['plugins']:
+                check_plugin_config(plugin, self.plugins)
+            smplrs[smplr_spec['daemons']] = smplr_spec
+        return smplrs
+
+    def build_plugins(self, config):
+        """
+        Generate plugins to load from a YAML config.
+        Return a dictionary keyed by the plugin's group name. Each dictionary entry
+        is a single plugin's configuration.
+        """
+        if 'plugins' not in config:
+            return None
+        if type(config['plugins']) is not dict:
+            raise TypeError(f'{LDMS_YAML_ERR}\n'
+                            f'store {DICT_ERR}\n'
+                            f'e.g. plugins:\n'
+                            f'       meminfo1 :\n'
+                            f'         name      : meminfo\n'
+                            f'         interval  : 1.0s\n'
+                            f'         config    : [ { schema : meminfo }, { ... : ... } ]\n')
+        plugins = {}
+        plugn_spec = config['plugins']
+        for plugn in plugn_spec:
+            if plugn in plugins:
+                raise ValueError(f'Duplicate plugin name "{plugin_name}". '
+                                 f'Plugin must be unique within a group.\n')
+            check_plugin_config(plugn, plugn_spec)
+            plugins[plugn] = plugn_spec[plugn]
+        return plugins
+
+    def __init__(self, client, name, cluster_config, args):
+        """
+        Build configuration groups out of the YAML configuration
+        """
+        self.client = client
+        self.name = name
+        self.args = args
+        self.cluster_config = cluster_config
+        self.advertisers = {}
+        self.prdcr_listeners = {}
+        self.daemons = self.build_daemons(cluster_config)
+        self.plugins = self.build_plugins(cluster_config)
+        self.aggregators = self.build_aggregators(cluster_config)
+        self.producers = self.build_producers(cluster_config)
+        self.updaters = self.build_updaters(cluster_config)
+        self.stores = self.build_stores(cluster_config)
+        self.samplers = self.build_samplers(cluster_config)
+
+    def ldmsd_arg_list(self, local_path, dmn_grp, dmn):
+        start_list = [ 'ldmsd' ]
+        for ep in self.daemons[dmn_grp][dmn]['endpoints']:
+            if self.daemons[dmn_grp][dmn]['endpoints'][ep]['maestro_comm'] is True:
+                ep_ = self.daemons[dmn_grp][dmn]['endpoints'][ep]
+                start_list.append('-x')
+                start_list.append(f'{ep_["xprt"]}:{ep_["port"]}')
+                auth = check_opt('auth', ep_)
+                if auth:
+                    auth_plugin = check_opt('plugin', ep_['auth'])
+                    auth_opt = check_opt('conf', ep_)
+                    start_list.append('-a')
+                    start_list.append(auth_plugin)
+                    if auth_opt:
+                        if len(auth_opt.split('=')) < 2:
+                            auth_opt = f'conf={auth_opt}'
+                        start_list.append('-A')
+                        start_list.append(auth_opt)
+        start_list.append('-c')
+        start_list.append(f'{local_path}/{dmn}.conf')
+        start_list.append('-r')
+        start_list.append(f'{local_path}/{dmn}.pid')
+        start_list.append('-l')
+        start_list.append(f'{local_path}/{dmn}.log')
+        start_list.append(f'-F')
+        return start_list
+
+    def write_advertisers(self, dstr, dmn_grp, dname, auth_list):
+        if dmn_grp not in self.advertisers:
+            return dstr, auth_list
+        ad_grp = self.advertisers[dmn_grp]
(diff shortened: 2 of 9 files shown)


tom95858:
@nick-enoent could you take a look at these build failures?

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -792,6 +792,235 @@ int __process_config_file(const char *path, int *lno, int trust,
...
+	size_t char_cnt = 0;
+	while (fgets(cstr, sizeof(cstr), fp) != NULL) {
+		printf("%s", cstr);
+		char_cnt += sizeof(cstr);
+		if (char_cnt >= 1024)
+			cfg_str = (char *)realloc(cfg_str, char_cnt - 256);
+		if (lineno)
+			strcat(cfg_str, cstr);
+		else
+			snprintf(cfg_str, sizeof(cstr), cstr);
```
tom95858, line 1014: @nick-enoent this sizeof(cstr) is 8, not 256 like you intended. This is what is breaking the build.

nick-enoent:
> @nick-enoent could you take a look at these build failures?

Done.

ldms/src/ldmsd/ldmsd_config.c
nick-enoent: Addressed in the latest commit.

narategithub (approved):
:+1:

tom95858:
@nick-enoent could you please remove the [wip] from the description and I will accept this?

morrone:
@nick-enoent How do I the following in the YAML config?

```
option --set_memory 100MB # and other command line options, of course
metric_sets_default_authz uid=0 gid=lcstaff perm=440
```

morrone (changes requested):
First problem: There are at least two stray non-printing characters output by ldmsd_yaml_parser at the end of its output.

Next problem: For sampler plugins, it seems pretty broken. I am only seeing it include the last config specified config. Consider this yaml:

```
daemons:
- names: hetchy1018
  hosts: hetchy1018
  endpoints:
  - names: "hetchy1018-mgmt"
    ports: 411
    xprt: sock
    auth:
      name: munge_domain0
      plugin: munge

samplers:
- daemons: hetchy1018
  plugins:
  - meminfo

plugins:
  meminfo:
    name: meminfo
    interval: 5s
    offset: 0s
    config:
    - producer: hetchy1018
      instance: hetchy1018/meminfo
      component_id: 1018
      uid: 0
      gid: lcstaff
      perm: 0440
      job_set: hetchy1018/job_info

```
and the output is only this:
```
$ ldmsd_yaml_parser --daemon_name hetchy1018 --ldms_config test.yaml
auth_add name=munge_domain0 plugin=munge
listen xprt=sock port=411 auth=munge_domain0
load name=meminfo
config name=meminfo job_set=hetchy1018/job_info
start name=meminfo interval=5000000 offset=0
```
Notice that "config" line for meminfo is missing a lot of things.

morrone:
I have some design issues that I'll bring up in the discussion #1496.

nick-enoent:
> First problem: There are at least two stray non-printing characters output by ldmsd_yaml_parser at the end of its output.
> 
> Next problem: For sampler plugins, it seems pretty broken. I am only seeing it include the last config specified config. Consider this yaml:
> 
> ```
> daemons:
> - names: hetchy1018
>   hosts: hetchy1018
>   endpoints:
>   - names: "hetchy1018-mgmt"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
> 
> samplers:
> - daemons: hetchy1018
>   plugins:
>   - meminfo
> 
> plugins:
>   meminfo:
>     name: meminfo
>     interval: 5s
>     offset: 0s
>     config:
>     - producer: hetchy1018
>       instance: hetchy1018/meminfo
>       component_id: 1018
>       uid: 0
>       gid: lcstaff
>       perm: 0440
>       job_set: hetchy1018/job_info
> ```
> 
> and the output is only this:
> 
> ```
> $ ldmsd_yaml_parser --daemon_name hetchy1018 --ldms_config test.yaml
> auth_add name=munge_domain0 plugin=munge
> listen xprt=sock port=411 auth=munge_domain0
> load name=meminfo
> config name=meminfo job_set=hetchy1018/job_info
> start name=meminfo interval=5000000 offset=0
> ```
> 
> Notice that "config" line for meminfo is missing a lot of things.

@morrone Thanks for giving this a test drive! The non printing character you're seeing is a null terminating '\0' when the string is passed back to ldmsd. I'll add a patch that addresses this when using the ldmsd_yaml_parser outside of ldmsd.

Good catch on the sampler config string bug, I'll introduce a patch for this today.

morrone:
> Thanks for giving this a test drive! The non printing character you're seeing is a null terminating '\0' when the string is passed back to ldmsd. I'll add a patch that addresses this when using the ldmsd_yaml_parser outside of ldmsd.

I think it would be best to never do that. I can't think of a reason that ldmsd would need a terminating null character. The end of the input stream is more than enough.

ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -0,0 +1,47 @@
...
+                        help="Daemon name to generate configuration from YAML file")
+    parser.add_argument("--version", metavar="VERSION",
+                        help="The OVIS version for the output syntax (4 or 5), default is 4",
+                        default=4)
+    parser.add_argument("--debug", action="store_true",
+                        help="Enable debug information")
+    args = parser.parse_args()
+    if not args.debug:
+        import sys
+        sys.tracebacklimit=0
```
morrone, line 28: We should _not_ do this. The backtrace is our only way to reasonably debug things when there are bugs in the code.

morrone (changes requested):

ldms/python/ldmsd/parser_util.py
```diff
@@ -0,0 +1,1005 @@
...
+                f"'1.5S' - 1.5 seconds\n"\
+                f"'2us'  - 2 microseconds\n"\
+                f"'3m'   - 3 minutes\n"\
+                f"'1h'   - 1 hour\n"\
+                f"'1d'   - 1 day\n"\
+                f"\n"
+    if type(interval_s) == int or type(interval_s) == float:
+        return interval_s
+    if type(interval_s) != str:
+        print(f"{error_str}")
```
morrone, line 84: Error message need to go to stderr, not stdout. I'm not make this comment every where it happens in the code, but it needs to be fixed everywhere that it happens.

ldms/python/ldmsd/parser_util.py
```diff
@@ -0,0 +1,1005 @@
...
+        if check_opt('environment', self.daemons[grp][dname]):
+            if type(self.daemons[grp][dname]['environment']) is not dict:
+                print(f'Error: Environment variables must be a yaml key:value dictionary\n')
+                sys.exit(22)
+            for attr in self.daemons[grp][dname]['environment']:
+                dstr += f'env {attr}={self.daemons[grp][dname]["environment"]}\n'
+        return dstr
+
+    def write_sampler(self, dstr, smplr_grp, sname):
+        if smplr_grp not in self.samplers:
```
morrone, line 786: "self.samplers" may be None if there is not a samplers section in the yaml. This will throw an exception, so the None case needs to be properly handled.

ldms/python/ldmsd/parser_util.py
```diff
@@ -0,0 +1,1005 @@
...
+        start_list.append('-c')
+        start_list.append(f'{local_path}/{dmn}.conf')
+        start_list.append('-r')
+        start_list.append(f'{local_path}/{dmn}.pid')
+        start_list.append('-l')
+        start_list.append(f'{local_path}/{dmn}.log')
+        start_list.append(f'-F')
+        return start_list
+
+    def write_advertisers(self, dstr, dmn_grp, dname, auth_list):
```
morrone, line 667: A whoooole bunch of these methods take "dstr" as an option. However, as far as I can tell not one of them actually needs to read the contents of that dstr, except that they do "dstr +=". I think it would make a lot of sense not do that, because the "+=" can always be done in the caller.

So unless I am missing something, I would recommend we redo all of these to not take "dstr" as an parameter.

morrone (changes requested):
I'll also add that that this PR is missing a documentation component.

morrone:
Container should  not be required in the YAML. If "container" is missing, just put ```container=""``` in the output.

morrone:
[hetchy_agg.yaml.gz](https://github.com/user-attachments/files/17670198/hetchy_agg.yaml.gz)

I hacked a quick fix to allow the code to progress when there are no samplers in a yaml, but when I process the attached file, I get nothing in the output.

```
$ ldmsd_yaml_parser --ldms_config hetchy_agg.yaml --daemon_name hetchy14
$
```

ldms/python/ldmsd/ldmsd_yaml_parser
nick-enoent: I agree with you, and that's how it was originally. However, there were some in the community that wanted the option to not see a debug error message, hence the "--debug" option.

ldms/python/ldmsd/parser_util.py
nick-enoent: Some methods return a tuple, rather than solely a string. This was done to remain consistent in how class methods are called/returned to avoid confusion, but I'm fine with going with your suggestion if that's what everyone wants.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -771,6 +771,214 @@ int __process_config_file(const char *path, int *lno, int trust,
...
+		}
+		strcat(cfg_str, cstr);
+	}
+	status = pclose(fp);
+	if (status)
+		goto err;
+	return cfg_str;
+err:
+	if (cfg_str) {
+		printf("ERROR: %s", cfg_str);
```
morrone, line 976: We don't want all of cfg_str printed when there is an error. We just want to know what the error was from ldmsd_yaml_parser. When ldmsd_yaml_parser is fixed to write its errors to stderr, then we don't even need to worry about capturing that error message in ldmsd; it will already have gone to the right place.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -771,6 +771,214 @@ int __process_config_file(const char *path, int *lno, int trust,
...
+	char cstr[256];
+        char *cfg_str = malloc(512);
+	*cfg_str = '\0';
+	snprintf(command, sizeof(command), "ldmsd_yaml_parser --ldms_config %s --daemon_name %s 2>&1", path, dname);
+	fp = popen(command, "r");
+	if (!fp)
+		printf("Error in yaml_parser\n");
+	size_t char_cnt = 0;
+	int status;
+	while (fgets(cstr, sizeof(cstr), fp) != NULL) {
```
morrone, line 960: The logic here looks broken in a few different ways to me. Once we cross over 512 bytes, it'll start realloc()ing, every single time, but only increasing the buffer size by 1 byte. But then we strcat something that is potentially alot larger than one byte larger.

But lets forget that. Fundamentally using fgets()/strlen()/strcat() are the wrong way to read a file into a buffer. This needs to be rewritten to use fread(). We can directly fread() into our buffer (no extra copies needed), and resize only when the buffer fills between fread()s.

morrone (changes requested):

ldms/python/ldmsd/parser_util.py
morrone: You can still do += after breaking apart the tuple. I'd really like to see it consistent in the other direction.

Sending the suggests to me, as a programmer, that the code is going to use/modify that string in some way. If it doesn't do that, then you need to explicitly document the contract you are making with the caller "I won't modify this, only append new data".

But the method doesn't take that as a parameter, you save yourself some documentation. It is impossible for the method to change something it doesn't have. Its just good programming practice to only pass a method the things it actually needs to accomplish its goal.

ldms/python/ldmsd/parser_util.py
nick-enoent: Addressed in 33deed8

ldms/python/ldmsd/parser_util.py
```diff
@@ -0,0 +1,1005 @@
...
+                f"'2us'  - 2 microseconds\n"\
+                f"'3m'   - 3 minutes\n"\
+                f"'1h'   - 1 hour\n"\
+                f"'1d'   - 1 day\n"\
+                f"\n"
+    if type(interval_s) == int or type(interval_s) == float:
+        return interval_s
+    if type(interval_s) != str:
+        print(f"{error_str}")
+        sys.exit(22)
```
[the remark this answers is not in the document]
nick-enoent: Addressed in 33deed8

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -771,6 +771,214 @@ int __process_config_file(const char *path, int *lno, int trust,
...
+}
+
+char *__process_yaml_config_file(const char *path, const char *dname)
+{
+	FILE *fp;
+	char command[256];
+	char cstr[256];
+        char *cfg_str = malloc(512);
+	*cfg_str = '\0';
+	snprintf(command, sizeof(command), "ldmsd_yaml_parser --ldms_config %s --daemon_name %s 2>&1", path, dname);
```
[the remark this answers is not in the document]
nick-enoent: Addressed in 33deed8

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -771,6 +771,214 @@ int __process_config_file(const char *path, int *lno, int trust,
...
+	fp = popen(command, "r");
+	if (!fp)
+		printf("Error in yaml_parser\n");
```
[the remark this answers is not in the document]
nick-enoent: Addressed in 33deed8

ldms/src/ldmsd/ldmsd_config.c
nick-enoent: All good points, thanks for the review. This should also be taken care of in 33deed8. I'll implement fixes for some of the other issues you mentioned today.

morrone:
@nick-enoent, as far as I can tell we are missing a way to do this:

```
# TCP networking on managment network
listen host=ehetchy1001 port=411 xprt=sock auth=munge_domain0
# The localhost socket is mainly needed for jobid info from the job scheduler
listen host=localhost port=411 xprt=sock auth=munge_domain0
```
Note the "host=" part of the lines. We need something under the each "endpoints" mapping of the yaml config to allow us to set the "host=" option. At the least, that isn't documented. I guessed "host" and "hosts", but those didn't do anything.

morrone:
I can confirm the sampler config key/value pairs are now making it into the output.

Another thing that I'll suggest is that we should probably be printing warnings (to stderr) when there are unrecognized keys in the yaml. At least any time we can reasonably do so. I recognize that knowing every sampler plugin's valid config options is probably out of scope, for instance. But we should make an effort to report them when we can. Otherwise we'll just silently not do what the user wanted because of things like simple typos, and users will be pretty frustrated. And we probably want those to be warnings instead of errors to support forward-compatibility (new keys added to a config for a future version of the code that hasn't been installed yet).

I'll wait for feedback on why the aggregator yaml I supplied here resulted in no output before my next round of testing. I don't know if I did something wrong, or if the parser has a bug. But I'm stuck on that at the moment.

nick-enoent:
> I can confirm the sampler config key/value pairs are now making it into the output.
> 
> Another thing that I'll suggest is that we should probably be printing warnings (to stderr) when there are unrecognized keys in the yaml. At least any time we can reasonably do so. I recognize that knowing every sampler plugin's valid config options is probably out of scope, for instance. But we should make an effort to report them when we can. Otherwise we'll just silently not do what the user wanted because of things like simple typos, and users will be pretty frustrated. And we probably want those to be warnings instead of errors to support forward-compatibility (new keys added to a config for a future version of the code that hasn't been installed yet).
> 
> I'll wait for feedback on why the aggregator yaml I supplied here resulted in no output before my next round of testing. I don't know if I did something wrong, or if the parser has a bug. But I'm stuck on that at the moment.

@morrone I just went over your hetchy-agg.yaml, and it's a bit of a configuration error that's exposing a bug. 

In your example, you defined the "hetchy14" daemon twice, once in the first "daemons" list entry, and again as a single daemon configuration, with identical arguments aside from the name/endpoints. You then added the first "daemons" list entry endpoints as "peers" to "hetchy14", effectively adding "hetchy14" as a producer to itself. I'm not aware of a use case for this, but I could be wrong.

All that said, this exposes a bug - we should at least get an error or a warning when this type of configuration event occurs, and certainly not an empty string. If we do need to support this use case, I need to modify the code to support it.

morrone:
> In your example, you defined the "hetchy14" daemon twice

OK, I fixed that, but I still don't get any output. See attached.
[hetchy_agg.yaml.gz](https://github.com/user-attachments/files/17684547/hetchy_agg.yaml.gz)

morrone:
Whoops! OK, I needed to give the right daemon label at the command line too. :)

But with the most recent yaml that I provided, I am getting this:

```
hetchy1001:~/tmp$ ldmsd_yaml_parser --ldms_config hetchy_agg.yaml --daemon_name hetchy14-agg --debug
Traceback (most recent call last):
  File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 849, in write_aggregator
    dstr = self.write_stores(dstr, group_name)
  File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 918, in write_stores
    strgp_add += f'schema={store_group[store]["schema"]}'
KeyError: 'schema'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/g/g0/morrone/local/ldms/bin/ldmsd_yaml_parser", line 35, in <module>
    ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
  File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 940, in daemon_config
    dstr = self.write_aggregator(dstr, grp, dname)
  File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 853, in write_aggregator
    raise Exception('Agg config Error: '+str(e)+' Line:'+str(ec.tb_lineno))
Exception: Agg config Error: 'schema' Line:849

```
The line number here is bogus. There is no line 849 in the file.

My normal non-aggregator config file doesn't have a "schema" entry anywhere in it, so I'm not sure why the yaml would need one. That seems like a bug.

Also, the exception inside an exception might need attention.

nick-enoent:
> Whoops! OK, I needed to give the right daemon label at the command line too. :)
> 
> But with the most recent yaml that I provided, I am getting this:
> 
> ```
> hetchy1001:~/tmp$ ldmsd_yaml_parser --ldms_config hetchy_agg.yaml --daemon_name hetchy14-agg --debug
> Traceback (most recent call last):
>   File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 849, in write_aggregator
>     dstr = self.write_stores(dstr, group_name)
>   File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 918, in write_stores
>     strgp_add += f'schema={store_group[store]["schema"]}'
> KeyError: 'schema'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/g/g0/morrone/local/ldms/bin/ldmsd_yaml_parser", line 35, in <module>
>     ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
>   File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 940, in daemon_config
>     dstr = self.write_aggregator(dstr, grp, dname)
>   File "/g/g0/morrone/local/ldms/lib/python3.6/site-packages/ldmsd/parser_util.py", line 853, in write_aggregator
>     raise Exception('Agg config Error: '+str(e)+' Line:'+str(ec.tb_lineno))
> Exception: Agg config Error: 'schema' Line:849
> ```
> 
> The line number here is bogus. There is no line 849 in the file.
> 
> My normal non-aggregator config file doesn't have a "schema" entry anywhere in it, so I'm not sure why the yaml would need one. That seems like a bug.
> 
> Also, the exception inside an exception might need attention.

@morrone Sorry it took me a bit to get back to you, I stepped out for a late lunch. I just pushed a change that addresses this exact issue. The line 849 you're seeing is from the parser_util.py file. ldmsd_yaml_parser import the YamlCfg class from parser_util.py so that going forward, maestro can import the YamlCfg class and implement a maestro-specific subclass. As to the exception inside the exception, that's an easy fix, and we can just print to stderr, but we would have to use sys.exit() at that point, as the try/except catch allows the program to continue running.

morrone:
I can confirm that [14bb1a8](https://github.com/ovis-hpc/ldms/pull/1461/commits/14bb1a8cb1f6fdab31eef011d3be4b72f7a24c7a) got rid of the ```KeyError: 'schema'``` issue.

Now looking at the output, I'm seeing something that might be an issue:
```
prdcr_add name=hetchy1-mgmt host=hetchy1 port=411 xprt=sock type=active reconnect=60000000 auth=munge_domain0
prdcr_start_regex regex=.*
prdcr_add name=hetchy2-mgmt host=hetchy2 port=411 xprt=sock type=active reconnect=60000000 auth=munge_domain0
prdcr_start_regex regex=.*
```
Is prdcr_start_regex really a thing? In my current non-yaml config files I'm using:
```
prdcr_add name=hetchy1 host=ehetchy1 type=active xprt=sock port=411 auth=munge_domain0 interval=10000000
prdcr_start name=hetchy1
prdcr_add name=hetchy2 host=ehetchy2 type=active xprt=sock port=411 auth=munge_domain0 interval=10000000
prdcr_start name=hetchy2
```

Also, for some reason the ldmsd_yaml_parser is taking around 30 seconds to process this smallish aggregator yaml. Are you seeing that too? I checked "top" and it doesn't seem to be using much CPU at all. It is almost like there is a sleep() in there somewhere.

morrone:
> As to the exception inside the exception, that's an easy fix, and we can just print to stderr, but we would have to use sys.exit() at that point, as the try/except catch allows the program to continue running.

That is not what I'm looking for. We _do_ want to reraise an exception, not call sys.exit(). Looking at the code more, here is what I think could be improved:

```Exception: Agg config Error: 'schema' Line:849```

Just saying "Line:849" is confusing to the user (me). There is no signal here that it is talking about a source code line number rather than the line in the config file where the problem occurred.  I think at the very least the line number needs to be paired with a source code file name to clear that up. We might also consider using the "traceback" module, and including just one line function deep of the call back, e.g.:

```s = traceback.format_stack(limit=1)```

And we can get rid of the "During handling of the above exception, another exception occurred:" by adding "from None"

```   raise Exception() as None```

**Comment by nick-enoent** [comment]
> I can confirm that [14bb1a8](https://github.com/ovis-hpc/ldms/pull/1461/commits/14bb1a8cb1f6fdab31eef011d3be4b72f7a24c7a) got rid of the `KeyError: 'schema'` issue.
> 
> Now looking at the output, I'm seeing something that might be an issue:
> 
> ```
> prdcr_add name=hetchy1-mgmt host=hetchy1 port=411 xprt=sock type=active reconnect=60000000 auth=munge_domain0
> prdcr_start_regex regex=.*
> prdcr_add name=hetchy2-mgmt host=hetchy2 port=411 xprt=sock type=active reconnect=60000000 auth=munge_domain0
> prdcr_start_regex regex=.*
> ```
> 
> Is prdcr_start_regex really a thing? In my current non-yaml config files I'm using:
> 
> ```
> prdcr_add name=hetchy1 host=ehetchy1 type=active xprt=sock port=411 auth=munge_domain0 interval=10000000
> prdcr_start name=hetchy1
> prdcr_add name=hetchy2 host=ehetchy2 type=active xprt=sock port=411 auth=munge_domain0 interval=10000000
> prdcr_start name=hetchy2
> ```
> 
> Also, for some reason the ldmsd_yaml_parser is taking around 30 seconds to process this smallish aggregator yaml. Are you seeing that too? I checked "top" and it doesn't seem to be using much CPU at all. It is almost like there is a sleep() in there somewhere.

The longest the ldmsd_yaml_parser takes on my setup is usually about 4-5 seconds. I'll run some tests with your example config to see if I can replicate the long parse time. 

> I can confirm that [14bb1a8](https://github.com/ovis-hpc/ldms/pull/1461/commits/14bb1a8cb1f6fdab31eef011d3be4b72f7a24c7a) got rid of the `KeyError: 'schema'` issue.
> 
> Now looking at the output, I'm seeing something that might be an issue:
> 
> ```
> prdcr_add name=hetchy1-mgmt host=hetchy1 port=411 xprt=sock type=active reconnect=60000000 auth=munge_domain0
> prdcr_start_regex regex=.*
> prdcr_add name=hetchy2-mgmt host=hetchy2 port=411 xprt=sock type=active reconnect=60000000 auth=munge_domain0
> prdcr_start_regex regex=.*
> ```
> 
> Is prdcr_start_regex really a thing? In my current non-yaml config files I'm using:
> 
> ```
> prdcr_add name=hetchy1 host=ehetchy1 type=active xprt=sock port=411 auth=munge_domain0 interval=10000000
> prdcr_start name=hetchy1
> prdcr_add name=hetchy2 host=ehetchy2 type=active xprt=sock port=411 auth=munge_domain0 interval=10000000
> prdcr_start name=hetchy2
> ```
> 
> Also, for some reason the ldmsd_yaml_parser is taking around 30 seconds to process this smallish aggregator yaml. Are you seeing that too? I checked "top" and it doesn't seem to be using much CPU at all. It is almost like there is a sleep() in there somewhere.

I haven't seen the long ldmsd_yaml_parser time - the longest it usually takes is around 5 seconds for a file that contains an entire clusters configuration. Is this using the example config file you posted here?

I looked at the other issue you noted here regarding prdcr_start_regex this morning and have implemented a fix for this as well. I'll push it shortly. The current implementation shouldn't affect this configuration from functioning, but it will produce log message warnings for the producers that have already been started, which is obviously something we don't want.

**Comment by morrone** [comment]
I can't reproduce the long running time right now. I am going to assume there something unrelated to the python code wrong on the system at the time.

@nick-enoent Let me know when you believe that you've worked through the outstanding issues and I'll take another pass.

**Comment by emdonat** [comment]
Hi Chris,

Would the endpoint names be the correct target for the host= parameter in the listen command generated by the -y option?  It seems like this would function well but that would mean less flexibility in the name...  Right now, it is pulling host and dropping that in the host= parameter which does, indeed, cause it to not correctly use both of the endpoints.

```
daemons:
- names: "hetchy[1-2,14,201-202,1002-1018]"
  hosts: "hetchy[1-2,14,201-202,1002-1018]"
  endpoints:
  - names: "hetchy[1-2,14,201-202,1002-1018]-mgmt"
    ports: 411
    xprt: sock
    auth:
      name: munge_domain0
      plugin: munge
  - names: "hetchy[1-2,14,201-202,1002-1018]-fabric"
    ports: 411
    xprt: fabric
    auth:
      name: munge_domain0
      plugin: munge

- names: "hetchy14-agg"
  hosts: "hetchy14-agg"
  endpoints:
  - names: "hetchy14-mgmt"
    ports: 411
    xprt: sock
    auth:
      name: munge_domain0
      plugin: munge

  - names: "hetchy14-fabric"
    ports: 411
    xprt: fabric
    auth:
      name: munge_domain0
      plugin: munge


aggregators:
- daemons: "hetchy14-agg"
  peers:
  - daemons: "hetchy[1-2,14,201-202,1002-1018]"
    endpoints: "hetchy[1-2,14,201-202,1002-1018]-mgmt"
    reconnect: 20s
    type: active
    updaters:
      - mode: pull
        interval: "1.0s"
        offset: "10ms"
        sets:
          - regex: "*."
            field: inst
  - daemons: "hetchy[1-2,14,201-202,1002-1018]"
    endpoints: "hetchy[1-2,14,201-202,1002-1018]-mgmt"
    reconnect: 20s
    type: active
    updaters:
      - mode: pull
        interval: "1.0s"
        offset: "10ms"
        sets:
          - regex: "*."
            field: inst
stores:
  sos-meminfo :
    daemons : "hetchy14"
    container : ldms_data
    schema : meminfo
    flush : 10s
    plugin : store_sos1

plugins:
  store_sos1 :
    name : store_sos
    config : ["path=/opt/DATA"]
```
Outputs:
```
hetchy14-agg
auth_add name=munge_domain0 plugin=munge
listen xprt=sock port=411 auth=munge_domain0
listen xprt=fabric port=411 auth=munge_domain0
prdcr_add name=hetchy1-mgmt host=hetchy1 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1-mgmt
prdcr_add name=hetchy2-mgmt host=hetchy2 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy2-mgmt
prdcr_add name=hetchy14-mgmt host=hetchy14 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy14-mgmt
prdcr_add name=hetchy201-mgmt host=hetchy201 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy201-mgmt
prdcr_add name=hetchy202-mgmt host=hetchy202 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy202-mgmt
prdcr_add name=hetchy1002-mgmt host=hetchy1002 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1002-mgmt
prdcr_add name=hetchy1003-mgmt host=hetchy1003 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1003-mgmt
prdcr_add name=hetchy1004-mgmt host=hetchy1004 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1004-mgmt
prdcr_add name=hetchy1005-mgmt host=hetchy1005 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1005-mgmt
prdcr_add name=hetchy1006-mgmt host=hetchy1006 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1006-mgmt
prdcr_add name=hetchy1007-mgmt host=hetchy1007 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1007-mgmt
prdcr_add name=hetchy1008-mgmt host=hetchy1008 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1008-mgmt
prdcr_add name=hetchy1009-mgmt host=hetchy1009 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1009-mgmt
prdcr_add name=hetchy1010-mgmt host=hetchy1010 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1010-mgmt
prdcr_add name=hetchy1011-mgmt host=hetchy1011 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1011-mgmt
prdcr_add name=hetchy1012-mgmt host=hetchy1012 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1012-mgmt
prdcr_add name=hetchy1013-mgmt host=hetchy1013 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1013-mgmt
prdcr_add name=hetchy1014-mgmt host=hetchy1014 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1014-mgmt
prdcr_add name=hetchy1015-mgmt host=hetchy1015 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1015-mgmt
prdcr_add name=hetchy1016-mgmt host=hetchy1016 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1016-mgmt
prdcr_add name=hetchy1017-mgmt host=hetchy1017 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1017-mgmt
prdcr_add name=hetchy1018-mgmt host=hetchy1018 port=411 xprt=sock type=active reconnect=20000000 auth=munge_domain0
prdcr_start name=hetchy1018-mgmt
updtr_add name=updtr_0 auto_interval=True interval=1000000 offset=10000
updtr_prdcr_add name=updtr_0 regex=.*
updtr_start name=updtr_0
updtr_add name=updtr_1 auto_interval=True interval=1000000 offset=10000
updtr_prdcr_add name=updtr_1 regex=.*
updtr_start name=updtr_1
```

This means that unless we capture some other field for host (interface as you suggested or reuse endpoint:name) we lose the ability to assign multiple endpoints to a single host.

**Comment by morrone** [comment]
> Would the endpoint names be the correct target for the host= parameter in the listen command generated by the -y option? It seems like this would function well but that would mean less flexibility in the name... 

I don't think it would quite allow the full degree of configuration completeness. Right now it is possible that one might want to have multiple listen sockets on the same interface (maybe to allow different auth or different xprt). For that we need different labels for the same interface names.

> Right now, it is pulling host and dropping that in the host= parameter which does, indeed, cause it to not correctly use both of the endpoints.

That was probably the intent, but in my testing there was no "host=" parameter at all on the "listen" line. So that was one bug I am proposing to fix.

**Comment by tom95858** [comment]
@morrone, @nick-enoent, @emdonat Do we have any chance of getting this in 4.4.5? We're basically at the end of the rope on 4.4.5 and I think there are users who want to play with it. If we extend a week, we can do the testing next week, however, there still seem to be syntax concerns from some users. Can we accept what we have now with the idea that we can add improvements later; or is -y doomed for 4.4.5?

**Comment by tom95858** [comment]
PS: @morrone, @nick-enoent, @emdonat, @jennfshr we could put in the release notes that the -y syntax in 4.4.5 is '_provisional_' and may change in subsequent releases.

**Comment by morrone** [comment]
I just stumbled across the man page "ldmsd_setgroup". Are set groups something we need to incorporate into the yaml spec?

**Comment by morrone** [comment]
Looking at the ```ldmsd_controller``` man page, it says:

```
LISTEN COMMAND SYNTAX
   Instruct ldmsd to listen to a port
       listen port=PORT xprt=sock|rdma|ugni|fabric [host=HOST] [auth=AUTH_REF]
[cut]
       host=HOST
              An optional hostname or IP address to bind. If not given, listen to all addresses (0.0.0.0 or PORT).
```

So my proposal of adding ```daemons->endpoints->interfaces``` is incomplete. We also need to allow the interfaces to be left out, which would should result in "host=" _not_ being specified in the "listen" line for ldmsd. It should _not_ use ```daemons->hosts``` when ```daemons->endpoints-interfaces``` is absent.

But that leaves us with another issue. We need to allow the user to decide which IP/hostname to use in the ``aggregators->peers```. Right now that is specified only by reference to things in the daemons section. When the daemon leaves out _interfaces_, it might not have the correct IP/hostname listed anywhere.

So this brings me to another proposal that I was thinking about anyway. This just solidifies in my mind that we should do it.

I would propose that we allow specifying the complete endpoint information under ```aggregations->peers``` rather than only specifying by reference to things in ```daemons```.

To elaborate on why I wanted that even before this use case: It always seemed strange to me, as a person always configuring a single daemon at a time in the yaml, that I needed to give the specification for many daemons. If I could just list the endpoints in detail under ```aggregators->peers```, then I wouldn't need to have more than one daemon in the daemons section. The daemon that the file is configuring.

So make these revised proposals:

- [ ] Introduce an optional _interfaces_ key under the endpoints mapping. If the _interfaces_ key is present, the daemon will attempt to listen on the hostname or IP address specified in the value of _interfaces_.  If _interfaces_ is not present, the daemon will attempt to listen on all interfaces.
- [ ] Remove the daemons key from the ```aggregators->peers``` mappings.
- [ ] Create ```aggregators->peers->interfaces```. If _interfaces_ is a string, it will must match a label in ```daemons->endpoints->names```. Otherwise, it _interfaces_ is a mapping, it must have the follwing keys: hosts, xprt, port, auth (etc.) and directly describe the peer interfaces.

Note, I have no idea if this is all compatible with the umpteen different ways aggregators and "peers" can connect to each other. We might need to make a list of all the ways they can do that so we can cover everything.

**Comment by emdonat** [comment]
Hi Chris,  I think that we have work that will provide for:

- [ ]  Introduce an optional interfaces key under the endpoints mapping. If the interfaces key is present, the daemon will attempt to listen on the hostname or IP address specified in the value of interfaces. If interfaces is not present, the daemon will attempt to listen on all interfaces.

Except the term will be hosts.  This will allow for specification of a hostname or IP address for the listen.  Absent this the listen will be on `0.0.0.0` all ip addresses.  The incorrect `hosts=` in producer_add will be remedied as well to be that host or IP as specified in the `daemons->endpoints->hosts` key.

Work is being done on the man pages as well for release.

ldmsd_setgroup is something that we will not carry forward now as it is being discussed for deletion.

The daemons key will require more thought and it is likely appropriate to kill it but given the compressed timeline and that it is being used as a structural component in the current configuration, it may be a wart that we have to accept for this iteration.  I don't believe that its existence will cause failures but is just increased verbosity.  

The work on `aggregators->peers->interfaces` is possibly an extension of functionality and can be captured more or less in the daemons above for now and is likely not able to be completed for this release but would be able to be discussed for a future release.

That seems like an appropriate coverage of functionality and matches your assertion above of "I am skeptical that it can make it for 4.4.5. I don't think it should land, even provisionally, without: -Fixing the known bugs/deficiencies -Fully documenting the specification in a man page as part of the same PR", which I agree with 100%.

Nick is working on the man pages now and should have something here for us soon.

**Comment by nick-enoent** [comment]
> Is there a commit to this that will need testing? I want to plan time this week.

Hey Evan, I've added code to support CLI options in the YAML file and new endpoints "hosts" functionality we discussed. These should be the last commits that need testing.

**Comment by morrone** [comment]
> > Is there a commit to this that will need testing? I want to plan time this week.
> 
> Hey Evan, I've added code to support CLI options in the YAML file and new endpoints "hosts" functionality we discussed. These should be the last commits that need testing.

I think the "Update host behavior" patch differs a bit from what we decided. We were supposed to have something like:

```    bind_all: True```

under ```daemons->interfaces```, right? With the current commit's implementation, when all of the "hosts" lines are missing, the aggregators->endpoints section won't have anything to reference to be able to connect to sampler daemons.

I think we agreed on something more like this:

```
"hosts" can be specified in a daemon endpoint.
If no "hosts" key is specified in endpoints, it defaults to the higher level daemon host (one of these _must_ be present)
"bind_all" (a boolean) can be specified in a daemon enpoint. "bind_all" defaults to False. If "bind_all" is True, the daemon will bind to all available interfaces rather than just the hostnames/IPs listed in "hosts". Other daemons (e.g. aggregators) will use the hostnames/IPs listed in "hosts" to connect to this "endpoint".
```

**Comment by nick-enoent** [comment]
> > > Is there a commit to this that will need testing? I want to plan time this week.
> > 
> > 
> > Hey Evan, I've added code to support CLI options in the YAML file and new endpoints "hosts" functionality we discussed. These should be the last commits that need testing.
> 
> I think the "Update host behavior" patch differs a bit from what we decided. We were supposed to have something like:
> 
> ` bind_all: True`
> 
> under `daemons->interfaces`, right? With the current commit's implementation, when all of the "hosts" lines are missing, the aggregators->endpoints section won't have anything to reference to be able to connect to sampler daemons.
> 
> I think we agreed on something more like this:
> 
> ```
> "hosts" can be specified in a daemon endpoint.
> If no "hosts" key is specified in endpoints, it defaults to the higher level daemon host (one of these _must_ be present)
> "bind_all" (a boolean) can be specified in a daemon enpoint. "bind_all" defaults to False. If "bind_all" is True,
>   the daemon will bind to all available interfaces rather than just the hostnames/IPs listed in "hosts". Other daemons
>   (e.g. aggregators) will use the hostnames/IPs listed in "hosts" to connect to this "endpoint".
> ```

For the bind_all option, do you mean it would bind to all available interfaces on a given node? i.e. interfaces configured outside the scope of the yaml file? I think I may have misunderstood what you were getting at in that conversation, and thought I could address binding the interfaces to a hostname without introducing the bind_all boolean.

The way it's currently implemented, it first looks if hosts are defined in "endpoints", and if they are, that hostname is used. If there are no hosts defined in an endpoint, it references the higher level "daemons" hosts and uses that hostname. If there are no hosts defined in either "daemons" or "endpoints", it will set the hostname for the interfaces to 0.0.0.0. 

I'm not opposed to introducing bind_all, I just want to make sure we're on the same page on how to implement it.

**Comment by nick-enoent** [comment]
> > > Is there a commit to this that will need testing? I want to plan time this week.
> > 
> > 
> > Hey Evan, I've added code to support CLI options in the YAML file and new endpoints "hosts" functionality we discussed. These should be the last commits that need testing.
> 
> I think the "Update host behavior" patch differs a bit from what we decided. We were supposed to have something like:
> 
> ` bind_all: True`
> 
> under `daemons->interfaces`, right? With the current commit's implementation, when all of the "hosts" lines are missing, the aggregators->endpoints section won't have anything to reference to be able to connect to sampler daemons.
> 
> I think we agreed on something more like this:
> 
> ```
> "hosts" can be specified in a daemon endpoint.
> If no "hosts" key is specified in endpoints, it defaults to the higher level daemon host (one of these _must_ be present)
> "bind_all" (a boolean) can be specified in a daemon enpoint. "bind_all" defaults to False. If "bind_all" is True,
>   the daemon will bind to all available interfaces rather than just the hostnames/IPs listed in "hosts". Other daemons
>   (e.g. aggregators) will use the hostnames/IPs listed in "hosts" to connect to this "endpoint".
> ```

The more I think about it, the more I agree with @morrone. I'll implement bind_all, but leave the other functionality as is until I get feedback on whether the top level "daemons" hosts should be required or not.

**Comment by nick-enoent** [comment]
> I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter? My understanding is that that would be the final truth in order to specify the target system. You may recall that there could be instances where there are multiple routes from an aggregator to a sampler (a storage network and a compute HSN as an example).

The prdcr_add host= will use whichever "host" is defined in the endpoints section that the "peer" is referencing. Currently the parser does not support defining an interface in daemons>endpoints>hosts for prdcr_add exclusively. I think in that event, you would define a separate "endpoints" dictionary, and reference that in aggregators>peers>endpoints

**Comment by emdonat** [comment]
> > I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter? My understanding is that that would be the final truth in order to specify the target system. You may recall that there could be instances where there are multiple routes from an aggregator to a sampler (a storage network and a compute HSN as an example).
> 
> The prdcr_add host= will use whichever "host" is defined in the endpoints section that the "peer" is referencing. Currently the parser does not support defining an interface in daemons>endpoints>hosts for prdcr_add exclusively. I think in that event, you would define a separate "endpoints" dictionary, and reference that in aggregators>peers>endpoints

Ok understood and tested as functioning as expected.  Thanks!

**Comment by emdonat** [comment]
> > I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter? My understanding is that that would be the final truth in order to specify the target system. You may recall that there could be instances where there are multiple routes from an aggregator to a sampler (a storage network and a compute HSN as an example).
> 
> The prdcr_add host= will use whichever "host" is defined in the endpoints section that the "peer" is referencing. Currently the parser does not support defining an interface in daemons>endpoints>hosts for prdcr_add exclusively. I think in that event, you would define a separate "endpoints" dictionary, and reference that in aggregators>peers>endpoints

This seems to work correctly on the agg side but the sampler now will have an additional listener that will listen on an interface that doesn't necessarily exist.

```
daemons:
- names: "hetchy[1-2,14]"
  hosts: "hetchy[1-2,14]"
  endpoints:
  - names: "hetchy[1-2,14]-mgmt"
    hosts: "hetchy[1-2,14]-mgmt"
    ports: 411
    xprt: sock
    auth:
      name: munge_domain0
      plugin: munge
  - names: "hetchy[1-2,14]-fabric"
    ports: 411
    xprt: fabric
    auth:
      name: munge_domain0
      plugin: munge
  - names: "hetchy[1-2,14]-ib"
    hosts: "hetchy[1-2,14]-ib"
    ports: 411
    xprt: sock
    auth:
      name: munge_domain0
      plugin: munge

- names: "hetchy14-agg"
  hosts: "hetchy14-agg"
  endpoints:
  - names: "hetchy14-mgmt"
    ports: 411
    xprt: sock
    auth:
      name: munge_domain0
      plugin: munge
  - names: "hetchy14-fabric"
    ports: 411
    xprt: fabric
    auth:
      name: munge_domain0
      plugin: munge

aggregators:
- daemons: "hetchy14-agg"
  peers:
  - daemons: "hetchy[1-2,14]"
    endpoints: "hetchy[1-2,14]-mgmt"
    reconnect: 20s
    type: active
    updaters:
      - mode: pull
        interval: "1.0s"
        offset: "10ms"
        sets:
          - regex: "*."
            field: inst
  - daemons: "hetchy[1-2,14]"
    endpoints: "hetchy[1-2,14]-ib"
    hosts: "hetchy[1-2,14]-ib"
    reconnect: 20s
    type: active
    updaters:
      - mode: pull
        interval: "1.0s"
        offset: "10ms"
        sets:
          - regex: "*."
            field: inst

stores:
  sos-meminfo :
    daemons : "hetchy14"
    container : ldms_data
    schema : meminfo
    flush : 10s
    plugin : store_sos1

plugins:
  store_sos1 :
    name : store_sos
    config : ["path=/opt/DATA"]
  meminfo1:
    name: meminfo
    interval: 1s
    config:
    - schema: meminfo
      component_id: "${COMPONENT_ID}"
      producer: "${HOSTNAME}-10001"
      perm: "0777"


samplers:
- daemons: "hetchy[1-2,14]"
  plugins: [ "meminfo1" ]
```
`ldmsd_yaml_parser --ldms_config ./hetchy.yaml --daemon_name hetchy14 --generate-config-path ./ --debug`
```
auth_add name=munge_domain0 plugin=munge
listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-mgmt
listen xprt=fabric port=411 auth=munge_domain0 host=hetchy14
listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-ib
load name=meminfo
config name=meminfo schema=meminfo component_id=${COMPONENT_ID} producer=${HOSTNAME}-10001 perm=0777 instance=bitzer-admin/meminfo
start name=meminfo interval=1000000
```

So the scenario can exist where I have a host definition on hetchy14-agg that indicates a target ip address (e.g. hetchy1-ib) but that isn't necessarily named hetchy1-ib on hetchy1.  I think that we covered this on Friday maybe?  I hope that I am not misrememering?

**Comment by emdonat** [comment]
Is the man page documentation a separate item and commit somewhere?

**Comment by nick-enoent** [comment]
> > > I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter? My understanding is that that would be the final truth in order to specify the target system. You may recall that there could be instances where there are multiple routes from an aggregator to a sampler (a storage network and a compute HSN as an example).
> > 
> > 
> > The prdcr_add host= will use whichever "host" is defined in the endpoints section that the "peer" is referencing. Currently the parser does not support defining an interface in daemons>endpoints>hosts for prdcr_add exclusively. I think in that event, you would define a separate "endpoints" dictionary, and reference that in aggregators>peers>endpoints
> 
> This seems to work correctly on the agg side but the sampler now will have an additional listener that will listen on an interface that doesn't necessarily exist.
> 
> ```
> daemons:
> - names: "hetchy[1-2,14]"
>   hosts: "hetchy[1-2,14]"
>   endpoints:
>   - names: "hetchy[1-2,14]-mgmt"
>     hosts: "hetchy[1-2,14]-mgmt"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
>   - names: "hetchy[1-2,14]-fabric"
>     ports: 411
>     xprt: fabric
>     auth:
>       name: munge_domain0
>       plugin: munge
>   - names: "hetchy[1-2,14]-ib"
>     hosts: "hetchy[1-2,14]-ib"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
> 
> - names: "hetchy14-agg"
>   hosts: "hetchy14-agg"
>   endpoints:
>   - names: "hetchy14-mgmt"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
>   - names: "hetchy14-fabric"
>     ports: 411
>     xprt: fabric
>     auth:
>       name: munge_domain0
>       plugin: munge
> 
> aggregators:
> - daemons: "hetchy14-agg"
>   peers:
>   - daemons: "hetchy[1-2,14]"
>     endpoints: "hetchy[1-2,14]-mgmt"
>     reconnect: 20s
>     type: active
>     updaters:
>       - mode: pull
>         interval: "1.0s"
>         offset: "10ms"
>         sets:
>           - regex: "*."
>             field: inst
>   - daemons: "hetchy[1-2,14]"
>     endpoints: "hetchy[1-2,14]-ib"
>     hosts: "hetchy[1-2,14]-ib"
>     reconnect: 20s
>     type: active
>     updaters:
>       - mode: pull
>         interval: "1.0s"
>         offset: "10ms"
>         sets:
>           - regex: "*."
>             field: inst
> 
> stores:
>   sos-meminfo :
>     daemons : "hetchy14"
>     container : ldms_data
>     schema : meminfo
>     flush : 10s
>     plugin : store_sos1
> 
> plugins:
>   store_sos1 :
>     name : store_sos
>     config : ["path=/opt/DATA"]
>   meminfo1:
>     name: meminfo
>     interval: 1s
>     config:
>     - schema: meminfo
>       component_id: "${COMPONENT_ID}"
>       producer: "${HOSTNAME}-10001"
>       perm: "0777"
> 
> 
> samplers:
> - daemons: "hetchy[1-2,14]"
>   plugins: [ "meminfo1" ]
> ```
> 
> `ldmsd_yaml_parser --ldms_config ./hetchy.yaml --daemon_name hetchy14 --generate-config-path ./ --debug`
> 
> ```
> auth_add name=munge_domain0 plugin=munge
> listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-mgmt
> listen xprt=fabric port=411 auth=munge_domain0 host=hetchy14
> listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-ib
> load name=meminfo
> config name=meminfo schema=meminfo component_id=${COMPONENT_ID} producer=${HOSTNAME}-10001 perm=0777 instance=bitzer-admin/meminfo
> start name=meminfo interval=1000000
> ```
> 
> So the scenario can exist where I have a host definition on hetchy14-agg that indicates a target ip address (e.g. hetchy1-ib) but that isn't necessarily named hetchy1-ib on hetchy1. I think that we covered this on Friday maybe? I hope that I am not misrememering?

I believe that the target ip address is defined in the endpoints section of hetchy1. That said, this example was solely for the configuration of an aggregator daemon, i.e. hetchy14, so the configuration for the sampler isn't in this YAML file. All it contains is _how_ to talk to a sampler LDMSD, that we are assuming is configured elsewhere

**Comment by nick-enoent** [comment]
> > > I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter? My understanding is that that would be the final truth in order to specify the target system. You may recall that there could be instances where there are multiple routes from an aggregator to a sampler (a storage network and a compute HSN as an example).
> > 
> > 
> > The prdcr_add host= will use whichever "host" is defined in the endpoints section that the "peer" is referencing. Currently the parser does not support defining an interface in daemons>endpoints>hosts for prdcr_add exclusively. I think in that event, you would define a separate "endpoints" dictionary, and reference that in aggregators>peers>endpoints
> 
> This seems to work correctly on the agg side but the sampler now will have an additional listener that will listen on an interface that doesn't necessarily exist.
> 
> ```
> daemons:
> - names: "hetchy[1-2,14]"
>   hosts: "hetchy[1-2,14]"
>   endpoints:
>   - names: "hetchy[1-2,14]-mgmt"
>     hosts: "hetchy[1-2,14]-mgmt"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
>   - names: "hetchy[1-2,14]-fabric"
>     ports: 411
>     xprt: fabric
>     auth:
>       name: munge_domain0
>       plugin: munge
>   - names: "hetchy[1-2,14]-ib"
>     hosts: "hetchy[1-2,14]-ib"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
> 
> - names: "hetchy14-agg"
>   hosts: "hetchy14-agg"
>   endpoints:
>   - names: "hetchy14-mgmt"
>     ports: 411
>     xprt: sock
>     auth:
>       name: munge_domain0
>       plugin: munge
>   - names: "hetchy14-fabric"
>     ports: 411
>     xprt: fabric
>     auth:
>       name: munge_domain0
>       plugin: munge
> 
> aggregators:
> - daemons: "hetchy14-agg"
>   peers:
>   - daemons: "hetchy[1-2,14]"
>     endpoints: "hetchy[1-2,14]-mgmt"
>     reconnect: 20s
>     type: active
>     updaters:
>       - mode: pull
>         interval: "1.0s"
>         offset: "10ms"
>         sets:
>           - regex: "*."
>             field: inst
>   - daemons: "hetchy[1-2,14]"
>     endpoints: "hetchy[1-2,14]-ib"
>     hosts: "hetchy[1-2,14]-ib"
>     reconnect: 20s
>     type: active
>     updaters:
>       - mode: pull
>         interval: "1.0s"
>         offset: "10ms"
>         sets:
>           - regex: "*."
>             field: inst
> 
> stores:
>   sos-meminfo :
>     daemons : "hetchy14"
>     container : ldms_data
>     schema : meminfo
>     flush : 10s
>     plugin : store_sos1
> 
> plugins:
>   store_sos1 :
>     name : store_sos
>     config : ["path=/opt/DATA"]
>   meminfo1:
>     name: meminfo
>     interval: 1s
>     config:
>     - schema: meminfo
>       component_id: "${COMPONENT_ID}"
>       producer: "${HOSTNAME}-10001"
>       perm: "0777"
> 
> 
> samplers:
> - daemons: "hetchy[1-2,14]"
>   plugins: [ "meminfo1" ]
> ```
> 
> `ldmsd_yaml_parser --ldms_config ./hetchy.yaml --daemon_name hetchy14 --generate-config-path ./ --debug`
> 
> ```
> auth_add name=munge_domain0 plugin=munge
> listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-mgmt
> listen xprt=fabric port=411 auth=munge_domain0 host=hetchy14
> listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-ib
> load name=meminfo
> config name=meminfo schema=meminfo component_id=${COMPONENT_ID} producer=${HOSTNAME}-10001 perm=0777 instance=bitzer-admin/meminfo
> start name=meminfo interval=1000000
> ```
> 
> So the scenario can exist where I have a host definition on hetchy14-agg that indicates a target ip address (e.g. hetchy1-ib) but that isn't necessarily named hetchy1-ib on hetchy1. I think that we covered this on Friday maybe? I hope that I am not misrememering?

Are you positing that we should include a boolean that lets the parser know that we don't want to include a listen for a given endpoint?

**Comment by nick-enoent** [comment]
> Is the man page documentation a separate item and commit somewhere?

I'm in the process of writing it, so it will be a separate commit. The current documentation we have is included in the google drive https://drive.google.com/file/d/1LwwmKsFKCIeV1FsIgZ_l3JFYUO_VHTHP/view but it needs to be updated with our agreed upon changes.

**Comment by emdonat** [comment]
> > > > I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter? My understanding is that that would be the final truth in order to specify the target system. You may recall that there could be instances where there are multiple routes from an aggregator to a sampler (a storage network and a compute HSN as an example).
> > > 
> > > 
> > > The prdcr_add host= will use whichever "host" is defined in the endpoints section that the "peer" is referencing. Currently the parser does not support defining an interface in daemons>endpoints>hosts for prdcr_add exclusively. I think in that event, you would define a separate "endpoints" dictionary, and reference that in aggregators>peers>endpoints
> > 
> > 
> > This seems to work correctly on the agg side but the sampler now will have an additional listener that will listen on an interface that doesn't necessarily exist.
> > ```
> > daemons:
> > - names: "hetchy[1-2,14]"
> >   hosts: "hetchy[1-2,14]"
> >   endpoints:
> >   - names: "hetchy[1-2,14]-mgmt"
> >     hosts: "hetchy[1-2,14]-mgmt"
> >     ports: 411
> >     xprt: sock
> >     auth:
> >       name: munge_domain0
> >       plugin: munge
> >   - names: "hetchy[1-2,14]-fabric"
> >     ports: 411
> >     xprt: fabric
> >     auth:
> >       name: munge_domain0
> >       plugin: munge
> >   - names: "hetchy[1-2,14]-ib"
> >     hosts: "hetchy[1-2,14]-ib"
> >     ports: 411
> >     xprt: sock
> >     auth:
> >       name: munge_domain0
> >       plugin: munge
> > 
> > - names: "hetchy14-agg"
> >   hosts: "hetchy14-agg"
> >   endpoints:
> >   - names: "hetchy14-mgmt"
> >     ports: 411
> >     xprt: sock
> >     auth:
> >       name: munge_domain0
> >       plugin: munge
> >   - names: "hetchy14-fabric"
> >     ports: 411
> >     xprt: fabric
> >     auth:
> >       name: munge_domain0
> >       plugin: munge
> > 
> > aggregators:
> > - daemons: "hetchy14-agg"
> >   peers:
> >   - daemons: "hetchy[1-2,14]"
> >     endpoints: "hetchy[1-2,14]-mgmt"
> >     reconnect: 20s
> >     type: active
> >     updaters:
> >       - mode: pull
> >         interval: "1.0s"
> >         offset: "10ms"
> >         sets:
> >           - regex: "*."
> >             field: inst
> >   - daemons: "hetchy[1-2,14]"
> >     endpoints: "hetchy[1-2,14]-ib"
> >     hosts: "hetchy[1-2,14]-ib"
> >     reconnect: 20s
> >     type: active
> >     updaters:
> >       - mode: pull
> >         interval: "1.0s"
> >         offset: "10ms"
> >         sets:
> >           - regex: "*."
> >             field: inst
> > 
> > stores:
> >   sos-meminfo :
> >     daemons : "hetchy14"
> >     container : ldms_data
> >     schema : meminfo
> >     flush : 10s
> >     plugin : store_sos1
> > 
> > plugins:
> >   store_sos1 :
> >     name : store_sos
> >     config : ["path=/opt/DATA"]
> >   meminfo1:
> >     name: meminfo
> >     interval: 1s
> >     config:
> >     - schema: meminfo
> >       component_id: "${COMPONENT_ID}"
> >       producer: "${HOSTNAME}-10001"
> >       perm: "0777"
> > 
> > 
> > samplers:
> > - daemons: "hetchy[1-2,14]"
> >   plugins: [ "meminfo1" ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ldmsd_yaml_parser --ldms_config ./hetchy.yaml --daemon_name hetchy14 --generate-config-path ./ --debug`
> > ```
> > auth_add name=munge_domain0 plugin=munge
> > listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-mgmt
> > listen xprt=fabric port=411 auth=munge_domain0 host=hetchy14
> > listen xprt=sock port=411 auth=munge_domain0 host=hetchy14-ib
> > load name=meminfo
> > config name=meminfo schema=meminfo component_id=${COMPONENT_ID} producer=${HOSTNAME}-10001 perm=0777 instance=bitzer-admin/meminfo
> > start name=meminfo interval=1000000
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So the scenario can exist where I have a host definition on hetchy14-agg that indicates a target ip address (e.g. hetchy1-ib) but that isn't necessarily named hetchy1-ib on hetchy1. I think that we covered this on Friday maybe? I hope that I am not misrememering?
> 
> I believe that the target ip address is defined in the endpoints section of hetchy1. That said, this example was solely for the configuration of an aggregator daemon, i.e. hetchy14, so the configuration for the sampler isn't in this YAML file. All it contains is _how_ to talk to a sampler LDMSD, that we are assuming is configured elsewhere
> 
> Are you positing that we should include a boolean that lets the parser know that we don't want to include a listen for a given endpoint?

In the included yaml there is an aggregator defined to run as hetchy14-agg and an additional ldmsd set to run as a sampler as hetchy14.  I will let Chris M reply on the implementation because I believe he was the major driver in our talk on Friday and I don't want to muddy anything further.

**Comment by morrone** [comment]
> I just want to confirm that it is expected that we can't specify an interface in daemons>endpoints>hosts for the prdr_add host= parameter?

Correct. One of the things I proposed earlier in this thread was to add that. But on our recent phone conversation no one seemed to be on board with that. It does mean that we have to have a "daemon" configuration section even for daemons that a config file won't necessarily be configuring (just so other parts of the config have access to that information).  (And an apology to @emdonat , I think you started to ask a question about that towards the end of the call, and I was like "I'm not proposing that", but you were probably asking a broader question and I was stuck in the immediate detail suggestion...I did of course propose that at one point, but I dropped it when folks were opposed to it.)

So we agreed on the "bind_all" solution instead (which technically doesn't cover 100% of the possible configuration space, but is very likely complete enough in practice, so i won't even get into the details on that).

> So the scenario can exist where I have a host definition on hetchy14-agg that indicates a target ip address (e.g. hetchy1-ib) but that isn't necessarily named hetchy1-ib on hetchy1. I think that we covered this on Friday maybe? I hope that I am not misrememering?

I'm sorry, I am not following this. The port numbers at least need fixing to make things work (admittedly made the same mistake one of the early examples that I shared, but I think I corrected those in the follow up example). I don't understand what you are saying about hetchy1-ib not existing on hetchy1. That just sounds like a configuration error. If we put in hostnames that don't exist the configuration won't work. What I am I missing?

**Comment by emdonat** [comment]
> > So the scenario can exist where I have a host definition on hetchy14-agg that indicates a target ip address (e.g. hetchy1-ib) but that isn't necessarily named hetchy1-ib on hetchy1. I think that we covered this on Friday maybe? I hope that I am not misrememering?
> 
> I'm sorry, I am not following this. The port numbers at least need fixing to make things work (admittedly made the same mistake one of the early examples that I shared, but I think I corrected those in the follow up example). I don't understand what you are saying about hetchy1-ib not existing on hetchy1. That just sounds like a configuration error. If we put in hostnames that don't exist the configuration won't work. What I am I missing?

Ok so maybe I am creating an issue out of thin air...  My position is that we should be able to just let the routing to the needful on any system to ensure it uses the correct outgoing interface on the aggregator.  It sounds like you do too so - if so this is solved/not an issue.

**Comment by morrone** [comment]
> Ok so maybe I am creating an issue out of thin air... My position is that we should be able to just let the routing to the needful on any system to ensure it uses the correct outgoing interface on the aggregator. It sounds like you do too so - if so this is solved/not an issue.

I don't think we ever have _direct_ control of outbound routing from a program. We can only select the target hostname/IP, which might result in a different outbound route. On our clusters at least, we _do_ have things set up that way. Nodes each have multiple hostnames, one for each network they are on, so using a different hostname on for the same node will result in a different network being used to get to that node.

Since you have questions and the proposed implementation doesn't have the bind_all, I am concerned that we didn't really get on the same page like I thought. Should I take a stab at rewording the need for either bind_all or target hostname/IP under aggregators->peers->endpoints?

**Comment by nick-enoent** [comment]
> > Ok so maybe I am creating an issue out of thin air... My position is that we should be able to just let the routing to the needful on any system to ensure it uses the correct outgoing interface on the aggregator. It sounds like you do too so - if so this is solved/not an issue.
> 
> I don't think we ever have _direct_ control of outbound routing from a program. We can only select the target hostname/IP, which might result in a different outbound route. On our clusters at least, we _do_ have things set up that way. Nodes each have multiple hostnames, one for each network they are on, so using a different hostname on for the same node will result in a different network being used to get to that node.
> 
> Since you have questions and the proposed implementation doesn't have the bind_all, I am concerned that we didn't really get on the same page like I thought. Should I take a stab at rewording the need for either bind_all or target hostname/IP under aggregators->peers->endpoints?

I implemented the bind_all, and am currently testing it. I also made it a requirement that at least daemons->hosts or daemons->endpoints->hosts exists. 

In the event of `bind_all : True` in daemons->endpoints, the ldmsd will listen on 0.0.0.0, but the aggregator will either use the host in daemons->hosts or daemons->endpoints->hosts when utilizing prdcr_add host=

That said I'm happy to have a discussion to make sure we've cleared the air.

**Comment by morrone** [comment]
bind_all sees to be working.

I wonder if we should just leave off the "host=" when in the output instead of doing "host=0.0.0.0", but it probably doesn't matter in practice.

```
daemons:
    environment:
      LDMSD_MEM_SZ: 100MB
```
 is generating this:

```
env LDMSD_MEM_SZ={'LDMSD_MEM_SZ': '100MB'}
```

I can't find documentation for the "env" command, but I am guessing that is not quite right.

**Comment by morrone** [comment]
> Thanks for jumping on this. I pushed a fix to address the environment issue you experienced.

Yes, that looks good now, thanks.

File: ldms/python/ldmsd/ldmsd_yaml_parser [hunk 11]
```diff
@@ -0,0 +1,47 @@
...
+import yaml
+import errno
+import argparse
+from ldmsd.parser_util import *
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="LDMS Monitoring Cluster Configuration")
+    parser.add_argument("--ldms_config", metavar="FILE", required=True,
+                        help="The ldmsd load balance domain configuration file. "
```
nichamon, line 12: @nick-enoent I believe the help string of `--ldms-config` may confuse readers as this is in the `ldmsd` proper. It is the argument that receives the path to the YAML configuration file, isn't it? My point is that this has nothing to do with load balancing in Maestro.

ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -0,0 +1,47 @@
...
+                        "This will not start the maestro "
+                        "load balancer")
+    parser.add_argument("--generate-config-path", metavar="STRING", required=False,
+                        default=False)
```
nichamon, line 16: Maybe you want to add the help string for `--generate-config-path`?

ldms/python/ldmsd/ldmsd_yaml_parser
nick-enoent: Good catch, thanks Mon! I updated this in [9b4113a](https://github.com/ovis-hpc/ldms/pull/1461/commits/9b4113a2e362def38eca1543eb71d2e326483e58)

ldms/python/ldmsd/ldmsd_yaml_parser
nick-enoent: Updated in [9b4113a](https://github.com/ovis-hpc/ldms/pull/1461/commits/9b4113a2e362def38eca1543eb71d2e326483e58)

tom95858:
@nick-enoent could you please squash these commits?

nichamon (approved):

nick-enoent:
> @nick-enoent could you please squash these commits?

Squashed. Running build tests, and this should be good to go.

PR #1462, opened 2024-10-09
Fix SIGSEGV in ldms_stream due to NULL sbuf

narategithub:
The code in `__remote_client_cb()` mistakenly assumed that `sbuf` is always available. 'sbuf' could be NULL for the stream publishing originating from the local process. In such case, we shall not queue the stream message and return with the error (application back-pressure).


tom95858:
@narategithub why would the subsystem publish no data?

PR #1463, opened 2024-10-09
Ignore LDMS_XPRT_EVENT_SEND_QUOTA_DEPOSITED event in ldmsd

nichamon:


nichamon:
@narategithub While testing the profiling compatibility fix, I hit an assert() when an aggregator received the SEND_QUOTA_DEPOSITED event in `__agg_routine()`. FYI, I didn't run into this assertion when I tested the profiling code a couple of weeks back.

tom95858 (changes requested):

tom95858:
@narategithub did you review this? Who is supposed to respond to this event?

tom95858:
@narategithub is this the event that a stalled stream client would use to re-send a message it got an EWOULDBLOCK/EAGAIN on?

narategithub:
@tom95858 I think @nichamon handle this correctly on `ldmsd`. The stream-related logic that handle the quota deposit was baked into LDMS library. This is just an event to let application know about quota deposition after LDMS library handled it so that the application may try to re-publish the message that got back-pressured. Currently in `ldmsd`, no one uses this event.

narategithub:
> @narategithub is this the event that a stalled stream client would use to re-send a message it got an EWOULDBLOCK/EAGAIN on?

Yes.

If the publisher is an application outside `ldmsd`, it will just use this event to try to re-send (or use ask ACK like slurm_notifier).

I think we don't handle the case that the publisher is a plugin inside `ldmsd` yet :sob: .

PR #1465, opened 2024-10-11
Perform validity checks on set name

tom95858:
The set name provided to ldms_set_create cannot contain characters that cannot be encoded in a quoted string. Prior to this patch it was possible to create a set that would crash ldms_set_dir because the result is encoded as a JSON object and the instance name as a JSON string.


nichamon (approved):

narategithub (approved):
:+1:

baallan:
This change should include an accompanying man pages change about what is allowed for schema/instance/metric names, unless that's in a concurrent PR.

PR #1480, opened 2024-10-25
b4.4 deprecation fix: fix warnings about using reconnect instead of interval

baallan:
the following bit of sed came handy in updating scripts:
```
sed -i -E -e 's%(^prdcr_add) (.*) interval=%\1 \2 reconnect=%' *
```


baallan:
need to make a cherrypick version of this for 4.5

tom95858:
@baallan, I think @valleydlr would prefer if the examples used "reconnect_interval" instead of "reconnect". @nichamon would you please confirm that we support either "reconnect" or "reconnect_interval" in b4.4?

baallan:
This patch is formulated based on the deprecation message seen. If we want something else, we need to update the deprecation message. @nichamon let me know if there's a final conclusion for 4.4 and if it will differ for main.

tom95858:
@baallan, there has been discussion about renaming 'interval' --> 'sample_interval'. @nichamon, It seems like we should have this discussion in the UG meeting on Monday. The advocates of this change argue that there would be symmetry in the naming. By that logic, however, name would become 'producer_name', etc... In other words, a context free grammar. I see that as fundamentally reductive, but let's have the discussion.

baallan:
I would be in favor of minimizing breaking changes to the spelling of existing things other than the ambiguous 'interval' until an ldms 5.0 where we might decide to use full spelling of words and more context clue prefixes instead of all the arbitrary letter omissions. If somewhere in 4.x we want to introduce a bunch of aliasing tables (prdcr|producer) which do not break any existing configuration scripts, that's fine by me.

baallan:
@nichamon this is ready to merge after some patch that adds reconnect_interval is merged.

emdonat:
Work will be done by Ben A. prior to merge.

baallan:
> Work will be done by Ben A. prior to merge.

work has been done by Ben A. now using reconnect. merge away. @tom95858

PR #1481, opened 2024-10-25
Address mixed use of ldmsd_log() and ovis_log()

### LINKED ISSUE #1471: Mixed Usage of ovis_log() and ldmsd_log() May Affect Log File Readability in b4.4
**Current Situation**
Throughout ldmsd, ldmsd_log() is the primary function used for error reporting. However, some ldmsd plugins and the auth plugins call ovis_log() instead - a result of code being back-ported from the top of tree where ovis_log() is used throughout. ldmsd_log() prints error messages to ldmsd's log file. ovis_log() prints the messages to stdout, and ldmsd redirects them to the log file.
**Issue**
Code inspection shows that this setup could potentially jumble error messages printed to the log file.
**Goal**
Ensure readable error messages are reported in the log file.

nichamon:
The pull request resolves #1471 .

### DIFF
--- a/lib/src/ovis_log/ovis_log.c
+++ b/lib/src/ovis_log/ovis_log.c
@@ -651,30 +651,25 @@ static int __log(ovis_log_t log, int level, char *msg,
 
 	int rc;
 	FILE *f;
+	char log_time[200] = "";
+
 	if (!log_fp)
 		f = stdout;
 	else
 		f = log_fp;
+
 	if (default_modes & OVIS_LOG_M_TS) {
-		rc = fprintf(f, "%lu.%06lu:", tv->tv_sec, tv->tv_usec);
+		snprintf(log_time, 199, "%lu.%06lu:", tv->tv_sec, tv->tv_usec);
 	} else if (default_modes & OVIS_LOG_M_DT) {
-		char dtsz[200];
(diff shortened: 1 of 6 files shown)


lib/src/ovis_log/ovis_log.c
```diff
@@ -945,6 +940,8 @@ int ovis_vlog(ovis_log_t log, int level, const char *fmt, va_list ap)
 		/* No workers, so directly log to the file. */
 		rc = __log(log, level, msg, &tv, &tm);
 		free(msg);
+		if (log_fp != OVIS_LOG_SYSLOG)
```
tom95858, line 943: I don't understand this condition.

nichamon: The if-statement prevents calling `fflush()` on an invalid `FILE *` stream, which is `OVIS_LOG_SYSLOG` (0x7). 

When `ovis_log_open()` is called with `syslog` as the path, I set `log_fp` to `OVIS_LOG_SYSLOG`. After investigation, I believe a better approach would be to let applications specify syslog output during initialization through the `modes` argument of the `ovis_log_init()` function. Since `modes` is a mask, we could add a new flag for syslog support. Then, the if-statement will check if the syslog mode is enabled.

The patch is going into b4.4. If we agree on chaning the syslog handling in ovis_log, I could implement the change for the top of tree.

PR #1485, opened 2024-10-29
wip: b4.4 bug: blob_stream_writer plugin: stop generating empty files

baallan:
For rollover periods where no messages arrive to store, remove the  empty output files instead of filling up the disk.


baallan:
this needs more testing.

PR #1486, opened 2024-10-29
b4.4 bug: trap failed json conversion in stream_client_dump

### LINKED ISSUE #1475: stream_client_dump causes ldmsd_controller to exit instead of returning an error message explaining why client dump isn't permitted.
if you are not root:
```
name                           bytes/sec    msg/sec      total bytes  msg count    first msg    last msg
------------------------------ ------------ ------------ ------------ ------------ ------------ ------------
darshanConnector (subscribed)
   published                              -            -            -            -            -            -
   received                               -            -            -            -            -            -
ldms-notify (subscribed)
   published                              -            -            -            -            -            -
   received                               -            -            -            -            -            -
slurm (subscribed)
   published                      16.664868     0.100164      1720331        10340   1729537684   1729640915
   received                       16.664868     0.100164      1720331        10340   1729537684   1729640915
syspapi_stream (subscribed)
   published                              -            -            -            -            -            -
   received                               -            -            -            -            -            -
_OVERALL_
   published                      16.664868     0.100164      1720331        10340   1729537684   1729640915
   received                               -            -            -            -            -            -
sock:localhost:412> stream_client_dump
Error converting Operation not permitted. to json object
stream          ctxt            cb_fn                                                       
--------------- --------------- ----------------------------------------------------------------------
'NoneType' object is not subscriptable
```

baallan:
fix #1475

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -1967,6 +1967,9 @@ class LdmsdCmdParser(cmd.Cmd):
         if not msg:
             raise RuntimeError("no response")
         obj = fmt_status(msg)
+        if obj is None:
+            print(msg)
+            return
         print(f"{'stream':15} {'ctxt':15} {'cb_fn':70}")
         print(f"{'-'*15} {'-'*15} {'-'*70}")
         for s in obj['streams']:


tom95858:
@baallan, I think we need to look at fixing this more generally. The fmt_status function is called a couple dozen times inside ldmsd_controller. The fmt_status() failure is the result of the server (*ldmsd*) returning an error instead of a valid result. Most of the code patterns are:

```Python
rc, msg = self.conn.some_status_function(...)
```
If the rc != 0, then the result is an error message and should not be assumed to be JSON. The caller can then return without attempting to decode the string. The issue with this is that this convention is not uniformly implemented in ldmsd_controller, e.g. in the code you changed. The bug was not that the JSON couldn't be decoded, but that the rc was not consulted before attempting to do so. To fix this properly, we need to look through the 23 fmt_status() call sites and ensure that the rc is always consulted.

baallan:
@tom95858 i did notice that fmt_status is used rather inconsistently, but that many sites had *some* kind of guard code. Picking a preferred handling style and refactoring the entire file to apply it is beyond the scope of both the reported bug and the fix. Let's not let better be the enemy of good in this case.    If you are closing this because ogc will be applying a better fix during the current release milestone, then that is fine.

PR #1487, opened 2024-10-29
b4.4 bug: Store csv lents

### LINKED ISSUE #1073: store_csv leaks s_handle->lents allocated at line 1516
By inspection, no free for lents occurs, and by test valgrind agrees.

baallan:
fixes #1073


nichamon:
Thanks for the patch! The fix looks good. I think the commit message could be more descriptive. Something like this might be helpful: ` store_csv: Free list entries on store closure`.

Issue #1443, opened 2024-09-06
ldmsd -t exits unless a log to be truncated is preexisting.


baallan:
in b4.4 at least starting with ldmsd -t generates 'Could not truncated the log file named '*'. errno=2

baallan:
@tom95858 or @nichamon please review

tom95858:
It looks like this is an issue where the implementation does not check if the file already exists.

baallan:
fixed in #1488

emdonat:
Resolved with #1488

PR #1488, opened 2024-10-29
b4.4 bug: deprecate -t since it is gone in main, mixes poorly with -l, and exits when there is no preexisting log

baallan:
truncating the log now works correctly whether the old log exists or not and independent of the ordering of -l and -t commands.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -131,7 +131,6 @@ int foreground;
 int cfg_cntr = 0;
 pthread_t event_thread = (pthread_t)-1;
 char *logfile;
-int log_truncate = 0;
 char *pidfile;
 char *bannerfile;
 
@@ -553,15 +552,6 @@ FILE *ldmsd_open_log()
 		return f;
 	}
 
-	if (log_truncate) {
-		int err = truncate(logfile, 0);
-		if (err) {
-			ldmsd_log(LDMSD_LERROR, "Could not truncate the log file named '%s'. errno=%d\n",
-				logfile, errno);
-			cleanup(12, "log truncate failed");
-		}
-	}
-
 	f = fopen_perm(logfile, "a", LDMSD_DEFAULT_FILE_PERM);
 	if (!f) {
 		ldmsd_log(LDMSD_LERROR, "Could not open the log file named '%s'\n",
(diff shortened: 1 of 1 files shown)


baallan:
fixes #1443

ldms/src/ldmsd/ldmsd.c
```diff
@@ -2116,7 +2116,7 @@ int ldmsd_process_cmd_line_arg(char opt, char *value)
 		}
 		break;
 	case 't':
-		log_truncate = 1;
+		/* already done log_truncate = 1; */
```
nichamon, line 2119: This will break the case that -t is given in configuration files, i.e., `ldmsd_process_cmd_line_arg()` is called by `__cmdline_options()` in ldmsd_request.c.

I believe an alternative solution is to call `ldmsd_open_log()` after processing the command line options in configuration files at line 2315 before the comment `/* Initialize LDMS */`. This could also avoid pre-scaning for `-t`.

nichamon (changes requested):
While the change fixes the order dependency of `-t` and `-l` when they are specified at the command line, `-t` won't be supported when it is given in configuration files.

ldms/src/ldmsd/ldmsd.c
baallan: @nichamon Good catch. The tricky bit is that deferring open-log until after processing all the options (and emitting all the associated warnings/errors to something other than the -l specified log) means the user doesn't get the diagnostics about other arguments.
What do we do instead? A prescan for -t in both main and __cmdline_options()?

tom95858:
@baallan, @nichamon this 'feature' has been removed from top of tree. Should we consider getting rid of this in 4.4? I know it's convenient to crush the log file at start up, but it can trivially be achieved in other ways with the new 'rm' command ;-)

baallan:
@tom95858 If it is gone from main, I'm fine with making it a no-op+warning  in 4.4. Just deleting it could break working scripts. But in theory this should be a community decision?

baallan:
@tom95858 revised this pr to deprecate -t.

nichamon:
@baallan Thanks for making the revisions. Given that your latest commit removes the `-t` option completely, would you consider dropping the first commit dcc50b5?

baallan:
@nichamon if we could just check the squash box on the pr, that should take care of it. I don't have any pull/merge controls on the github repo, however.

baallan:
@nichamon found the branch and squashed it.

PR #1490, opened 2024-10-30
[b4.4] Fix static decomposition

narategithub:
This is the same as #1456, but for b4.4.


tom95858:
@morrone, could you please check if this change affects your configuration?

morrone:
Was about to...

morrone:
@tom95858 this breaks decomposition. See issue #1495.

PR #1491, opened 2024-10-30
issue1489: for ldmsd_process_config_request, if hostname does not res…

jstile-lbl:
issue1489: for ldmsd_process_config_request, if hostname does not resolve skip host, non-fatal

ldmsd should not fail if a hostname is not resolved.


tom95858:
@jstile-lbl I think @nichamon has published a more complete fix for this. Could you please review her change and then close this PR if you think her change is sufficient?

tom95858:
@jstile-lbl please reopen if @nichamon fix isn't sufficient.

PR #1492, opened 2024-10-31
Make ldmsd continue startup on producer's hostname resolution failure

### LINKED ISSUE #1489: ldmsd aggregator dies when a hostname of a sampler node can't be resolved
ldmsd aggregator dies when a hostname of a sampler node can't be resolved 

```
ldmsd_prdcr_new: my_bad_hostname:6002 not resolved.
msg_no 4654: error 97: Error resolving hostname 'my_bad_hostname:'
```

I would like for ldmsd to continue working and simply kick the bad host out of the list.

Version: ovis-4.4.2

nichamon:
Change producer's hostname resolution failure at config time from an error to a warning, allowing ldmsd to continue starting up. The hostname will be resolved again before establishing the connection and retired at the reconnect interval if the resolution fails.

Resolves #1489

### DIFF
--- a/ldms/src/ldmsd/ldmsd_failover.c
+++ b/ldms/src/ldmsd/ldmsd_failover.c
@@ -440,6 +440,12 @@ int __failover_send_prdcr(ldmsd_failover_t f, ldms_t x, ldmsd_prdcr_t p)
 	if (rc)
 		goto cleanup;
 
+	/* Cache_ip */
+	snprintf(buff, sizeof(buff), "%d", p->cache_ip);
+	rc = ldmsd_req_cmd_attr_append_str(rcmd, LDMSD_ATTR_IP, buff);
+	if (rc)
+		goto cleanup;
+
 	/* Terminate the message */
 	rc = ldmsd_req_cmd_attr_term(rcmd);
 	if (rc)
@@ -2025,6 +2031,7 @@ int failover_cfgprdcr_handler(ldmsd_req_ctxt_t req)
 	char *perm = __req_attr_gets(req, LDMSD_ATTR_PERM);
 	char *auth = __req_attr_gets(req, LDMSD_ATTR_AUTH);
 	char *stream = __req_attr_gets(req, LDMSD_ATTR_STREAM);
+	char *cache_ip = __req_attr_gets(req, LDMSD_ATTR_IP);
 
 	uid_t _uid;
 	gid_t _gid;
@@ -2079,7 +2086,7 @@ int failover_cfgprdcr_handler(ldmsd_req_ctxt_t req)
 	}
 
 	p = ldmsd_prdcr_new_with_auth(name, xprt, host, atoi(port), ptype,
-			atoi(interval), auth, _uid, _gid, _perm);
+			atoi(interval), auth, _uid, _gid, _perm, atoi(cache_ip));
 	if (!p) {
 		rc = errno;
 		str_rbn_free(srbn);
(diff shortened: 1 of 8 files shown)


ldms/src/ldmsd/ldmsd_prdcr.c
```diff
@@ -845,6 +845,17 @@ static void prdcr_connect(ldmsd_prdcr_t prdcr)
 {
 	int ret;
 
+	if (prdcr_resolve(prdcr->host_name, prdcr->port_no, &prdcr->ss, &prdcr->ss_len)) {
+		if (0 == prdcr->not_resolved_cnt) {
+			ldmsd_log(LDMSD_LERROR, "Producer '%s' connection failed. Unable to resolve hostname '%s:%u'.\n",
+				prdcr->obj.name, prdcr->host_name,(unsigned) prdcr->port_no);
+		}
+		prdcr->not_resolved_cnt++;
```
tom95858, line 853: @nichamon, We should talk about this. I don't agree that we should only report this once. The service comes and goes and we only have one user who complained about the possibility of their configuration error being reported over and over. Let's consult with other users in the UG meeting on Monday.

tom95858:
@jstile-lbl, this is the change that modifies the handling of unresolved host names causing the daemon to exit. Please weigh in on wether the error should be reported just once or every reconnect interval. There will be a reconnect error anyway, it seems like the administrator should know why the reconnect failed.

johnstile:
By the time the administrator sees the logs, the original name resolution error may have rotated out, so I think the name resolution error should be present in each reconnect.

Additionally one could identify when the name resolution is fixed by filtering for these errors.

nichamon:
@nichamon will make the following changes to the patch:
- [ ] Make IP cache configurable. Caching IP is the default. This is to support cloud environment that the IPs of compute nodes change when compute nodes are up.
- [ ] If caching IP is enabled, ldmsd will not resolve sampler's hostname again after it is resolved
- [ ] Make ldmsd report error messages at every reconnect until the hostname can be resolved

nichamon:
The revision also addresses the problem that `gethostbyname()` is not thread-safe by using `getaddrinfo()`. There is one single thread that handles configuration files. However, there may be multiple threads that connect to sampler daemons.

nichamon:
@narategithub The patch touched the failover paths. Please help me make sure that I didn't miss anything.

ldms/src/ldmsd/ldmsd_failover.c
```diff
@@ -2079,7 +2086,7 @@ int failover_cfgprdcr_handler(ldmsd_req_ctxt_t req)
 	}
 
 	p = ldmsd_prdcr_new_with_auth(name, xprt, host, atoi(port), ptype,
-			atoi(interval), auth, _uid, _gid, _perm);
+			atoi(interval), auth, _uid, _gid, _perm, atoi(cache_ip));
```
narategithub, line 2089: what happened if `cache_ip` is `NULL`?

baallan: more significantly, afaict ip may be 64 bits, so atoi is not necessarily a valid  conversion.

nichamon: I added the `cache_ip` value to the producer information message sent by the buddy ldmsd, so it isn't NULL, the same as the other attributes.

tom95858, line 2089: @nichamon, did we check what would happen if we run a 4.4.5 ldmsd_controller/maestro against a 4.4.3 ldmsd? As long as it ignores the extra parameter I think we're fine, but it shouldn't crash.

PR #1493, opened 2024-10-31
Improve error reporting in ldmsd_controller

tom95858:
This change refactors ldmsd_controller to consistently handle failures from handle_args() and fmt_status(). This change
more consistently reports server-side errors such as 'Operation not permitted' which would previously result in a crash or confusing output.

There are still issues if ldmsd_controller is started without specifying a host/port/auth. In this mode, it is not connected and therefore nearly all commands will fail with a crash. The only use case I know of for this mode is to access the help which provides useful syntax information. We should consider removing it to avoid user confusion.

@narategithub has run this change through the ldms-test system test suite and this change does not cause any errors.


narategithub (approved):
:+1:

PR #1506, opened 2024-11-08
Set record_sampler non-array metric `array_len` spec to 0

narategithub:
Set record_sampler non-array metric `array_len` spec to 0 (this will be stored in the descriptor in the set) so that the consequent components (e.g.  store) shall fail if they incorrectly handle the length.


tom95858:
I'm sorry @narategithub this change  makes no sense. Is something missing?

Issue #1497, opened 2024-11-04
Add new alias to the 'interval' parameter in `start`, `prdcr_#`, and `updtr_#`


nichamon:
- Add 'sample_interval' as an alias for 'interval' in the 'start' command
- Add 'reconnect_interval' for 'prdcr_#' command
- Add 'update_interval' for 'updtr_#' command

baallan:
this should also account for the new advertiser_add.

nichamon:
The final decision for the new alias is:

- No change to the `interval` parameter in 'start'. `sample_interval` isn't used. The change was reverted or not merged.
- `reconnect` is used in 'prdcr_add, prdcr_start, and 'prdcr_start_regex',  'advertiser_add' and 'advertiser_start' command
- No change to the `interval` parameter in `updtr_add`.

The changes have been made. I'm closing the issue.

PR #1507, opened 2024-11-11
Add new aliases to the sample interval, reconnect interval, and update interval attributes

nichamon:


nichamon:
Resolves #1497

ldms/src/ldmsd/ldmsd_request_util.c
```diff
@@ -193,6 +195,7 @@ const struct req_str_id attr_str_id_table[] = {
 	{  "type",              LDMSD_ATTR_TYPE  },
 	{  "udata",             LDMSD_ATTR_UDATA  },
 	{  "uid",               LDMSD_ATTR_UID  },
+	{  "update_interval",  LDMSD_ATTR_INTERVAL  },
```
narategithub, line 198: Just a nit. Could you please fix the text alignment to match the surrounding text?

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -357,42 +357,66 @@ class LdmsdCmdParser(cmd.Cmd):
     def complete_daemon_exit(self, text, line, begidx, endidx):
         return self.__complete_attr_list('daemon_exit', text)
 
+    def __handle_reconnect_alias(reconnect_interval, reconnect, interval):
```
nick-enoent, line 360: Could we possibly move this function to ldmsd_communicator.py? I realize we don't have parser_util.py merged yet, or I'd suggest moving it there. This would be useful to be able to call when parsing YAML files for both the ldmsd -y option and maestro_ctrl

tom95858:
@nichamon  can we confirm that this does not any compatibly issues?  I'm not really asking @nichamon  I'm asking all the people who advocated for this change.

ldms/python/ldmsd/ldmsd_controller
tom95858: @nick-enoent  let's look at the API and decide if we need something else.

ldms/src/ldmsd/ldmsd_request_util.c
nichamon: > Just a nit. Could you please fix the text alignment to match the surrounding text?

Done

nichamon:
On Hold for Future Release (v4.5+)

This PR is being postponed to a future release due to potential impacts on existing configuration file generation scripts.

Impact:
- The 'interval' attribute will be marked as deprecated
- Without script modifications, warning messages will be generated whenever 'interval' is used

We'll revisit this change after v4.4.5 to ensure a smooth transition for all users.

tom95858:
This will be done in 4.5

PR #1510, opened 2024-11-13
[b4.4] record_sampler `with_name` option

narategithub:
Add `with_name` option to record_sampler to generate `dev[name]` device name in the set (for testing).


narategithub:
@tom95858 could you please merge this too? It is needed in the test (ldmsd_decomp_static_op_test)[https://github.com/ovis-hpc/ldms-test/blob/b4.4/ldmsd_decomp_static_op_test]

PR #1514, opened 2024-11-13
allow Json stream sampler 1 set per schema seen in the same stream

baallan:
Publishers on a given stream (such as spank plugin and netlink-notifier) produce multiple schemas for different event types, but the json_stream_sampler does not account for this.
This patch creates a set per schema and keeps a list of set instance names it has created so that they may be purged when no further messages on the same schema have occurred within an age (or timeout) window.
The included test scripting illustrates how the sets created by the json sampler may not ultimately arrive at a csv store.
Before the commit which adds the multiple schemas per stream functionality, 1 message of 9 appears in the store; with the complete patch, 4 messages (1 per schema) appear in the store.
This patch is in draft for comment and as it does not (yet?) segregate schemas to individual streams (the same schema name could be seen in multiple streams with different semantics in each stream). Comments on the implementation details and on how one might configure updaters to move the created sets properly from L0 to L1 to L2 are welcome.

### DIFF
--- a/ldms/scripts/examples/json_stream_sampler
+++ b/ldms/scripts/examples/json_stream_sampler
@@ -1,0 +1,96 @@
+export plugname=json_stream_sampler
+export dsname=$(ldms_dstat_schema_name mmalloc=1 io=1 fd=1 stat=1 auto-schema=1)
+export dstat_schema=$dsname
+export LDMSD_LOG_LEVEL=ERROR
+export LDMSD_LOG_TIME_SEC=1
+export LDMSD_EXTRA="-m 128m"
+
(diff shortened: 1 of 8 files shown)


tom95858:
@baallan, this has already been done. I will it in the tree today/tomorrow.

baallan:
> @baallan, this has already been done. I will it in the tree today/tomorrow.

@tom95858 is it on a branch i can review before we rush it in? or is it only targeted at main?

tom95858:
@baallan, there is an open pull request and you are specified as a reviewer on b4.4.

baallan:
superceded.

Issue #1429, opened 2024-08-08
ldmsd_decomposition.man needs update to document functional operations


morrone:
The new functional operations and related syntax (group, limit, index, etc.) need to be documented in ldmsd_decomposition.man.

tom95858:
@narategithub we should get this done for 4.4.5

narategithub:
> @narategithub we should get this done for 4.4.5

I'll add this on my list.

emdonat:
Resolved in pr #1515

PR #1515, opened 2024-11-14
[b4.4] static decomposition man update

narategithub:
Update `ldmsd_decomposition.man` to reflect the latest "src", "dst", "type", "fill" and "op".


narategithub:
This is for issue #1429 .

Issue #1512, opened 2024-11-13
SIGSEGV in ldmsd_row_cache()


morrone:
With ovis-ldms [v4.4.4_6toss](https://github.com/LLNL/ldms/tree/v4.4.4_6toss), I am now hitting a SIGSEGV:

```
(gdb) bt
#0  0x0000155553a91a84 in pthread_mutex_lock () from /usr/lib64/libpthread.so.0
#1  0x0000555555585a7b in ldmsd_row_cache (rcache=0x0, group_key=group_key@entry=0x1554ec005450, row_key=row_key@entry=0x1554ec011700, row=row@entry=0x1554ec005ce0) at ldmsd_row_cache.c:263
#2  0x00001555076c8555 in decomp_static_decompose (strgp=0x5555558aec70, set=0x1554ec005900, row_list=0x155506cbb6d0, row_count=0x155506cbb6cc) at decomp_static.c:1785
#3  0x00001555078ce410 in flex_decompose (strgp=0x555555841870, set=0x1554ec005900, row_list=0x155506cbb740, row_count=0x155506cbb73c) at decomp_flex.c:360
#4  0x000055555557dccf in strgp_decompose (prd_set=0x1554ec012a90, strgp=0x555555841870) at ldmsd_strgp.c:183
#5  strgp_update_fn (strgp=0x555555841870, prd_set=0x1554ec012a90) at ldmsd_strgp.c:206
#6  0x000055555557c4d0 in updtr_update_cb (t=<optimized out>, set=<optimized out>, status=0, arg=0x1554ec012a90) at ldmsd_updtr.c:284
#7  0x0000155555114f50 in __handle_update_data (ev=<optimized out>, ctxt=0x155500110e50, x=0x1555000dff70) at ldms_xprt.c:2338
#8  handle_zap_read_complete (ev=<optimized out>, zep=<optimized out>) at ldms_xprt.c:2425
#9  ldms_zap_cb (zep=<optimized out>, ev=<optimized out>) at ldms_xprt.c:2829
#10 0x000015550eef1e5c in process_sep_msg_read_resp (sep=0x1555000e0210) at zap_sock.c:733
#11 0x000015550eef46f5 in sock_read (ev=0x1555000f06b4) at zap_sock.c:1372
#12 sock_ev_cb (ev=0x1555000f06b4) at zap_sock.c:1152
#13 0x000015550eef539d in io_thread_proc (arg=0x1555000f0650) at zap_sock.c:1423
[...]

PR #1521, opened 2024-11-14
Fix SEGV when 'group' not specified.

narategithub:


narategithub:
This addresses issue #1512 .

Issue #1524, opened 2024-11-18
SOS store failure on schema kills entire daemon


bschwal:
Just wanted to open up a discussion about this. Currently if a schema changes and SOS is not able to store because of the change the entire LDMS daemon will fail. A daemon not running is a surefire way to alert the user that something is wrong but perhaps we want to add an option to the sos_store configuration to allow the schema store to fail, remove that storage policy, log an error, and let the daemon continue. This way we could still get the data from other schemas and ldms would continue working, albeit in a degraded state.

tom95858:
@bschwal please be more specific. What does "fail" mean? Does it crash, does it exit, does it log an error?

bschwal:
The daemon will throw an assert, the daemon will stop, and a coredump will be written.

bschwal:
No log messages other than "an assert failed in sos_store.c" are written.

tom95858:
Ben, please help me here. "an assert failed ..." does not exist. What assert? What line number? Where is the core dump? What machine?

bschwal:
Sorry I thought this was an intended design that we talked about a while ago so I wanted to discuss design changes rather than stating its an issue. I wasn't giving much info about the assert as a result. Here's the full story: 

The filesingle schema changed while a store for it was running on daemons on our cluster Shirley (I put an example coredump in your home dir). The meaningful lines from the gdb of the corefile are: 

#4  0x00007f4fbc4d2804 in __store_metric (obj=0x7f4e30b93780, attr=0x7f4e300589a0, set=0x7f4e244ce2c0, metric_type=LDMS_V_S64, mval=0x7f4fc50567d8, count=0)
    at ../../../../ldms/src/store/store_sos.c:1044
1044                            assert(0);
(gdb) up
#5  0x00007f4fbc4d36d2 in __store_basic (si=0x7f4e30b910b0, s=0x7f4e244ce2c0, metric_arry=0x7f4e30b90b30, metric_count=347)
    at ../../../../ldms/src/store/store_sos.c:1305
1305                    attr = __store_metric(obj, attr, s, metric_type, mval, count);
(gdb) p obj->schema->data->name
$1 = "filesingle_{CLUSTER}", '\000' <repeats 229 times>

The only output to logfile in this scenario is this line:
ldmsd: ../../../../ldms/src/store/store_sos.c:1044: __store_metric: Assertion `0' failed.

[...]

PR #1528, opened 2024-11-20
store_sos: Replace assertion with type mismatch error message

nichamon:


nichamon:
Address Issue #1524

baallan:
Store-metric also needs to check set metric name == expected name in existing store schema; so far we usually got lucky there was also a type conflict.

nichamon:
@tom95858 I can do that. However, I did some testing before I created the pull request using the `test_sampler` plugin. The new log messages got reported when:
A sampler daemon created a set of two metrics of U64 type using `test_sampler`. An aggregator collected and stored the set using `store_sos`. Then, I killed both daemons. I changed the type of a metric and restarted the sampler daemon and the aggregator. store_sos reported the log messages from the patch.

Besides the above case, I also tested another scenario where I only restarted the sampler daemon and let the aggregator running. The storage policy logic caught the schema change and reported an error. This case won't hit the code in `store_sos`.

The tests I made cover the case of a type change and verify that `store_sos` reported the log message. Do you think I should test with the `filesingle` plugin? Are there any specific changes to the `filesingle` plugin configuration I should try?

tom95858:
This has been superseded by PR# 1532. @nichamon, let's leave the asserts in because theoretically, those issues conditions should never happen unless the SHA256 happens to collide. I wonder if we can remove those checks altogether for efficiency's sake.

Issue #1533, opened 2024-11-22
slingshot_metrics samler function returns an error causing ldmsd to stop the sampler


johnstile:
The ldmsd shows the sampler was stopped, but the other samplers continue working:

Nov 19 10:19:53 nidXXXXXX ldmsd[NNNNN]: Tue Nov 19 10:19:53 2024: ERROR     : 'slingshot_metrics': failed to sample. Stopping the plug-in.

When cxi_read_counter fails it closes and returns an error.
The sample function in slingshot_metrics returns the error.
ldmsd disables the sampler.
The generation number stops incrementing.
ldmsd aggregators start generating thousands of "there is an outstanding update" messages due to the producer not
I was able to use ldms_controller to resume the sampler, so I'm not sure why it had an error (need console investigation, but the nic is alive). 

Proposed fix discussion for @morrone or @tom95858 

Never return an error from the sample function in slingshot_info or slingshot_metrics samplers,
instead log the error and return zero from the sample function to prevent ldmsd from stopping the sampler.

johnstile:
Every time we see the slingshot_metrics stop, The only kernel message I see is for  stopping automatic telemetry cache refresh.
``
Nov 22 03:28:55 nidXXXXXX ldmsd[92889]: Fri Nov 22 03:28:55 2024: ERROR     : 'slingshot_metrics': failed to sample. Stopping the plug-in.
``
There is a corresponding kernel messages
``
Nov 22 03:28:55 nidXXXXXX kernel: cxi_core 0000:21:00.0: cxi0[hsn0] stopping automatic telemetry cache refresh
``

tom95858:
@johnstile we probably need to get HPE involved to explain the kernel message; what it means and what causes it. Obviously, it's not terminal or restarting the sampler would not have worked, i.e. we would have restarted the sampler and it would have stopped right away with another error.

johnstile:
This is a graph showing the rate at which we see the oversampling, which is caused by the slingshot_metrics sampler deactivating, which shows a linear growth impact on monitoring the system.

![Screenshot_20241122_094408](https://github.com/user-attachments/assets/70703cc5-c034-4bd8-9cd5-bc89c36bf72d)

tom95858:
[...]

PR #1535, opened 2024-11-23
[b4.4] slingshot_metrics: don't return an error from sample() when a device's counter lookup fails

morrone:


morrone:
Closes #1533

tom95858:
@morrone, do we need a log message at cache_cxil_dev_get too?

tom95858:
@morrone, I guess the point is that when this fails, we need to know where and why so we can pass the issue on the HPE. With this change, we only know about 1 of the 3 failure modes.

morrone:
> @morrone, do we need a log message at cache_cxil_dev_get too?

No, when this very issue of a counter lookup fails, I call ```cxil_close_device(cache_cxil_dev[i]);```. It will be normal for the device to be missing from the cache until at least the next cache refresh.

> With this change, we only know about 1 of the 3 failure modes.

Assuming cache_cxil_dev_get() was your second failure mode, what is the third that you are wondering about?

>  I guess the point is that when this fails, we need to know where and why so we can pass the issue on the HPE.

This tells most of what we know. A read of a counter failed for a device. The console log message is going to tell HPE more than LDMS can.

The thing that I was considering is maybe making cache_cxil_device_list_get() non-fatal. But if that fires, it probably means you don't even have the cxi kernel module loaded. Do we print a message over and over again when things are potentially misconfigured? Or do we move in the direction of considering almost nothing fatal?  I guess I don't have a strong opinion either way on that.

tom95858:
@morrone, I don't have answers to any of those questions. I was suggesting them in case you did ;-). @jstile might you be able to test this on your system? It should minimize the time lost between recognizing and restarting a stopped sampler.

johnstile:
@morrone The patch logs the error, but does not include the return code.  Would that be helpful to find the cause?

johnstile:
As I watch the logs, after ldms starts the sampler we see a kernel message from cxi_core

`kernel: cxi_core 0000:c2:00.0: cxi0[hsn0] starting automatic telemetry cache refresh`

After ldmsd logged " 'slingshot_metrics': failed to sample. Stopping the plug-in.", we see the message from cxi_core

` kernel: cxi_core 0000:21:00.0: cxi0[hsn0] stopping automatic telemetry cache refresh`

These messages from cxi_core are purely informational and just means that some application started using telemetry and then stopped using it on the node.

PR #1537, opened 2024-11-26
Make ldmsd continue startup on producer's hostname resolution failure

nichamon:


nichamon:
The patch is ported from #1492 to main.

PR #1543, opened 2024-12-02
Add requests to enable/disable streams communication

tom95858:
This change adds a command to disable stream support in the ldmsd daemon. When streams communication is disabled, stream data cannot be published to or subscribed from the daemon.

The current default of enabled is maintained.

### DIFF
--- a/ldms/man/ldmsd_controller.man
+++ b/ldms/man/ldmsd_controller.man
@@ -12,18 +12,9 @@ ldmsd_controller \- a python program to configure an ldms daemon.
 ldmsd_controller> <cmd> [ <attr>=<value> ]
 
 .SH DESCRIPTION
-With LDMS (Lightweight Distributed Metric Service), the ldmsd
-can be configured via the ldmsd_controller.
-
-If ldms is built with --enable-readline, one can invoke the ldmsd_controller
-from the command line and obtain an input interface with feedback. In many
-instances, instances, however, it is prefered to execute scripts and send the
-output commands to an ldmsd instead.
+Configure and query the status of a running LDMSD daemon.
 
 .SH ENVIRONMENT
-Note: python2.6 with the additional installation of the argparse module
-OR python2.7 (which has the argparse module) is required.
-
 .TP
 PYTHONPATH
 <path_to_ovis_install>/lib[64]/pythonX.Y/site-packages/
@@ -402,6 +393,14 @@ attr=<value>
 The producer name. If none is given, the statuses of all producers are
 reported.
 .RE
+
+.SS Disable stream communication
+.BR stream_disable
+.br
+Disable the subscription and publication of data on a stream in this daemon.
+Once stream communication is disabled, all stream requests will result in an
+error of ENOSYS. Stream communication cannot be re-enabled without restarting
+the daemon.
 
 .SS Subscribe for stream data from all matching producers
 .BR prdcr_subsribe
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -109,6 +109,7 @@ int ldmsd_req_debug = 0; /* turn bits on / off using gdb or -L
 FILE *ldmsd_req_debug_file = NULL; /* change with -L or
 				    * ldmsd.c:process_log_config */
 
+static int stream_enabled = 1;
 static int cleanup_requested = 0;
 
 void __ldmsd_log(enum ldmsd_loglevel level, const char *fmt, va_list ap);
@@ -291,6 +292,7 @@ static int stream_unsubscribe_handler(ldmsd_req_ctxt_t reqc);
 static int stream_client_dump_handler(ldmsd_req_ctxt_t reqc);
 static int stream_new_handler(ldmsd_req_ctxt_t reqc);
 static int stream_status_handler(ldmsd_req_ctxt_t reqc);
+static int stream_disable_handler(ldmsd_req_ctxt_t reqc);
 
 static int listen_handler(ldmsd_req_ctxt_t reqc);
 
@@ -624,6 +626,9 @@ static struct request_handler_entry request_handler[] = {
 	[LDMSD_STREAM_STATUS_REQ] = {
 		LDMSD_STREAM_STATUS_REQ, stream_status_handler, XALL
 	},
+	[LDMSD_STREAM_DISABLE_REQ] = {
+		LDMSD_STREAM_DISABLE_REQ, stream_disable_handler, XUG | MOD
+	},
 
 	/* LISTEN */
 	[LDMSD_LISTEN_REQ] = {
@@ -685,6 +690,7 @@ int is_req_id_priority(enum ldmsd_request req_id)
 	case LDMSD_AUTH_ADD_REQ:
 	case LDMSD_CMDLINE_OPTIONS_SET_REQ:
 	case LDMSD_INCLUDE_REQ:
+	case LDMSD_STREAM_DISABLE_REQ:
 		return 1;
 	default:
 		return 0;
@@ -950,6 +956,12 @@ int validate_ldmsd_req(ldmsd_req_hdr_t rh)
 	return 1;
 }
 
+static int is_stream_request(int id)
+{
+	return (id >= LDMSD_STREAM_PUBLISH_REQ
+		&& id <= LDMSD_STREAM_STATUS_REQ);
+}
+
 int ldmsd_handle_request(ldmsd_req_ctxt_t reqc)
 {
 	struct request_handler_entry *ent;
@@ -958,6 +970,14 @@ int ldmsd_handle_request(ldmsd_req_ctxt_t reqc)
 	uid_t luid;
 	gid_t lgid;
 	mode_t mask;
+
+	if (!stream_enabled && is_stream_request(reqc->req_id)) {
+		reqc->errcode = ENOSYS;
+		(void)Snprintf(&reqc->line_buf, &reqc->line_len,
+			       "The stream service is DISABLED on this system.");
+		ldmsd_send_req_response(reqc, reqc->line_buf);
+		return 0;
+	}
 
 	if (LDMSD_CFG_TYPE_LDMS == reqc->xprt->type)
 		ldms = reqc->xprt->ldms.ldms;
@@ -7381,7 +7401,7 @@ static const char *__xprt_prdcr_name_get(ldms_t x)
 
 static int stream_publish_handler(ldmsd_req_ctxt_t reqc)
 {
-	char *stream_name;
+	char *stream_name = NULL;
 	ldmsd_stream_type_t stream_type = LDMSD_STREAM_STRING;
 	ldmsd_req_attr_t attr;
 	int cnt;
@@ -7736,6 +7756,15 @@ static int stream_new_handler(ldmsd_req_ctxt_t reqc)
 	return 0;
 }
 
+static int stream_disable_handler(ldmsd_req_ctxt_t reqc)
+{
+	stream_enabled = 0;
+	reqc->errcode = 0;
+	(void)Snprintf(&reqc->line_buf, &reqc->line_len, "OK");
+	ldmsd_send_req_response(reqc, reqc->line_buf);
+	return 0;
+}
+
 static int stream_status_handler(ldmsd_req_ctxt_t reqc)
 {
 	int rc;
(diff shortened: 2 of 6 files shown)


tom95858:
@morrone, @emdonat I'm going to modify this pull request so that the default enabled vs. disabled is controlled by a configure option. Does that work for everyone?

nick-enoent:
> @morrone, @emdonat I'm going to modify this pull request so that the default enabled vs. disabled is controlled by a configure option. Does that work for everyone?

I was about to ask about this. We'll need to add a YAML option to address this as well.

nichamon:
@tom95858 In `prdcr_subscribe` path, maybe aggregators could report the error message responded by the producer when it has  Streams disabled. I'm talking about having [ldmsd_prdcr.c:__on_subs_resp()](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_prdcr.c#L498) print the error message it receives in the response.

To support `stream_enable` and `stream_disable` commands in configuration files, could you please add the strings to this table: [ldmsd_request_util.c:req_str_id_table](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_request_util.c#L64C25-L64C41)?

Could you also add the commands to the[ ldmsd_controller man page](https://github.com/ovis-hpc/ldms/blob/b4.4/ldms/man/ldmsd_controller.man)?

nichamon (changes requested):
@tom95858 In `prdcr_subscribe` path, maybe aggregators could report the error message responded by the producer when it has  Streams disabled. I'm talking about having [ldmsd_prdcr.c:__on_subs_resp()](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_prdcr.c#L498) print the error message it receives in the response.

To support `stream_enable` and `stream_disable` commands in configuration files, could you please add the strings to this table: [ldmsd_request_util.c:req_str_id_table](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_request_util.c#L64C25-L64C41)?

Could you also add the commands to the[ ldmsd_controller man page](https://github.com/ovis-hpc/ldms/blob/b4.4/ldms/man/ldmsd_controller.man)?

morrone:
I would like to see most of the streams rpcs rejected even earlier in ldmsd_handle_request() when streams is disabled. Something like:

```
  if streams_disabled() && is_streams_rpc():
     return unimplemented_handler();
```

Or instead of unimplemented_handler(), it could be a new streams_disabled_handler().

tom95858:
@morrone, this has already been accomplished in top of tree. What you propose would require dipping down into the LDMS layer which is explicitly **_not_** what this streams implementation wanted to do. The top of tree implementation is in the LDMS layer like you suggest. The 4.4 implementation sits on top of LDMS and looks like an **application** to the system. This is why we re-implemented it. My only point is that if we **_really_** want all of this in 4.4 then we should back-port the top of tree streams implementation, not fuss about with an implementation that is going to be deprecated in 6 mos.

nichamon:
@tom95858 I believe @morrone referred to [`ldmsd_handle_request()`](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L953) in `ldmsd_request.c` in the `ldmsd` layer. The check he mentioned could be added right after the request ID validation [here](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L971).

nichamon:
@tom95858 I also believe we should add `stream_disable` to the priority list in [`is_req_id_priority()`](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L679) so that streams would be disabled before `ldmsd` listens on a listening port. This will ensure that `ldmsd` wouldn't handle any stream publish request after listening but before disabling Streams. What do you think? Let me know if you want me to make a patch and send it to you.

morrone:
> @tom95858 I believe @morrone referred to [`ldmsd_handle_request()`](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L953) in `ldmsd_request.c` in the `ldmsd` layer. The check he mentioned could be added right after the request ID validation [here](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L971).

Yes, that is what I was talking about.

> @tom95858 I also believe we should add `stream_disable` to the priority list in [`is_req_id_priority()`](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L679) so that streams would be disabled before `ldmsd` listens on a listening port. This will ensure that `ldmsd` wouldn't handle any stream publish request after listening but before disabling Streams. What do you think? Let me know if you want me to make a patch and send it to you.

That sounds good to me!

morrone:
By the way, @tom95858 , it doesn't look like the PR had checks for LDMSD_STREAM_CLIENT_DUMP_REQ or LDMSD_STREAM_NEW_REQ. Shouldn't those be blocked as well?

tom95858:
@morrone sure we can add those checks.

morrone:
> @morrone, this has already been accomplished in top of tree

I don't believe that it has.

tom95858:
@morrone, that may be you to explain why. Otherwise, we can't possibly resolve the issue.

morrone:
> @morrone, that may be you to explain why. Otherwise, we can't possibly resolve the issue.

See discussion #1548. When those questions have technical solutions, we'll have resolved the issues.

tom95858:
> @tom95858 In `prdcr_subscribe` path, maybe aggregators could report the error message responded by the producer when it has Streams disabled. I'm talking about having [ldmsd_prdcr.c:__on_subs_resp()](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_prdcr.c#L498) print the error message it receives in the response.
> 
> To support `stream_enable` and `stream_disable` commands in configuration files, could you please add the strings to this table: [ldmsd_request_util.c:req_str_id_table](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_request_util.c#L64C25-L64C41)?
> 
> Could you also add the commands to the[ ldmsd_controller man page](https://github.com/ovis-hpc/ldms/blob/b4.4/ldms/man/ldmsd_controller.man)?

I think this brings up another issue. What if streams are disabled and then subsequently enabled? Isn't it free to subscribe? In other words allowing the subscription to succeed but never delivering anything is "free". @morrone, what do you think?

tom95858:
@morrone, @nichamon could you please review this latest update and ensure that your concerns have. been addressed?

morrone:
> I think this brings up another issue. What if streams are disabled and then subsequently enabled? Isn't it free to subscribe? In other words allowing the subscription to succeed but never delivering anything is "free". @morrone, what do you think?

I don't think that we want clients maintaining stale subscriptions in the ldmsd when streams are disabled. In fact, when streams are disabled, the code should probably explicitly close and free resources associated with any outstanding subscribers.

oceandlr:
Is this about enabling/disabling it forever? Or about enabling or disabling particular streams dynamically?

morrone:
> Is this about enabling/disabling it forever? Or about enabling or disabling particular streams dynamically?

This ticket is about having one big switch that enables or disables the entire streams service in an ldmsd. When disabled, none of the streams related rpcs (other than enable/disable) will be usable on the ldmsd.

So "forever" isn't really a word that applies here. One _could_ choose to "enable" the streams service at any time (though expect that those of us that will disable the service will leave it off for the lifetime of the ldmsd process).

oceandlr:
Ah, ok. I am working on a test case of setting up and tearing down streams dynamically, rather than in the configuration, as they were done in 4.4.4.

tom95858:
> > I think this brings up another issue. What if streams are disabled and then subsequently enabled? Isn't it free to subscribe? In other words allowing the subscription to succeed but never delivering anything is "free". @morrone, what do you think?
> 
> I don't think that we want clients maintaining stale subscriptions in the ldmsd when streams are disabled. In fact, when streams are disabled, the code should probably explicitly close and free resources associated with any outstanding subscribers.

Ok, but the subscribers don't know that the publisher has disabled streams. This flows all the way up the forwarding chain. All that would be saved from this additional functionality is a few bytes of memory at the cost of a significant amount of complexity.

nichamon:
> @tom95858 In prdcr_subscribe path, maybe aggregators could report the error message responded by the producer when it has Streams disabled. I'm talking about having [ldmsd_prdcr.c:__on_subs_resp()](https://github.com/ovis-hpc/ldms/blob/ffd70a8ec11a7c6e7a913b7a87bfd566542268d0/ldms/src/ldmsd/ldmsd_prdcr.c#L498) print the error message it receives in the response.

I suggested this so that there would be a log message in the aggregator log file that `prdcr_subscribe` didn't success or complete. With out `__on_subs_resp()` reports the error message, there is no way we will know that streams are disabled on some sampler daemons.

nichamon:
> I think this brings up another issue. What if streams are disabled and then subsequently enabled? Isn't it free to subscribe? In other words allowing the subscription to succeed but never delivering anything is "free".

This is an interesting point. Initially, we added the stream_enable and stream_disable commands to address a specific use case where administrators want to permanently disable streaming features. Your question brings up additional scenarios to consider - specifically what happens when streams are temporarily disabled and then re-enabled.

In the case of temporary pausing, there's value in preserving the subscriber list while streams are disabled. This would eliminate the need for reconfiguration when streams are re-enabled. While I understand @morrone's suggestion to purge subscribers, I believe administrators should have the flexibility to maintain their subscriber configurations across disable/enable cycles.

I propose we keep the current implementation of this pull request (where ldmsd doesn't receive stream data or new subscribers while disabled) but add a new stream_unsubscribe command to ldmsd_controller. This would give administrators manual control over removing stale subscribers when needed.

ldms/man/ldmsd_controller.man
```diff
@@ -403,6 +394,21 @@ The producer name. If none is given, the statuses of all producers are
...
 
+.SS Enable stream communication
+.BR stream_enable
+.br
+Enable the subscription and publication of data on a stream in this daemon. When
+stream communication is disabled, all stream requests will result in an error
+of ENOSYS.
+
+.SS Disable stream communication
+.BR stream_enable
```
nichamon, line 405: A typo. This should read `stream_disable`.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -950,6 +961,12 @@ int validate_ldmsd_req(ldmsd_req_hdr_t rh)
 	return 1;
 }
 
+static int is_stream_request(int id)
+{
+	return (id >= LDMSD_STREAM_PUBLISH_REQ
+		&& id <= LDMSD_STREAM_STATUS_REQ);
```
nichamon, line 967: Should we add `PRDCR_SUBSCRIBE` here?

nichamon (changes requested):

nichamon:
@tom95858 @morrone After further consideration of the motivation behind adding these commands, I believe we shouldn't allow administrators to enable Streams after they disable it. Allowing temporary disabling of Streams would introduce additional complexities that would need to be carefully handled. Since the disable action should be permanent, we should consider removing the `stream_enable` command entirely and document that `stream_disable` is permanent, requiring an ldmsd restart to re-enable Streams. What do you think?

morrone:
> @tom95858 @morrone After further consideration of the motivation behind adding these commands, I believe we shouldn't allow administrators to enable Streams after they disable it. Allowing temporary disabling of Streams would introduce additional complexities that would need to be carefully handled. Since the disable action should be permanent, we should consider removing the `stream_enable` command entirely and document that `stream_disable` is permanent, requiring an ldmsd restart to re-enable Streams. What do you think?

That is fine with me. I don't really have a use case for reenabling (we can always just restart the ldmsd with a different configuration to have streams enabled).

If we don't yet have your earlier suggestion of "I also believe we should add stream_disable to the priority list in [is_req_id_priority()](https://github.com/ovis-hpc/ldms/blob/fd189261297901305f90aeeb6ee489d167d3fc6d/ldms/src/ldmsd/ldmsd_request.c#L679) so that streams would be disabled before ldmsd listens on a listening port. This will ensure that ldmsd wouldn't handle any stream publish request after listening but before disabling Streams.", I think we should have that as well.

Either that or clean up any subscriptions at stream_disable() time, but @tom95858, you seem opposed to that on complexity grounds.

nichamon:
@tom95858 already added the commands to the priority list, so I think the patch is good to be merged after @tom95858 removes the `stream_enable` logic.

PR #1544, opened 2024-12-03
Standardize error response in three request handlers

nichamon:
Modify updtr_task, prdcr_hint, and set_route handlers to use ldmsd_send_req_response() for sending error messages back to requesters. This function automatically handles message length calculation, removing the need for manual length calculations in these handlers.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -4373,9 +4373,9 @@ static int updtr_task_status_handler(ldmsd_req_ctxt_t reqc)
 	if (name) {
 		updtr = ldmsd_updtr_find(name);
 		if (!updtr) {
+			reqc->errcode = ENOENT;
 			cnt = snprintf(reqc->line_buf, reqc->line_len, "updtr '%s' not found", name);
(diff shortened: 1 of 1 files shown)


nichamon:
This change is based on code inspection rather than an observed issue. It standardizes the error response handling in these handlers by using ldmsd_send_req_response() for consistent message length calculation.

tom95858:
@nichamon does anyone still use ldmsd_send_error_reply?

nichamon:
> @nichamon does anyone still use ldmsd_send_error_reply?

Yes, `ldmsd_process_config_request()` calls `ldmsd_send_error_reply()` to send an error when an error occurs while preparing the request context `ldmsd_req_ctxt reqc`.

tom95858:
@nichamon, I'm a little bit concerned that this code is essentially untested unless we force the error path. Have we tried to do that?

nichamon:
> @nichamon, I'm a little bit concerned that this code is essentially untested unless we force the error path. Have we tried to do that?

@tom95858 Yes, I tested by instrumenting the code to and force it to go through the error paths. While I tested the change in the set_route path, I found bugs in the python code that handles the `set_route` command. The fix is in PR #1552.

PR #1550, opened 2024-12-09
reStructuredText, ReadTheDocs and `ldms_stream` manpage

narategithub:
As I was writing `ldms_stream` documentation, I think that this might be a good point to start moving the in-source-tree documentation to `.rst` format so that we can get both `ReadTheDocs` and `manpage` outputs. So, here it is. This pull request contains:

- `README.md` -> `docs/README.rst` (rename & modify formatting)
  - `README.rst` top-level README file is a soft link to `docs/README.rst`
- add ReadTheDocs support (`.readthedocs.yaml`)
  - see [ldms-try.readthedocs.io](https://ldms-try.readthedocs.io/en/latest/) for example output
- add `rst2man` build support for rst documents
- add `docs/ldms_stream.rst`


narategithub:
@tom95858 This is now ready for your review.

PR #1556, opened 2024-12-12
update test scripts to used reconnect

baallan:
somehow there were lingering reconnect_interval and interval. this fixes them.


baallan:
@tom95858 not sure what went wrong in the last attempt, but with this patch the scripts finally get to use 'reconnect' for the prdcr_add interval. Must not have pushed when i thought i did. For 4.4.

PR #1558, opened 2024-12-12
Add 'cache_ip' to the optional attribute list of prdcr_add

nichamon:


nichamon:
@tom95858  I caught this when I worked on PR #1557 .

PR #1559, opened 2024-12-12
Add 'cache_ip' to the optional attribute list of prdcr_add

nichamon:


nichamon:
I created this by mistake.

PR #1560, opened 2024-12-12
Add the column header 'Type' to the prdcr_status output

nichamon:


nichamon:
@tom95858  I caught this when I worked on PR #1557 .

PR #1562, opened 2024-12-13
Fix how thread_stats and updtr_status handle 'reset' in ldmsd_controller

nichamon:


nichamon:
@tom95858 The bugs had been already fixed on the 'main' branch.

PR #1565, opened 2024-12-16
Enchance permission string parsing

morrone:
Enhance the permission string parsing from PR #1564.

Add full checking of the string permission representation: (r|-)(w|-)-(r|-)(w|-)-(r|-)(w|-)- )
Add checking of an octal-number-in-a-string, at least verifying that the number looks like octal, and doesn't have excess digits.
(We could perhaps also verify that the "x" bits are not set, but do not do so in this patch.)


nick-enoent:
@morrone Thanks for throwing this together! I tweaked it to remove python3 octal support as some people found that confusing. I appreciate your help putting this together =)

PR #1567, opened 2024-12-18
fix errno report in  blobwriter, stop warning about repeat subscribe

baallan:
fix errno misreported in blobwriter, and stop warning about repeat subscribe, which happens a lot (50000 out of 55000 log lines in a recent case)

### DIFF
--- a/ldms/src/ldmsd/ldmsd_stream.c
+++ b/ldms/src/ldmsd/ldmsd_stream.c
@@ -346,8 +346,11 @@ ldmsd_stream_subscribe(const char *stream_name,
 	}
 	LIST_FOREACH(cc, &s->s_c_list, c_ent) {
 		if (cc->c_cb_fn == cb_fn && cc->c_ctxt == ctxt) {
+#if 0
+			/* this is a high frequency common case */
 			msglog("The client %p is already subscribed to "
 			       "stream %s\n", cc, stream_name);
+#endif
 			errno = EEXIST;
 			pthread_mutex_unlock(&s->s_lock);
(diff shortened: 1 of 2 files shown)


baallan:
@tom95858 sorry for the late patch; things turned up in logs yesterday.

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -346,8 +346,11 @@ ldmsd_stream_subscribe(const char *stream_name,
 	}
 	LIST_FOREACH(cc, &s->s_c_list, c_ent) {
 		if (cc->c_cb_fn == cb_fn && cc->c_ctxt == ctxt) {
+#if 0
+			/* this is a high frequency common case */
```
tom95858, line 350: This looks like we're hiding a configuration error and it has nothing to do with this fix.

ldms/src/sampler/blob_stream/blob_stream_writer.c
```diff
@@ -246,6 +246,7 @@ static void fclose_and_spool(FILE* *f, char* *fname, int final)
 			sprintf(rbuf, "%s/spool/%s", dirn, base);
 			err = rename(*fname, rbuf);
 			if (err) {
+				err = errno;
```
tom95858, line 249: This code doesn't do anything. The function returns void.

tom95858: @baallan, ok I see that you're setting it so that the error message is correct.

PR #1569, opened 2025-01-07
Use peer's hostname and listen port for advertised producer's names

nichamon:
Currently, an advertised producer's name is the name of the advertiser on the peer. This creates confusion. The patch constructs the producer name as <hostname>:<port>, where hostname is the peer hostname and port is the listening port. If there are multiple listening ports, ldmsd uses the first listening port.


nichamon:
The change is the same as #1557 but for `main`.

PR #1570, opened 2025-01-07
YAML configuration file support for OVIS-4.5

nick-enoent:
Adds ldmsd_yaml_parser and ldmsd "-y" option to parse YAML files when starting a LDMSD Support for rails-specific parameters in YAML configuration file Update usage documentation in ldmsd_controller to include "ip" for LdmsdCmdParser class method "do_prdcr_listen_add"

### DIFF
--- a/ldms/man/Makefile.am
+++ b/ldms/man/Makefile.am
@@ -13,6 +13,7 @@ ldmsd.man \
 ldms-plugins.man \
 ldmsd_exits.man \
 ldmsd_controller.man \
+ldmsd_yaml_parser.man \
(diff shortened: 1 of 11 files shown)


nichamon:
@nick-enoent Could you please add the `-n` option back to ldms/man/ldmsd.man? We should add the new `-y` option to the man page too.

nick-enoent:
> @nick-enoent Could you please add the `-n` option back to ldms/man/ldmsd.man? We should add the new `-y` option to the man page too.

Thanks Mon! Good catch, I'll add that in.

PR #1572, opened 2025-01-09
[b4.4] Add Cython3 support (still support Cytnon 0.29)

narategithub:
This patch addressed issues with Cython3. The patch does not break Cython 0.29 compatibility.

Cython3 became the stock package on Ubuntu 24.04. The change in Cython language policy requires explicit `except` directive on the C function pointer type; otherwise, the function pointer type was assumed `noexcept` but the actual function implementations without `except` directive is assumed `except *`, making them not able to be assigned as callback functions.

In addition, Cython 3.0.8 (stock package on Ubuntu 24.04) has Enum generation bug (multiple declarations of same Enum). When `ldms_xprt_event_type type` enum type is explicitly referred to in `ldms_xprt_event` struct, Cython 3.0.8 somehow generated the same Enum "type_to_py" functions twice -- leading to compilation error.  By declaring the `type` field as `int type`, the redundant "type_to_py" functions of the same enum disappeared. The fix for this particular Cython3 bug was fixed in Cython 3.0.9
(https://github.com/cython/cython/pull/5905). Since Ubuntu 24.04 shipped with Cython 3.0.8, I think we have to stick with this work-around for now.

### DIFF
--- a/ldms/python/ldms.pxd
+++ b/ldms/python/ldms.pxd
@@ -196,7 +196,7 @@ cdef extern from "ldms.h" nogil:
     struct ldms_xprt:
         pass
     ctypedef ldms_xprt *ldms_t
-    ctypedef void (*ldms_log_fn_t)(const char *fmt, ...)
+    ctypedef void (*ldms_log_fn_t)(const char *fmt, ...) except *
     struct ldms_timestamp:
         uint32_t sec
         uint32_t usec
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 I sniff test this on Ubuntu 24.04 (Cython3) and ldms-dev container (Ubuntu 22.04; Cython 0.29).

PR #1575, opened 2025-01-15
Multiple configuration plugin support

tom95858:
* Converted plugins to configuration objects
* Added multiple-configuration support to samplers:
  * procnetdev2
  * meminfo
  * avro_kafka store
  * csv store
  * sos store
* Converted all samplers to support new plugin infrastructure
* Changed configuration object reference counting to use ref_t


ldms/src/core/ldms.h
```diff
@@ -2787,14 +2787,38 @@ int ldms_set_unpublish(ldms_set_t set);
...
+ *
+ * Used by applications to take a reference on a metric set. An
+ * application should use this interface if it caches a pointer
+ * to the set in a data strucure. Use ldms_set_put_ to drop a
+ * reference obtained with this interface
+ *
+ * @param set The set handle
+ * @return ldms_set_t The set handle
+ */
+ldms_set_t ldms_set_get_(ldms_set_t set);
```
morrone, line 2810: Maybe this should just be "ldms_set_get" and the following merged with the existing "ldms_set_put". As far as I can tell, the only reason these things ldms_set_put_ is different from ldms_set_put is to put a different string in the tracing. But if we really care about the string in the tracing then these generic ldms_set_get_/ldms_set_put_ functions should probably work more like ref_get/ref_put and allow a string an additional string option.

morrone (changes requested):
I like the general approach of adding a "context" pointer. While I would probably prefer to see it broken out of ldmsd_plugin (because a plugin doesn't need to be told the things it already knows about itself, e.g. its own function pointers), I do understand that this is a less invasive change to the code base.

Maybe we would add a flag to ldmsd_plugin so that a plugin can advertise whether or not it is multi-instance capable.

In store_avro_kafka.c, is "close_store()" part of the deprecated open_store()/get_ucontext()/flush_store() interface? If so, we can probably just delete it's contents like the other functions. Or maybe we should have another PR to remove that deprecated interface?

morrone:
@tom95858 Could you give me a pointer to where the plugin's inst_name field is set? I'm not seeing it. Or maybe that is still WIP?

One thing I might suggest is that we use a different variable name than "inst_name" for the plugin instance name. With both metric sets and plugin instance using the same variable name "inst_name", it is difficult to search for what one is looking for because the name is used so often in the code for two different things.

morrone:
OK, I think I see, this adds a new "inst" keyword to the "load" command, right? But isn't that backwards? Shouldn't "name" become the name of the plugin instance, and we should have a new keyword "plugin" that gives the real name of the plugin? Otherwise, how would the other commands like "config", "term" , etc. work? They would all need to have an "inst" name added as well, correct?

What I'm suggesting would seem to be consistent with other places in the existing configuration:

{noformat}
strgp_add name=ibmad_records_sampler plugin=store_csv container=ibmad schema=ibmad decomposition=/etc/ldmsd/decomp_sonar.json                                                                                                                                                                
strgp_prdcr_add name=ibmad_records_sampler regex=.*                                                                                                                                                                                                                                          
strgp_start name=ibmad_records_sampler                                                                                                                                                                                                                                                       
strgp_add name=procnetdev2 plugin=store_csv container=procnetdev2 schema=procnetdev2 decomposition=/etc/ldmsd/decomp_sonar.json                                                                                                                                                              
strgp_prdcr_add name=procnetdev2 regex=.*                                                                                                                                                                                                                                                    
strgp_start name=procnetdev2              
{noformat}

We could do similar for plugins:

{normat}
load name=myprocnetdev plugin=procnetdev2
config name=myprocnetdev producer=foo instance=foo/bar
start name=myprocnetdev interval=5000000 offset=0
load name=otherprocnetdev plugin=procnetdev2
config name=otherprocnetdev producer=foo instance=foo/baz
start name=otherprocnetdev interval=5000000 offset=0
{noformat}

And presumably the first time that the plugin code learns the plugin instance name is at config(), since this currently reuses the existing API functions.

But maybe I'm missing something in the code.

tom95858:
I agree with you that name should be the instance name. But to preserve backward compatibility, I made it so that all configuration files would continue to work even with the new 4.5. Let's discuss... If 'instance' is not specified, it defaults to 'name'.

morrone:
> I agree with you that name should be the instance name. But to preserve backward compatibility, I made it so that all configuration files would continue to work even with the new 4.5. Let's discuss... If 'instance' is not specified, it defaults to 'name'.

Right now, the "name" for a plugin must match an actual plugin name, right? So I think we could preserve backward compatibility by making it so that "name" & "plugin" mean "plugin instance name" & "plugin name" respectively. If "plugin" is not specified, then "name" will mean _both_.

So it achieves the same thing I think, unless I'm missing some use case. But it means only "load" needs the additional "plugin" line. If we add "inst" to load, then config and start also need "inst" added.

morrone:
Just a heads up: I'm working on implementing some of the changes that I suggested (in particular implementing the context pointer without the context_size). I'll try to share the wip soon (even before it works) just so you now what I'm thinking.

ldms/src/sampler/procnetdev2/procnetdev2.c
```diff
@@ -410,37 +437,62 @@ static int sample(struct ldmsd_sampler *self)
...
 
-static struct ldmsd_sampler procnetdev2_plugin = {
+static struct ldmsd_sampler procnetdev2_sampler = {
 	.base = {
 		.name = SAMP,
 		.type = LDMSD_PLUGIN_SAMPLER,
 		.term = term,
 		.config = config,
 		.usage = usage,
+		.context_size = sizeof(struct procnetdev2_s),
```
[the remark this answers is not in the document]
tom95858: I disagree; then every plugin has to handle the memory management. With this approach, the infrastructure is responsible for allocating and freeing the memory and if wants to optimize the allocation by appending it to the tail of the base allocation, it can. With your approach, we have to ensure that every single plugin does this properly; as opposed to a single line of code in the base.

ldms/src/ldmsd/ldmsd.h
```diff
@@ -894,62 +909,83 @@ int process_config_file(const char *path, int *lineno, int trust);
...
+	int (*flush)(ldmsd_store_handle_t sh);
+	void *(*get_context)(ldmsd_store_handle_t sh);
+	int (*store)(ldmsd_store_handle_t sh, ldms_set_t set, int *, size_t count);
+	int (*commit)(ldmsd_strgp_t strgp, ldms_set_t set, ldmsd_row_list_t row_list, int row_count);
 };
 
-struct ldmsd_sampler {
+typedef struct ldmsd_sampler {
 	struct ldmsd_plugin base;
+	ldms_set_t (*get_set)(struct ldmsd_sampler *self);
```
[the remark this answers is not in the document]
tom95858: Sure...

ldms/src/sampler/procnetdev2/procnetdev2.c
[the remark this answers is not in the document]
morrone: We already need to reply on plugins getting everything right. We should be be helping the plugins by dealing with the harder stuff. Automating simple things through byzantine syntax makes the plugin API harder to use. 

In other words, its _simple_ memory management to alloc a buffer at the entry place (currently get_plugin()) and free it in the exit place (currently term()). We should not replace simple things with custom unexpected and confusing infrastructure that takes documentation to explain. It is also very easy to review code that does allocations and frees at the same level of the code.

We should put effort into creating infrastructure to automate the complex and/or tedious stuff. Not replacing simple, straight forward things with unusual interfaces that take extra effort to learn and understand.

morrone:
Now that I understand the arrangement of the data structures a bit better, I think I can work with the configuration object changes. There is still a lot to do, but you can see some of my wip over here: [morrone-multi-instance2](https://github.com/morrone/ldms/tree/morrone-multi-instance2).

In particular, you might look at ldms/src/sampler/procnetdev2/procnetdev2.c in that branch. I reworked things to clean up the plugin API's use of struct ldmsd_plugin. Now that is entirely owned by the plugin. It is not a mix of the plugin's and ldmsd's information. The plugin's "context" is split out into a separate pointer, opaque to ldmsd. Instead of passing a pointer to the struct ldmsd_plugin in each API call, ldmsd instead just passes the context pointer.

I think it is a good first pass at cleaning up the plugin API. It makes the ldmsd side code a lot cleaner too. But I think there is a lot more that I can still improve. I'll keep working on it.

morrone:
@tom95858 Did you have a plan for how plugin instance names on "load" line with interact with the existing "instance" on the "config" line?

It is kind of confusing to have both, I think. Maybe we should be retiring the "config" line's instance=?

ldms/src/ldmsd/ldmsd.h
```diff
@@ -931,6 +931,7 @@ typedef struct ldmsd_plugin {
 	int (*config)(struct ldmsd_plugin *self, struct attr_value_list *kwl, struct attr_value_list *avl);
 	void (*term)(struct ldmsd_plugin *self);
 	const char *(*usage)(struct ldmsd_plugin *self);
+	ldmsd_cfgobj_t cfgobj; /* cfgobj associated to instance */
```
morrone, line 934: @tom95858 This is a thing that I don't think we should just land. This moves us in the wrong direction. We need to be _removing_ ldmsd's internal state from the plugin's API struct, not adding more. This is where we're currently failing to have good abstraction, encapsulation, and laying in the plugin API.

In my code (which granted, is still WIP) already remove _all_ of ldmsd's internal state from that struct. This is going to introduce even more that I'll have to work to back out.

I think we should remove this commit from this PR.

tom95858:
@morrone, completely disagree. This allows us to treat all objects the same, which is the point of the change. Now samplers, stores, auth, etc... are all proper configuration objects that are all treated in the same way. Reference counting, etc... This is why your change ldmsd_plugin_sample() is not appropriate because not all plugins are samplers and therefore they don't all have a sample interface.

morrone:
> @morrone, completely disagree. This allows us to treat all objects the same, which is the point of the change. Now samplers, stores, auth, etc... are all proper configuration objects that are all treated in the same way. Reference counting, etc... This is why your change ldmsd_plugin_sample() is not appropriate because not all plugins are samplers and therefore they don't all have a sample interface.

I think you might need to look a little closer at the code. ldmsd_plugin_sample() makes absolutely no assumption that all of the plugins are the same and have a sampler interface. It works directly with this PR's new cfgobj code.

And we can "treat all objects the same" without poluting the plugin's API struct with ldmsd internals. It is by no means at all required to put that point in the plugin API's struct.  There's no requirement that we violate good abstraction techniques to achieve that. We can both keep ldmsd's internals out of the plugins' API struct, and also acheive your goals.

I think we really need to have some conversations before we land any of this.

tom95858:
@morrone, my point is only that the interface name ldmsd_*plugin*_sample implies that all plugins have a sample interface.

morrone:
> @morrone, my point is only that the interface name ldmsd__plugin__sample implies that all plugins have a sample interface.

I don't know what you are talking about. Where exactly are you seeing "ldmsd__plugin__sample", and why does it imply that all plugins have a sample interface??

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -316,8 +316,8 @@ int ldmsd_load_plugin(char* inst_name, char *plugin_name, char *errstr, size_t e
 	struct ldmsd_plugin *api = load_plugin(plugin_name);
```
morrone, line 316: Love the switch to "plugin=" instead of "instance="!

But this "api = load_plugin(plugin_name);" now needs to be moved down after the "if (!plugin_name)" and before the "switch (api->type)" line.

ldms/src/sampler/netlink/netlink-notifier.c
```diff
@@ -794,7 +794,7 @@ static char *get_username(const uid_t uid, forkstat_t *ft)
 static void uid_name_info_free(forkstat_t *ft)
 {
 	size_t i;
-	if (!ft || !ft->uid_name_info)
```
morrone, line 797: Probably accidentally included in this commit?

ldms/python/ldmsd/ldmsd_communicator.py
```diff
@@ -1639,26 +1639,26 @@ def store_time_stats(self, name=None, reset = False):
...
+        plugin- The plugin name. If None, 'name' is used
 
         Returns:
         A tuple of status, data
         - status is an errno from the errno module
         - data is an error message if status != 0 or None
         """
-        if not instance:
-            instance = name
+        if not plugin:
```
morrone, line 1655: I think we should drop this check. It is redundant, since ldmsd is has the logic to handle when plugin is not provided. In fact having this duplicated logic here makes it difficult to test if the server code is doing what it should, because ldmsd_controller won't let us send without "plugin" set.

What I did was remove this check, and only optionally add include the "LDMSD_Req_Attr" for plugin when plugin is not None.

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -292,18 +292,18 @@ class LdmsdCmdParser(cmd.Cmd):
...
+        plugin= The configuration name for the plugin (default = <name>)
         """
         arg = self.handle_args('load', arg)
         if arg:
             name = arg['name']
-            if arg['inst']:
-                inst = arg['inst']
+            if arg['plugin']:
+                plugin = arg['plugin']
             else:
```
morrone, line 302: We should drop this "else" and just leave plugin as None. Otherwise this is triply redudantly handling what the server already handles. This stops us from checking the ldmsd_communicator.py is handling a missing "plugin=" correctly, which is in turn stopping us for checking that the server is doing it right.

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -292,18 +292,18 @@ class LdmsdCmdParser(cmd.Cmd):
...
-                inst = name
-            rc, msg = self.comm.plugn_load(name, inst)
+                plugin = name
+            rc, msg = self.comm.plugn_load(name, plugin)
             if rc:
-                print(f'Error loading plugin {arg["name"]} as {arg["inst"]}: {msg}')
+                print(f'Error loading plugin {arg["name"]} as {arg["plugin"]}: {msg}')
```
morrone, line 306: Might want to change this message a bit. "plugin X as Y" probably reads to a user as "X" being the real plugin name, and "Y" being the instance name.

I'd maybe suggest this:

```
   print(f'Error loading instance {arg["name"]} from plugin {arg["plugin"] if arg["plugin"] else arg[["name"]}: {msg}')
```

morrone (changes requested):

morrone:
@tom95858 I made a first pass at rebasing my #1600 PR on the current state of this PR. One thing I did was consolidate the remaining chagnes I suggest about the handling of "plugin=" into one commmit. So you might want to cherry-pick 700deb91a25df4c0daf86d28c73e642e564272a4 into your PR.

ldms/src/sampler/netlink/netlink-notifier.c
tom95858: You are right, it belongs in a different commit. I wanted to get it all compilable/testable...which it is. But it definitely could be cleaned up. I discovered this when I built it on AlmaLinux 9.5, which is the latest.

ldms/src/ldmsd/ldmsd_config.c
tom95858: Does this match what you were looking for? I liked you suggestion and think it makes sense. I agree that instance is way overloaded. I have a patch that will be up shortly that addresses that issue.

morrone: Yes, that change to plugin= instead of instance= is what I was looking for! (And I implmented over in PR #1600). You can get the rest of the little changes for just this plugin= stuff in commit da2c868e6ad8a4bdf93c05fdb88dbcff1080495a.

ldms/src/sampler/netlink/netlink-notifier.c
tom95858: I'll put this in a different commit, but the code doesn't compile with gcc 11.4 without it (AlmaLinux 9.5)

tom95858:
> @tom95858 Could you give me a pointer to where the plugin's inst_name field is set? I'm not seeing it. Or maybe that is still WIP?
> 
> One thing I might suggest is that we use a different variable name than "inst_name" for the plugin instance name. With both metric sets and plugin instance using the same variable name "inst_name", it is difficult to search for what one is looking for because the name is used so often in the code for two different things.

I changed it to cfg_name. I agree inst_name collides with the set usage and could be confusing.

ldms/src/ldmsd/ldmsd.h
tom95858: Removed...I think it was detritus.

morrone: I _think_ it was added solely to give json_stream_sampler access to the ref in the cfgobj. But yes, I didn't immediately see how that ref would be necessary. I reverted those commits in my last rebase too (and made a note to myself to revisit them later when I figured out why the ref get/put was happening in json_stream_handler). But if you don't see a reason for it either, it makes things easier to just get rid of it. Thanks!

morrone:
I have a general question/concern about locking semantics as a result of this patch stack. This introduces a a "struct ref_s ref" to struct ldmsd_cfgobj. But it also still leaves "pthread_mutex_t lock" in the same struct. So which thing is now protecting the struct? We probably need to settle on one or the other. I think that means getting rid of the older pthread_mutex_t lock.

We might also consider replacing ldms's custom ref counter code with C11's standardized operations from stdatomic.h. Granted, we lose ldms's optional tracking (debugging) code that requires a string identifier in every get/put call. But frankly, it is not clear to me that those string identifiers are being maintained as strictly as they should be (not as man puts as gets for a particular string). And the string means many different accessor functions are being written for exactly the same ref counter. This makes it a lot harder than it needs to be to review and audit the code to for ref count correctness. I am leaning towards the extra code obfuscation around the refs outweighing the benefits of the custom ref tracking code.

tom95858:
> I have a general question/concern about locking semantics as a result of this patch stack. This introduces a a "struct ref_s ref" to struct ldmsd_cfgobj. But it also still leaves "pthread_mutex_t lock" in the same struct. So which thing is now protecting the struct? We probably need to settle on one or the other. I think that means getting rid of the older pthread_mutex_t lock.
> 
> We might also consider replacing ldms's custom ref counter code with C11's standardized operations from stdatomic.h. Granted, we lose ldms's optional tracking (debugging) code that requires a string identifier in every get/put call. But frankly, it is not clear to me that those string identifiers are being maintained as strictly as they should be (not as man puts as gets for a particular string). And the string means many different accessor functions are being written for exactly the same ref counter. This makes it a lot harder than it needs to be to review and audit the code to for ref count correctness. I am leaning towards the extra code obfuscation around the refs outweighing the benefits of the custom ref tracking code.

The ref_t and the lock serve different purposes. The reference ensures that no client of the structure still has a reference to it to protect use-after-free problems. The mutex ensures that state changes to the object are consistent.

morrone:
> The ref_t and the lock serve different purposes. The reference ensures that no client of the structure still has a reference to it to protect use-after-free problems. The mutex ensures that state changes to the object are consistent.

OK, makes sense.

What state changes are we protecting against? If state changes are an issue, we might need a lot more locking around read accesses that we aren't currently doing.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -311,11 +311,12 @@ void ldmsd_store___del(ldmsd_cfgobj_t obj)
  */
 int ldmsd_load_plugin(char* cfg_name, char *plugin_name, char *errstr, size_t errlen)
 {
-	struct ldmsd_plugin *api = load_plugin(plugin_name);
+	struct ldmsd_plugin *api;
+	if (!plugin_name || !cfg_name)
```
morrone, line 315: I think it is good to only do the checking for the optional "plugin=" field in one place. But this change leaves the handling to the optional plugin= argument to the client rather than the server.

Older clients (e.g. maestro, ldmsd_controller), won't work with this approach of handling an optional "plugin=" on the client side, and it being required on the server side.

If we want backwards compatibility, the optional handling of "plugin=" should happen on the server (ldmsd) side. Clients should just pass through whatever arguments they have (i.e. always send "name=", only send "plugin=" when it is available).

morrone:
@tom95858 how about cherry picking commit 9c823794c4110561ca551750c9aeae7aaf6e6600 from PR #1615 on to your patch stack here in PR #1575? It fixes the struct/typdef name mismatch [that I mentioned earlier](https://github.com/ovis-hpc/ldms/pull/1575#issuecomment-2635489969), and makes the names better reflect their ldmd_cfgobj derivation.

It probably aligns with your changing variables names from inst_name to cfg_name too.

tom95858:
@morrone, I have added a number of your requested changes and added an attribution to you, i.e. the From: in the commit log. Can you take a look and see what I may have missed, but maybe we're close?

morrone:
> @morrone, I have added a number of your requested changes and added an attribution to you, i.e. the From: in the commit log. Can you take a look and see what I may have missed, but maybe we're close?

I'm not from lbl.gov. :) I'll take a closer look tomorrow.

If you do a cherry-pick, it retains the git authorship. Or if you are want to change the author of the most recent commit, you can do something like:

```
git commit --amend --author="Christopher J. Morrone <morrone2@llnl.gov>" --no-edit
```

Or I can split back apart the two commits that got munged together for you, if you like.

morrone:
Hey, @tom95858 . I decided to clean up the last commit by splitting it in two. I pushed a temporary PR for you to use to get it.

You can swap it in for your multi-instance-wip branch by doing (assuming ovis-hpc/ldms repo is named "origin"):

```
git checkout multi-instance-wip
git fetch origin pull/1616/head
git reset --hard FETCH_HEAD
```

Then you can force-push that to ovis-hpc/ldms.

morrone:
I'm not sure that the logic in ldmsd_communicator.py:plugn_load() is quite right yet. I think we want to make this change:

```
-        if not plugin:
-            plugin = name
-        req = LDMSD_Request(
-                command_id=LDMSD_Request.PLUGN_LOAD,
-                attrs=[
-                    LDMSD_Req_Attr(attr_id=LDMSD_Req_Attr.NAME, value=name),
-                    LDMSD_Req_Attr(attr_id=LDMSD_Req_Attr.PLUGIN, value=plugin)
-                ])
+        attrs = []
+        attrs.append(LDMSD_Req_Attr(attr_id=LDMSD_Req_Attr.NAME, value=name))
+        if plugin is not None:
+           attrs.append(LDMSD_Req_Attr(attr_id=LDMSD_Req_Attr.PLUGIN, value=plugin))
+        req = LDMSD_Request(command_id=LDMSD_Request.PLUGN_LOAD, attrs=attrs)
```

morrone:
And ldmsd_config.c:ldmsd_load_plugin() still has this:

```
        if (!plugin_name || !cfg_name)
                return EINVAL;
        api = load_plugin(plugin_name);
```

We want the server side to set plugin_name = cfg_name when plugin_name == NULL. We don't want the client to do that check.

Making that swap will preserve backwards compatibility with older clients.

And we probably need parser_util.py updated as well so ldmsd_yaml_parser can handle the new option plugin= option for samplers and stores. Maybe we'll want @nick-enoent's help with that. I don't know if he'll need to do anything in maestro too, or if it will just benefit from whatever change we make in parser_util.py.

tom95858:
@morrone so doing it this way requires the server to handle the case where the PLUGIN attribute is not present which means that both the client and the server need the logic. I _thought_ your point was to put the checking in only one place. However, if we do it lilke this then older clients will work with the newer server.

morrone:
> @morrone so doing it this way requires the server to handle the case where the PLUGIN attribute is not present which means that both the client and the server need the logic. I _thought_ your point was to put the checking in only one place. However, if we do it lilke this then older clients will work with the newer server.

What I was saying is that this (psuedo-code) test should only happen in one place, and that place should be the server:

```
if plugin not provided
  plugin = name
```

The updated client code in this PR should work the way that I showed: if "plugin=" was supplied, then send along the plugin attribute. If not, that is fine, the server will deal with it being absent. But _don't_ replicate the "if plugin not provided, plugin = name" logic in the client. There is no need for that. Not only is that redundant, but it makes testing the server logic hard. In point of fact, you missed a bug in the server logic probably because the client was replicating the logic of the server and masking the server bug.

So both the client and the server do not need the logic. I'm proposing checking in one place. And the better place is the server side, because that allows backwards compatibility with clients that didn't send the "plugin=" attribute in the past.

PR #1576, opened 2025-01-15
Bug fix when "stores" is not present in YAML config

nick-enoent:
Resolves error when "stores" is returned as None, rather than an empty dictionary in the event no "stores" are present in the YAML configuration


tom95858:
@emdonat could you please confirm this is fixed?

PR #1585, opened 2025-01-21
Correct stat() error check in decomposition library loading

nichamon:
The original code incorrectly used !stat() which would continue the loop when the library was failed to be loaded (stat returns 0). Change to continue searching for the library when stat() returns a non-zero. Otherwise, an error message is reported.


nichamon:
I found this bug when I tested b4.4 for v4.4.5 release. The decomposition libraries were installed in /<PREFIX>/lib64, and I had PATH set to <PREFIX>/lib;<PREFIX>/lib64. ldmsd reported a false positive decomposition library loading error.

nichamon:
After I created this pull request, I found out that the bug has been fixed on 'main'. The fix on 'main' gives more information on the error, so I'm porting that back to b4.4.

PR #1588, opened 2025-01-23
Add AVRO/Serdes support in LDMS Stream

narategithub:


narategithub:
Associated test case:  https://github.com/narategithub/ldms-test/blob/main-dev/ldms_stream_avro_ser_test

ldms/src/core/ldms.h
```diff
@@ -1429,18 +1531,22 @@ int ldms_stream_publish_file(ldms_t x, const char *stream_name,
...
 	LDMS_STREAM_EVENT_UNSUBSCRIBE_STATUS, /* reporting unsubscription status */
 	LDMS_STREAM_EVENT_CLOSE, /* reporting stream client close event.
 				  * This is the last event to deliver from a
 				  * client. */
+
+	LDMS_STREAM_EVENT_LAST, /* The last enumeration; not a real event */
 };
 
+/**
+ * TODO DOC ME
```
tom95858, line 1546: @narategithub we probably should doc this interface or remove the comment.

ldms/src/core/ldms.h
```diff
@@ -1506,6 +1617,12 @@ ldms_stream_subscribe(const char *stream, int is_regex,
 		      ldms_stream_event_cb_t cb_fn, void *cb_arg,
 		      const char *desc);
 
+/* TODO Just try it out .. to be revised. */
```
tom95858, line 1620: Remove TODO?

tom95858 (changes requested):

ldms/src/core/ldms.h
```diff
@@ -1381,8 +1381,35 @@ void ldms_qgroup_info_free(ldms_qgroup_info_t qinfo);
...
+ *
+ * \retval s The symbol (e.g. "LDMS_STREAM_STRING").
+ */
+const char *ldms_stream_type_sym(ldms_stream_type_t t);
+
+typedef struct ldms_stream_client_s *ldms_stream_client_t;
+typedef struct json_entity_s *json_entity_t;
+
+/* from avro/schema.h */
+typedef struct avro_obj_t *avro_schema_t;
```
morrone, line 1405: Please, please don't do this. We should get this definition by doing #include <avro/schema.h>. Please don't cut-and-paste definitions from external packages.

I can only presume that this is done to avoid having ldms.h have ever-present dependencies on external avro packages. But the solution to this problem is to just introduce a separate header. Make building avro support optional if we want the dependency to be optional.

narategithub: @morrone your presumption is correct. I first did "ldms_avro.h" and "ldms_avro.h". It has only one API, so I just put it here as it is more convenient and I was thinking that this `typedef` doesn't care about the definition of the structure. But, if `avro` somehow changes redefine `avro_schema_t` to something else and we're done, which is bad.

Thanks for catching this, I'll put the avro bits into its own thing.

morrone, line 1405: Please don't cut-and-paste definitions from external headers. We should have the proper #include <avro/schema.h> (and other includes) instead.  Maybe that means making the inclusion conditional and/or splitting the public stream API out into a header other than ldms.h.

ldms/src/core/ldms.h
```diff
@@ -1429,18 +1531,25 @@ int ldms_stream_publish_file(ldms_t x, const char *stream_name,
...
+/**
+ * \brief String symbol of event type \c t for printing.
+ *
+ * \retval s The string symbol of the given event type (e.g. "LDMS_STREAM_EVENT_RECV").
+ * \retval "UNKNOWn" If the event type \c t is out of range.
+ */
+const char *ldms_stream_event_type_sym(enum ldms_stream_event_type t);
+
 /* For stream data delivery to the application */
 struct ldms_stream_recv_data_s {
```
morrone, line 1554: This struct now needs a lot more documentation.

ldms/src/core/ldms.h
```diff
@@ -1451,7 +1560,12 @@ struct ldms_stream_recv_data_s {
 	uint32_t data_len;
 	const char *name; /* stream name */
 	const char *data; /* stream data */
-	json_entity_t json; /* json entity */
+
+	json_entity_t json; /* json entity (LDMS_STREAM_JSON type) */
```
morrone, line 1564: Why is the json field present at the same time as the the avro fields? Shouldn't that be a union to make it clear that it will be one or the other?

How would the user of the API go about freeing this struct? Is there a helper function to do that?

morrone (changes requested):
I don't really understand why this API is trying to embed partially talking to a schema registry. Why force the use of schema registry at all?

It seems like it would make a lot more sense to leave the decision of whether or not to use a schema registry up to the user. Leave the avro encoding/decoding up to the users. 

I would suggest that the ldms stream API just accept the pre-encoded data. Maybe it also adds a field to give the schema name (and maybe schema version?) for the associated data.

narategithub:
@tom95858 `ldms_stream_avro_ser` now becomes its own library and is ready to merge.

@morrone I insist that we keep `AVRO_SER` instead of `AVRO` since the payload has to be framed by libserdes. If the application wants to publish the Avro/Serdes payload directly, it call `ldms_stream_publish()` with `AVRO_SER` stream type. I put a remark on the API doc.

stream API / struct doc will be added in another PR. I want to contain this PR within stream Avro/Serdes support.

morrone:
> @morrone I insist that we keep `AVRO_SER` instead of `AVRO` since the payload has to be framed by libserdes. If the application wants to publish the Avro/Serdes payload directly, it call `ldms_stream_publish()` with `AVRO_SER` stream type. I put a remark on the API doc.
> 
> stream API / struct doc will be added in another PR. I want to contain this PR within stream Avro/Serdes support.

I think I understand where you are coming from. But I think this maybe reflects a misunderstanding of what Avro is, and what role libserdes plays.

Apache Avro is a published serialization format. It is a standard. You could compare Avro to json.  If you compare libraries between avro and json, one near equivalent to libserdes might be libjansson. You wouldn't consider naming the json encoded stream "json_jansson", right? Because what library was used to generate the json encoded data in the stream isn't relevant. The encoding is json.

Likewise, we shouldn't name the avro stream "avro_ser" because it doesn't matter that what library was used to create the avro-encoded data. The encoding is avro. Avro is a published standard, and any library that supports avro can be used to decode that avro data.

In other words, the payload has been "framed" (i.e. serialized) by _avro_. libserdes is just one particular library implementation that helps do that.

morrone:
@narategithub Also, I don't think you have adequately explained why you are embedding forced schema registry usage in the API. I don't think that belongs in the API. It makes the API overly complicated. I would suggest that we leave it up to the user decide if they want to use a schema registry, or agree on a schema through other means. Just allow the user to generate the avro data blob and pass the encoded data blob to the ldms stream API.

I'm not really understanding why this is much more than just adding an avro type to ldms_stream_type_t.

Even if you, for some reason, think that libserdes is too hard to use and you want to provide additional wrappers for ease of use, you should make those helpers separate from the ldms stream API. I would more prefer to see the use look like this:

1) Encode avro data using Narate's helper function (or whatever library one likes, for that matter)
2) Call ldms_stream_publish()

In other words, we don't need ldms_stream_publish_avro_ser(). That combines too many layers of operation into one big, single-purpose, overly restrictive way of doing avro encoding. This takes power aware from the user. It decides that they _must_ use a registry (which is _not_ a general requirment of avro), and not only that, they have to use the exact _type_ of registry that _you_ have chosen to hard-code into the API's implementation.

morrone:
@narategithub  OK! I think I finally understand what you were trying to convey with all the "Avro/Serdes" and "LDMS_STREAM_AVRO_SER".

So here's what I now believe is really being done in this PR:

It adds ldmsd stream support for Avro encoded messages _that are encapsulated in Confluent Platform framing (cp1)_. 

Note that "libserdes"/"serdes"/"ser" in _no way_, implies, on its own, Confluent Platform framing! So the current documentation is pretty misleading.

So I would suggest some changes based on your approach:

1) Stop saying "Avro/Serdes", and instead say "Avro/CP1" or "Avro+CP1" or "Avro+CP1 framing" (something along those lines).
2) Rename LDMS_STREAM_AVRO_SER with LDMS_STREAM_AVRO_CP1.
3) Somewhere clearly document and explain that "CP1" framing means "Confluent Platform Framing".
4) If you leave initial serdes handle configuration to the user, you _must_ document that they need to have "deserializer.framing" nad/or "serializer.framing" set to "cp1" for this library to work.

I really can't stress enough how important it is communicate the dependency on "Confluent Platform framing (cp1)". To my understanding, this is a framing format specific to the Confluent company. It isn't even a documented format (or at least very easy to find), although open source implementations obviously exist, so we could figure it out.

In summary, please stop calling all of this "serdes", and instead call it what it is: cp1 framing of avro encoded message.

morrone:
Just to reiterate a bit, and maybe a bit for @tom95858.

The messages in ldms that are currently labelled "LDMS_STREAM_AVRO_SER", are encoded as avro plus cp1 framing.

If I google "avro ser". I don't learn anything directly about the necessary framing on top of avro encoding.

If I google "avro cp1", the google AI tells me that cp1 is "Avro CP1 is a Confluent Platform (CP) framing format used in Avro serialization and deserialization. It's the default framing format in the libserdes library". That is super useful in understanding the message format in ldms.

And like I proposed, we should additionally have words that explain what "Confluent Platform Framing (CP1)" is.

And here is why it isn't sufficient to only tell the users of the LDMS_STREAM_AVRO_SER messaging to "use libserdes". This PR says:

```
 *     If you wish to just publish bytes of Avro/Serdes encoded data, simply
 *     call \c ldms_stream_publish() the encoded data with \c stream_type being
 *     \c LDMS_STREAM_AVRO_SER.
```

I can use Acro/Serdes quite easily to generate messages that either have CP1 framing, or _do not have CP1 framing_. Only one of those will be the correct format for the message in ldms. That is why it is insufficient to tell the user to use libserdes. We _must_ document what the message encoding is.

And heck, what is to stop a user from generating avro+cp1 encoded messages using an entirely different software stack then libavro-c/libavro-c++/libserdes/etc. What if the user's application is written in java, go, rust, etc? Maybe the avro+cp1 message is entirely encoded in the native language, and only the final step has the C linkage to publish through ldms? How would they know that the messages need CP1 framing?

And I would, of course, further argue that giving the ldms message a name that fully reflects its encoding (_AVRO_CP1), rather than being named after just one library that implements CP1 (libserdes) would logically follow as well.

PR #1590, opened 2025-01-24
Prevent race between update scheduling and set deletion

nichamon:
Fix a race condition where a set delete request from a peer could invalidate maps while an update is being scheduled. Hold the lock during the entire update submission to ensure maps remains valid.


nichamon:
The patch has been tested that it doesn't create deadlock, and the crash has gone away.

PR #1591, opened 2025-01-27
Add parameter to account for local mode when using --generate-config-path

nick-enoent:


nick-enoent:
@tom95858 This PR is good to merge.

PR #1592, opened 2025-01-28
Prevent race between look-up and set deletion

nichamon:
Fix a race condition where a set delete request from a peer could invalidate maps while ldmsd is handling a rendezvous lookup and is about to submit a remote read request.  an update is being scheduled. Check if the remote and local map handles of the set are valid and hold the lock during the entire read submission to ensure maps remains valid.

This is a corner case. The time window between the times the server deletes the set and responds to a lookup request is very small. In practice, this could happen when a set is very short live.


nichamon:
During a resiliency test, an L2 crashed in the lookup's read complete because the set's data was garbage.

PR #1593, opened 2025-01-29
Add the yyjson library

evanmcc:
The existing JSON generation code uses sprintf and allocates quite a bit. The open source yyjson library, via limiting allocation, copying, and specialized type pretty-printing code, claims to be able to hit gigabytes- per-second rates both with encode and decode.

Switching to this code in the storage plugin critical section more than doubles the per-process workload we can handle at our site.

This is a pretty basic version of this PR.  Two bits of further work that'd probably be a good idea:
1. I punted on doing the timestamps because it was quicker not to work it out for the proof of concept.  It might be worth the time, since we generate a lot of timestamps, to work out how to do this conversion using the yyjson primitives, but the worthwhile-ness would require some careful measurements.
2. Ideally, yyjson would not be a hard requirement, but I am too much of a libtool neophyte to know the right way to do this.  I've probably also done something wrong, autotools suggestions and fixes welcomed.

Finally, this compiles but it isn't 100% identical to the code we tested.  I haven't had the opportunity to test it in our setup yet.

### DIFF
--- a/ldms/src/store/avro_kafka/store_avro_kafka.c
+++ b/ldms/src/store/avro_kafka/store_avro_kafka.c
@@ -51,7 +51,8 @@ typedef struct aks_handle_s
 	serdes_t *serdes;	    /* The serdes handle */
 	enum {
 		AKS_ENCODING_JSON = 1,
-		AKS_ENCODING_AVRO = 2
+		AKS_ENCODING_AVRO = 2,
+        AKS_ENCODING_YYJSON = 3
 	} encoding;
 	char *topic_fmt;	   /* Format to use to create topic name from row */
 	char *topic_name;
@@ -60,7 +61,7 @@ typedef struct aks_handle_s
 static const char *_help_str =
     "   config name=store_avro_kafka [path=JSON_FILE]\n"
     "       encoding=MODE"
-    "            MODE is one of JSON or AVRO (default)."
+    "            MODE is one of JSON or YYJSON or AVRO (default)."
     "       kafka_conf=PATH"
     "            Path to a file in Apache Kafka format containing key/value\n"
     "            pairs defining Kafka configuration properties. See\n"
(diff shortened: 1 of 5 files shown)


ldms/src/store/avro_kafka/store_avro_kafka.c
```diff
@@ -60,7 +61,7 @@ typedef struct aks_handle_s
 static const char *_help_str =
     "   config name=store_avro_kafka [path=JSON_FILE]\n"
     "       encoding=MODE"
-    "            MODE is one of JSON or AVRO (default)."
+    "            MODE is one of JSON or YYJSON or AVRO (default)."
```
morrone, line 64: I don't think this should be a configurable mode. yyjson is a json implementation, like libjansson, not a different encoding. It somes out as json regardless of the library used to to do the encoding.

If yyjson is going to be a _required_ external library (and that _is_ how this patch implements it in configure.ac), and it is a lot faster, then I would say just replace libjansson with yyjson.

On the other hand, if we want to make yyjson an optional external library, then we should have conditional compilation of the backend that implements the json mode. If only libjansson is available, compile against that. If yyjson is available, compile against that.

evanmcc: That's reasonable.  The existing plugin doesn't use jansson (which in our tests is actually slower than the sprintf-based code in the ldmsd_decomp.c), at least using the json encoding mode.  

If you can give me some guidance about libtool library presence testing and what defines it puts in place, I can change it so that it's conditional to libyyjson being on the system.  It should be pretty easy to make it a seamless replacement.

evanmcc:
@morrone I think I did what you wanted in [8bbfa17](https://github.com/ovis-hpc/ldms/pull/1593/commits/8bbfa1795788bd431bda42007e05d792c4971912).  Let me know if this works for you.

I probably also owe you an update to the README wrt building but I think that might have to wait till next week.

morrone (changes requested):

emdonat:
Clearing a milestone from this as we have to have a deeper discussion about where this might end up and how it will be built against (requirement vs Chris' suggestion of building against if present).

evanmcc:
Sorry, I haven't pushed this forward because there were some concerns about accepting it.  I'll try to address all the comments and get it rebased.  Let me know what you all decide.

morrone (approved):

tom95858:
I think we should consider using this code in some paths. But I'm going to close this for now. If someone wants to consider it for future work, please open an issue. It _is_ faster and scales better, but the API is more complicated  and in most of our use cases, not a critical factor.

PR #1594, opened 2025-01-30
Multi-instance plugin support (WIP)

morrone:
Builds on PR #1575. We switch to having the "self" ldmsd_plugin struct owned and allocated by the plugins, and allocated dynamically by them for each instance (if the plugin supports multi-instance).

This makes results in the allocation and ownership of the "context" pointer being much cleaner and more straight forward. Instread of having ldmsd allocate the context pointer on behalf of a plugin, the plugin just allocates it itself. This also makes it much clearer where the context (and ldmsd_plugin) need to be freed: in also in the plugin. We choose term() as the location that is agreed upon to free those allocations.


morrone:
@tom95858 This code isn't nearly finished (doesn't even compile), but:

Maybe take a look at ldms/src/ldmsd/ldmsd_config.c:load_plugin_instance() and ldms/src/ldmsd/ldmsd_config.c:unload_plugin_instance() in this PR.

I think we need to be creating a generic plugin wrapper instance sooner in the code than we have been. We need to catch things like the handle from dl_open, which we weren't doing before. And by having this instance struct sooner, we don't need to violate layering by shoving things in the API struct that the plugin returns from get_plugin().

I don't entirely understand where you were going with your cfgobj stuff though, so I'm not sure how to harmonize the two approaches yet.

morrone:
I have a new approach in the works.

PR #1595, opened 2025-01-30
Fix `ldms_xprt_connect_by_name()` wrong xprt type handling in sync mode

narategithub:
The application-facing xprts are now rails. However `ldms_xprt_connect_by_name()` handled it as xprt and consequently corrupt the rail structure. This patch moves the related code into `ldms_rail.c` and handle it as rail accordingly.


nichamon:
@narategithub Do we need this fix in b4.4?

narategithub:
`b4.4` does not have rail. So, this is not needed in `b4.4`.

nichamon (approved):

PR #1606, opened 2025-02-11
Development after multi-instance plugin change

narategithub:
@tom95858 here you are.


narategithub:
@tom95858 I just rebased this. The rebased code also tested with `agg_test`, `ldmsd_sampler_exclusive_thread_test`, and `ldmsd_flex_decomp_test`.

ldms/src/store/store_csv_common.c
baallan, line 1: @narategithub it's questionable whether the "split general timestamp values into two columns" is correct behavior. The split into 2 parts of the primary metric set time stamp is a legacy behavior from NCSA that is unneeded for any other time or for most clients. 

I suggest that we *not* generate 2 columns per timestamp in either data or header (othe r than for the legacy primary metric set time stamp), making the correct fix to be get rid of the extra column, not add the extra header.

tom95858: @baallan, I agree with you, however, the code is currently broken to the extent that the header doesn't match the data. Let's just get this in and if you/CSV users decide they want to change it they can.

baallan: @tom95858 given that it's currently broken, why not just fix it the right way now? the original timestamp with all the special column split handling isn't a 'normal' metric anyway.

tom95858: @baallan, because you didn't say how to "fix" it. Only that you feel it is broken. Aren't there countless scripts that eat CSV? Don't they assume that the extra column is present? Why did those scripts ever work? What I am suggesting is that we reasonably ask these questions, answer them, and then post a well considered change. Right how, we have a mix and @narategithub makes the row output consistent.

tom95858:
@narategithub is this ready to go? Did it pass the tests? If so, let's get it merged, and then we can start considering @morrone's API changes.

narategithub:
@tom95858 yes. Let's merge.

PR #1608, opened 2025-02-17
Remove set_route command functionality

nichamon:
The set_route command is currently broken due to locking issues:
- Returns while holding both global producer config lock and producer lock
- Makes requests to other producers while holding locks, creating race conditions
- Info structure contains pointers to both set and prd_set

No active issues or users have been reported for this functionality. Removing this code simplifies the locking model and eliminates potential race conditions.


tom95858:
@nichamon, I already have code to remove this in main.

PR #1609, opened 2025-02-18
Remove the ldms/ldms_utils tree

morrone:
The files in ldms/ldms_utils/ all look pretty old. A number of them look like they wouldn't work any more. Can we just delete the entire tree?

Is there anything in the ldms/ldms_utils/ tree that we should keep?


tom95858:
I agree, let's give folks a chance to ruminate before we pull the trigger.

PR #1615, opened 2025-02-18
Rename ldmsd_store_t and ldmsd_sampler_t

morrone:
Rename "struct ldmsd_sampler_inst" to "struct ldmsd_cfgobj_sampler".
Rename ldmsd_sampler_t to ldmsd_cfgobj_sampler_t (to match previous).
Rename "struct ldmsd_store_inst" to "struct ldmsd_cfgobj_store".
Rename ldmsd_store_t ldmsd_cfgobj_store_t (to match previous).

These changes better reflect that these struct are derived from
the base struct ldmsd_cfgobj.

The typedef names now directly match the underlying struct, with
just "_t" appended to the name. Previously the typedef names did
not match the struct names.


tom95858:
@morrone, I like this change a lot. I'm going to try and pull this change on top of the current mutli-cfg stuff and then force push it. I will likely rebase and squash some of the changes to make the change history a bit prettier.

morrone:
Tom pulled this changed into his tree, so I closed this PR.

PR #1618, opened 2025-02-21
Create new plugin API abstractions

morrone:
Develop better plugin API abstractions.
    
    Make read-only ldmsd's access to the API function pointer structs: struct ldmsd_plugin,
    struct ldmsd_store, and struct ldmsd_sampler. The plugins now own those structs.
    
    Introduce the ldmsd_plug_api.[ch] files, which most importantly include the accessor
    functions for the new ldmsd_plug_handle_t, which is a opaque to the plugins.
    
    ldmsd passes an opaque ldmsd_plug_handle_t to the plugins in each plugin API call.
    
    Plugins may use the accessor functions to get and set things through the handle.
    
    In particular, a plugin may use ldmsd_plug_context_set() to record its local context
    pointer in one call, and use ldmsd_plug_context_get() to retrieve its local context
    pointer.
    
    Functions calls constructor() and destructor() are added to the plugin API to simplify
    plugin multi-instance support.
    
    The filesingle sampler is upgraded to multi-instance support.


morrone:
This builds, but has virtually no testing yet. I share the draft now just so folks know where I'm heading and can provide early feedback.

morrone:
OK, I am considering this PR done feature wise. It now needs testing and bug fixing.

morrone:
@tom95858 I've rebased PR#1618 on top of the main branch. The latest at this time is commit 3f757c28d7.

morrone:
@tom95858 I fixed the compile error in the rebase. Lastest commit is now 315de81eb.

morrone:
Included in PR #1698

PR #1621, opened 2025-02-27
Fix how ovis_json:json_dict_build() use va_arg()

nichamon:
Before the change, json_dict_build() recursively calls functions to create a dictionary. It passes a va_list object to the nested function. The function calls va_arg(). According to the va_arg man page, after the function returns, the va_list object is undefined.

The patch passes a va_list pointer to its nested function, so the va_list properly moves when the inner function calls va_arg().


nichamon:
@tom95858 Please close this pull request if you have already incorporated the change in your branch.

PR #1624, opened 2025-03-06
Fix incorrect cast in ovis_json

tom95858:
The symptom here was a garbage value for the sample count.

The JSON_INT is passed as an int, however, the ovis_json logic was using uint64_t to pull the value from the stack. This resulted in garbage values on some platforms.


baallan:
@tom95858 this fix to json was incorrect; see #1626.

PR #1631, opened 2025-03-19
Fix SIGSEGV in ldms_stream introduced by "hops" info

narategithub:


ldms/src/core/ldms_stream.c
```diff
@@ -380,12 +380,24 @@ __remote_client_cb(ldms_stream_event_t ev, void *cb_arg)
 	op_ctxt->op_type = LDMS_XPRT_OP_STREAM_PUBLISH;
 	op_ctxt->stream_pub_profile.hop_num = _ev->pub.hop_num;
 	op_ctxt->stream_pub_profile.recv_ts = _ev->pub.recv_ts;
+
+	struct ldms_stream_hop *hops;
+	uint32_t hop_cnt;
+
+	if (_ev->sbuf) {
```
nichamon, line 387: Is `_ev->sbuf` null when it is a local publish?

ldms/src/core/ldms_stream.h
```diff
@@ -129,6 +129,7 @@ struct ldms_stream_client_s {
 /* the full stream message */
 struct ldms_stream_full_msg_s {
 	struct ldms_addr src;
+	uint32_t pad0;
```
nichamon, line 132: Just out of curiosity, why are we adding pads in `ldms_stream_full_msg_s` and `ldms_stream_full_msg_s`?

nichamon (approved):

ldms/src/core/ldms_stream.h
tom95858: I think it is to ensure alignment of the payload on a 32b boundary. But I would suggest we also add one or more 64b "reserved" fields as well to allow us to add additional data to the header for unforeseen features in the future that won't break backward compatibility.

PR #1632, opened 2025-03-20
Introduce event-based storage

nichamon:


nichamon:
@tom95858  The remaining work I plan for the pull request is to add a section to the `store_time_stats` output to display the percentage of each stage:decomposition, waiting for a storage worker, waiting idly in the worker queue, and the actual storing time.

baallan:
@nichamon @emdonat What problem is this solving? Where is the design document/discussion?

nichamon:
@tom95858 I added the number of store operations per second and the overall percentages of each stage to the store_time_stats. Here is the updated store_time_stats output.

image

baallan:
Not really a satisfactory answer in the meeting. did we end up with a write-up in the discussion pages or elsewhere?

tom95858:
@nichamon, given that we've pushed 4.5.1 out, should we consider accepting this?

nichamon:
> @nichamon, given that we've pushed 4.5.1 out, should we consider accepting this?

Before we merge this, I'd like to see how much it helps with reducing the storage bottleneck, especially with Rails. If there's no rush, how about we plan to include this in the next release and perform performance comparisons between versions with and without this patch?

The patch introduces a new config variable (number of storage threads) that adds some complexity. However, I believe decoupling storage from IO threads would benefits scalability and configuration (Using data transfer rate to determine rail sizes and using storage workload to determine number of storage threads). With more time, we could revisit the default value of the number of storage threads, which the current default is 1.

ldms/src/ldmsd/ldmsd_event.c
```diff
@@ -0,0 +1,377 @@
...
+
+#define STRG_WORKER_WAIT	0
+#define STRG_WORKER_SIGNAL	1
+
+TAILQ_HEAD(strg_worker_wait_list, strg_worker_wait_cond);
+struct strg_worker_wait_list strgw_wait_list = TAILQ_HEAD_INITIALIZER(strgw_wait_list);
+pthread_mutex_t strgw_wait_list_lock = PTHREAD_MUTEX_INITIALIZER;
+
+/* The function blocks until it can get a worker. */
+struct strg_worker *strg_worker_get(void)
```
morrone: I would like to see this function rewritten without the goto statements.

tom95858:
@nichamon, is this ready to merge? If so, please fix the merge conflicts.

nichamon:
@tom95858 I want to hold this off for the next release so we could do more testing, so I'll convert this to a draft now.

tom95858:
@nichamon, what is the status of this pull request? We need to merge it sooner rather than later or we will end up continuously rebasing it.

nichamon:
> @nichamon, what is the status of this pull request? We need to merge it sooner rather than later or we will end up continuously rebasing it.

@tom95858 I just rebased and fixed a few bugs after testing the patch on my machine. I haven't tested it at scale; however, it's ready to be merged now.

tom95858:
@nichamon, could you please look at `strg_worker_get', and addresses @morrone's comments. I think that this could be modified to avoid the goto, except for perhaps the retry path. But the goto's may make this code a bit more difficult to follow.

nichamon:
@tom95858 @morrone The function has been refactored to remove all `goto`.

ldms/man/ldmsd_event_based_storage.rst
```diff
@@ -0,0 +1,374 @@
...
+ARCHITECTURE
+============
+
+Worker Thread Pool
+------------------
+
+The event-based storage system maintains a configurable pool of worker threads
+dedicated to storage operations. Each worker thread has its own scheduler and
+event queue, allowing storage events to be processed in parallel while
+maintaining ordering guarantees per worker.
```
morrone: Is this ordering of any significance? If we are more-or-less randomly distributing events across workers, then have we not already lost ordering?

ldms/man/ldmsd_event_based_storage.rst
```diff
@@ -0,0 +1,374 @@
...
+
+When a metric set is updated, LDMSD creates a storage event containing the
+necessary context for processing:
+
+* **Set snapshot**: A point-in-time copy of the metric set to preserve data
+  integrity. The snapshot ensures that storage operations access consistent data
+  even if the live set is updated during processing.
+* **Storage policy reference**: Pointer to the storage policy that defines how
+  data should be stored
+* **Producer set reference**: Reference to the producer-consumer relationship
```
morrone: I don't know what "producer-consumer relationship" refers to here.

ldms/man/ldmsd_event_based_storage.rst
```diff
@@ -0,0 +1,374 @@
...
+  tracking the source of the data
+* **Row list (for decomposed storage)**: For storage plugins supporting
```
morrone: Should decomposition maybe happen in the worker instead of before the worker?

ldms/man/ldmsd_event_based_storage.rst
```diff
@@ -0,0 +1,374 @@
...
+
+2. **Snapshot creation**: Before processing any storage policies, a snapshot of
+  the metric set is created if there are active storage policies
+
+3. **Event creation**: For each applicable storage policy, a storage event
+  context is created containing the snapshot and policy information
+
+4. **Worker acquisition**: LDMSD attempts to acquire an available worker
+  thread. If all workers are at capacity (queue full), the requesting thread
+  blocks on a condition variable until a worker becomes available
```
morrone: What happens to the data collection process when the requesting threads block? Do metric sets stop getting live updates? Do snapshots from the live updates stop getting created?

morrone: Maybe expand some too on this aspect of the design. Why are callers responsible for acquiring a worker? Why don't we just place all events into a single shared queue? We already seem to have fairly heavy weight locking in the critical path (pthread_mutexes) with this solution, so there might not be a performance hit for doing that.

ldms/man/ldmsd_event_based_storage.rst
```diff
@@ -0,0 +1,374 @@
...
+
+5. **Event queuing**: The storage event is added to the acquired worker's event
+  queue
+
+6. **Worker processing**: The worker thread processes the event from its queue:
+
+   * Retrieves the set snapshot and storage policy details
+   * For decomposed storage: calls the decomposition function to transform the snapshot into rows
```
morrone: Oh, I guess the worker _does_ call the decomposition. I guess I'm not sure why earlier the documentation claims that the event already contains a list of the decomposed rows. Do one of these documentation points maybe need updating?

ldms/src/ldmsd/ldmsd_event_strg.c
```diff
@@ -0,0 +1,490 @@
...
+	static int worker_idx = 0;
+	struct strg_worker *w = NULL;
+	int attempts;
+	int current_idx;
+	int should_wait;
+
+	errno = 0;
+
+	/* Check if there are already threads waiting */
+	pthread_mutex_lock(&strgw_wait_list_lock);
```
morrone, line 336: It is not entirely clear to me why we need this is in the critical path. We don't care if this thread wastes some time looking for workers if all it is going to do is go to sleep waiting. Unless we think strg_worker_try_aquire() is maybe too heavy weight to try when the workers are busy? But those increments/decrements will be happening in parallel anyway with the workers, so I guess it isn't obviously to me that that would be an issue.

So maybe we could just start this function at:

270:	/* Try to find an available worker */

morrone (commented):
There are still things I am working on understanding. But here are some of my initial questions and comments.

ldms/src/core/ldms.h
```diff
@@ -2901,6 +2902,90 @@ ldms_set_t ldms_set_create(const char *instance_name,
...
+ * - ldms_set_snapshot_get() - Get a reference to a snapshot
+ * - ldms_set_snapshot_put() - Release a reference to a snapshot
+ *
+ * Do NOT use ldms_set_delete() or ldms_set_put() with snapshots.
+ *
+ * The standard ldms_metric_* and ldms_set_info_* functions can be used to access metric
+ * and set information from a snapshot.
+ *
+ * Snapshot properties:
+ * - Always have an array cardinality of 1. If the original set has a larger
```
morrone, line 2938: Maybe you could elaborate on what "array cardinality" means? I have a vague recollection that metric sets can actually have multiple copies in memory for some reason or other, and I guess this is related to that. But why does that cardinality exist in the first place? Are we making that functionality moot by avoiding making snapshots of earlier versions in the array?

My understanding of this area is _very weak_. I am envisioning that maybe a sampler samplers at a rate of 1sec, and has an array cardinality of 5 on its metric sets. Then the aggregator can aggregate only every 5 seconds and get all 5 seconds worth of data all at once. In which case we would want to snapshot each of the entries in the array, not just the "current" one.

But maybe I'm imagining all of that, and that isn't the way the arrays work at all. :)

ldms/man/ldmsd_event_based_storage.rst
nichamon: Yes, global ordering across workers is not preserved because storage workload is distributed across workers. The statement clarifies that within each worker's queue, events are processed first-come-first-serve.

ldms/man/ldmsd_event_based_storage.rst
nichamon: Good catch - the wording was confusing. The producer set reference is for tracking the data source for statistics purposes. 

For your information, the rst file was a mix of user and developer documentation. I've revised it to be user-focused, discussing configuration and LDMSD performance statistics for tuning the worker pool size and queue depths. Architecture and design rationale are now documented in comments in ldmsd_event_strg.c.

ldms/man/ldmsd_event_based_storage.rst
nichamon: Decomposition happens in the IO thread to guarantee samples are processed in arrival order. If decomposition occurred in workers, there's no guarantee that Sample 1 would be decomposed before Sample 2 for a given set - they could be assigned to different workers and processed out of order. While unlikely, this isn't guaranteed and matters for operations like rate calculation that depend on sequential processing.

ldms/man/ldmsd_event_based_storage.rst
nichamon: For the blocking behavior, when all workers are at capacity, the IO thread blocks in the update callback. This has the same behavior as when storage plugins take a long time to store data to databases or send data to third parties (e.g., Kafka) in the current synchronous storage - the update callback doesn't complete, so no new update is scheduled for that set and no new snapshots are created. The key difference with event-based storage is the configurable buffer (storage_threads × storage_queue_depth) that allows the system to absorb bursts before blocking occurs.

A central queue would have the same backpressure behavior when full - IO threads would still block waiting for space. Both approaches require limiting queue depth to prevent unbounded memory growth, and both result in the same backpressure when limits are reached.

ldms/man/ldmsd_event_based_storage.rst
nichamon: This section is outdated. The decomposition is processed on the IO threads. Thanks for catching this.

ldms/src/ldmsd/ldmsd_event_strg.c
nichamon: The wait list check enforces FIFO ordering among competing threads. Without it:

- Thread A finds all workers full, begins creating its wait entry
- Worker becomes available
- Thread B arrives, sees available worker via round-robin, takes it
- Thread A finishes creating wait entry, starts waiting

With the check, Thread B immediately joins the wait queue behind Thread A. The check requires a mutex lock/unlock to just verify if the wait list is empty, which happens once per worker acquisition attempt.

ldms/src/core/ldms.h
nichamon: LDMS applications, including LDMSD, have no knowledge of metric set arrays. Applications always interact with the current array index, and LDMS is designed so that data from different indices is delivered seamlessly through separate update completions. When a snapshot is created, it captures the current state of the metric set - which is the current array index at that moment. Each array index is delivered as a separate update completion, so each gets snapshotted independently when its update arrives.

ldms/man/ldmsd_event_based_storage.rst
tom95858: @nichamon, I think @morrone is asking a different question. "Does it matter if the storage events are out of order?" The answer, I think, is it depends. If you are stuffing it on the Kafka bus, maybe. If you're storing to a CSV file, then the timestamps would be out of order in the file; does that matter? If you are storing it in indexed storage, then the answer is "probably not."

nichamon:
I'm closing this PR as it has been incorporated in #2219.

PR #1635, opened 2025-03-26
Fix __APPEND macro to prevent infinite loop of reallocation

nichamon:
The macro was causing an indefinite loop when the length of the string to be appended equaled the size of the remaining buffer. In this case, LDMS_ROUNDUP(len-sz, __APPEND_SZ) evaluted to zero, resulting in reallocation with the same buffer size repeatedly.

This patch ensures that the buffer is always reallocated with a larger size when len == sz, preventing the infinite loop condition.


nichamon:
@tom95858 I have updated the code according to your discussion.

PR #1637, opened 2025-03-26
Refactor internal dictionaries of YAML cluster configuration

nick-enoent:
Support for subsets of hostlist expressions
Normalize ldmsd_yaml_parser parameter names
Add error handling if mutually exclusive parameters are provided


tom95858:
@nick-enoent could you confirm that u have fixed this. @emdonat can we ask that you confirm with independent testing?

nick-enoent:
> @nick-enoent could you confirm that u have fixed this. @emdonat can we ask that you confirm with independent testing?

I've tested this with subsets of hostlist expressions/advertisers/producer listeners/maestro/etc. That said, I'd like @emdonat to test some of his scripts with this PR before we merge.

PR #1639, opened 2025-03-27
Add "Interval" and "Offset" back to `plugn_status` command

narategithub:
The "Interval" and "Offset" sampling information has been (accidentally?) removed from the `plugn_status` command. This patch added them back. For storage plugin, the values of "Interval" and "Offset" are "-".

NOTE: Currently there is no other command to obtain the sampling interval.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -1588,12 +1588,15 @@ class LdmsdCmdParser(cmd.Cmd):
         rc, msg = self.comm.plugn_status(arg['name'])
         if rc == 0:
             plugins = fmt_status(msg)
-            print("Config Name  Plugin                   Type         libpath")
-            print("------------ ------------------------ ------------ ------------")
+            print(f"{'Config Name':12}", f"{'Plugin':24}", f"{'Type':12}",
+                  f"{'Interval':12}", f"{'Offset':12}", f"{'libpath':12}")
+            print('-'*12, '-'*24, '-'*12, '-'*12, '-'*12, '-'*12)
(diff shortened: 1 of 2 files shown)


tom95858:
I don't think they were accidentally removed. The plugn_status command applies to  all plugin types, but the values in question only apply to sampler plugins. So I think we need a strategy to have either sampler_status/store_status/auth_status commands or have some way to have the plugn_status command to post configuration information (i.e. avl), but note that this doesn't handle default values. But publishing sample interval and offset for a store is incorrect.

narategithub:
@tom95858 currently there is no command to obtain "interval" and "offset" information. Do you suggest that we have another command? Or do we split the "plugn_status" output into sections according to their types?

tom95858:
Yes, I'm suggesting that we have store_status, sampler_status, etc... The plugn_status command is actually kind of useless except that it shows you what has been loaded. But having plugn_status return interval and offset for a storage plugin is very confusing to the user. I am guessing that the motivation for your change is that our test infrastructure is using that command to query configuration values to ensure they are set properly.

I think we need to push this to 4.5.2 so that we can discuss because I think what we really need is to have a sampler policy object similar to the storage policy object. The MI patch effectively implements that but we are lacking the configuration query commands to access the plugin configuration specific data (except what was specified in config()).

What I was trying to suggest is that the plugn_status command could dump the avl/kvl of the plugin configuration instance. I was trying to point out that this doesn't necessarily solve the test problem if default values are involved because there will not be an avl/kvl value present in the config string.

narategithub:
Got it. Thanks! I'll skip that test case for now.

PR #1646, opened 2025-03-28
dcgm_sampler: reinitialize the dcgm library when needed

### LINKED ISSUE #1636: DCGM Sampler problems
We worked with LANL to enable DCGM sampler on their A100 system, and found a couple issues that were a struggle to get through, though in retrospect it was a pretty obvious omission.  We feel that some extra checks in the code and documentation in the manpage could save others from the same problem(s):

1. DCGM sampler will fail with `Host Engine is disconnected or invalid` errors when the `nvidia-dcgm.service` (i.e., /`usr/bin/nv-hostengine`) isn't running); Can we improve checking and messaging when failures occur?
2. When DCGM plugin fails to run, the daemon doesn't die, but it also doesn't permit the other samplers to work.

morrone:
In sample(), initialize the dcgm library if it is not currently initailized. When gpu_sample() fails in sample(), shutdown the dcgm library usage.

Fixes #1636

### DIFF
--- a/ldms/src/sampler/dcgm_sampler/dcgm_sampler.c
+++ b/ldms/src/sampler/dcgm_sampler/dcgm_sampler.c
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: (GPL-2.0 OR BSD-3-Clause)
  */
 #include <stdlib.h>
+#include <stdbool.h>
 #include <limits.h>
 #include <string.h>
 #include <dirent.h>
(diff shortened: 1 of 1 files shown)


baallan:
@morrone we've tested the combination of this and #1623 on our gpu system hops by installing just the rebuilt dcgm_sampler rpm. Looks good! Would like to get this into toss as soon as possible.

@tom95858 let's get #1623, #1646 merged please.

morrone:
Things aren't landing for `b4.4` any longer, so I am closing this PR. An equivalent patch landed on `main`.

PR #1665, opened 2025-04-11
Replace .rst auto-copying workflow with build-time symlinks  

nichamon:
This commit removes the GitHub workflow actions auto-copying mechanism 
and implements a solution directly in docs/conf.py that runs during    
ReadTheDocs build time.

The change addresses permission issues during GitHub workflow execution
and prevents modifications to contributors' branches. It also prevents 
duplication of all .rst files in two locations.

The PR also removes the .rst files in the docs/rst_man directory.


nichamon:
@narategithub Could you please review the `docs.yml` whether it started jobs at appropriate Github steps or not. Thanks!

morrone:
I've only skimmed this, but it looks like it is a major improvement!

Snell1224:
Looks good to me! Just need to add in the RTD_WEBHOOK_SECRET variable to the repo so the RTD webpage is triggered to build under "Generate Webhook Signature and Trigger RTD" in docs.yml.

Thank you @nichamon!

Snell1224 (approved):

PR #1666, opened 2025-04-11
Remove readthedocs workflow while we find a better solution

morrone:
I think we need to remove the docs.yml (readthedocs) workflow until we find a better solution. It is trying to push commits to the repository, and we really don't want it to do that. And currently it is causing build test failures for PRs.


morrone:
I didn't see #1665. Closing.

PR #1667, opened 2025-04-11
Add buffer length validation in __make_mdesc to silence compiler warnings

### LINKED ISSUE #1054: __make_mdesc compiler warning
ldms.c is triggering the following warning for me:

```
ldms.c: In function ‘__make_mdesc’:
ldms.c:1457:2: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
  strncpy(vd->vd_name_unit, md->name, name_len);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ldms.c:1456:13: note: length computed here
  name_len = strlen(md->name) + 1;
             ^~~~~~~~~~~~~~~~
ldms.c:1460:3: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
   strncpy(&vd->vd_name_unit[name_len], md->unit, unit_len);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ldms.c:1459:14: note: length computed here
   unit_len = strlen(md->unit) + 1;
              ^~~~~~~~~~~~~~~~
```

This warning seems valid to me. There doesn't seem to be checking to make sure that the destination is large enough. But I don't understand the macro magic going on here. I am not clear on how one would know what the available size of the destination is to add that check.

nichamon:
The patch adds remaining buffer size tracking and assertions to validate string copy operations in __make_mdesc(). While the memory allocation is correct in ldms_set_create() which calculates the total required space, including names and units, the compiler cannot determine this relationship. These explicit checks document the safety of the string operations and prevent the compiler warnings without changing the functionality.

This maintains the existing memory management approach while making the safety guarantees explicit to both the compiler and future code readers.

Resolve #1054

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -1635,10 +1635,18 @@ void __make_mdesc(ldms_mdesc_t vd, ldms_mdef_t md, off_t *value_off)
 	vd->vd_flags = md->flags;
 	vd->vd_array_count = __cpu_to_le32(md->count);
 	name_len = strlen(md->name) + 1;
-	strncpy(vd->vd_name_unit, md->name, name_len);
+	/*
+	 * The memory safety of these string operations is guaranteed by the allocation
+	 * process in compute_set_sizes() and ldms_set_create(). The 'meta' buffer is
+	 * pre-allocated with enough space to hold all metric descriptors, including
+	 * their variable-length name and unit strings. Each 'vd' (ldms_value_desc)
+	 * points to a section of this buffer, and its vd_name_unit flexible array member
+	 * has sufficient space allocated for both the name and unit strings.
+	 */
+	memcpy(vd->vd_name_unit, md->name, name_len);
 	if (md->unit) {
 		unit_len = strlen(md->unit) + 1;
-		strncpy(&vd->vd_name_unit[name_len], md->unit, unit_len);
+		memcpy(&vd->vd_name_unit[name_len], md->unit, unit_len);
 	} else {
 		unit_len = 0;
 	}
@@ -1648,7 +1656,6 @@ void __make_mdesc(ldms_mdesc_t vd, ldms_mdef_t md, off_t *value_off)
 		(*value_off) += __ldms_value_size_get(md->type, md->count);
 	} else
 		vd->vd_data_offset = 0; /* set after all metrics defined */
-
 }
 
 void __init_rec_array(ldms_set_t set, ldms_schema_t schema)
@@ -1723,15 +1730,15 @@ ldms_set_t ldms_set_create(const char *instance_name,
 	}
 
 	hsz = heap_sz;
-	int ssz = compute_set_sizes(instance_name, schema,
+	int total_mm_sz = compute_set_sizes(instance_name, schema,
 				    &set_array_card, &meta_sz,
 				    &array_data_sz, &hsz);
-	if (!ssz)
+	if (!total_mm_sz)
 		return NULL;
 
 	__ldms_schema_finalize(schema);
 
-	meta = mm_alloc(meta_sz + array_data_sz);
+	meta = mm_alloc(total_mm_sz);
 	if (!meta) {
 		errno = ENOMEM;
 		return NULL;


ldms/src/core/ldms.c
```diff
@@ -1627,18 +1627,24 @@ static size_t compute_set_sizes(const char *instance_name, ldms_schema_t schema,
 	return *meta_sz + *array_data_sz;
 }
 
-void __make_mdesc(ldms_mdesc_t vd, ldms_mdef_t md, off_t *value_off)
+void __make_mdesc(ldms_mdesc_t vd, ldms_mdef_t md, off_t *value_off, size_t remaining)
```
morrone, line 1630: Right before this function add:

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-overflow"

ldms/src/core/ldms.c
```diff
@@ -1648,7 +1654,6 @@ void __make_mdesc(ldms_mdesc_t vd, ldms_mdef_t md, off_t *value_off)
 		(*value_off) += __ldms_value_size_get(md->type, md->count);
 	} else
 		vd->vd_data_offset = 0; /* set after all metrics defined */
-
 }
```
morrone, line 1657: Right after the function __make_mdesc add this pragma to restore normal diagnostics:

#pragma GCC diagnostic pop

morrone (commented):
I like it! It helps me understand what is going on a bit more.

The only sad thing is that, at least with my version, gcc isn't convinced. But with the assert()s in place, I'm OK with adding pragma's around __make_mdesc to quiet the warning. I put the suggested pragma lines in in-line comments.

nichamon:
> I like it! It helps me understand what is going on a bit more.
> 
> The only sad thing is that, at least with my version, gcc isn't convinced. But with the assert()s in place, I'm OK with adding pragma's around __make_mdesc to quiet the warning. I put the suggested pragma lines in in-line comments.

@morrone Thanks for trying the patch and letting me know that your gcc still complains about this. Did it give the same warning as in the issue #1054?

morrone:
> > I like it! It helps me understand what is going on a bit more.
> > The only sad thing is that, at least with my version, gcc isn't convinced. But with the assert()s in place, I'm OK with adding pragma's around __make_mdesc to quiet the warning. I put the suggested pragma lines in in-line comments.
> 
> @morrone Thanks for trying the patch and letting me know that your gcc still complains about this. Did it give the same warning as in the issue #1054?

Yes, it was the same warning.

baallan:
@nichamon @morrone I think i followed the logic defining where 'remaining' comes from, so the asserts here are only defending against ourselves in set layout calculations, not defending against bad user or plugin developer input, correct?

That seems ok, but we have other places like 'ldms_metric_modify' of the same file where we assert on a perfectly likely range error from a plugin writer. Buggy plugins taking down the daemon shouldn't be supported with assert-- we need in those cases meaningful error messages and ignoring of the plugin's request.

morrone:
> @nichamon @morrone I think i followed the logic defining where 'remaining' comes from, so the asserts here are only defending against ourselves in set layout calculations, not defending against bad user or plugin developer input, correct?

A plugin coding error and a bad user are very different things in my mind.

A plugin runs in the same address space as everything else in ldmsd. There is really nothing to stop a plugin halting ldmsd or scrambling it memory at any time. So I'm OK with an assert on a buggy buffer calculation bug from a plugin.

If just data from a user can trigger the assert, then I totally agree that would need to be fixed.

baallan:
>  Either we trust compute_set_sizes(..., &meta_sz, ...) or we don't.

Seems like we have reason to assume that we don't, historically. If we do, we compile with -DNDEBUG.

tom95858:
> > Either we trust compute_set_sizes(..., &meta_sz, ...) or we don't.
> 
> Seems like we have reason to assume that we don't, historically. If we do, we compile with -DNDEBUG.

If that is true, which I don't think it is, then we should stop wringing our hands and where it is appropriate trust the input parameters.

nichamon:
I've simplified the patch to focus on what's needed to silence the compiler warnings in issue #1054: replacing `strncpy()` with `memcpy()`. The detailed comment explains why the memory operations are safe based on how `compute_set_sizes()` works.

Regarding the validation checks, I understand the point about assertions providing safety. However, I've decided to remove them as they add complexity without contributing to fixing the compiler warnings. If there are specific concerns about `compute_set_size()` reliability, we could track them in separate issues.

This approach keeps the code cleaner while still addressing issue #1054  and providing documentation about the memory safety guarantees.

PR #1673, opened 2025-04-17
Documentation Fixes

Snell1224:
Added a rule to trigger RTD for all files under the docs/ directory (i.e. images, conf.py, etc.), removed any thing Flux related, updated README.rst and separated out trigger-rtd into its own workflow for PR merges


narategithub (approved):
The `build-test-rtd` and `trigger-rtd` workflows look good to me. :+1:

.github/workflows/trigger-rtd.yml
```diff
@@ -0,0 +1,29 @@
+name: Trigger RTD on Versioned Updates
+
+on:
+  push:
+    branches:
+      - main
+      - 'b[0-9]+.[0-9]+'
```
nichamon, line 7: Why did you add the `b[0-9]+.[0-9]+`?

docs/README.rst
```diff
@@ -34,46 +34,71 @@ For any code or plugin-specific documentation (e.g., manual pages), please use t
...
+To preview documentation changes through a pull request (PR):
+
+1. **Submit a PR** with your changes to the official repository.
+2. Read the Docs will automatically detect the pull request and generate a preview.
+3. On the PR page, a comment from Read the Docs will appear with a link to the **preview documentation build**.
+4. Click the link to review the rendered documentation, including a **side-by-side comparison** ("View docs as diff") between the current version and your proposed changes.
+
+For more details, please visit the `Read the Docs PR Preview Documentation <https://docs.readthedocs.com/platform/stable/pull-requests.html#features>`_.
+
+This is especially useful for maintainers to verify how changes will appear before merging.
```
nichamon, line 76: This section assumes that `ovis-hpc/ldms` is configured to enable the readthedocs preview document feature, right?

Personally, I would like to have this feature enabled because it is more convenient on the contributors than setting up their own readthedocs project to review the documentation changes. However, I am not sure that the LDMS team is aware of this change to the pull request creation/review process.

My point is that when this PR is merged. This section isn't true until the feature is enabled.

What do you think?

.github/workflows/trigger-rtd.yml
Snell1224: I was following the same setup as the existing GitHub workflows for the PR push section. I figured it made sense to include it since future branches and updates would need to build their own version of the docs too. But I’m totally fine removing it if there’s a better way to handle version changes.

docs/README.rst
Snell1224: Yes, that’s correct—and until we’re able to get this working, it should be removed.

I agree that we should have this feature enabled. For the LDMS team, I don’t think they’ve had a chance to contribute to the new RTD setup yet, so they’re not familiar with the PR process yet so it will be fine to remove this for now, and I can add it back in once everything’s properly enabled.

Also, I realized I misunderstood how RTD handles builds. RTD will automatically **build a preview** when a **PR is opened and the webhook is triggered**. However, RTD **won’t update the latest version until the PR is merged**—at that point, the workflow **will trigger the webhook again** and RTD **will build the latest version with the changes from the **merged PR**. 

I did some testing to double-check this behavior and make sure that a trigger for open PRs gives us a build of the **preview only** and that the merge takes care of building the latest version. Now I'm wondering if we should revert back to the original GitHub workflow that triggers on pull requests....

We can discuss further next week. Until then, I'll make the changes to the README.rst and remove the info about PR previews for now.

Snell1224:
@nichamon @tom95858 PR is ready to be merged. Docs workflow has been reverted and branch is up to date with top of tree.

PR #1679, opened 2025-04-21
Test-related workflows update

narategithub:
- ~~Remove CentOS 7 support~~
  - RHEL 7 is still breathing ... keep CentOS 7 build tests.
- Add build test on Alma 8
- Add build test on Alma 9 (currently 9.5)
- Add build test on openSUSE 15 (currently 15.6)
- Replace `build-test-ubuntu` (using `ovishpc/ovis-ubuntu-build` image) with `build-test-ubuntu-22.04` (using `ovishpc/ldms-dev-ubuntu-2204` image) so that the distro is explicitly listed.
- Add `build-test-ubuntu-24.04`.
- Update `4.3.3-compat-test` to use Alma 8 (`ovishpc/ldms-dev-alma-8` image).


baallan:
@narategithub is there a reason for dropping centos7 / an alternate rhel7-alike os to be tested instead?  TOSS4/rhel7 isn't going away as a production environment for at least a year.

emdonat:
> @narategithub is there a reason for dropping centos7 / an alternate rhel7-alike os to be tested instead? TOSS4/rhel7 isn't going away as a production environment for at least a year.

Do we have RHEL7 type systems that will also be upgrading to 4.5.1 and versions moving forward?

emdonat:
Out of the meeting today - capturing the need for SLES build tests along with slingshot.

narategithub:
RHEL 7 is still breathing .. so build-test on centos7 is added back.

build-test on openSUSE 15 is also added.

baallan:
@emdonat the projected course for os upgrades is cts1/cts2 machines (everything snl has except i think eldorado) are quite likely to jump from toss4/rhel7 to toss6/rhel9, with toss5/rhel8 only being used where it is required because toss4 kernels don't support new hardware.  But i would not stake large bets on skipping toss5 since something might conceivably delay the release of toss6.

morrone:
> @narategithub is there a reason for dropping centos7 / an alternate rhel7-alike os to be tested instead? TOSS4/rhel7 isn't going away as a production environment for at least a year.

TOSS4 is RHEL8-based. TOSS5 is RHEL9-based.

baallan:
> TOSS4 is RHEL8-based. TOSS5 is RHEL9-based.

blarrrrrg. I sit corrected re rhel7; thanks Chris. Dump that rhel7 goop, @narategithub . I stand by the point that some systems may skip TOSS5 and go to TOSS6/Rhel10 from toss4. So maybe 6 months from now we can add a rhel10-alike to the testing.

PR #1680, opened 2025-04-22
Refactor ldms_xprt to use ovis_ref and fix rail & xprt references

nichamon:
The pull request is a set of patches:

- Refactor ldms_xprt to use ovis_ref
- Fix ldms_xprt & rail references in the advertisement path on aggregator
- Modify ref.h so that it unlocks ref objects before calling assert() when _REF_TRACK_ is enabled


nichamon:
@tom95858 This PR also fixes the xprt leak I found. The leak only occurs with advertisements.

@narategithub Could you please review this, especially the reference names in `ldms.pyx`? I tested it with `ldmsd_controller`, and it exits gracefully.

PR #1685, opened 2025-04-28
prevent building sos-dependent store plugins without sosdb

baallan:
This adds store_app  darshan  proc_store to the existing list of store plugins that will only be built if sosdb was configured into ldms.


tom95858:
@baallan, I'm sorry, but I'm not understanding something. sosdb is the python plugin for Sos. But it is not used by any of the C plugins.

baallan:
@tom95858 the sosdb c libraries are a dependency of several stores. this change has nothing to do with python.

PR #1689, opened 2025-04-29
Get rid of the `env var=$(cmd)` command value expansion in config file

narategithub:


nichamon:
@narategithub Could you please give background on this change? It seems that the patch removes a configuration capability, which allows `ldmsd` execute a bash command in `$(cmd)` and cache the exit value in environment variable `var`. I support removing this capability because it could lead to security risk as it makes `ldmsd` execute bash commands in LDMS configuration file. If I remember correctly, this was added to support specific usages for convenient setup and deployment a few years back.

I want to make sure that this has been communicated or will be communicated with the community to prevent surprise breakage in LDMS deployment and configuration.

tom95858:
@morrone, @baallan, @jennfshr please confirm that no one is using this feature.

baallan:
I can confirm that the env configuration command is not used in the genders-scripted ldmsd service managment for toss systems.
I cannot provide any insight re older cray systems where it might have been used.
@valleydlr may know something.

jstile-lbl:
I rely on this a bit

`
env HOSTNAME=$(hostname)                                                        
env COMPONENT_ID=$(python3 -c 'import ipaddress; print(int(ipaddress.IPv4Address("127.0.0.1")))')
env PROCTYPE=$(cat /sys/devices/system/cpu/modalias | cut -d ':' -f 3 | tr ',' '_')
env LDMS_PROCNETDEV_IFACES=$(ip -o link show | awk -F':' '{print $2}' |paste -s -d,)
 `

baallan:
@jstile-lbl Is there a systemd service involved at a higher level? if so, there is a way to migrate all of that you listed into the service file.  Separately, for the long term I recommend replacing the use of procnetdev plugin with one that autodetects the interfaces present and active; this would lead to a change in the database schema for network data. We run the 'procnet' plugin on my systems.

jstile-lbl:
@baallan 

I am actually using the procnetdev2 sampler, which does not use the variable LDMS_PROCNETDEV_IFACES, so I can remove that one.   Do you still recommend migrating to 'procnet'?   I'll need to find an example.

Replacing the others sounds manageable during system boot.

baallan:
@jstile-lbl If you're on procnetdev2, switching to procnet would be a matter of preference. Both produce a database record per device instead of a wide schema dependent on the devices found which is harder to manage. So most likely no need to disturb your current practice.

notes on the differences:
 - procnet config would look like:  "config name=procnet producer=${host} instance=${host}/procnet"
    with additional options per site practice (common to most plugins) for perm, component id, job id. It has an optional exclude_ports argument described in the man page.
 - procnet makes a metric set per interface instead of a single metric set with multiple records.
 - procnet does not update the individual metric set if there is no traffic on the polled interface during the interval; this leads to a sparse time-series in storage for minimally used interfaces.
 - procnet allows an exclusion list (names of interfaces to be ignored; some sites have nodes with lots of extra unused ports).

PR #1692, opened 2025-04-30
Fix parsing brittleness in lustre_client for b4.4

baallan:
This also adds a test mode option for reading fake sample data without root access and two options to include new/supplementary data in the schema.

This patch cures lustre_client data loss for lustre 2.15 client metrics that do not end in .sum but do get collected from lustre metric rows that do contain sum and sumsq columns in the raw.

### DIFF
--- a/ldms/src/sampler/lustre_client/lustre_client_general.c
+++ b/ldms/src/sampler/lustre_client/lustre_client_general.c
@@ -59,6 +59,23 @@ static char *llite_stats_uint64_t_entries[] = {
         NULL
 };
 
+
+/* the following get added to the schema if extra215 config option is present. */
+static const char *extra215_llite_stats_uint64_t_entries[] = {
+        "read",
+        "write",
+        "opencount",
+        "openclosetime",
+        NULL
+};
+
+/* the following exist at least as far back as 2.15 lustre clients */
+static const char *extratimes_llite_stats_uint64_t_entries[] = {
+	"start_time", /* this truncates the data to seconds; it's ok. */
+	"elapsed_time", /* this truncates the data to seconds; it's ok. */
+        NULL
+};
+
 int llite_general_schema_is_initialized()
 {
         if (llite_general_schema != NULL)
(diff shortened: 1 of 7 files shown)


baallan:
@jennfshr @bschwal this should fix the lustre data loss anywhere running 4.4.x and lustre 2.15

baallan:
Included is an update to the lustre_client static test and the man page. The test_path option is potentially useful when lustre yet again changes the root location of their performance data.

tom95858:
@morrone, please let me know if/when this is OK to merge

ldms/src/sampler/lustre_client/lustre_client_general.c
```diff
@@ -212,21 +252,31 @@ static int llite_stats_sample(const char *stats_path,
...
                         if (index == -1) {
-                                log_fn(LDMSD_LWARNING, SAMP ": llite stats metric not found: %s\n",
-                                       str1);
-                        } else {
+				/* non-sum metric, possibly */
+				str1[base_name_len] = '\0';
+				index = ldms_metric_by_name(general_metric_set, str1);
+				if (index != -1) {
+					ldms_metric_set_u64(general_metric_set, index, val1);
+				}/* else {
```
morrone, line 273: I'm not a fan of commented-out dead code.

ldms/src/sampler/lustre_client/lustre_client_general.c
```diff
@@ -212,21 +252,31 @@ static int llite_stats_sample(const char *stats_path,
                             str1, &val1, &val2);
                 if (rc == 2) {
```
morrone, line 253: I think it would be simpler to understand if we start, in sscanf, to catch the units inside the square brackets. Then we can just make new condition that is basically:

`if units == "usecs" and count == 4`

(4 because catchings the units string will add another field. The other numbers will need to be bumped too.)


Granted, the numbers change once we start parsing the units string.

morrone (approved):
I am good with landing it the way it is.

morrone:
FYI, I landed this in [v4.4.5_5toss](https://github.com/LLNL/ldms/releases/tag/v4.4.5_5toss).

emdonat:
@morrone Do you have changes to this for 4.5.1?  Is this something you wanted to add to to port forward?

PR #1698, opened 2025-05-05
Plugin-API Refactor

morrone:
This request implements what is likely very close to the final refactor. This version adds the plugin-handle that @morrone requested as well as some additional features, including:

The ovis_log is automatically registered when the plugin is loaded and stuffed into the plugin_handle. This handle can be retrieved by the plugin with ldmsd_plugin_log_get.
The multi-config capable plugins use the plugin-api to retrieve their configuration name, context, etc...
The legacy plugins are not affected by this change other than they receive the plugin-handle instead of the cfgobj
This change has been tested with both legacy-samplers and multi-config-samplers.

(Note: this is a reworked history of PR#1671 at commit 62a68139b. The end result is identical. The contents of the commits have been rearranged to fix up some rebasing errors.)


tom95858:
@morrone, Ok are we ready to merge?

morrone:
> @morrone, Ok are we ready to merge?

Ready.

PR #1699, opened 2025-05-05
Always dlclose, and more API changes (remove get_plugin() and "name" from the API)

morrone:
This PR stack begins with a refactoring patch "[Compine generic plugin tacking, introduce unload_plugin()](https://github.com/ovis-hpc/ldms/commit/ef16aa95f54aa59af2fa9ac23d15fea484fea106)". It combines libpath and the generic base api pointer together into one struct to simplify what we need to remember in the cfgobj structs, and what we have to pass around. It also helps reduce duplication between the sampler and store code paths. It also give us a convenient, cleaner, generic struct (struct ldmsd_plugin_generic) to add more variables in the subsequent patches. Finally, it give us a cleanup function "unload_plugin() to match up with "unload_plugin()".

Next [dlclose() each handle we receive from dlopen()](https://github.com/ovis-hpc/ldms/commit/f3fdacd261a9ced76451f122b56467094a35ab55) employs the new struct lmsd_plugin_generic to remember the handle returned by dlopen(), and then calls dlclose() in unload_plugin().

Then [Replace get_plugin() with ldmsd_plugin_interface symbol](https://github.com/ovis-hpc/ldms/commit/eec9e3ecf8b09afff39e4744703382f8e7b6d355) makes a bolder API change by eliminating the "get_plugin()" call from the API entirely. More detail on the rationale in the commit message.

Finally, [Remove "name" from the plugin API structs](https://github.com/ovis-hpc/ldms/commit/0abd8d2c063d97c4242b522115602e63a5aa33b7) changes the API by removing the "name" field from the plugin api struct. "name" was, in practice, redundant. It gave opened up an opportunity for human error to have inconsistent plugin names (between the plugin naming of the .so file, and what the plugin internally reports), which would have confusing consequences for users. Removing the redundant "name" field removes that possibility for error, and makes one less thing that plugin authors need to think about.

Update (May 7, 2025):

A commit has been added to remove term() from the plugin API.

Update (May16, 2025):

Added minor cleanup commit named "Improve constructor call points".

### DIFF
--- a/ldms/src/ldmsd/ldmsd.h
+++ b/ldms/src/ldmsd/ldmsd.h
@@ -1009,6 +1009,7 @@ struct ldmsd_store {
 struct ldmsd_plugin_generic {
         struct ldmsd_plugin *api;
         char *libpath;
+        void *dl_handle; /* handle that was returned by dlopen() */
 };
 
 typedef struct ldmsd_cfgobj_sampler *ldmsd_cfgobj_sampler_t;
--- a/ldms/src/ldmsd/ldmsd.h
+++ b/ldms/src/ldmsd/ldmsd.h
@@ -1002,7 +1002,7 @@ struct ldmsd_store {
 
 /**
  * struct ldmsd_plugin_generic is used to track generic information common to
- * all plugins that implement the ldms get_plugin() function.
+ * all plugins that implement the ldmsd_plugin_interface symbol.
  * NOTE: This is an ldmsd internals data structure.
  *       Not to be used from inside plugins.
  */
@@ -1163,8 +1163,6 @@ ldmsd_store_close(ldmsd_cfgobj_store_t store, ldmsd_store_handle_t sh)
 	if (store->api->close)
 		store->api->close(store, sh);
 }
-
-typedef struct ldmsd_plugin *(*ldmsd_plugin_get_f)();
 
 /* ldmsctl command callback function definition */
 typedef int (*ldmsctl_cmd_fn_t)(char *, struct attr_value_list*, struct attr_value_list *);
--- a/ldms/src/ldmsd/ldmsd.h
+++ b/ldms/src/ldmsd/ldmsd.h
@@ -962,7 +962,6 @@ char *process_yaml_config_file(const char *path, const char *dname);
 #define LDMSD_CFG_FILE_XPRT_MAX_REC 8192
 /* This struct is owned by the plugin. ldmsd should never modify the contents. */
 typedef struct ldmsd_plugin {
-	char name[LDMSD_MAX_PLUGIN_NAME_LEN]; /* plugin name (e.g. meminfo) */
 	enum ldmsd_plugin_type {
 		LDMSD_PLUGIN_OTHER = 0,
 		LDMSD_PLUGIN_SAMPLER,
@@ -1008,6 +1007,7 @@ struct ldmsd_store {
  */
 struct ldmsd_plugin_generic {
         struct ldmsd_plugin *api;
+        char *name;
         char *libpath;
         void *dl_handle; /* handle that was returned by dlopen() */
 };
--- a/ldms/src/ldmsd/ldmsd.h
+++ b/ldms/src/ldmsd/ldmsd.h
@@ -999,6 +999,18 @@ struct ldmsd_store {
 		      ldmsd_row_list_t row_list, int row_count);
 };
 
+
+/**
+ * struct ldmsd_plugin_generic is used to track generic information common to
+ * all plugins that implement the ldms get_plugin() function.
+ * NOTE: This is an ldmsd internals data structure.
+ *       Not to be used from inside plugins.
+ */
+struct ldmsd_plugin_generic {
+        struct ldmsd_plugin *api;
+        char *libpath;
+};
+
 typedef struct ldmsd_cfgobj_sampler *ldmsd_cfgobj_sampler_t;
 typedef struct ldmsd_sampler {
 	struct ldmsd_plugin base;
@@ -1007,6 +1019,7 @@ typedef struct ldmsd_sampler {
 
 struct ldmsd_cfgobj_store {
 	struct ldmsd_cfgobj cfg;
+	struct ldmsd_plugin_generic *plugin;
 	struct ldmsd_store *api;
 
 	/* Set to 1 if the plugin has been configured */
@@ -1015,7 +1028,6 @@ struct ldmsd_cfgobj_store {
 	/* List of strgp that are using this store */
 	LIST_HEAD(, ldmsd_strgp) strgp_list;
 
-	char *libpath;
 	/* Private context pointer, managed by plugin */
 	void *context;
 	/* ovis_log handle to use when logging plugin messages */
@@ -1030,12 +1042,12 @@ typedef struct ldmsd_sampler_set {
 
 struct ldmsd_cfgobj_sampler {
 	struct ldmsd_cfgobj cfg;
+	struct ldmsd_plugin_generic *plugin;
 	struct ldmsd_sampler *api;
 
 	/* Set to 1 if the plugin has been configured */
 	int configured;
 
-	char *libpath;
 	/* Private context pointer, managed by plugin */
 	void *context;
 	/* ovis_log handle to use when logging plugin messages */
@@ -1061,8 +1073,7 @@ struct ldmsd_cfgobj_sampler {
 #define LDMSD_JOBID "job_id"
 
 ldmsd_cfgobj_sampler_t ldmsd_sampler_add(const char *name,
-					struct ldmsd_sampler *api,
-					const char *libpath,
+					struct ldmsd_plugin_generic *plugin,
 					ldmsd_cfgobj_del_fn_t __del,
 					uid_t uid, gid_t gid, int perm);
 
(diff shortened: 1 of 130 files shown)


ldms/src/ldmsd/ldmsd.h
```diff
@@ -1028,12 +1041,12 @@ typedef struct ldmsd_sampler_set {
...
 	struct ldmsd_cfgobj cfg;
+	struct ldmsd_plugin_generic *plugin;
 	struct ldmsd_sampler *api;
 
 	/* Set to 1 if the plugin has been configured */
 	int configured;
 
-	char *libpath;
 	/* Private context pointer, managed by plugin */
 	void *context;
```
tom95858, line 1051: Should the context pointer be moved to the ldmsd_plugin_generic structure too?

ldms/src/ldmsd/ldmsd.h
```diff
@@ -997,6 +996,20 @@ struct ldmsd_store {
...
 };
 
+
+/**
+ * struct ldmsd_plugin_generic is used to track generic information common to
+ * all plugins that implement the ldmsd_plugin_interface symbol.
+ * NOTE: This is an ldmsd internals data structure.
+ *       Not to be used from inside plugins.
+ */
+struct ldmsd_plugin_generic {
```
tom95858, line 1006: I had a similar thought but was going to make a ldmsd_cfgobj_generic and put everything that you have in here and also context, ldmsd_cfgobj and the new configured 'flag'. So then the cfgobj_sampler/store is almost all common

morrone: @tom95858 I've had thoughts like that myself. If by "cfgobj_generic" you are refering to the base "struct ldmsd_cfgobj", that is used as the base for may things that are not plugins: ldmsd_prdcr, ldmsd_prdcr_listen, ldmsd_updtr, ldmsd_stgrp, etc. So putting generic plugin state in there wouldn't really work well.

If you are talking about adding a new middle layer of cfgobj between the normal base of ldmsd_cfgobj and ldmsd_cfgobj_sampler/ldms_cfgobj_store, yes that could work. Something like this:

```
   struct ldmsd_cfgobj_generic_plugin {
      struct ldmsd_cfgobj obj;
      /* real generic plugin stuff */
  }
  struct ldmsd_cfgobj_sampler {
      struct ldmsd_cfgobj_generic_plugin;
      /* sampler stuff */  
}
```

But I think we can take this refactoring in steps. Now that we have a clean abstraction in the plugin API, we can refactor and reimplement this code all we like without breaking plugin compatibility.

(P.S. moved this comment from another ticket, since it seemed like a better conversation for here.)

morrone: Actually, maybe it won't be that different. I'll take a stab at changing ldmsd_plugin_generic to a cfgobj.

morrone: I changed my mind. There are more little details that need to change around startup. We maybe need to move the call to ldmsd_cfgobj_new_with_auth() earlier into load_plugin(), but then it would just need to allocate max(sizeof(ldmsd_cfgobj_stampler_t, ldmsd_cfgobj_store_t)...a bit more refactoring than I had hoped.

I think we should leave that work to a follow-on after this PR.

ldms/src/ldmsd/ldmsd.h
morrone: Actually, thinking about this again, I wonder if it wouldn't be better to leave this where it is.

Are we always going to keep the code organized the way it is, with plugins redundantly opened for each instance? We really only need to dlopen once per plugin, and instances can share the same information. It would be easier to stop redundantly opening plugins in the future if we keep per-instance stuff in the per-instance cfgobj (e.i. cfgobj_sampler, cfgob_store), and let the generic plugin struct remain specific to the generic plugin information.

morrone:
@tom95858 I am pretty confident that I could also remove term() in this PR as well, if you are amenable to that. I am proposing a straight forward bug-for-bug compatible conversion from term() to destructor() in all the plugins. Granted there, _are_ existing bug in term() implmentations, but I things we can leave the bug fixing as follow on work.  My get_plugin() removal introduced a lot of constructor() usage, after all.

ldms/src/ldmsd/ldmsd.h
baallan: > and the new configured 'flag'.

@tom95858 can you elaborate on what "the new configured 'flag'" is?

Are we changing to an api where there's exactly one config call per instance between load and start, and if you want a big multiline configuration you must get it by putting it in a file which is passed by name to the plugin? or perhaps by putting it in an attribute whose right side is a json object string without carriage returns? I'm thinking I could live with that.

morrone: @baallan I know this one. :) Tom added a new "unint64_t flags" to the "struct ldmsd_plugin", which is the base API plugin struct. The first (currently only) flag in that field is the first bit:

```
#define LDMSD_PLUGIN_MULTI_INSTANCE 1 /* Set in flags to indicate MI capability */
```

This allows each plugin to declare whether or not they are multi-instance capable. Many of the plugins will likely not be upgraded to be multi-instance by the v4.5.1 release, and even if we convert most, there may always be some that are not multi-instance capable for some external reason. If LDMSD_PLUGIN_MULTI_INSTANCE is _not_ set, ldmsd will throw an error when a second instance of that plugin is loaded.

Technically, at the moment when LDMSD_PLUGIN_MULTI_INSTANCE is not set, ldmsd will throw an error if "name=" and plugin=" are not the same for that plugin. That is sort of a quick-and-dirty substitute for checking if an instance of the plugin is already loaded, and I would like to change that. Granted, it is at least backwards compatible. But I don't see any reason that we couldn't allow users to give even a single plugin instance a name of their own creation.

morrone: Wait, no I might be wrong. Thats what I get for trying to be helpful. :) He might have been talking generically about a new flags in the store and sampler cfgobj to note when they are "in use", or in this case he referred to them as "configured".

baallan: > Wait, no I might be wrong. Thats what I get for trying to be helpful. :) He might have been talking generically about a new flags in the store and sampler cfgobj to note when they are "in use", or in this case he referred to them as "configured".

You're making me feel better about having asked... :)   hmmmm

tom95858:
@morrone, @baallan 

1. The config name vs. plugin name check is done at "load" time. If the plugin is not MI capable, we force the plugin name and the configuration name to be equal
2. The MI config time check is different. It is a check that is enforced by the infrastructure.

Some plugins supported a kind of multi-config internally, specifically the test_sampler and I think CSV. I think I need to investigate if these checks broke some of our plugins (e.g. CSV, PAPI)

morrone:
The force-push to fd02535 is just a rebase on main.

ldms/src/ldmsd/ldmsd.h
morrone: I think we agree that ldmsd_plugin_generic works for now, and we can always refactor further in a follow-on PR.

morrone:
@tom95858 What do we need to get this landed?

morrone:
This force-push is just a simple rebase on main to make it conflict-free again.

tom95858:
@morrone, I am going to test this as well as I can this weekend.

tom95858:
@morrone, there are a few samplers that don't build (e.g. kokkos). I'm trying to test this before I merge it. I'm not sure how short we are on the changes we might be missing. It could just be the one sampler or it could be a lot. Can you take a look?

morrone:
> @morrone, there are a few samplers that don't build (e.g. kokkos). I'm trying to test this before I merge it. I'm not sure how short we are on the changes we might be missing. It could just be the one sampler or it could be a lot. Can you take a look?

I don't have kokkos to build against. Show me the error message and I'll get it fixed.

morrone:
@tom95858 I'm guessing it was missing the ldmsd_plug_api.h include. But if that wasn't it, just share the error message and I'll fix it up.

morrone:
I see some more that need the ldmsd_plug_api.h header. I'm adding them now.

morrone:
@tom95858 OK, I added bunch of missing includes of ldmsd_plug_api.h. Although that may not have really been the issue, because I see you snuck "#include ldmsd_plug_api.h" into ldmsd.h. I'll probably want to remove that in a follow up PR.

If that wasn't it its probably something obvious if I can see the error message.

morrone:
@tom95858 OK, give the latest force-pushed branch a try.

I couldn't get sos to build for me, and gave up on that. But I added "--enable-kokkos" to one of the automated build tests that is able to build sos. I used that to get the error message. I've now fixed that in all the stores where it was an issue.

morrone:
I fixed some bugs in the rdc_sampler build too. Thats all that I can find at the moment. Let me know if you see any other build errors.

morrone:
I just added a "Remove redundant plugin log creation" patch to this stack. I would have posted it separately, but it is going to wind up with conflicting context if I do that. If you don't want that commit in this PR I can always back it out and submit a new PR for that when this PR lands.

tom95858:
@morrone, I think it's fine. Let's just get it merged...

PR #1700, opened 2025-05-05
Minor api change cleanup

morrone:
A collection of minor clean up changes after the landing of PR#1698. See commit messages in the stack for details.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -153,15 +153,13 @@ static struct ldmsd_plugin_generic *load_plugin(const char *plugin_name)
 		goto err_1;
 	}
 
-	ldmsd_plugin_get_f pget;
-	pget = dlsym(plugin->dl_handle, "get_plugin");
-        if (!pget) {
+        plugin->api = dlsym(plugin->dl_handle, "ldmsd_plugin_interface");
+        if (!plugin->api) {
 		ovis_log(config_log, OVIS_LWARNING,
 			"The library, '%s',  is missing the ldmsd_plugin_interface "
 			"symbol.", plugin_name);
                 goto err_2;
 	}
-	plugin->api = pget();
 
 	/* Verify the following API contract:
 	 * - If flags & MULTI_INSTANCE is !0, then
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -164,15 +164,13 @@ static struct ldmsd_plugin_generic *load_plugin(const char *plugin_name)
 	/* Verify the following API contract:
 	 * - If flags & MULTI_INSTANCE is !0, then
 	 *   -- both constructor() and destructor() must be defined
-	 *   -- term() must not be defined
 	 */
 	if (plugin->api->flags && LDMSD_PLUGIN_MULTI_INSTANCE) {
-		if (!plugin->api->constructor || !plugin->api->destructor
-                    || plugin->api->term) {
+		if (!plugin->api->constructor || !plugin->api->destructor) {
 			ovis_log(config_log, OVIS_LERROR,
 				 "The library, '%s', has the multi-instance flag, "
 				 "but is missing either the constructor or "
-				 "destructor functions, or has term defined.\n",
+				 "destructor functions.\n",
 				 plugin_name);
 			goto err_2;
 		}
@@ -342,8 +340,6 @@ void ldmsd_sampler___del(ldmsd_cfgobj_t obj)
 	ldmsd_cfgobj_sampler_t samp = (void*)obj;
 	if (samp->plugin->api->destructor)
 		samp->plugin->api->destructor(obj);
-	else if (samp->plugin->api->term)
-		samp->plugin->api->term(obj);
         unload_plugin(samp->plugin);
 	ldmsd_cfgobj___del(obj);
 }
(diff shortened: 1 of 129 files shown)


tom95858:
@morrone, this looks good on its face. Let me test it tomorrow and we'll get it merged.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -209,10 +209,7 @@ ldmsd_sampler_add(const char *cfg_name,
 	sampler->api = api;
 	sampler->libpath = strdup(libpath);
 	sampler->thread_id = -1; /* stopped */
-	if (sampler->api->base.constructor != NULL
-		&& sampler->api->base.destructor != NULL)
-	{
-		/* This plugin is multi-instance capable */
+	if (sampler->api->base.constructor != NULL) {
```
tom95858, line 212: I think with the cfgobj_generic and/or plugin_generic you are proposing, these code paths can be combined, specifically config can be combined for sampler and store.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -328,6 +328,10 @@ void ldmsd_sampler___del(ldmsd_cfgobj_t obj)
 void ldmsd_store___del(ldmsd_cfgobj_t obj)
 {
 	ldmsd_cfgobj_store_t store = (void*)obj;
+	if (store->api->base.destructor)
```
baallan, line 331: @morrone Is it always the case that the implementer's choice of destructor and term would be mutually exclusive? Seems like there may be reasonable scenarios where the multiple instances might share some state that persists outside the instance lifecycle (i.e. until term).  Or is term only defined and used for plugins advertising as singletons and wanting the framework to enforce that?    Did we end up with a header documenting all this carefully, or no?

morrone: term() is now deprecated, and destructor() should be used instead. I'm hoping to actually delete term() from the API before the release, and offered to do the code fore that. Some feel like it is too complicated and intially turned me down. 

My guess is that they are concended that term() is buggy in many plugins. And I would agree. But I think that we can do a straight-forward bug-compatible refactoring from term() to destructor(), and leave further cleanup and refactoring to follow on work.

morrone: > Seems like there may be reasonable scenarios where the multiple instances might share some state that persists outside the instance lifecycle (i.e. until term). 

OH! I see what you mean. As of the new "cfgobj" code and the multi-configuration implementation of plugins, there is not really any such thing as "outside the instance lifecycle", at least not as it pertains to term. (And I did not implement this part, I'm just relaying my understanding from working the code recently.)

Every time that a user issues a "load" command, even, for the same plugin, the plugin .so is dlopen()ed, and cfgobj is malloced to track that instance.

In other words, the code is re-opening the same plugin library (calling dlopen() everytime) for each "load" command.

Likewise, when the user issues the matching "term" command for an _instance_ of a plugin, the destructor() or term() function of that _instance_ will be called. The term command doesn't apply to _all_ instances.

It is certainly redundant to dlopen() the same library multiple times for different instances. Maybe some day we will refactor the code to change that. Maybe not. :)

In the API, we do not currently offer any API calls that are triggered only once globally per plugin library. The way it is designed, all calls are now per-instance.

baallan: I would venture a guess that most plugins which actually bothered to implement a non-empty term are probably correct modulo some memory leaks. So with that in mind, renaming term to destructor and carrying any potentially buggy code forward seems to be fine.   Re a 'done with plugin/all instances gone' callback, i guess that's what the dtor hook that should be called as part of a last dlclose/unlink is used for. It's on the developer, not on us. We could document that they should not be managing ldms-defined structures in their dtor as ldms libraries might get unloaded before them during shutdown if they have no outstanding dlopens.

baallan: Re term being deprecated, i agree then it should just be gone. There's such a huge pile of semantic changes that it's fairly minor in comparison to tell anyone with a 3rd party plugin "here's the list of function signature changes (including renames)".

tom95858: Just to be sure we're all talking about the same thing. Consider that a storage plugin may have multiple configuration instances *and* have multiple references to it. So think of store_csv has multiple instances with different configurations. Each of those plugin configurations is referred to by multiple storage policies.

morrone: > From my current understanding, term is different than destructor. Destructor destroys the plugin configuration instance, term unloads the plugin.

Not really, no.. destructor() is called at unload time, when the user runs the "term" command. It would be an add destructor() that didn't tear everything down.

morrone: > Just to be sure we're all talking about the same thing. Consider that a storage plugin may have multiple configuration instances _and_ have multiple references to it. So think of store_csv has multiple instances with different configurations. Each of those plugin configurations is referred to by multiple storage policies.

I'm not sure why you bring that up here. "references" don't really seem apropos to the destructor/term discussion. Did I miss something?  By "references" in this case, since you talk about a store, I believe you are referring to strgps? Those do of course have their own cfgobjs and depend on the store cfgobjs. A store's destructor/term would of course only be called once those dependencies are gone.

morrone:
OK, based on this discussion, I am just going to go head and add a commit to this patch stack that removes term() as well.

morrone:
> OK, based on this discussion, I am just going to go head and add a commit to this patch stack that removes term() as well.

Sorry, I meant I'll add that to the #1699 stack. It is more appropriate for that (touches many many plugins).

morrone:
I think #1699 should land before this PR. Then I can rebase and update this PR onto that one.

morrone:
The removal of term() in PR #1699 resulted in most of the changes in this PR disappearing. The one minor commit that remained here ("Improve constructor call points") has been added to PR #1699, so I am closing this PR #1700.

PR #1701, opened 2025-05-06
Add some reserved space in struct ldmsd_plugin

morrone:
Adding some reserved space to struct ldmsd_plugin will allow us to add furture calls without breaking backwards compatibility.


ldms/src/ldmsd/ldmsd.h
```diff
@@ -981,6 +981,8 @@ typedef struct ldmsd_plugin {
 
 	/* Deprecated, use destructor() */
 	void (*term)(ldmsd_plug_handle_t handle);
+
+	void *reserved[8]; /* reserved for future use */
```
tom95858, line 985: I think this is a good idea.

Issue #1694, opened 2025-05-02
zap_fabric.c won't compile against libfabric 2.1.0


morrone:
On the main branch as of commit 82f2ab13533f5fef2fa381c4bdabb5ec6685108f, the zap fabric subsystem will not compile.

```
  CC       libzap_fabric_la-zap_fabric.lo
zap_fabric.c: In function ‘__map_addr’:
zap_fabric.c:240:7: error: variable ‘mr_mode’ has initializer but incomplete type
  240 |  enum fi_mr_mode mr_mode = ep->fi->domain_attr->mr_mode;
      |       ^~~~~~~~~~
zap_fabric.c:240:18: error: storage size of ‘mr_mode’ isn’t known
  240 |  enum fi_mr_mode mr_mode = ep->fi->domain_attr->mr_mode;
      |                  ^~~~~~~
zap_fabric.c:241:13: warning: 'FI_MR_SCALABLE' is deprecated
  241 |  if (!mr_mode || (mr_mode & FI_MR_SCALABLE)) /* SCALABLE MR mode */
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

It appears that the "enum fi_mr_mode" no longer exists.

tom95858:
@narategithub can you please take a look at this.

morrone:
@tom95858 @narategithub, I think that the "enum fi_mr_mode" symbol really is _gone_ in libfabric 2.1.0. I checked out the code myself an grepped for it, and I didn't find it (well, some man pages mention it, but I suspect they forgot to update the man pages).

I don't think changing build paths will help with this. I think we have an API change to deal with.

morrone:
@emdonat I don't know if this should hold up 4.5.1 or not. The summary for this one is that the libfabric API changed, and ldms no longer has buildable libfabric zap support.

morrone:
I'm adding the 4.5.1 milestone to this one on the assumption that we want 4.5.1 to have working libfabric support.

jennfshr:
I'm now encountering this with our latest move to toss4.8-10, where libfabric bumped to 2.1.  I'm looking for a solution presently, since I *must* rebuild due to libosm incompatibilities with our package we constructed from various bugfixes/patches off v4.4.5.  Just FYI, me too.

tom95858:
[...]

PR #1702, opened 2025-05-06
Fix zap fabric build

narategithub:
This pull request addressed the issue of zap fabric failed to build when libfabric was installed in a location that is not in the library lookup path. The build failed due to the following reasons:

- `@LIBFABRIC_LDFLAGS@` was not added to zap fabric LIBADD.
- `@LIBFABRIC_INCDIR_FLAGG@` was not set in the `OPTION_WITH_CHECK` macro.
- `@LIBFABRIC_CFLAGS@` was also incorrect (missing `-I` in front of the include path).

NOTE: If `libfabric` was installed in the default lookup locations (e.g. /usr or /usr/local), we will not see the issues.


tom95858:
This may resolve issue #1694

morrone:
> This may resolve issue #1694

@tom95858  That doesn't seem likely. libfabric _is_ installed in the default lookup locations on our systems. And our issue is one of libfabric's API changing. ldmsd uses symbols that no longer exist in the newer libfabric code.

PR #1703, opened 2025-05-06
Convert ref_get/ref_put to use atomic compare-and-swap

morrone:
Convert ref_get/ref_put to use atomic compare-and-swap

Eliminate the race condition between ref_put() hitting zero and
freeing memory, while another thread ref_get()'s the counter
back to one.
    
Using atomic compare-and-swap (__atomic_compare_exchange() in gcc) allows
us to atomically prevent ref_get() from succeeding after the counter has
hit zero.
    
This commit only makes it _possible_ to avoid the race condition. Any
consumer of ref_get() will need to check the return code so that it
can avoid using the memory that is in the process of being freed.

### DIFF
--- a/lib/src/ovis_ref/ref.h
+++ b/lib/src/ovis_ref/ref.h
@@ -1,5 +1,6 @@
 #ifndef _REF_H_
 #define _REF_H_
+#include <stdbool.h>
 #include <sys/queue.h>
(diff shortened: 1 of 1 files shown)


morrone:
This one is still a work in progress.

PR #1705, opened 2025-05-07
Expand the name column in the log_status output

nichamon:


tom95858:
@nichamon I looked into this and I have a couple of observations that we might want to address with this pull request. 

1. There is both a 'stream' and an 'ldms.stream' log tag. It is not clear what those two are. But we should probably have something called ldms.message and ldmsd.stream respectively. One for the LDMSD Stream API and one for the LDMS Message API
2. We should probably put "ldmsd." in front of the tags that have nothing, i.e. config --> ldmsd.config, producer --> ldmsd.producer, sampler --> ldmsd.sampler
3. It's not clear what xprt.ldms is
4. xprt.zap should probably be ldms.zap and xprt.zap.sock should probably be ldms.zap.sock

What do you think?

tom95858:
@nichamon, I am wrong. This is just a python change. Let's make the changes I suggested in a separate pull request.

PR #1709, opened 2025-05-08
Allow empty string value for strgp container=

morrone:
We intentionally do not check if the value is an emptry string. With stores, like store_avro_kafka, the server configuration is more complicated than a single field, and happens in the store, not in the strgp container= line. This function needs to allow returning an emptry string value to support that behavior.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request_util.c
+++ b/ldms/src/ldmsd/ldmsd_request_util.c
@@ -1164,8 +1164,11 @@ char *ldmsd_req_attr_str_value_get_by_id(ldmsd_req_ctxt_t req, uint32_t attr_id)
 	ldmsd_req_attr_t attr = ldmsd_req_attr_get_by_id(req->req_buf, attr_id);
 	if (!attr)
 		return NULL;
-	if (attr->attr_value[0] == '\0')
-		return NULL;
+        /* We intentionally do not check if the value is an emptry string.
+           With stores, like store_avro_kafka, the server configuration is more
+           complicated than a single field, and happens in the store, not in
+           the strgp container= line. This function needs to allow returning
+           an emptry string value to support that behavior. */
 	return str_repl_env_vars((char *)attr->attr_value);
 }
 


nichamon:
@tom95858 I just looked into this path. and you added the check of `attr_value[0]='\0'` with an API change commit. I'm wondering if you have added this due to specific crashes during your tests?

In the avro_kafka store man page, there is an example of:
`strgp_add name=... container=   plugin=avro_kafka`

If @tom95858 added the check due to a crash, we probably should find a solution that fixes the crash, and users can specify that they want to use the default hostname:port for avro_kafka store.

By the way, why do we ask users to add `container= ` without the value? Why don't we let them omit the `container` attribute completely?

morrone:
> By the way, why do we ask users to add `container= ` without the value? Why don't we let them omit the `container` attribute completely?

That would definitely be better for us.

tom95858:
@morrone, I'm OK with removing the container= I just want to ensure that our stores don't crash.

nick-enoent:
> @morrone, I'm OK with removing the container= I just want to ensure that our stores don't crash.

Would this entail adding a check for the storage plugin to see if the given plugin required the container parameter?

morrone:
> > @morrone, I'm OK with removing the container= I just want to ensure that our stores don't crash.
> 
> Would this entail adding a check for the storage plugin to see if the given plugin required the container parameter?

I don't know the details of how it would happen. Perhaps each store plugin, when appropriate, can be updated to throw an error when a strgp tries to attach without passing in "container".

nichamon:
**Recommendation: Option 1.2** (Keep `container` mandatory with non-empty values). I'm also okay with **Option 1.1** since no issues have been reported with v4.4 and it would allow us to merge this PR without further modifications.

I believe we have these options:

1. Keep `container` mandatory for `strgp_add`
    1.1. Allow `<attr_name>=` in configuration files (the same as in v4.4)
      - Pros:
         - Minimal code changes
         - Backward compatible with existing configs
         - Can merge PR as-is
      - Cons:
         - Less explicit; @tom95858 noted no crashes but the check was added for safety

    1.2. Make it mandatory that <attr_value> can't be an empty string
      - Pros:
         - Explicit configuration (e.g., `container=default` vs `container=`  `container=default` is more obvious that ldmsd or the plugin will use the default behavior.)
         - Stricter syntax prevents potential empty-string issues
      - Cons:
         - Requires modifying existing configuration files
         - Update in the arvo_kafka man page

2. Make `container` attribute optional for `strgp_add`
   - Pros:
     - `container` attribute can be omitted completely when it isn't needed
   - Cons:
     - Requires removing `container=` from existing configs
     - Requires additional NULL checks in storage plugins

I prefer 1.2 for its explicitness, but 1.1 is a pragmatic choice that allows immediate progress. Option 2 requires too many code changes across plugins.

morrone:
I have a draft PR #1738 that tries to implement "2. Make container attribute optional for strgp_Add".

morrone:
See PR #1738.

PR #1713, opened 2025-05-14
Remove non-existent plugin types

morrone:
Remove from "enum ldmsd_plugin_type" (inside struct ldmsd_plugin)
the plugin types that do no currently exist. None of the following
are valid plugin types, of type "struct ldmsd_plugin" at this time:

         LDMSD_PLUGIN_AUTH,
         LDMSD_PLUGIN_DECOMP
         LDMSD_PLUGIN_OTHER

The only valid plugin types at this time are:

        LDMSD_PLUGIN_SAMPLER
        LDMSD_PLUGIN_STORE


morrone:
We are changing the plugin API, so we don't need to maintain compatibility of this particular enum. Although I suppose I could start the counting at 1 instead of 0 if folks prefered. I don't think the "_OTHER" value should be in the enum, since that is not currently a valid plugin type.

Is there a way to add exceptions to the enum compatibility test?

morrone:
I updated the patch to leave in "LDMSD_PLUGIN_OTHER" just to quiet the automated test. It isn't really a valid plugin type, so I would recommend that we remove that as well in a follow up PR.

tom95858:
@morrone, I would suggest getting rid of all of them or none of them. The one I think is most egregious is "OTHER" and that's the one we're not removing. So why not remove them all, avoid the confusion and update the build workflow?

morrone:
> @morrone, I would suggest getting rid of all of them or none of them. The one I think is most egregious is "OTHER" and that's the one we're not removing. So why not remove them all, avoid the confusion and update the build workflow?

I'm fine with removing OTHER. You'll see above that it removing OTHER _was_ part of my origin all commit.

What is the right way to get the enum compatibility checks to to not complain about it?

nichamon:
Regarding the enum compatibility test, removing OTHER from `.github/enum-check/*-enum.list` files will prevent the test from complaining. The check is in place to prevent unintentional changes to any enum.

nichamon (approved):

PR #1720, opened 2025-05-15
Fix `--with-munge` configure option

narategithub:
- Remove `--enable-munge` and `--disable-munge` (the default was --disable-munge)

- Without specifying anything, the configure script will default to "check", and build the dependant (currently only `ldms_auth_munge`) if `libmunge` is present. If `libmunge` is not present, the dependent will not be built.

- `--with-munge=no` will skip munge check and disable the build of the dependent.

- `--with-munge=yes` will check libmunge and exited with an error if libmunge is not found.

- `--with-munge=<MUNGE_PREFIX>` will build the dependent with libmunge in the munge installation prefix (e.g. /opt/munge).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -543,8 +543,7 @@ OPTION_WITH([cray-hss-devel], [CRAY_HSS_DEVEL],[/usr])
 OPTION_ARIES_MMR
 
 dnl munge
-OPTION_DEFAULT_DISABLE([munge], [ENABLE_MUNGE])
-OPTION_WITH([munge], [MUNGE])
+OPTION_WITH_CHECK([munge], [MUNGE], [munge.h], [munge], [munge_encode])
 
 LDMSD_PLUGIN_LIBPATH=${pkglibdir}
 AC_SUBST(LDMSD_PLUGIN_LIBPATH)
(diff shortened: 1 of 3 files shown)


tom95858:
@narategithub It's not clear to me what we are trying to do here.

morrone (changes requested):
This looks good, but it looks like there is a use of ENABLE_MUNGE_TRUE over in ldms/src/test/ldms-static-test-list.sh.in that probably will need updating in this PR.

morrone:
> @narategithub, @morrone: what concerns me is this syntax:
> 
> ```
> --with=no
> --with=yes
> --with=<path>
> ```
> 
> Is this commonly used?

Yes, that is standard with autoconf. Users might more often use these versions:

```
--with-munge  # equivalent in autoconf to --with-munge=yes
--without-munge  # equivalent in autoconf to --with-munge=no
--with-munge=<path>
```

https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/External-Software.html

Of course this assumes that ldms's custom OPTION_* macros maintain normal autoconf semantics. And actually I think the default "check" case might be broken in OPTION_WITH_CHECK. But that is probably a discussion for a new issue rather than this PR.

morrone:
> Of course this assumes that ldms's custom OPTION_* macros maintain normal autoconf semantics. And actually I think the default "check" case might be broken in OPTION_WITH_CHECK. But that is probably a discussion for a new issue rather than this PR.

Nevermind, That look OK too. I saw that the common "withval=check" was missing in the not-supplied case of AC_ARG_WITH, but the subsequence test of withval matches on withval=check _or_ withval empty. So that is fine too.

morrone (approved):

PR #1723, opened 2025-05-16
Fix json_entity_dump() in ovis_json

nichamon:
The jbuf pointer is not updated correctly after the buffer has been reallocated when the entity contains dictionaries or lists. The patch makes sure that the jbuf handle points to the address returned by realloc().


nichamon:
The pull request ports commit https://github.com/ovis-hpc/ldms/commit/7ac016f6846cd5bb74c2e50f9cd61c6f8d4d6d68 from `b4.4`.

PR #1724, opened 2025-05-16
add 'plugin_keywords' to yaml config dictionary

baallan:
This lets the user specify yaml congruent with the keyword (as opposed to attr/val pair) syntax in the ldmsd plugin configuration script language. A single keyword or a list of keywords (as a string array) is allowed.
```
config:
  foo: bar
  plugin_keywords: ["timing", "bob"]
or
config:
  foo: bar
  plugin_keywords: timing
```
to support the existing ldmsd configuration language which allows keywords (as distinct from attr=value pairs).
The results of the above would be
config ... foo=bar timing bob
or
config ... foo=bar timing

The attribute name 'plugin_keywords' is reserved for ldmsd yaml use, as it seems unlikely to be used by future plugins. It seems, on the other hand, that some future plugin writer could easily come up with a parameter 'keywords'.

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -187,9 +187,21 @@ def parse_to_cfg_str(cfg_obj):
     cfg_str = ''
     for key in cfg_obj:
         if key not in INT_ATTRS:
-            if len(cfg_str) > 1:
-                cfg_str += ' '
-            cfg_str += key + '=' + str(cfg_obj[key])
+            if key == "plugin_keywords":
+                words = cfg_obj[key]
+                if isinstance(words, str):
+                    if len(cfg_str) > 1:
+                        cfg_str += ' '
+                    cfg_str += words
+                else:
+                    for w in words:
+                        if len(cfg_str) > 1:
+                            cfg_str += ' '
+                        cfg_str += str(w)
+            else:
+                if len(cfg_str) > 1:
+                    cfg_str += ' '
+                cfg_str += key + '=' + str(cfg_obj[key])
     return cfg_str
 
 def parse_yaml_bool(bool_):


baallan:
This bug fix allows the user to avoid hacks like appending keywords to the value of an attr/value parameters. Thus we could now add (separately) a check that validates the value in a pair as not containing whitespace.

morrone:
YAML already allows for keys that are valueless. I think this would be better:

```
config:
  foo: bar
  timing:
  bob:
```

This is also equivalent in YAML to:

```
config:
  foo: bar
  timing: NULL
  bob: NULL
```

Either and explicit NULL value or the absence of a value means "a key with no value" in YAML. That seems to be directly equivalent to the "keywords" in ldmsd config language.

I think that our converter should just handle NULL value, if it doesn't already.

baallan:
> YAML already allows for keys that are valueless. I think this would be better:
> 
...
> Either an explicit NULL value or the absence of a value means "a key with no value" in YAML. That seems to be directly equivalent to the "keywords" in ldmsd config language.
> 
> I think that our converter should just handle NULL value, if it doesn't already.

As presently coded, you end up with (at best) ".... key= ....", instead of "... key ...", which kicks key into the av list instead of the keyword list. Hence the proposal. At worst, the python yaml1.1 implementation dislikes valueless keys.

This patch is just part of a broader issue first surfaced here over 2 years ago around conf syntax that does not map to dictionaries or yaml as perpetrated by python: 
https://github.com/ovis-hpc/maestro/issues/74

emdonat:
> YAML already allows for keys that are valueless. I think this would be better:
> 
> ```
> config:
>   foo: bar
>   timing:
>   bob:
> ```
> 
> This is also equivalent in YAML to:
> 
> ```
> config:
>   foo: bar
>   timing: NULL
>   bob: NULL
> ```
> 
> Either an explicit NULL value or the absence of a value means "a key with no value" in YAML. That seems to be directly equivalent to the "keywords" in ldmsd config language.
> 
> I think that our converter should just handle NULL value, if it doesn't already.

In Python it would be a NoneType and a branch in the same section of code could handle it and also preserve the ability to use what is a normal YAMLism (if we don't then we still leave the timing=None issue in place?).  I think that 

> > YAML already allows for keys that are valueless. I think this would be better:
> 
> ...
> 
> > Either an explicit NULL value or the absence of a value means "a key with no value" in YAML. That seems to be directly equivalent to the "keywords" in ldmsd config language.
> > I think that our converter should just handle NULL value, if it doesn't already.
> 
> As presently coded, you end up with (at best) ".... key= ....", instead of "... key ...", which kicks key into the av list instead of the keyword list. Hence the proposal. At worst, the python yaml1.1 implementation dislikes valueless keys.
> 
> This patch is just part of a broader issue first surfaced here over 2 years ago around conf syntax that does not map to dictionaries or yaml as perpetrated by python: [ovis-hpc/maestro#74](https://github.com/ovis-hpc/maestro/issues/74)
> 
> Also assuming yaml containing "key: " as intentional for the keyword syntax overlooks the possibility that it came from a jinja template where the substitution evaluated to empty.

In Python it would be a NoneType and a branch in the same section of code could handle it and also preserve the ability to use what is a normal YAMLism (if we don't then we still leave the timing=None issue in place?).  This would also allow a string of None in a kv pair to evaluate correctly (timing: "None" -> timing=None).

Jinja templates are not a mandate for creating the conf so engineering around a dropped config there seems out of scope maybe.

morrone:
> As presently coded, you end up with (at best) ".... key= ....", instead of "... key ...", which kicks key into the av list instead of the keyword list. Hence the proposal. At worst, the python yaml1.1 implementation dislikes valueless keys.

Right, we should change the way it is coded to handle yaml NULL values as valueless, and just do "... key ...".

This is _not_ a problem in python. It is just a problem in parser_util.py's handling of NULL.

> Also assuming yaml containing
"key: "
as intentional for the keyword syntax overlooks the possibility that it came from a jinja template where the substitution evaluated to empty.

That would be bug in the jinja template. We don't need to concern ourselves with bugs in this issue.

If we _want_ to have an empty-string value for a key, YAML supports that as well:

```
config:
  foo: ""
  bar:
```

Emptry-string could generate "foo=" while NULL (or implied NULL) generates "bar". But again, if the jinja generates the wrong thing, that is just a bug in the jinja template.

baallan:
@emdonat @morrone Of course, "key: " could just be a bug in the jinja template. The problems are:
 - it's likely to be a common bug based on typos.
 - it leads to something that parses but probably (without an error in many cases) misconfigures a plugin
 
Setting up our jinja users for silent mistakes is not a great way to maintain our reputation for ease of use (which after all was the point of going to yaml).

baallan:
for reference in the conf input:
foo=
will pass the input parser and C will see a NULL parameter value rather than an empty string "".
foo="" will pass the parser and C will see a string with value (if expressed as a C source code string) "\"\"".

emdonat:
> @emdonat @morrone Of course, "key: " could just be a bug in the jinja template. The problems are:
> 
> * it's likely to be a common bug based on typos.
> * it leads to something that parses but probably (without an error in many cases) misconfigures a plugin
> 
> Setting up our jinja users for silent mistakes is not a great way to maintain our reputation for ease of use (which after all was the point of going to yaml).

It seems like there is a fundamental question that needs answering.  How do we/LDMS as a consumer of a yaml file want to treat the following:

```
foo:
bar: value
```

Right now we are converting `foo:` to be added into the string building as foo=None.  I think that we agree that this isn't useful and needs fixing.  The options seem to be that we: 
- take that to mean a keyword as the consumers of the file
- continue to let it just resolve to a NoneType that is cast to a string as "None"
- add it to the string builder as `foo=`

Additionally if we are adding a new keyword `plugin_keywords` it doesn't seem to address an instance where `foo:` is provided other than leaving it to convert to `foo=None`.  I would suggest that we can't necessarily guarantee that `foo=None` loudly fail and is also a likely configuration mistake.

morrone:
> for reference in the conf input: foo= will pass the input parser and C will see a NULL parameter value rather than an empty string "". foo="" will pass the parser and C will see a string with value (if expressed as a C source code string) """".

I didn't say anything about putting ```foo=""``` on the config command line for C. I gave an example of an empty string in YAML. We can take the following YAML, which says "the value is an empty string":

```
  config:
    foo=""
```

and generate this string for the C code, which tells the C code, "the value is an empty string":

```
  foo=
```

morrone:
> @emdonat @morrone Of course, "key: " could just be a bug in the jinja template. The problems are:
> 
> * it's likely to be a common bug based on typos.
> * it leads to something that parses but probably (without an error in many cases) misconfigures a plugin
> 
> Setting up our jinja users for silent mistakes is not a great way to maintain our reputation for ease of use (which after all was the point of going to yaml).

I think that is an issue that can, and should be addressed elsewhere inside of ldmsd. It isn't something we should work around in the YAML configuration.

Frankly, I've never really understood why there are separate argument lists sent to plugins for keys with and keys without values. We could just have the value be NULL when a key is provided by itself on the "config" command line. And if they were unified, then all the plugins would more easily be able to throw errors when a key has a value that it shouldn't, or a key is missing a value when it is required. But since the same key can appear in either list, most plugins only every check the one they expect and leave the other to silently be ignored.

_That_ is the real root problem that you are identifying here. That ldmsd might silently accept a valueless keyword. But again, that is a problem elsewhere in the code, not in the yaml.

morrone:
> Right now we are converting `foo:` to be added into the string building as foo=None. I think that we agree that this isn't useful and needs fixing. The options seem to be that we:
> 
> * take that to mean a keyword as the consumers of the file
> * continue to let it just resolve to a NoneType that is cast to a string as "None"
> * add it to the string builder as `foo=`
> 
> Additionally if we are adding a new keyword `plugin_keywords` it doesn't seem to address an instance where `foo:` is provided other than leaving it to convert to `foo=None`. I would suggest that we can't necessarily guarantee that `foo=None` loudly fail and is also a likely configuration mistake.

I completely agree.  I don't think the parser should generate the string "None" in the output unless the literal string "None" appeared in the YAML. I think the first "take that to mean a keyword as the consumers of the file" option is the way to go, and your proposed code change looked sane to me (not having looked much at the surrounding code).

I think only ```foo=""``` in the YAML should become ```foo=``` in the string builder, and I think your patch would do that as well.

baallan:
> I think only `foo=""` in the YAML should become `foo=` in the string builder, and I think your patch would do that as well.

@morrone @emdonat 
So I would propose to revise the patch to handle config: section inputs as follows:
```
foo:
bar: 1
baz: ["a", "b"]
bob: []
barry: ""

config ....   foo bar=1 baz=a baz=b bob= barry=
```
This leaves only the corner case of repeated keyword uncovered:
```
config .... foo foo foo
```
Then any plugin which fails to use the ldmsd_plugattr api or its own hand-coded checks of avl/kwl content to trap input templating errors is just buggy. We have a lot of those already anyway. 

Does anyone have a yaml syntax suggestion for how repeated keywords should be rendered as yaml input? One that isn't essentially  the original plugin_keywords proposal?

morrone:
> @morrone @emdonat So I would propose to revise the patch to handle config: section inputs as follows:

Great, I think we are on the same page then for the things we've been discussing so far. Your suggestion adds array handling in the discussion. I don't think I am opposed to that approach.

Duplicate keys on the "config" line is another thing that I'm not fond of in ldms. I don't think that most new plugin authors know that is possible, and probably few handle it sanely. But we have it, and I am not seeing an issue with your proposal.

> This leaves only the corner case of repeated keyword uncovered:
> 
> ```
> config .... foo foo foo
> ```

You _want_ to be able to do that? To be consistent, that would be any of the following:

```  foo: [NULL, NULL, NULL] ```
```  foo: [ ~, ~, ~ ]``'

or using a long-form sequence:

```
  foo:
    - NULL
    -
    - ~
```

Why do we even allow this sort of thing in ldms?

**Comment by baallan** [comment]
> Why do we even allow this sort of thing in ldms?

I don't know the history of the 'keyword' syntax in ldms, but it allows darn near anything to the plugin authors, e.g.
```
# for folk in love with repeated -v:"
config ...... verbose verbose verbose
# the above people need to learn about our new logging control system.

# for folk with their own custom plugin config syntax not based on a=v pairs:
config name=oddplugin .... opt1 foo opt2 bar opt3 baz optarray4 0 1 0 0
config name=oddplugin .... dopt1 quid dopt2 pro dopt3 quo optarray7 0 1 0 0 1 0 1
```
Back in the days of wild plugins for cray and MSRs, this may have seen more usage.
Interestingly, what keyword doesn't allow is mixing the two while preserving order, e.g.
```
config name=cursedplugin ... kw1 kw2 kw3 a=v kw4 kw5
```
which pulls a=v out of the keyword syntax sequence and puts it in avl.

**Comment by nick-enoent** [comment]
I don't love the idea of hacking around a dictionary to achieve this effect. The config key is a dictionary of lists, where each list calls its own "config" command. It accepts both strings and dictionaries. @baallan you can achieve your desired configuration by passing a string (as a list) to the config command. 

e.g. 

`config: [ "foo=bar timing bob"  ]`

or 

`config :
  - "foo=bar timing bob" `

I'm fine with adjusting this to make more sense for the community, but I'm not sure passing NULL as a key value is the way to do it.

**Comment by morrone** [comment]
> e.g.
> 
> `config: [ "foo=bar timing bob" ]`

I am not sure that I am a fan of that. We are trying to design a yaml configuration file that abstracts aways the underlying  implementation details. Doing so allows us to change the implementation as we desire in the future with minimal impact to the YAML configuration language. But this seems to telegraph the underlying ldmsd config language to the YAML configuration without any abstraction.

**Comment by nick-enoent** [comment]
> > e.g.
> > `config: [ "foo=bar timing bob" ]`
> 
> I am not sure that I am a fan of that. We are trying to design a yaml configuration file that abstracts aways the underlying implementation details. Doing so allows us to change the implementation as we desire in the future with minimal impact to the YAML configuration language. But this seems to telegraph the underlying ldmsd config language to the YAML configuration without any abstraction.

Just to clarify, this is how it works now. I'm not proposing it should work this way in perpetuity. This was introduced because we had a few users who wanted to be able to call the configure command multiple times on the same plugin (along with being able to use strings for the configure command).

**Comment by baallan** [comment]
The current 'string' configuration (instead of a dictionary) is essentially a free-form use-at-your-own-risk escape hatch for handling multiline and keyword syntax. It's good to have when there is no alternative (as was the case up to now), but shouldn't be the go-to solution in most cases. The function write_sampler() in the python conf generator provides all sorts of defaults if there's a dictionary under "config", but just pastes exactly what the user (mis-)specified when it's a string. I would like to be able to write the entire specification for all plugins without resorting to an escape hatch back to the conf language.

**Comment by nick-enoent** [comment]
> The current 'string' configuration (instead of a dictionary) is essentially a free-form use-at-your-own-risk escape hatch for handling multiline and keyword syntax. It's good to have when there is no alternative (as was the case up to now), but shouldn't be the go-to solution in most cases. The function write_sampler() in the python conf generator provides all sorts of defaults if there's a dictionary under "config", but just pastes exactly what the user (mis-)specified when it's a string. I would like to be able to write the entire specification for all plugins without resorting to an escape hatch back to the conf language.

That's fine, and I think we can approach this in a cleaner fashion. That said, we should socialize this with the user group, but I don't think this should land in 4.5.

**Comment by baallan** [comment]
> That's fine, and I think we can approach this in a cleaner fashion. That said, we should socialize this with the user group, but I don't think this should land in 4.5.

This is a bug fix to make our production plugins work at snl when 4.5.1 is released. Now is the time. The folk in the conversation *are* the user group, for all practical purposes, with respect to this bug.

**Comment by baallan** [comment]
@nick-enoent brought to my attention that i shortcut the list portion of the config syntax, which allows for multiple config lines on the same plugin. The following example emphasizes more obviously that a sampler (given the C library's list of keywords syntax) could have a custom config language:
```
  test_sampler:
    name: test_sampler
    interval: 1s
    config:
    - schema: test_sampler
      component_id: "${COMPONENT_ID}"
      producer: "${HOSTNAME}-10001"
      perm: "0644"
      instance: "${host}/job"
      config_append: "a=1 a=2 a=3 d e f_array_3 0.1 0 1 1"
      # config_append becomes a reserved attribute name plugin authors can't use themselves.
    - baz: ["a","b"]
    - timing:
    - "set glop 2.75"
 ```
could be rendered as:
```
config name=test_sampler schema=test_sampler ... instance=${host}/job a=1 a=2 a=3 d e f_array_3 0.1 0 1 1
config name=test_sampler timing
config name=test_sampler baz=a baz=b
config name=test_sampler set glop 2.75
```

This lets the configurer get the goodness of the dictionary and defaulting for common stuff and allows straight pass-through of baroque plugin-defined languages/keywords as a config_append string, while also giving the configurer the ability to opt out of the 'goodness' and just pass one or more handcrafted pure string lines.

baallan:
@nick-enoent after some meeting chatter, the near-concensus seems to be:

- For really messy stuff with repeated keywords (not repeated attributes) or numbers as keywords, just use the existing "whole line" string trapdoor:
```
- "schema=test_sampler component_id=${COMPONENT_ID} producer=${HOSTNAME}-10001 perm=0644 instance=${host}/job d e f_array_4 0.1 0 1 1"
```
- We should add support for yaml keys with the null value producing keyword ( a instead of a=) output
- We should add support for yaml keys with a list value producing repeated attr/value iterated over the value list.
- SNL will contribute a patch handling the 2 previous points.
- SNL is compiling a survey of what config language (attr/value vs keyword vs repeat vs ...) is used for each plugin to answer:
     - how bad is the legacy syntax stuff (repeat attrs or keywords)
     - is it practical to switch to a dictionary-style syntax where legacy syntax exists
     - are all the options given validated in the plugin?
- After a suitable period of refactoring plugins to make them yaml dictionary friendly, the C plugin api could shift (or expand: register config_dict(obj) instead of config(kwl, avl)) from the "lib/src/ovis_util/util.h: av_list" to something yaml/json tree/list-like.

baallan:
@morrone @emdonat @nick-enoent the revised patch (with revised description and test at top of PR page) for the yaml sampler config handling of list and empty attributes is here.

ldms/python/ldmsd/parser_util.py
```diff
@@ -183,12 +183,42 @@ def bin_search(slist, x):
...
+                for i in val:
+                    if len(cfg_str) > 1:
+                        cfg_str += ' '
+                    if i is None:
+                        cfg_str += str(key)
+                        continue
+                    if isinstance(i, (str, int, float)):
+                        cfg_str += key + '=' + str(i)
+                    else:
+                        raise TypeError(f'Error: parse_to_cfg_str not expecting list value of' + str(i)+' for key '+  key )
```
nick-enoent, line 218: The "i" variable can referenced directly in the string with `{i}` as you're using a formatted string in the TypeError. Breaking the string and appending `str(i)` is unnecessary.

ldms/python/ldmsd/parser_util.py
```diff
@@ -183,12 +183,42 @@ def bin_search(slist, x):
...
+                continue
+            raise TypeError(f'Error: parse_to_cfg_str not expecting value ' + str(val)+' for key '+  key )
```
nick-enoent, line 220: Same thing here

nick-enoent (changes requested):

ldms/python/ldmsd/parser_util.py
```diff
@@ -183,12 +183,42 @@ def bin_search(slist, x):
     return None
 
 def parse_to_cfg_str(cfg_obj):
+    """ convert element(s) from a config list dictionary to a conf line.
+    string literal elements are handled elsewhere.
+    a:b  --> a=b
+    a:""  --> a=
```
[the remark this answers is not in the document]
baallan: > there are plugins that use av_value(x) != NULL as a boolean test instead of the keyword syntax.
@nick-enoent 
Sorry, the !=NULL check corresponds to X=anything_non_empty; there are plugins using this.

Test (av_idx_of(avl, x)!= -1) could be used to check for the mere existence of X= with no restriction on the value.
A quick grep check suggests there are not, at present, such plugins.  

We have no direct way in the configuration language to specify the non-null empty string (c: "\0") as the conf language has no quoting mechanism. I haven't reverified whether in conf processing ' x=${undefined_env_var} ' yields null or an empty string as the av_value.

nick-enoent:
@baallan Thanks for making those changes - can you change the commit message so we don't have confusion going forward? Otherwise I think this is good to go.

nick-enoent:
> @baallan Thanks for making those changes - can you change the commit message so we don't have confusion going forward? Otherwise I think this is good to go.

I see you did, thanks!

nick-enoent:
@baallan I think the only thing we're missing is updated documentation concerning the configuration vernacular.

baallan:
> @baallan I think the only thing we're missing is updated documentation concerning the configuration vernacular.

@nick-enoent 
that's just in ldmsd_yaml_parser.rst, right?

baallan:
> @baallan I think the only thing we're missing is updated documentation concerning the configuration vernacular.

@nick-enoent i added a commit to update the doc for config.

nick-enoent (approved):

baallan:
@tom95858 this is ready to merge and definitely wanted in the release. @nick-enoent approved.
@emdonat @morrone please comment now or forever hold your peace.

emdonat:
This is good to go.

PR #1738, opened 2025-05-19
Make the strgp configuration parameter "container" optional

morrone:
Make the strgp configuration parameter "container" optional so that store_avro_kafka can use its internally configured server information.

Add checks to the rest of the stores to make sure that container is not NULL.


tom95858:
@morrone, I guess I'm a bit reluctant to modify 17 stores to verify that container= is specified, so that one doesn't have to ignore the value.

tom95858:
@morrone, I am looking into this and agree that at the very least you should be able to specify an empty value. I'll try to get this done today.

PR #1743, opened 2025-05-20
Define destructor, and add macros for TEL_PATH_CXI and RH_PATH_CXI

jstile-lbl:
Changes
---
* Define macros for file paths to telemetry (TEL_PATH_CXI) and retry handler (RH_PATH_CXI).
* Fill out the destructor function.

### DIFF
--- a/ldms/src/sampler/cxi_sampler/cxi_sampler.c
+++ b/ldms/src/sampler/cxi_sampler/cxi_sampler.c
@@ -1,0 +1,598 @@
+/* -*- c-basic-offset: 8 -*-
+ * Copyright (c) 2025 Lawrence Berkley National Lab, All rights reserved.
+ * Copyright (c) 2025 Open Grid Computing, Inc. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the BSD-type
+ * license below:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *      Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ *
+ *      Redistributions in binary form must reproduce the above
+ *      copyright notice, this list of conditions and the following
+ *      disclaimer in the documentation and/or other materials provided
+ *      with the distribution.
+ *
+ *      Neither the name of Sandia nor the names of any contributors may
+ *      be used to endorse or promote products derived from this software
+ *      without specific prior written permission.
+ *
+ *      Neither the name of Open Grid Computing nor the names of any
+ *      contributors may be used to endorse or promote products derived
+ *      from this software without specific prior written permission.
+ *
+ *      Modified source versions must be plainly marked as such, and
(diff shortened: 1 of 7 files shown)


ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -578,10 +574,63 @@ static int constructor(ldmsd_plug_handle_t handle)
...
+        ldms_record_delete(cxi->rh_rec);
+
+    // Free base data
+    base_del(cxi->base);
+
+    // Finally, free the main structure
+    free(cxi);
 }
 
 static struct ldmsd_sampler cxi_plugin = {
```
morrone, line 633: Just FYI, if you want to make this compatible with the latest API changes (which is probably why you removed .base.name), you will also need to change this struct definition to:

```struct ldmsd_sampler ldmsd_plugin_interface = {```

And then delete the get_plugin() function below.

File: ldms/src/sampler/cxi_sampler/cxi_sampler.c [hunk 2]
```diff
@@ -578,10 +574,63 @@ static int constructor(ldmsd_plug_handle_t handle)
...
+            for (int j = 0; j < cxi->rh_files[i].count; j++) {
+                free(cxi->rh_files[i].names[j]);
+            }
+        }
+        free(cxi->rh_files);
+    }
+
+    // Free LDMS records
+    if (cxi->tel_rec)
+        ldms_record_delete(cxi->tel_rec);
```
morrone, line 622: I see why this looks like the right thing to do. We really need to fix the documentation for this function (and maybe even tweak the records API...).

Once records are added to a schema, they are owned by the schema. They will be freed when schema is freed. So calling ldms_record_delete() here could very well cause a segfault.

In fact, I would suggest removing "tel_rec and "rh_rec" from struct cxi_s. We can just use function-scope variables in create_metric_set(). After the record is attached to the schema, I don't think we ever need to (or maybe even _can_) do anything with it.

morrone (changes requested):

ldms/src/sampler/cxi_sampler/cxi_sampler.c
tom95858: @morrone that has all been done in cxi_sampler ... and actually tested.

ldms/src/sampler/cxi_sampler/cxi_sampler.c
tom95858: @morrone, yes we should distinguish between a record definition and a record instance. It is confusing.

tom95858: @morrone, we need to have the record definition to create new record instances, however, we could look them up in the schema instead of caching them in the CXI instance.

morrone: > @morrone, we need to have the record definition to create new record instances, however, we could look them up in the schema instead of caching them in the CXI instance.

We would only need the record definition again if we need to generate a new schema. To add record instances in a metric set, we only need the index number of the record from the schema. Those are recorded in cxi_s in:

```
	int tel_rec_mid;
	int rh_rec_mid;
```

morrone: And, actually, I don't think a record definition _can_ be reused in a second schema. Because the record definition becomes owned by that one schema once it is attached to it.

PR #1756, opened 2025-05-24
Add a sampler of CXI telemetry and retry-hander data

jstile-lbl:
Rquest to add Add a sampler of CXI telemetry and retry-hander data

Extend cxi_sampler sampler
    
    1. Define macros for file paths to telemetry (TEL_PATH_CXI) and retry handler (RH_PATH_CXI)
    2. Fill out the destructor function for cxi_sampler
    3. Define ldmsd_plugin_interface to avoid warnings during startup.
    4. Add 2 optional config filters to collect subsets of retry-handler and telemetry counters.
    5. Update cxi_sampler documentation with counters configuration option


ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -0,0 +1,754 @@
...
+	rh_rec = ldms_record_create("rh_record");
+	rc = ldms_record_metric_add(rh_rec,
+				    "iface_name", "",
+				    LDMS_V_CHAR_ARRAY, 32);
+
+	for (i = 0; i < cxi->rh_files[0].count; i++) {
+		rc = ldms_record_metric_add(rh_rec, cxi->rh_files[0].names[i],
+					    "", LDMS_V_D64, 0);
+		if (rc < 0) {
+			rc = errno;
```
morrone, line 430: Add ```ldms_record_delete(rh_rec)```. Not yet owned by the schema.

ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -0,0 +1,754 @@
...
+					 ldms_record_heap_size_get(tel_rec) * cxi->iface_count);
+	cxi->tel_list_mid = rc;
+	if (rc < 0) {
+		ovis_log(cxi->log, OVIS_LERROR,
+			 "Error %d creating tel_list.\n", errno);
+		goto err;
+	}
+
+	/* List of per interface retry handler records */
+	rh_rec_mid = ldms_schema_record_add(cxi->schema, rh_rec);
```
morrone, line 450: Need to check return code, and on error call ```ldms_record_delete(rh_rec)```.

ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -0,0 +1,754 @@
...
+		cxi->tel_list_mval = ldms_metric_get(cxi->set, cxi->tel_list_mid);
+		rh_rec_mval = ldms_record_alloc(cxi->set, rh_rec_mid);
+		tel_rec_mval = ldms_record_alloc(cxi->set, tel_rec_mid);
+
+		rc = ldms_list_append_record(cxi->set, cxi->rh_list_mval, rh_rec_mval);
+		if (rc) {
+			ovis_log(cxi->log, OVIS_LERROR,
+				 "Error %d appending record to rh_list.\n",
+				 rc);
+			goto err;
```
morrone, line 483: At this point, and any following goto err, we have also called base_set_new_heap(). Which means we need a second label, such as "err2", in which we call base_set_delete(cxi->set). (and set cxi->set == NULL)

ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -0,0 +1,754 @@
...
+				 "Error %d appending record to tel_list.\n",
+				 rc);
+			goto err;
+		}
+	}
+	return 0;
+
+err:
+	if (cxi->base)
+		base_schema_delete(cxi->base);
```
morrone, line 498: Should probably set ```cxi->base == NULL```  so we can't double call base_schema_delete() in destructor()

morrone (changes requested):
Making good progress!

Besides the minor things I commented about inline, I think there are two bigger things I'd like to see in this sampler:

1) Adjust dynamically to the number of cxi interfaces found (slinghost_metrics sampler does this)
2) Allow a configuration file (or two files) listing the metrics we want to monitor (slingshot_metrics does this).

The reason for the latter is about having a predictable ldms schema. Offering a regex to pattern-match files seems nice, but might be a problem in the future unless the regexes used are very tightly written. When upgrading to a new version of SHS (Slingshot Host Software), we might get new metrics that match one of our existing regular expressions. That would cause an unexpected schema change.

By listing the desired metrics out in a configuration file(s), while tedious, it ensures that the schema only changes when a human decides that they want the schema to change.

Also:

3) Don't use atof() to parse the integer values in the files. Switch from _D64 to _U64 to store the metrics in the ldms schema.

ldms/src/sampler/cxi_sampler/cxi_sampler.c
morrone: The multiple labels for errors should only be used in for cases where there is a strict stack (LIFO) of allocations in a function. Which is a fairly common idiom. So if we do allocations A, B, C, D, then the error labels can free in them in the reverse order D, C, B, A. Each can have a label so we can jump to the right point in freeing things depending on how far we have gotten with allocations.

```
   A = malloc()
   if (not A) goto err1;
   B = malloc()
   if (not B) goto err2;
   C = malloc()
   if (not C) goto err3;
   D = malloc()
   if (not D) goto err4;
   return success
err4:
  free C
err3:
  free B
err2:
   free A
err1:
   return failure
```

But we can't do that here because records aren't directly freeable once they have been passed to a schema. So that stack-like manner of allocations and frees in the error path doesn't apply here.

ldms/src/sampler/cxi_sampler/cxi_sampler.c
morrone: Only if we set cxi->base to NULL after base_schema_delete(). Otherwise the pointer will be non-NULL, and the code in destructor() will try to access memory that has already been freed.

tom95858:
I don't have access to the platform, but I thought I there were doubles in the test telemetry data files.  The rh files were all counters.

morrone (approved):

morrone:
I think that addresses all of the must-have changes that I listed, as of 1f5a536. For production use, I think I would need the other two wishlist items that I mentioned: dynamically adjust the list to the interfaces available (as they come and go), and allow configuration of metrics from a file or files. But those wishlist things could be done in follow-up PRs. This one could land as it is.

@tom95858 @nichamon, someone with admin access needs to approve @jstile-lbl's PR for running the automated github workflows.

jstile-lbl:
@morrone @tom95858 I have added the configuration parameters to read counter filters from files for telemetry and retry handler metrics

jstile-lbl:
squashed commits with a rebase.

morrone:
It looks like you accidentally did a merge again instead of a rebase. You can get your branch back to pre-merge state by doing this:

```
git checkout jstile-lbl/cxi_sampler
# make sure your tree is clean, no work in progress
git reset --hard b8a4c8414
```

Then you can rebase onto main like this:

```
git checkout jstile-lbl/cxi_sampler
git fetch origin
git rebase --onto origin/main HEAD^^ #(notice two carets '^' because there are two commits on your branch)
git push -f jstile-lbl jstile-lbl/cxi_sampler
```

morrone:
How are things going addressing the compiler warnings that the github checks are revealing?

jstile-lbl:
I haven't been able to meet with Tom.  Some of the errors are in the
recommended code, so I'm not sure.

jstile-lbl:
> It looks like you accidentally did a merge again instead of a rebase. You can get your branch back to pre-merge state by doing this:
> 
> ```
> git checkout jstile-lbl/cxi_sampler
> # make sure your tree is clean, no work in progress
> git reset --hard b8a4c8414
> ```
> 
> Then you can rebase onto main like this:
> 
> ```
> git checkout jstile-lbl/cxi_sampler
> git fetch origin
> git rebase --onto origin/main HEAD^^ #(notice two carets '^' because there are two commits on your branch)
> git push -f jstile-lbl jstile-lbl/cxi_sampler
> ```

One of the tests told me this file had a merge conflict, which is what the merge addressed.

if I revers the resolution to the merge conflict,  then how do I address the merge conflict?

tom95858:
@jstile-lbl the way to do this without the merge commit is to rebase. I typically do git rebase -i which allows you to interactively apply your commits, reorder them, combine them, etc... The rebase step will "stall" when it detects a merge conflict and allow you to choose how you want to combine the conflicting changes. Take yours, take theirs, combine the two, etc... After you've resolved the conflict you do a git rebase --continue and it will move on to the next commit

tom95858:
e

> > 0@1743569213.088677609
> 
> No words associated with this comment, so I'll assume it meant "Oh, right, the counters are integers." :)

Those are definitely not integers. They are floating point numbers. That said, they are from my "test files" they are not from an actual system. I assumed they were copied from the telemetry data on a running system, but I don't know for sure.

morrone:
> e
> 
> > > 0@1743569213.088677609
> > 
> > 
> > No words associated with this comment, so I'll assume it meant "Oh, right, the counters are integers." :)
> 
> Those are definitely not integers. They are floating point numbers. That said, they are from my "test files" they are not from an actual system. I assumed they were copied from the telemetry data on a running system, but I don't know for sure.

We are only concerned with the counter, which is the number _before_ the `@` symbol. And that is the only part that the code parses. The counters before the `@` symbols are integers, not floating point.

The number after the `@` symbol, the only one that contains a decimal point, is the timestamp. We don't parse that. If we _did_ parse that, we would want to read that in as two integers (the normal way to handle high resolution timestamps), not treat it as a floating point number. Any way we look at it, we shouldn't be using floating point here.

John already fixed this.

ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -0,0 +1,861 @@
...
+			ovis_log(cxi->log, OVIS_LERROR,
+				 "Error %d appending record to tel_list.\n",
+				 rc);
+			goto err2;
+		}
+	}
+	return 0;
+err2:
+    if (cxi->set) {
+        base_set_delete(cxi->base);
```
morrone, line 576: We still need to set `cxi->set = NULL;` to avoid the double free that would happen in destructor().

ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -0,0 +1,861 @@
...
+    }
+err1:
+	if (cxi->base)
```
morrone, line 579: We should probably switch this test to `if (cxi->schema)`, and then set `cxi->schema = NULL` to avoid the double free ithat would otherwise happen in destructor().

morrone (changes requested):

morrone:
I don't know what you tried there with git, but something went very wrong.  Maybe talk us through the commands that you are trying?

When you think you have things in the right shape, it is always good to verify by looking at the tree with something like `git log --oneline --graph`.You'll want to look at the tree of commits with something like "git log --graph" to make sure things are coming out correct. Your commits should be a straight set on top of main with no branches or merges.

jstile-lbl:
We decided to start a new pull request because once we solved the merge conflict
with ldms/src/sampler/Makefile.am, the commit history is not linear, and we can't complete the pull request until the conflict is solved. The  change is simple and I'm not clear why it can't merge  the change.

So we closed this pull request and opened  https://github.com/ovis-hpc/ldms/pull/1787

PR #1758, opened 2025-05-27
decomp_flex warning instead of error for unhandled set

narategithub:


nichamon:
I think this change makes sense. The warning message when there's no default behavior to fall back on helps prevent silent failures. I believe this case would commonly occur when users configure decomposition for specific set schema digests but don't provide a decomposition scheme for wildcards. The repeated warning messages will be helpful for catching unexpected set schema digests.

Issue #1716, opened 2025-05-14
Port 0840a25756d6310f03c2e3e7d95ed3d91df66fef from b4.4 to main


nichamon:
commit 0840a25756d6310f03c2e3e7d95ed3d91df66fef
Author: Tom Tucker <tom@ogc.us>
Date:   Thu Sep 26 17:28:01 2024 -0500

    Fix crash in ldms_ls when dumping decomposition

    Empty lists would cause ldms_ls to crash with the -d option.

nichamon:
The issue has been addressed by #1776 .

PR #1776, opened 2025-06-03
Fix crash in ldms_ls when dumping decomposition

nichamon:
Empty lists would cause ldms_ls to crash with the -d option.


nichamon:
This addressed #1716 .

Issue #1649, opened 2025-04-03
lustre samplers needing sys/kernel/debug need to document that fact in the rst pages.


baallan:
he man pages for lustre2_client, lnet_stats, lustre_client, and lustre_mdc should all mention that mounting debugfs is needed for the related stats files to exist.

baallan:
also applies to 4.4.6 if there is one.

emdonat:
Are there proposed changes for these?

tom95858:
The Lustre samplers rely on certain files being present in /sys. These files only appear if debugfs is mounted. If debugfs is not mounted, the sampler will fail with ENOENT errors and the data will not be available.

baallan:
#1780 attempts to address this issue.

morrone:
Patch landed.

PR #1780, opened 2025-06-04
add a note about debugfs for lustre plugins.

baallan:
This adds a reminder of the need for debugfs and root when using the lustre plugins to the related man pages.


baallan:
this is an attempt to address #1649

morrone (approved):

PR #1782, opened 2025-06-04
Port the ldmsd controler patches from b4.4

nichamon:


nichamon:
The patch has been rebased and the conflicts were resolved.

PR #1783, opened 2025-06-04
Fix static string size bug in store_timescale

nichamon:
There is a compile error when built with -Wall -Werror. The host in the format is a static string that is 100 bytes. That with the other elements of the format couuld exceed 128 resulting in truncated output.


nichamon:
This pull request is to port 5aae285 from b4.4 branch.

tom95858:
@nichamon, this code is broken. This fixed the compiler warning but we need to fix it or delete it. We should canvas who uses this sampler because I think it is a candidate for removal

PR #1794, opened 2025-06-06
Improve prdcr_stream_statua and stream_status

nichamon:
The patch replaces 'msg/sec' and 'msg count' with 'count/sec' and 'count' to provide more accuarate column headers and prevent any confusions with ldms_msg.


nichamon:
The patch changes the column headers from:
image


to:

image

PR #1798, opened 2025-06-07
Make parser_util.py:check_opt() ignore None interval

nichamon:


nick-enoent (commented):
Thanks for this fix! I think the `else: return None` is unnecessary as it should drop through and return spec[attr] with your check. But that's more of a nitpick.

nick-enoent (approved):

nichamon:
> Thanks for this fix! I think the `else: return None` is unnecessary as it should drop through and return spec[attr] with your check. But that's more of a nitpick.

You're right. I removed the `else:return None` logic. Thanks for pointing that out.

PR #1802, opened 2025-06-09
Fix the units in ldmsd_controller:store_time_stas

nichamon:
The statistics of min, max, and average in store_time_stats are in microseconds, but the column headers indicate that they are seconds. The patch fixes this.

The patch also changes the printed timestamp format from 4 decimal places to show only the second part.


nichamon:
The commit was pull out from ##1782 as it isn't part of the porting effort.

PR #1804, opened 2025-06-10
cxi_sampler: Introduce the virtual metric derived:link_restarts

morrone:
Introduce the virtual metric derived:link_restarts.  When specified in tel_counters or tel_counters_file, the cxi_sampler watches the link_restarts timestamp files and keeps a count of the changes in the metric named "derived:link_restarts". This allows us to count up to 10 link restarts per sample interval, which is good enough in practice for relatively quick sampling rates (say 5s or faster) to detect floods of link restart issues.

### DIFF
--- a/ldms/src/sampler/cxi_sampler/cxi_sampler.c
+++ b/ldms/src/sampler/cxi_sampler/cxi_sampler.c
@@ -76,8 +76,16 @@
 
 struct files_s {
 	int count;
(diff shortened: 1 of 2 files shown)


tom95858:
@jstile-lbl Did you get a chance to review this?

jstile-lbl:
Sorry i messed this. Looks like a nice enhancement!

PR #1807, opened 2025-06-12
Address quota and rx_rate unit in the input string and prdcr_status output

nichamon:


nichamon:
@nick-enoent @narategithub @tom95858 When I put the change to support, for example, '1k' as a value of quota or rx_rate. I forgot to add my changes in Python in the pull request.

nick-enoent (approved):

PR #1808, opened 2025-06-12
Curated Metric Set Sampler

nick-enoent:


ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,635 @@
...
+	}
+}
+#endif
+
+static int parse_cmetrics(char *path) {
+	int i;
+	int rc = 0;
+	json_t *cfp, *cfg_metrics;
+	json_error_t jerr;
+	cfp = json_load_file(path, 0, &jerr);
```
morrone, line 398: Please, please, please do not use json for this configuration file. Going forward I hope that we can avoid introducing any new json configration files into ldms. Gradually I'd like to convert the existing ones (e.g. decomposition config file) from a json parser to a  yaml parser.

And for anyone that just loves writing json config files, keep in mind that yaml is a strict superset of json. So a yaml parser can read both yaml _and_ json files.

ldms/src/sampler/curated_metrics/ldms-sampler_curated_metrics.rst
```diff
@@ -0,0 +1,112 @@
...
+| Within ldmsd_controller or a configuration file:
+| config name=curated_metrics [ <attr>=<value> ]
+
+DESCRIPTION
+===========
+
+With LDMS (Lightweight Distributed Metric Service), plugins for the
+ldmsd (ldms daemon) are configured via ldmsd_controller or a
+configuration file. The curated_metrics plugin provides memory info from
+/proc/curated_metrics.
```
nichamon, line 27: You may want to update the description to match of what data the curated metric plugin provides.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+		return 0;
+	}
+	return ENOMEM;
+}
+
+static void destructor(ldmsd_plug_handle_t handle)
+{
+	curated_metrics_t mi = ldmsd_plug_ctxt_get(handle);
+	struct iface_ent *e1, *e2;
+	TAILQ_INIT(&mi->netdevq);
```
nichamon: No need to call TAILQ_INIT() here.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+		       "%s: The schema '%s' could not be created, errno=%d.\n",
+		       __FILE__, mi->base->schema_name, errno);
+		rc = errno;
+		goto err;
+	}
+
+	mi->mf = fopen(procmeminfo, "r");
+	if (!mi->mf) {
+		ovis_log(mi->log, OVIS_LERROR, "Could not open the " SAMP " file "
+				"'%s'...exiting sampler\n", procmeminfo);
```
nichamon: A nit .. when `fopen()` returns NULL, it sets `errno`. You could include `strerror(errno)` in the log message to provide more information to users.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+		return ENOENT;
+	}
+
+	/* Location of first metric from proc/meminfo file */
+	mi->metric_offset = ldms_schema_metric_count_get(schema);
+
+	// Derive Free idle mem %
```
nichamon: Is the comment relevant? It seems the loop handles all metrics in `metric_list`.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+	mi->rec_def_idx = ldms_schema_record_add(schema, rec_def);
+	if (mi->rec_def_idx < 0) {
+		rc = -mi->rec_def_idx;
+		goto err3;
+	}
+
+	mi->netdev_list_mid = ldms_schema_metric_list_add(schema, "netdev_list", NULL, mi->heap_sz);
+	if (mi->netdev_list_mid < 0) {
+		rc = -mi->netdev_list_mid;
+		goto err2;
```
nichamon: Should this be `goto err3`?

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+	}
+	fclose(mi->nf);
+	mi->set = base_set_new_heap(mi->base, mi->heap_sz);
+	if (!mi->set) {
+		rc = errno;
+		goto err2;
```
nichamon, line 612: `goto err3`?

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+}
+
+static int sample(ldmsd_plug_handle_t handle)
+{
+	int rc;
+	int metric_cnt;
+	int *metric_no = &metric_cnt;
+
+	curated_metrics_t mi = ldmsd_plug_ctxt_get(handle);
+	if (!mi->base->set) {
```
nichamon: Are you verifying if the plugin instance has been configured? If I understand correctly, maybe we can verify `!mi->base` or both `!mi->base && ~mi->base->set`? This is to prevent ldmsd crashing when the sampler plugin instance starts before it has been configured.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+	mi->nf = fopen(netdev, "r");
+	if (!mi->nf) {
+		ovis_log(mi->log, OVIS_LDEBUG, "Could not open /proc/net/dev file ");
+		return ENOENT;
+	}
+	lh = ldms_metric_get(mi->base->set, mi->netdev_list_mid);
+
+	fseek(mi->nf, 0, SEEK_SET);
+	s = fgets(lbuf, sizeof(lbuf), mi->nf);
+	s = fgets(lbuf, sizeof(lbuf), mi->nf);
```
nichamon: Duplicated line.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+        if (!mi->sf) {
+                ovis_log(mi->log, OVIS_LERROR, "Could not open the " SAMP " file "
+                                "'%s'...exiting sampler\n", procstat);
+                rc = ENOENT;
+		goto out;
+        }
+
+	fseek(mi->sf, 0, SEEK_SET);
+	while (1 == fscanf(mi->sf, "%s", tok)) {
+		printf("%s\n", tok);
```
nichamon: Probably line 785 is left over from debugging sessions?

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+
+	// Derive Free idle mem %
+	//
+	int metric_list_sz = sizeof(metric_list) / sizeof(metric_list[0]);
+	ldms_record_t rec_def;
+	mi->rec_metric_len = 0;
+	for (i = 0; i < metric_list_sz; i++) {
+		if (mi->collect_list[i].collect) {
+			if (metric_list[i].not_rec) {
+				rc = ldms_schema_metric_add(schema, metric_list[i].name, metric_list[i].vtype);
```
nichamon: Just an FYI, you can call `ldms_schema_metric_add_with_unit()` for the metrics that have unit.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+		}
+		rc = 0;
+                /* Strip the colon from metric name if present */
+                i = strlen(metric_name);
+                if (i && metric_name[i-1] == ':')
+                        metric_name[i-1] = '\0';
+
+		if (strcmp(metric_name, "Active") == 0) {
+			rc = sscanf(lbuf, "%s %"PRIu64, metric_name, &(mi->minfo.active));
+			if (rc != 2 && rc != 3) {
```
nichamon: We are scanning two items: `%s` and `%PRIu64`, `sscanf()` can return at most 2. `rc !=3` is leftover from an old code?

nichamon (changes requested):

ldms/src/sampler/curated_metrics/curated_metrics.c
nick-enoent: Hey @nichamon , thanks for looking through this! This is to iterate over the first two lines in the /proc/net/dev file, as the first two lines are headers for the data.

ldms/src/sampler/curated_metrics/curated_metrics.c
nick-enoent: That's exactly what this is, thanks =)

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+		"				The 'metrics' keyword value is a list of metrics in the curated_metrics sampler to include.\n"
+		"				The 'ifaces' keyword value is a list of interfaces to include.\n";
+}
+
+static int config(ldmsd_plug_handle_t handle, struct attr_value_list *kwl, struct attr_value_list *avl)
+{
+	curated_metrics_t mi = ldmsd_plug_ctxt_get(handle);
+	int rc;
+
+	mi->log = ldmsd_plug_log_get(handle);
```
morrone: It is little better to get a copy of the log handle in a single time in constructor() rather than here in config().

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+#include "sampler_base.h"
+
+#include <jansson.h>
+#include <assert.h>
+
+#define PROC_MEMINFO "/proc/meminfo"
+#define PROC_LOADAVG "/proc/loadavg"
+#define PROC_STAT "/proc/stat"
+#define NET_DEV "/proc/net/dev"
+#define SAMP "curated_metrics"
```
morrone: This define should be removed.

We don't need to manually insert SAMP or the instance or plugin name in the log messages if we use the log handle provided by `ldmsd_plug_log_get(handle)`. That log handle is preconfigured to include the plugin name and plugin instance name in the output.

ldms/src/sampler/curated_metrics/curated_metrics.c
```diff
@@ -0,0 +1,1147 @@
...
+static char *procmeminfo = PROC_MEMINFO;
```
morrone: I don't see a good reason to have both macros and variable versions of the same string literal I think you should just pick one or the other and use them consistently throughout the code. And if you go with variables, it should really be "const char *", not just "char *".

ldms/src/sampler/curated_metrics/curated_metrics.c
nick-enoent: Hey @morrone, good point, and I agree. I've been updating code today and this was one of the first things I changed, alongside suggested changes from @nichamon

ldms/src/sampler/curated_metrics/curated_metrics.c
nick-enoent: This is leftover code, thanks @nichamon

ldms/src/sampler/curated_metrics/curated_metrics.c
morrone: There should be code comments to explain that. And you probably shouldn't do "s =", because that misleads the reader into thinking that you are planning to use the returned value for some purpose, but then that isn't used for anything.

ldms/src/sampler/curated_metrics/curated_metrics.c
morrone: Nichamon is probably right. But using this method of trying to infer initialization rather than explicitly stating it has led to bugs over time in other plugins in ldms.

I think it it would be best have a boolean in the instance context (curated_metric_t) that _explicitly_ states whether the instance is fully initialized or not.  Have the "initialized" boolean initialized to false at allocation time. Whenever the structure is considered initialized, set the value to true.

ldms/src/sampler/curated_metrics/curated_metrics.c
morrone: Looking at this again, it lookes like the configuration file is just a list of strings, right? There is precedent in other plugins to just use a simple text file for configuration, one string per line. That would be better too than using json, I think.

tom95858:
@nick-enoent I think this is ready to be converted from a draft...yes?

PR #1809, opened 2025-06-12
ovis_log: Improve the default logging format

morrone:
The default logging level and default mode are probably not the best default values. Probably the two most common ways that folks will run ldmsd is: on the command line when they first start playing with ldms, from systemd when ldmsd is using in production. In neither of these cases do genearlly want a timestamp to be prefixed to the log messages. For instance systemd logs to journald by default, and journald does its own timestamping. So right now when I use journalctl to list the ldmsd log, I see double time stamps like so:

```
Jun 12 12:17:42 hetchy1 ldmsd[2815186]: Thu Jun 12 12:17:42 2025: INFO      : Listening on sock:411 using `sock` transport and `munge` authentication
```
this is because on the command line ldmsd is doing:
```
$ ldmsd -F -c ./samp.conf
Thu Jun 12 12:12:41 2025:  WARNING: : The option `-F` is deprecated. ldmsd now ALWAYS runs in the foreground.
```
After we get rid of the timestamp (by default, folks who want it can always set an option to get timestamps) we have:

```
$ ldmsd -F -c ./samp.conf
  WARNING: : The option `-F` is deprecated. ldmsd now ALWAYS runs in the foreground.
```
It is weird to have double colons, so the code is updated to have "name" be NULL by default instead of an empty string. That way we can easier check if name is unset, and then conditionally print the name:

```
$ ldmsd -F -c ./samp.conf
  WARNING: The option `-F` is deprecated. ldmsd now ALWAYS runs in the foreground.
```

And in the course of updating all of that a few other fixes were made:

Change the default logging mode to OVIS_LOG_M_TS_NONE (no time string).

Introduce the environment varable OVIS_LOG_DATE_TIME to enable logging mode OVIS_LOG_M_DT (the previous default).

Call ovis_log_init() much sooner to get "ldmsd" set as the name in the default log level. This allows the message to be:

```
$ ldmsd -F -c ./samp.conf
  WARNING: ldmsd: The option `-F` is deprecated. ldmsd now ALWAYS runs in the foreground.
```

Introduce the ovis_log_default_set_level() function to allow ldmsd to change the log level of the default log _after_ calling ovis_log_init().

Raise the default log level to include OVIS_LWARNING.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -1730,12 +1730,14 @@ int ldmsd_process_cmd_line_arg(char opt, char *value)
 					"specified to %s. Ignore the new value %s\n",
 					ovis_log_level_to_str(ovis_log_get_level(NULL)), value);
 		} else {
-			is_loglevel_thr_set = 1;
 			log_level_thr = ovis_log_str_to_level(value);
 			if (log_level_thr < 0) {
+				ovis_log(NULL, OVIS_LERROR, "Unrecognized log level '%s'\n", value);
 				log_level_thr = OVIS_LERROR;
 				return EINVAL;
 			}
+			is_loglevel_thr_set = 1;
+			ovis_log_default_set_level(log_level_thr);
 		}
 		break;
 	case 'F':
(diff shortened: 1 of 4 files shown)


baallan:
@morrone It seems like there's probably some man page changes that go with this but not in evidence.

And I disagree that command-line users of ldms are not interested in time stamps. Especially users of aggregators who are trying to figure out when things are going on.

morrone:
> @morrone It seems like there's probably some man page changes that go with this but not in evidence.
> 
> And I disagree that command-line users of ldms are not interested in time stamps. Especially users of aggregators who are trying to figure out when things are going on.

I did not state that there is _no_ interest, only that it probably isn't the thing they want when they start playing with ldmsd, and it probably should not be the default behavior.

For command line use especially, one might argue that time stamps don't need to be built into ldmsd, because one can always pipe the output to the "ts" command and get much more control over the time stamp format.

But in this PR I am not removing timestamp support, only changing the default behavior.

tom95858:
@nichamon, could you please review this. My comments are:
- should the default logging mode should be _M_DT instead of _M_NONE (like it is now)
- should we leave the default log level at LERROR

Otherwise, I agree that it would be nice to be able to disable the timestamps that end up duplicated in syslog.

PR #1810, opened 2025-06-15
Make advertised producers subscribe to stream automatically

nichamon:


nichamon:
@tom95858 @narategithub I forgot to create a pull request for this change last week.

PR #1811, opened 2025-06-15
Fix the default value of rail, quota, rx_rate of advertised producers

nichamon:


nichamon:
@tom95858 I found this issue over the weekend. Currently, without the patch, prdcr_status has inconsistent behavior when reporting rail, quota, and rx_rate values:

- **When values ARE specified** in `prdcr_listen_add`: Reports correctly
- **When values are NOT specified** in `prdcr_listen_add`: Incorrectly reports as 0

The correct behavior for unspecified values should be:

- **quota and rx_rate**: Should display -1 (indicating unlimited)
   - **Active producer**: -1 (indicating unlimited)
   - **Passive producer**: Inherit the limits from the listening endpoint.
- **rail**: Should display different values depending on producer type:
   - **Active producer**: 1
   - **Passive producer**: Use the same rail size as specified on the peer (since the peer initiates the connection)

The patch fixes this inconsistency so that prdcr_status reports the proper default values instead of showing 0 for unspecified parameters.

PR #1812, opened 2025-06-16
Fix prdcr_listen path in parser_util.py

nichamon:
There is a typo 'rails' in the prdcr_listen path. The correct parameter is 'rail' in the singular form. Without the patch, ldmsd_yaml_parser will not include the 'rail' value in the prdcr_listen_add line.


nick-enoent (approved):
@nichamon Thanks for catching this!

PR #1820, opened 2025-06-17
fix reconnect/interval for prdcr examples on main

baallan:


baallan:
This ports https://github.com/ovis-hpc/ldms/commit/c60765b1f90e6fa12368638ce7cb9c6c106dd331 forward from 4.4

PR #1833, opened 2025-06-22
Free lookup_peer when receiving a set_delete reply

nichamon:


nichamon:
I have sniff tested the patch, but it should be tested more extensively, especially for a race condition between the set_delete path and the disconnection path.

tom95858:
@nichamon should this be marked as a draft?

nichamon:
> @nichamon should this be marked as a draft?

I changed it to 'ready' as @baallan confirmed that the leak in the lookup path identified by Valgrind was gone after applying this patch. But looking back at my comment, I realize I haven't actually done that additional race condition testing I mentioned between the set_delete and disconnection paths. Should I flip it back to draft until I can test that properly?

tom95858:
@nichamon, can we just accept this and then if you find a race we'll fix it in a different pull request?

nichamon:
> @nichamon, can we just accept this and then if you find a race we'll fix it in a different pull request?

Sounds good to me! Thank you for merging this.

Issue #1752, opened 2025-05-21
decomposition complains endlessly about unrecognized hashes


morrone:
There has been a change to decomposition on main that results in ldmsd endlessly reporting a decomposition error "2" each time it encounters an unrecgonized hash.

I think this may have been a side effect of commit 8008a0f7e4ea0111466f3993de40c60bc261654a that changed the flex decomposition's return code from 0 to ENOENT here:

```
# ldms/src/decomp/flex/decomp_flex.c:flex_decompose()
+       if (0 == n_decomp) {
+               /* does not match; apply the default */
+               if (!dcfg->default_decomps)
+                       return ENOENT;
```

And an example of the endless error messages:

```
Wed May 21 16:10:35 2025:    ERROR: store: decompose error: 2 for set 'hetchy1017/lslide-ffff99da367ba800'
Wed May 21 16:10:35 2025:    ERROR: store: decompose error: 2 for set 'hetchy1011/lslide-ffff9232c95fa000'
Wed May 21 16:10:35 2025:    ERROR: store: decompose error: 2 for set 'hetchy1014/lslide-ffff909be06b7800'
Wed May 21 16:10:35 2025:    ERROR: store: decompose error: 2 for set 'hetchy1008/lslide-ffff9ab532d80000'
Wed May 21 16:10:35 2025:    ERROR: store: decompose error: 2 for set 'hetchy1007/lslide-ffff9d39b9698800'
```

I'd like it to go back to the previous behavior (just ignore unrecognized hashs).

Alternatively, it might print a warning a single time for each unrecognized _hash_.

emdonat:
resolved in #1758

morrone:
I am reopening this issue. I am not really satisfied with changing the ERROR to a WARNING.  I can't have these warnings constantly streaming from all of the ldmsd, but I _do_ want normal warnings. And I don't want to have to enter hashes into my decomp configuration that I don't want converted into rows for the store.

So I still need some way to get back to the behavior we had on b4.4.

I would say that if you really want to make this a warning, then the code _needs_ to keep track of which hashes is has warned about, and only warn ONCE for each hash. Printing thousands of warnings on repeat is just not a reasonable thing to do at the level WARNING.

Alternatively, it can print at some higher log level like DEBUG.

emdonat:
Got it - @narategithub would it be acceptable to repush with LDEBUG?

tom95858:
@morrone, I think this is resolved. Can you confirm.

PR #1835, opened 2025-06-23
Change decomp_flex from WARN to DEBUG when no default rule

nichamon:
Address #1752


nichamon:
I believe the log level could be either INFO or DEBUG with two different reasons: users intentionally and unintentionally don't specify the default rule or a rule that handles the hash.

If it is intentional, INFO level is too high as they are aware that there are no rules to handle the hash. On the other hand, if it is by mistake, INFO or even WARN would deliver the message better.

I decided to use DEBUG in this patch because the DEBUG could be turned on if something feels off. INFO will spill the messages repeatedly. We can't turn off the log per messages.

PR #1837, opened 2025-06-24
Fix parsing brittleness in lustre_client

morrone:
This adds a test mode option for reading fake sample data without root access and two options to include new/supplementary data in the schema.

Forward port to main of PR #1692.

### DIFF
--- a/ldms/src/sampler/lustre_client/lustre_client_general.c
+++ b/ldms/src/sampler/lustre_client/lustre_client_general.c
@@ -62,6 +62,23 @@ static char *llite_stats_uint64_t_entries[] = {
         NULL
 };
 
+
+/* the following get added to the schema if extra215 config option is present. */
+static const char *extra215_llite_stats_uint64_t_entries[] = {
+        "read",
+        "write",
+        "opencount",
+        "openclosetime",
+        NULL
+};
+
+/* the following exist at least as far back as 2.15 lustre clients */
+static const char *extratimes_llite_stats_uint64_t_entries[] = {
+	"start_time", /* this truncates the data to seconds; it's ok. */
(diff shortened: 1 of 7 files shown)


ldms/src/sampler/lustre_client/lustre_client_general.c
```diff
@@ -62,6 +62,23 @@ static char *llite_stats_uint64_t_entries[] = {
         NULL
 };
 
+
+/* the following get added to the schema if extra215 config option is present. */
+static const char *extra215_llite_stats_uint64_t_entries[] = {
```
baallan, line 67: @morrone  we have additionally (per lustre source) that I just overlooked in the b4.4 attempt at this.
```
fallocate
pcc_attach
pcc_detach
pcc_auto_attach
pcc_hit_bytes
pcc_attach_bytes
hybrid_noswitch
hybrid_writesize_switch
hybrid_readsize_switch
hybrid_read_bytes.sum
hybrid_write_bytes.sum
```
Now would be a good time to add them to extra215_llite_stats_uint64_t_entries, as I doubt very much anyone has yet deployed the sampler with extra215 enabled.
It would be nice if 4.5 lustre_client sampler was ready out of the box instead of needing another version for recent lustres.

morrone: @baallan OK, I'll add those.

What is the idea behind adding start_time and elapsed_time? I'm not sure how those would be useful in a monitoring database.

baallan: start_time is the base for various cumulative stats, which may be in itself interesting if it changes only when a device is started/mounted. As it may not be of interest to some (since we historically left it out or it wasn't there in the dawn of lustre) I made it optional.

Issue #1840, opened 2025-06-26
Bug in dcgm_sampler only works if config sets  use_base=1


jstile-lbl:
In v4.5.1-rc.1, a check was added to check dcgm_sampler config line for for **use_base** = [0|1].

If I do not set use_base=1 in the config, I end up with an uninitialized 

```
    ...
    DEBUG: xprt.ldms: __send_lookup_reply(): x 0x7f674c010e90: sharing ... remote lookup ctxt 0x7fa8506c1c90
    DEBUG: xprt.ldms: __send_lookup_reply(): x 0x7f674c010e90: sharing ... remote lookup ctxt 0x7fa8506c2830
    DEBUG: xprt.ldms: __send_lookup_reply(): x 0x7f674c010e90: sharing ... remote lookup ctxt 0x7fa8506c0de0
    DEBUG: xprt.ldms: __send_lookup_reply(): x 0x7f674c010e90: sharing ... remote lookup ctxt 0x7fa8506c1460
    DEBUG: xprt.ldms: __send_lookup_reply(): x 0x7f674c010e90: sharing ... remote lookup ctxt 0x7fa8506c2d50
    DEBUG: xprt.ldms: __send_lookup_reply(): x 0x7f674c010e90: sharing ... remote lookup ctxt 0x7fa8506c2ff0
    DEBUG: sampler.dcgm_sampler.dcgm_sampler: dcgm_sampler sample() called
Segmentation fault (core dumped)
```

```
gdb -c core.ldmsd_wkr_0.1977737.nid001005 $(which ldmsd)
...
Core was generated by `/opt/ovis-ldms/sbin/ldmsd -x sock:6002 -c /opt/ovis-ldms/etc/ldms/nersc-ldmsd.s'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f677dd0e207 in gpu_sample () at dcgm_sampler.c:367
367             ldms_set_t set_old = base->set; /* this should be null */
[Current thread is 1 (Thread 0x7f677e556700 (LWP 1977741))]
Missing separate debuginfos, use: zypper install ovis-ldms-debuginfo-4.5.1_rc.1-1.x86_64
(gdb) 
(gdb) dir /builds/nersc/csg/ovis/ldms/src/sampler/dcgm_sampler
Source directories searched: /builds/nersc/csg/ovis/ldms/src/sampler/dcgm_sampler:$cdir:$cwd
(gdb) tb
[...]

PR #1844, opened 2025-06-26
fix base pointer misuse in dcgm_sampler no use_base configuration

baallan:
this also adds the simple integration test case for the no sampler_base configuration of dcgm_sampler.


baallan:
fixes #1840

Issue #1846, opened 2025-06-30
ldms/python/ldms.pyx requires python-3.8 feature in threads module


jstile-lbl:
SLES 15-sp5 has python 3.6

```
cat /etc/os-release 
NAME="SLES"
VERSION="15-SP5"
VERSION_ID="15.5"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp5"
DOCUMENTATION_URL="https://documentation.suse.com/
"
```

ldms/python/ldms.pyx calls **threading.get_native_id()** function, introduced in Python 3.8. 

It provides the operating system's native thread ID (TID), which is distinct from the Python-assigned thread identifier obtained via threading.get_ident().

On a system that doesn't offer a native thread ID, one could default to what is available.

Possible code change:

```
diff -Naru ./ldms/python/ldms.pyx ./ldms/python/ldms.pyx.b
--- ./ldms/python/ldms.pyx      2025-06-25 07:32:30.832329224 -0700
+++ ./ldms/python/ldms.pyx.b    2025-06-30 14:59:17.724292319 -0700
@@ -4353,7 +4353,10 @@
         uid = ev.recv.cred.uid
         gid = ev.recv.cred.gid
         perm = ev.recv.perm
-        tid = threading.get_native_id()
+        if sys.version_info >= (3, 8):
+            tid = threading.get_native_id()
+        else:
+            tid = threading.get_ident()
         obj = MsgData(name, src, tid, uid, gid, perm, is_json, data,
                          raw_data,
                          ldms.ldms_msg_type_e(ev.recv.type))
```

This worked for me

I was able to inject a messages

```
cat ./ldms_msg_publish.py
#!/usr/bin/env python3
import sys
sys.path.append('/opt/ovis-ldms/lib/python3.6/site-packages')
from ovis_ldms import ldms
ldms.init(16*1024*1024)
x = ldms.Xprt("sock", "munge")
x.connect("localhost", "6002")
x.msg_publish("nersc", "This is a test")
```

And read the message

```
cat ./ldms_msg_subscribe.py
#!/usr/bin/env python3

import time
import sys
sys.path.append('/opt/ovis-ldms/lib/python3.6/site-packages')
from ovis_ldms import ldms

mc = ldms.MsgClient(".*", True)

[...]

PR #1847, opened 2025-06-30
https://github.com/ovis-hpc/ldms/issues/1846

jstile-lbl:
ldms/python/ldms.pyx requires the threading.get_native_id() function, which is available in Python 3.8 and newer.

For older Python versions (e.g., 3.6 on SLES15sp5), use threading.get_ident() instead.

This patch adds compatibility for these older Python versions.

### DIFF
--- a/ldms/python/ldms.pyx
+++ b/ldms/python/ldms.pyx
@@ -4353,7 +4353,11 @@ cdef class MsgData(object):
         uid = ev.recv.cred.uid
         gid = ev.recv.cred.gid
         perm = ev.recv.perm
-        tid = threading.get_native_id()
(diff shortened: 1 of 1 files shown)


morrone (changes requested):
The first line of the commit message needs to be a short, descriptive comment rather than a URL. Otherwise, it seems reasonable to me.

morrone:
And FYI, you can add a line at the end of your commit message like this:

```
Fixes #1846
```

That automatically links the PR to the issue, and will auto-close the issue when the PR lands. Here's some info on that:

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

Issue #1815, opened 2025-06-16
The lustre plugin (Plugin_lustre_ost and Plugin_lustre_mdt) using incorrect path for OSDs


punjabimunda47:
I've setup the lustre_ost and lustre_mdt plugins correctly based on their documentation:
example:
```
load name=lustre_ost
config name=lustre_ost
start name=lustre_ost interval=1000000
```

In the logs, it shows the plugin is having trouble finding the osd metrics:
```
WARNING   : lustre_ost unable to open /proc/fs/lustre/osd-zfs/lustre-OST0000/filesfree
WARNING   : lustre_ost unable to open /proc/fs/lustre/osd-zfs/lustre-OST0000/filestotal
WARNING   : lustre_ost unable to open /proc/fs/lustre/osd-zfs/lustre-OST0000/kbytesavail
WARNING   : lustre_ost unable to open /proc/fs/lustre/osd-zfs/lustre-OST0000/kbytesfree
WARNING   : lustre_ost unable to open /proc/fs/lustre/osd-zfs/lustre-OST0000/kbytestotal
```

Manually searching in the filesystem I can see the files are actually located in `/sys/fs/lustre`:
```
[root@lustre-oss-00 ~]# ll /sys/fs/lustre/osd-zfs/lustre-OST0000/*
...
 /sys/fs/lustre/osd-zfs/lustre-OST0000/filesfree
 /sys/fs/lustre/osd-zfs/lustre-OST0000/filestotal
...
 /sys/fs/lustre/osd-zfs/lustre-OST0000/kbytesavail
 /sys/fs/lustre/osd-zfs/lustre-OST0000/kbytesfree
 /sys/fs/lustre/osd-zfs/lustre-OST0000/kbytestotal
...
```
From the docs, the plugin only searches in `/proc/fs/lustre`

I'm testing this on a virtualized lustre with two mdts and two osts.

ldms_ls'ing the sampler gives me some info, though as expected, some is missing:
```
[root@lustre-oss-00 ~]# ldms_ls -x sock -p 10001 -l                                                                                                       
lustre-oss-00.local/lustre_ost/lustre-OST0000: consistent, last update: Mon Jun 16 15:03:05 2025 -0600 [2844us]                                           
M u64        component_id                               0                                                                                                 
M char[]     fs_name                                    "lustre"                                                                                          
M char[]     ost                                        "lustre-OST0000"                                                                                  
D u64        read_bytes.sum                             104857600                                                                                         
[...]

PR #1849, opened 2025-07-02
Update lustre_mdt/lustre_ost samplers to handle new lustre data paths

morrone:
Update the lustre_mdt and lustre_ost samplers to handle the change in metrics directory locations in newer Lustre releases.

Also merge the lustre_mdt, lustre_ost, and lustre_client subdirectories into one "lustre_newer" subdirectory. Then start sharing some code between the three samplers.


baallan:
@morrone I'm guessing that this PR is supposed to land after #1848 (due to the zillion overlaps)?

baallan:
Since there's a lot of refactoring going on, can you expand (perhaps with a common config option name/implementation) the other affected lustre plugins to include test_path similar to lustre_client?

baallan:
@morrone Does this PR fix #1815? (I suspect so)

morrone:
> @morrone I'm guessing that this PR is supposed to land after #1848 (due to the zillion overlaps)?

There are a few minor conflicts, which is why this PR is based on that PR's commit. If this lands first, it'll have that commit and the other PR will be a no-op. But generally, yes, we would land PR #1848 first.

> Since there's a lot of refactoring going on, can you expand (perhaps with a common config option name/implementation) the other affected lustre plugins to include test_path similar to lustre_client?

We could could that in a follow-on PR. I am not going to be able to get to that in this one.

> @morrone Does this PR fix #1815? (I suspect so)

Yes. The "lustre: search for osd dir in multiple places" commit has "Fixes #1815" in the commit comment.  github has automatically made a note of that connection in PR #1815. When this lands, github will also automatically close that commit because of the "Fixes" annotation.

tom95858:
@morrone, can you please resolve these conflicts. I think #1848 broke you.

PR #1851, opened 2025-07-07
pll-ldms-static-test/ldms-static-test new features and cases

baallan:
This adds portability and checking to pll-ldms-static-test, yaml to ldms-static-test, doc updates, and new cases derived from testing 4.5.1.

### DIFF
--- a/ldms/scripts/Makefile.am
+++ b/ldms/scripts/Makefile.am
@@ -17,6 +17,17 @@ bin_SCRIPTS += ovis-roll-over.py
 bin_SCRIPTS += ldms-csv-anonymize
 bin_SCRIPTS += ldms-csv-export-sos
 bin_SCRIPTS += ldmsd-check-env
+bin_PROGRAMS = ldms_cpu_grind
+bin_PROGRAMS += ldms_barrier
+
+ldms_cpu_grind_SOURCES = ldms_cpu_grind.c
+ldms_cpu_grind_LDADD = -lm
+#ldms_cpu_grind_LDFLAGS = $(AM_LDFLAGS) -lpthread
+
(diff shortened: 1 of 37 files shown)


baallan:
There are no changes to existing ldms-related python or C sources.

tom95858:
@baallan, @emdonat think the testing is great, however, I'm not sure the unit test infrastructure is appropriate for scalability testing. We have the ldms-test infrastructure for that and it provides for the distributed collection and reporting of test results.

One minor nit is that there are two empty files checked in. What do you think?

baallan:
The empty files is part of how it works/detects misconfiguration.
It's an integration test structure, not really a unit test structure.
Unfortunately, even if supported, we can't automatically publish test results to outside snl.
Does tada support slurm and bare metal? this does (and has for years: we're just adding more result sanity checks, synchronization, and more cases today).

tom95858:
> The empty files is part of how it works/detects misconfiguration. It's an integration test structure, not really a unit test structure. 

Ok, thanks for the explanation.

> Unfortunately, even if supported, we can't automatically publish test results to outside snl.

Of course, TADA does not publish "outside."

> Does tada support slurm and bare metal? this does (and has for years: we're just adding more result sanity checks, synchronization, and more cases today).

I think it's fantastic to add more testing. I'm really thinking about how we move forward so that our scalability tests are a push-button instead of a fire drill. Given that one of your commits said "scalable" it tweaked this nerve.

baallan:
@tom95858 yes, we're working on automating Jim's launches and checks with portability to any slurm (and eventually flux) installation of any size. implementing his eyeball norms is nontrivial, but the recent expansions of the status commands coverage are helping.

PR #1857, opened 2025-07-15
Remove LDMS_MSG_LAST and LDMS_MSG_EVENT_LAST

nick-enoent:
Add ldms_msg_publish and ldms_msg_subscribe command

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -6132,9 +6132,9 @@ const char *ldms_msg_type_sym(ldms_msg_type_t t)
 		[LDMS_MSG_AVRO_SER] = "LDMS_MSG_AVRO_SER",
 	};
 
-	if (t < LDMS_MSG_LAST)
+	if (t <= LDMS_MSG_AVRO_SER)
 		return tbl[t];
-	return "UNKNOWN";
+	assert(NULL=="Invalid ldms_msg type");
 }
 
 const char *ldms_msg_event_type_sym(enum ldms_msg_event_type t)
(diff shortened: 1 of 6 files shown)


baallan:
Why do we need to get rid of these? Do we have something against writing iterations over these enums with an endpoint that doesn’t change even if the enum expands? @nick-enoent

baallan:
> Why do we need to get rid of these? Do we have something against writing iterations over these enums with an endpoint that doesn’t change even if the enum expands? @nick-enoent

If the issue is extra cases, the fix for that is enum element x_last=x_avro and document x last is sentinel that changes to new final value when enum expands.

tom95858:
> Why do we need to get rid of these? Do we have something against writing iterations over these enums with an endpoint that doesn’t change even if the enum expands? @nick-enoent

Because it forces the programmer to add cases to their switch statements for events that will never occur to avoid the 'missing enum in switch statement' warnings. Specifically,

``` C
switch (ev->data) {
    ...
    case LDMS_MSG_EVENT_LAST:   /* By the way this can never occur */
        break;
}
```

If you feel strongly that we should have these, then we could add something like this:

``` C
enum ldms_msg_event_e {
    ...
    LDMS_MSG_EVENT_WHICHEVER_EVENT_IS_LAST,
    LDMS_MSG_EVENT_LAST = LDMS_MSG_EVENT_WHICHEVER_EVENT_IS_LAST
};
```

nick-enoent:
> Why do we need to get rid of these? Do we have something against writing iterations over these enums with an endpoint that doesn’t change even if the enum expands? @nick-enoent

LDMS_MSG_LAST is not a type that's used, and LDMS_MSG_EVENT_LAST is not an event that will occur. They also require additional cases in switch statements that utilize the ldms_msg protocol.

ldms/src/core/ldms.c
```diff
@@ -6145,9 +6145,10 @@ const char *ldms_msg_event_type_sym(enum ldms_msg_event_type t)
 		[LDMS_MSG_EVENT_SUBSCRIBE_STATUS] =
 			"LDMS_MSG_EVENT_SUBSCRIBE_STATUS",
 		[LDMS_MSG_EVENT_UNSUBSCRIBE_STATUS] =
-			"LDMS_MSG_EVENT_UNSUBSCRIBE_STATUS",
+			"LDMS_MSG_EVENT_UNSUBSCRIBE_STATUS"
 	};
-	if (t < LDMS_MSG_EVENT_LAST)
+	if (t <= LDMS_MSG_EVENT_CLIENT_CLOSE) {
```
tom95858, line 6150: @nick-enoent please change this to:
```C
if (t < sizeof(tbl) / sizeof(tbl[0])) {
```
If we do it that way, we'll never need to change this code if we add another event.

PR #1868, opened 2025-07-23
Disable the ldms_msg API by default

tom95858:
This change adds a new command 'msg_enable' that will enable the LDMS message service. It also changes the behavior in that the msg service can be enabled after it has been disabled. This used to be prohibited, however, the service is now disabled by default, therefore it must be permitted to enable the disabled message service.

Also added are configuration MSG_ENABLE_REQ and API ldms_msg_enable() and associated logic.

The msg_stats and msg_client_stats commands will not return data when the service is disabled and instead report an error indicating the the LDMS message service is disabled.


tom95858:
@morrone could you take a quick spin through this and make certain we've got everything we think we need?

morrone:
> @morrone could you take a quick spin through this and make certain we've got everything we think we need?

Looks reasonable to me.

Issue #1869, opened 2025-07-23
Bug in Python check version during installation


Logan590:
The Python check version return something like "Python 3.12 older than Python 3.6". There is no issue with python 3.8. My guess is that the check is doing this kind of comparison:
 3.12 < 3.6 => not ok
but 
3.8 > 3.6. => ok.

morrone:
Assuming addressed by #1870. @Logan590 please open a new ticket with more specific error message if the latest on `main` is still giving you trouble.

PR #1870, opened 2025-07-23
Fix ax_python_module_version.m4 file

### LINKED ISSUE #1854: Python3.12 incompatibility with ax_python_module_version.m4
Hi, the distutils module is not supported in Python 3.12+, and therefore, to conduct the argparse version check, the python cmd need to be adjusted from distutils.version: StrictVersion to packaging.version: Version.

```patch
diff --git a/m4/ax_python_module_version.m4 b/m4/ax_python_module_version.m4
index 0ffb7b09..fcd35bd3 100644
--- a/m4/ax_python_module_version.m4
+++ b/m4/ax_python_module_version.m4
@@ -28,7 +28,7 @@
 AC_DEFUN([AX_PYTHON_MODULE_VERSION], [
     AX_PYTHON_MODULE([$1], [required], [$3])
     AC_MSG_CHECKING([for version $2 or higher of $1])
-    $PYTHON -c "import sys, $1; from distutils.version import StrictVersion; sys.exit(StrictVersion($1.__version__) < StrictVersion('$2'))" 2> /dev/null
+    $PYTHON -c "import sys, $1; from packaging.version import Version; sys.exit(Version($1.__version__) < Version('$2'))" 2> /dev/null
     AS_IF([test $? -eq 0], [], [
         AC_MSG_RESULT([no])
         AC_MSG_ERROR([You need at least version $2 of the $1 Python module.])
```

I'm reporting it here for recording and reminding purposes.  I'll devise a solution (unless someone else wants to take a whack at it) as soon as possible.

jennfshr:
Closes #1854 

Condition based on 3.12+ python versions to use `from packaging.version import Version` in detecting module version since distutils is not supported in 3.12 python.

This may also resolve the issue #1869, but I'd need to see the config.log of the error encountered to know that for sure.

### DIFF
--- a/m4/ax_python_module_version.m4
+++ b/m4/ax_python_module_version.m4
@@ -28,10 +28,31 @@
 AC_DEFUN([AX_PYTHON_MODULE_VERSION], [
     AX_PYTHON_MODULE([$1], [required], [$3])
     AC_MSG_CHECKING([for version $2 or higher of $1])
-    $PYTHON -c "import sys, $1; from distutils.version import StrictVersion; sys.exit(StrictVersion($1.__version__) < StrictVersion('$2'))" 2> /dev/null
+
+    python_fullversion=`${PYTHON} -c "import sys; print(sys.version)" | sed q`
+    python_majorversion=`echo "$python_fullversion" | sed '[s/^\([0-9]*\).*/\1/]'`
+    python_minorversion=`echo "$python_fullversion" | sed '[s/^[0-9]*\.\([0-9]*\).*/\1/]'`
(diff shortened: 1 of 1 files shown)


morrone:
I am not necessarily opposed to this. But I wanted to point out that the only place this macro is used is here:

```
AX_PYTHON_MODULE_VERSION(["argparse"], 1.1, "python")
```

And that looks like it may be left over from the days before argparse was in necessarily in python standard library. With argparse in the standard library, its version might not ever change again.

Argparse was added to the standard library in 3.2, and we require 3.6. So I think we should now remove that AX_PYTHON_MODULE_VERSION test for the argparse version.

PR #1873, opened 2025-07-25
memory misuse fixes to ldmsd_updtr.c round 1

baallan:
fixing leaks or initialization problems.


baallan:
@tom95858: updated commit message and description

PR #1874, opened 2025-07-25
memory misuse fixes to netlink-notifier.c round 1

baallan:
remove leak due to shadow variable.


baallan:
@tom95858 : updated commit message and description

PR #1875, opened 2025-07-25
memory misuse fixes to ldmsd_stream.c round 1

baallan:
fix leak

### DIFF
--- a/ldms/src/ldmsd/ldmsd_stream.c
+++ b/ldms/src/ldmsd/ldmsd_stream.c
@@ -603,7 +603,7 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
 {
 	int rc, stream_type;
 	size_t data_len;
-	static char *buffer;
+	static char *buffer; /* why is buffer static? not reused */
 	size_t max_msg_len;
 	ldms_t x;
 	size_t cnt;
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -603,7 +603,7 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
 {
 	int rc, stream_type;
 	size_t data_len;
-	static char *buffer;
+	static char *buffer; /* why is buffer static? not reused */
```
tom95858: We should remove that comment. If we know it's not reused, why add that in a comment and leave the bug there?

ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -775,7 +777,9 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
 close_xprt:
 	ldms_xprt_close(x);
 err:
+	free(b);
 	free(buffer);
+	buffer = NULL;
```
tom95858: So this is only here if we assume that the buffer pointer is static, which we have already decided it should not be.

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_stream.c
baallan: fixed.

ldms/src/ldmsd/ldmsd_stream.c
baallan: removed 'static' and comment.

baallan:

PR #1876, opened 2025-07-25
memory misuse fixes to ldms_xprt.c round 1

baallan:


ldms/src/core/ldms_xprt.c
```diff
@@ -3131,8 +3136,13 @@ static void ldms_zap_cb(zap_ep_t zep, zap_event_t ev)
 						"Memory allocation failure.\n");
 				return;
 			}
-			if (zap_thrstat_ctxt_set(zep, thrstat, __thrstats_reset))
+			if (zap_thrstat_ctxt_set(zep, thrstat, __thrstats_reset)) {
 				free(thrstat);
+				ovis_log(xlog, OVIS_LCRIT,
```
tom95858: This logic introduces a race condition, multiple free, etc... The problem is that the logic was written assuming that  a given endpoint was only ever serviced on one thread, however, this is not true. Before the state of the transport is established and authentication is done, the endpoint is still assigned to the listening thread. This means that the thread stat assignment, etc... is potentially racing with another thread. This is definitely not a critical error, it is expected. You can see it trivially if you do 'nc -z localhost 10001'.

baallan:
@tom95858 : updated the commit message.
@narategithub : the pr initially failed on a compatibility test timeout in github (maybe gh infrastructure problem); if that persists, I'll need you to look at if the compatibility test itself is correct-- nothing changed here should break a correct test.

baallan:
@narategithub, @nichamon  it looks like the compatibility test is hanging. I don't see how anything in this PR should affect that, though. Please take a look.

baallan:
but maybe there's a locking thing i'm missing in ldms_zap_cb?

ldms/src/core/ldms_xprt.c
```diff
@@ -3131,8 +3136,13 @@ static void ldms_zap_cb(zap_ep_t zep, zap_event_t ev)
...
+                                                "zap_thrstat_ctxt_set failed.\n");
+				thrstat = NULL;
+				return;
```
nichamon: @baallan The workflow test failed due to the `return` here. The workflow was timed out because `ldms_ls` hang. `ldmsd` didn't send either `accept` or `reject` to `ldms_ls`.

I would like to suggest to remove the change here to move the pull request forward with the other fixes. I believe @tom95858 is working on the thrstat` allocation in the `ldms_zap_cb` path.

baallan: will do that.

baallan:
@tom95858 I removed the change in ldms_zap_cb per Mon. This is ready to go afaict.

baallan:
@tom95858 are you waiting on a review from @narategithub , or was @nichamon look sufficient?

ldms/src/core/ldms_xprt.c
```diff
@@ -3921,8 +3926,9 @@ int __ldms_remote_lookup(ldms_t _x, const char *path,
 				sizeof(struct ldms_request) + sizeof(*ctxt),
 				LDMS_CONTEXT_LOOKUP_REQ,
 				cb, arg, lu_path, flags);
+	free(lu_path);
```
tom95858: @baallan, I think this introduces a use-after-free bug. The lu_path is stuffed into the ctxt at ldms_xprt.c:349. I don't see where this path is copied. That's why the free was protected by the ctxt check.

ldms/src/core/ldms_xprt.c
```diff
@@ -4724,8 +4730,10 @@ struct ldms_thrstat_result *ldms_thrstat_result_get(uint64_t interval_s)
 
 	lres = calloc(1, sizeof(*lres) +
 		zres->count * sizeof(struct ldms_thrstat_result_entry));
-	if (!lres)
+	if (!lres) {
+		zap_thrstat_free_result(zres);
 		goto out;
+	}
```
tom95858, line 4736: It seems like it would be cleaner to just move the zres = zap_thrstat... allocation to after this check and avoid the free in the lres memory allocation failure path.

ldms/src/core/ldms_xprt.c
baallan: @tom95858  I didn't catch the consumption of the vararg. Will delete this change.

ldms/src/core/ldms_xprt.c
baallan: @tom95858 lres depends on zres->count; not sure how I could make the shift you suggest.

tom95858: @baallan, you're right Ben. That change is fine.

ldms/src/core/ldms_xprt.c
```diff
@@ -3922,7 +3928,6 @@ int __ldms_remote_lookup(ldms_t _x, const char *path,
 				LDMS_CONTEXT_LOOKUP_REQ,
 				cb, arg, lu_path, flags);
 	if (!ctxt) {
-		free(lu_path);
```
tom95858: @baallan, Ii think this change is incorrect. 

The lu_path is a strdup() of the path argument. This is passed to __ldms_remote_lookup() which takes this pointer and stuffs it in the ctxt structure.

So far, so good, however, if the calloc() of the ctxt structure fails, the lu_path is left unused and leaks the lu_path allocation. This is why the if (!ctxt) path frees lu_path.

This change also falls under the category of "if this fails the process is dying anyway because we're out of memory. In other words, it doesn't matter, other than quieting the tool.

That said, and out of curiosity, what tool gave you the advice that this needed to be removed, because I  believe that advice is incorrect.

ldms/src/core/ldms_xprt.c
```diff
@@ -1998,11 +2004,11 @@ void __process_dir_reply(struct ldms_xprt *x, struct ldms_reply *reply,
 	if (!ctxt->dir.cb)
 		return;
 
+	(void)clock_gettime(CLOCK_REALTIME, &start);
```
tom95858, line 2007: This looks good.

ldms/src/core/ldms_xprt.c
baallan: @tom95858 reviewing this on in the tool, i had marked it false positive because the tool didn't decode (because of the varargs) that lu_path was stored conditionally in the alloc function.  So i must have accidentally deleted a line here.
will put it back.

baallan: @tom95858 updated for the lu_path problem.

baallan:
@tom95858 leaving this in 4.5.1 as it fixes ldms_xprt_event_type_to_str which can affect messages from core or plugins.
rest of the nits fixed are unlikely to be issues in production afaict.

PR #1877, opened 2025-07-25
  fix potential use-after-free in ldms_msg_subscribe path

baallan:
fix potential use-after-free in ldms_msg_subscribe path.
also fixed potential missing va_end in ldsm_msg_publish path


baallan:

PR #1878, opened 2025-07-25
fix ldms_ls.c for case of ldms_record having no members (card()==-1)

baallan:


baallan:
removed from 451 milestone as nothing in the release testing actually tests this case.

PR #1879, opened 2025-07-25
fix doc bug in ldms.h

baallan:


ldms/src/core/ldms.h
```diff
@@ -3690,7 +3690,8 @@ extern const char *ldms_metric_type_to_str(enum ldms_value_type t);
 /**
  * \brief Get a metric type primitive base
  *
- * Returns the scalar ldms_value_type related to t.
+ * Returns the scalar ldms_value_type related to t, or LDMS_V_NONE
+ * when t is structure type (list, etc).
```
tom95858: etc... or what?

baallan: @tom95858 updated issue title and doc comment and updated commit message

PR #1882, opened 2025-07-25
stop dstring from expanding to unnecessary huge size

baallan:
in corner case, dstring buffer could get reallocated to 2Gb unnecessarily.


baallan:
@tom95858  updated commit message

PR #1883, opened 2025-07-25
 fix leaks in config() of cxi_sampler.c

baallan:


baallan:

ldms/src/sampler/cxi_sampler/cxi_sampler.c
```diff
@@ -849,9 +849,13 @@ static int config(ldmsd_plug_handle_t handle,
 		goto err;
 	}
 
+	free(tel_counters);
```
morrone, line 866: For this to be safe, don't wee need to properly initialize these variables too?

And when tel_counters is the result of av_value() rather than get_counter_names_from_file(), then this isn't the right thing to do, correct? More cleanup seems to be needed here.

baallan: yep. fixing.

baallan:
@morrone I updated to handle the av_value cases I missed before.

PR #1884, opened 2025-07-25
declaration of buffer w in app_sampler.c was insufficient; active if uid captured

baallan:
if uid is included in the data to be collected, a buffer write overrun by 1 may occur.

### DIFF
--- a/ldms/src/sampler/app_sampler/app_sampler.c
+++ b/ldms/src/sampler/app_sampler/app_sampler.c
@@ -968,7 +968,7 @@ int __line_uid(app_sampler_inst_t inst, ldms_set_t set, const char *linebuf,
 	/* populate `status_uid` and/or `status_*username` if changed. */
 	int n;
 	uint64_t x[4];
-	char w[4][20] = { {'\0'}, {'\0'}, {'\0'}, {'\0'} };
+	char w[4][21] = { {'\0'}, {'\0'}, {'\0'}, {'\0'} };
 	n = sscanf(linebuf, "%" SCNu64 " %" SCNu64 " %" SCNu64 " %" SCNu64,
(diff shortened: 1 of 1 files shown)


tom95858:
@baalllan could you please provide a bit more descriptive commit messages? It doesn't have to be a treatise, but "Fix memory misuse ... round 1" is not as helpful as it could be when we review the commit log later. Just something like "Static declaration of buffer w[][] in app_sampler  was insufficient." This comment applies to the last few pull requests. I think this work is great, but if you tweak the commit message, it would be even better.

baallan:
I will see what i can do; as you noticed, there were so many that i caved in to using automation with a unique (for patch tracking) but not highly detailed message.
________________________________
From: Tom Tucker ***@***.***>
Sent: Thursday, July 24, 2025 9:17 PM
Cc: Allan, Ben Andrew ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [ovis-hpc/ldms] memory misuse fixes to app_sampler.c round 1 (PR #1884)

[https://avatars.githubusercontent.com/u/7026652?s=20&v=4]tom95858 left a comment (ovis-hpc/ldms#1884)<https://github.com/ovis-hpc/ldms/pull/1884#issuecomment-3116246792>

@baalllan could you please provide a bit more descriptive commit messages? It doesn't have to be a treatise, but "Fix memory misuse ... round 1" is not as helpful as it could be when we review the commit log later. Just something like "Static declaration of buffer w[][] in app_sampler was insufficient." This comment applies to the last few pull requests. I think this work is great, but if you tweak the commit message, it would be even better.

tom95858:
@baallan, your automation may be faulty. The tools are not always correct. Some of these patches introduce bugs.

baallan:

PR #1885, opened 2025-07-25
 minimal fix to zap_event_str in zap.c (change range check)

baallan:
Valid zap events range from 1 to zap_event_last-1.

 Changing enum define to use  ZAP_EVENT_LAST=ZAP_EVENT_SEND_COMPLETE
 might be a good idea, but may touch other code or tests of inter-version enum compatibility

### DIFF
--- a/lib/src/zap/zap.c
+++ b/lib/src/zap/zap.c
@@ -201,7 +201,7 @@ enum zap_err_e zap_errno2zerr(int e)
 
 const char* zap_event_str(enum zap_event_type e)
 {
-	if ((int)e < 0 || e > ZAP_EVENT_LAST)
+	if ((int)e < 1 || e >= ZAP_EVENT_LAST)
 		return "ZAP_EVENT_UNKNOWN";
 	return __zap_event_str[e];
 }


lib/src/zap/zap.c
```diff
@@ -201,7 +201,7 @@ enum zap_err_e zap_errno2zerr(int e)
 
 const char* zap_event_str(enum zap_event_type e)
 {
-	if ((int)e < 0 || e > ZAP_EVENT_LAST)
+	if ((int)e < 0 || e >= ZAP_EVENT_LAST)
```
tom95858: ZAP_EVENT_LAST is not a valid event, it can never occur. It should not appear in our event naming arrays or any other debug info. This kind of thing should be:
```C
if (index < sizeof(array)/sizeof(array[0]))
    return array[index]
```
Adding invalid enum values to an enum definition is not good practice in my opinion.

tom95858: Your proposed change demonstrates why this code pattern is not a good one.

baallan:
@tom95858 ZAP_EVENT_LAST was already there; i just shrank the range of the array lookup code by changing the operator. It sounds like we should revise the enum definition so that ZAP_EVENT_LAST=\<last valid enum tag\> (as we have elsewhere, but not consistently).

I'm fine with making that revision, but i'm loathe to touch the zap API (visible enum) without someone telling me to. I don't know what enum compatibility checks will break where.  Might be better to hand this off to @narategithub. Might be good to apply the _LAST=\<last valid\> to all our enums that might ever reasonably expand (get rid of the 'invalid' _LAST items).

The sizeof array solution fails when someone expands the enum but not the array -- this has turned up in several current places by static array bounds checking.

PR #1886, opened 2025-07-25
memory leak fixes to util.c round 1

baallan:


baallan:
removed from milestone as av_add not used in anything that is tested.

PR #1888, opened 2025-07-25
memory misuse fixes to test_sampler.c round 1

baallan:


baallan:

PR #1890, opened 2025-07-25
    fix stream_csv_store.c potential per-message memory leak

baallan:
- _get_header_from_data can leak memory if user stream json data is misformatted.
    - open_streamstore can leak file descriptor during plugin config()


baallan:
@tom95858 i updated commit and description for this issue

PR #1892, opened 2025-07-25
 fix ovis_log leaks/wild pointers in log_register and open      

baallan:
- fix uninit pointer use in ovis_log_register error paths
    - fix memory ovis_log_level_to_str leak in ovis_log_list
    - fix unclosed files in __log_open error paths.

### DIFF
--- a/lib/src/ovis_log/ovis_log.c
+++ b/lib/src/ovis_log/ovis_log.c
@@ -405,7 +405,7 @@ ovis_log_t ovis_log_register(const char *subsys_name, const char *desc)
 		errno = EEXIST;
 		return NULL;
 	}
-	log = malloc(sizeof(*log));
+	log = calloc(1, sizeof(*log));
 	if (!log) {
 		errno = ENOMEM;
 		return NULL;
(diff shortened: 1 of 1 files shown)


lib/src/ovis_log/ovis_log.c
```diff
@@ -469,15 +469,16 @@ static int __get_log_info(ovis_log_t l, struct ovis_log_buf *buf)
...
+		level_s = strdup("default");
 	else
 		level_s = ovis_log_level_to_str(l->level);
 
 	rc = __buf_append(buf, "{\"name\":\"%s\","
 			        "\"desc\":\"%s\","
 			        "\"level\":\"%s\"}",
 				l->name, l->desc,
-				level_s);
+				level_s ? level_s : "out_of_memory");
```
tom95858, line 480: So your expectation is that this buffer append will succeed after a previous memory allocation has just failed?

baallan: The buffer that the is being appended here is largish to start with, and unlikely to require reallocation while being appended with the small format string in use here.

PR #1894, opened 2025-07-25
expand hint on correct use of jbuf appending in ovis_json.h 

baallan:


baallan:
@tom95858 i updated commit message

PR #1895, opened 2025-07-25
 Fix leaks and use-after-free in example code ldmsd_stream_subscribe.c

baallan:
noone should be copying ldmsd_stream_subscribe for new development (use ldms_msg instead), but this patch may be useful to anyone that has in the past.


baallan:
@tom95858 i updated commit line and description

PR #1898, opened 2025-07-25
 fix leaks and wild pointers in ldmsd_prdcr.c

baallan:
fix leaks and wild pointers in ldmsd_prdcr.c

    - make __ldmsd_xprt_ctxt_free not free invalid pointers to substructures when called with null.
    - initialize all members of ctxt to NULL to avoid free of wild pointers in error paths;
      this affects prdcr_connect_cb ldmsd_prdcr_subscribe_regex ldmsd_prdcr_unsubscribe_regex
    - fix leaks in ldmsd_prdcr_unsubscribe_regex error paths.


baallan:
@tom95858 updated commit message and description

PR #1899, opened 2025-07-25
memory leak fixes to ldmsd_plugattr.c 

baallan:


baallan:
@tom95858 i updated the commit message

PR #1900, opened 2025-07-25
stop use of uninit buffer in ldmsd_failover.c 

baallan:


baallan:
@tom95858 updated commit message per your request

PR #1901, opened 2025-07-25
fix leaks in ldmsd_row_to_json_avro_schema of ldmsd_decomp.c 

baallan:


baallan:
@tom95858 fixed commit message per your request

PR #1902, opened 2025-07-25
memory /pipe leak fixes to ldmsd_config.c round 1

baallan:


ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -1131,6 +1131,7 @@ char *__process_yaml_config_file(const char *path, const char *dname)
 	cfg_str[tbytes] = '\0';
 	int status;
 	status = pclose(fp);
+	fp = NULL;
 	if (status) {
```
tom95858: I don't think we care about the return code from pclose() at this point.  The error message is incorrect, the configuration file was read fine, the error occurred on close. The data read has not been processed yet. This whole if (status) should be removed.

baallan: @tom95858 removed error message per your suggestion and updated commit message as requested.

PR #1903, opened 2025-07-25
fix missing values for ldmsd_cfgobj_type_str() of ldmsd_cfgobj

baallan:
without this fix, callers of the type_str function can get back invalid pointers.


baallan:
@tom95858 updated commit message as requested.

PR #1905, opened 2025-07-25
     fix ldmsctl memory leak for every command handled 

baallan:
fix ldmsctl memory leak for every command handled

    also fixes uninitialized error code return


baallan:
@tom95858 updated commit message per request

PR #1907, opened 2025-07-28
Darshan stream store test

spwalto:
I added a test script for the `darshan_stream_store` plugin in `ldms-static-test.sh`. While testing it in an LDMSD config file, I ran into some errors. After digging into it, I found that `darshan_stream_store.c` was still using the `ldmsd_plugin` version of the `ldmsd_plugin_interface` struct. I updated it to use the correct `ldmsd_store ldmsd_plugin_interface` and added the missing `.type = LDMSD_PLUGIN_STORE` field.

I also noticed the kokkos_* and linux_proc_* store plugins had the same issue, so I went ahead and fixed those too.

Please let me know if anything looks off.


ldms/src/store/kokkos_appmon/kokkos_appmon.c
```diff
@@ -424,16 +424,23 @@ static int constructor(ldmsd_plug_handle_t handle)
...
-        .constructor = constructor,
-        .destructor = destructor,
+struct ldmsd_store ldmsd_plugin_interface = {
+	.base = {
+		.type = LDMSD_PLUGIN_STORE,
+		.config = config,
+		.usage = usage,
+		.constructor = constructor,
+		.destructor = destructor,
+	},
```
tom95858, line 445: @spwalto do none of the stores publish a store() callback function?

spwalto: From what I saw in the code, the only stores that didn’t publish a store() callback are the ones that were modified — `darshan`, `kokkos`, and `linux_proc_*`. I wasn’t sure if they were supposed to have one, so I just left them as-is and updated the function type to `ldmsd_store`.

baallan: @tom95858 these are all really just subscribers that store to sos from specific schema stuff. so there's no callback for a metric set handling.

tom95858: @baallan, ok, that's what I figured. They are driven of a receive callback, not a metric set update.

ldms/src/store/kokkos/kokkos_store.c
```diff
@@ -996,16 +996,23 @@ static int constructor(ldmsd_plug_handle_t handle)
 
 static void destructor(ldmsd_plug_handle_t handle)
 {
-	if (sos)
-		sos_container_close(sos, SOS_COMMIT_ASYNC);
-	if (root_path)
-		free(root_path);
+	ovis_log(mylog, OVIS_LDEBUG, "term %s\n", ldmsd_plug_cfg_name_get(handle));
```
tom95858, line 999: It looks like the indent on this is inconsistent.

Issue #1864, opened 2025-07-17
Should MAYBE_KOKKOS Be Treated Like Other SOS-Dependent Stores in Makefile.am?


spwalto:
I was enabling Darshan store support and ran into some confusion with `ldms/src/store/Makefile.am`. I noticed that `MAYBE_DARSHAN` is defined inside the `if ENABLE_SOS` block:

```
if ENABLE_SOS
 ...
MAYBE_DARSHAN = darshan
...
endif
```
Because of that, I didn’t realize at first that I needed to pass both `--enable-sos` and `--enable-darshan` to configure for the Darshan store to actually get built. There weren’t any messages or errors during the build to indicate this dependency—the build just quietly skipped Darshan.

But then I saw that `MAYBE_KOKKOS` is handled outside of the `ENABLE_SOS` block:
```
if ENABLE_KOKKOS 
MAYBE_KOKKOS = kokkos kokkos_appmon
endif
SUBDIRS += $(MAYBE_KOKKOS)
```
The thing is, Kokkos also depends on SOS, and if SOS isn’t enabled, the build will fail—unlike with Darshan, where it just gets skipped. That got me wondering:

Should we be handling `MAYBE_KOKKOS` the same way as the other SOS-dependent stores? Maybe it makes sense to put it inside the `if ENABLE_SOS` block. Also, it might be nice to add a message that the store will be skipped when someone enables something like Darshan or Kokkos without also enabling SOS—just to make things a little more user-friendly.

Curious what others think. Happy to help make these changes if this sounds good!

nick-enoent:
Hey @spwalto 

That makes sense to me, we should be consistent with plugins that have
dependencies.

Thanks for catching this.

I think an error message and a build fail would be better than a warning
when someone attempts to build with a plugin that doesn't have necessary
dependencies, but I'm open to discussion.

morrone:
I noted an issue with the kokkos configure logic in #1728 as well. Really, the logic in the configure.ac needs to be overhauled, not try to solve the interdependencies in the Makefile.

There are examples of setting variables from checking multiple dependencies, such as:

```
AM_CONDITIONAL([ENABLE_PAPI], [test "x$enable_papi" != xno -a "x$HAVE_LIBPAPI" = xyes -a "x$HAVE_LIBPFM" = xyes])
```

I think that would be a better approach.

spwalto:
@nick-enoent @morrone — thank you for the suggestions and info!

[...]

PR #1908, opened 2025-07-28
Fix SOS Dependency Checks and Fix -L Flag in Build Configuration

spwalto:
This fixes issue #1864 - In `ldms/src/store/Makefile.am`, the SOS-dependent plugins now follow the same format as the `if ENABLE_KOKKOS` block for consistency.

While testing, I ran into this error:
`libtool: error: require no space between '-L' and '../../../../ldms/src/core/libldms.la'.`
After digging into it, I found the issue was with `m4/options.m4`—it was setting` -L` flags even when the specified directories didn’t exist. In my case, I was using a local SOS build where the library lived under `build/sos/src/.libs/`, so the default `/lib `and `/lib64 `checks failed, and the script ended up setting `-L `with no path.

To fix this, I updated `options.m4` so that the `*_LIB64DIR_FLAG` and `*_LIBDIR_FLAG` are only set if the corresponding directories actually exist. Otherwise, they’re left empty.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -521,7 +521,7 @@ AM_CONDITIONAL([HAVE_TX2MON], [test "x$HAVE_TX2MON" = xyes])
 AM_CONDITIONAL([ENABLE_TX2MON], [test "$enable_tx2mon" = "yes"])
 
 # List of plugins that require SOS
-required_plugins="store-app proc-streams darshan kokkos"
+required_plugins="sos store-app proc-streams darshan kokkos"
 
 # Initialize an empty list to collect enabled plugins missing SOS
 missing=""
@@ -533,14 +533,10 @@ for plug in $required_plugins; do
   eval val=\$$varname
 
   if test -z "$val"; then
+    CHECK_SOS=1
     missing="$missing $plug,"
   fi
 done
-
-# Now check if SOS is enabled
-if test -n "$ENABLE_SOS_TRUE" && test -n "$missing"; then
-  AC_MSG_ERROR([The following features require --enable-sos:$missing. Please re-run configure with --enable-sos.])
-fi
 
 dnl Other libraries
 OPTION_DEFAULT_DISABLE([gemini-gpcdr], [ENABLE_GEMINI_GPCDR]) dnl use gpcdr for gemini (not mutually exclusive)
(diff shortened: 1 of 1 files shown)


tom95858:
@baallan could you please take a look at this?

baallan:
when tested under toss4, the m4 change isn't generating correct code for the gpcdlocal test or (afaict) the sosdb test. there's a bunch of other stuff changing in configure.ac as well, so it may be a little bit before i untangle my understanding of where the rest of the changes are intended to get us and what is just a  bug.

tom95858:
@baallan, @spwalto thanks for looking into this. Let me know when it's ready to go. Maybe convert to 'draft' for now?

tom95858:
@spwalto if you use the --with-sos=<your location>

Does the problem go away?

spwalto:
> @spwalto if you use the --with-sos=
> 
> Does the problem go away?

Yes, the LDMS builds successfully. I'll go ahead and remove the changes to options.m4. Thank you for looking into this!

tom95858:
@baallan, @spwalto is this ready to merge?

tom95858:
@baallan, @spwalto ...

> @baallan, @spwalto is this ready to merge?

Is this ready to merge?

baallan:
@spwalto, since @jennfshr is the most frequent snl packager these days, please check with her on this. Are her pipelines good with it?

configure.ac
```diff
@@ -520,21 +520,26 @@ AS_IF([test "x$enable_tx2mon" != xno],[
...
+  eval val=\$$varname
+
+  if test -z "$val"; then
+    missing="$missing $plug,"
+  fi
+done
+
+# Now check if SOS is enabled
+if test -n "$ENABLE_SOS_TRUE" && test -n "$missing"; then
+  AC_MSG_ERROR([The following features require --enable-sos:$missing. Please re-run configure with --enable-sos.])
```
tom95858: @spwalto, I think the error should be `--with-sos=PATH` is required, please re-run with this configure option.

The --enable-sos is not correct. We should probably remove it.

spwalto: Ah, yes — you’re right. I had forgotten about the changes made to the `Makefile.am`. I’ve updated the file to remove the `ENABLE_SOS_TRUE` check and changed the error message to output `--with-sos=PATH (or --with-sos=)` . The LDMS build & install was tested with the `--with-sos` flag and there were no issues.

spwalto:
@baallan @tom95858 I'll let you know when I have updates.

configure.ac
jennfshr: I was able to successfully build with these new changes on @spwalto fork.  This looks fine by me.

spwalto:
Pipelines all look good on our end. @tom95858, unless there’s something I missed in the latest updates, this PR should be good to merge.

PR #1910, opened 2025-07-29
slingshot_metrics sampler backport shs-12.0.0 support to be in b4.4

Matt8687:
Backport requisite slingshot_metric for release/shs-12.0.0 per commit f4be13e.
 * This cherry picks portions of the fore mentioned commit f4be13e, as not all changes in that commit are supported in v4.4.X

### DIFF
--- a/ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
+++ b/ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
@@ -293,27 +293,9 @@ static int use_counter(const char * const counter_name);
 static int use_counter_group(const char * const counter_group)
 {
         int i;
-        bool all_counters = false;
-        enum c_cntr_group group;
         int rc;
 
-        if (!strcmp(counter_group, "all")) {
-                all_counters = true;
-        } else if (!strcmp(counter_group, "ext")) {
-                group = C_CNTR_GROUP_EXT;
-        } else if (!strcmp(counter_group, "pi_ipd")) {
-                group = C_CNTR_GROUP_PI_IPD;
-        } else if (!strcmp(counter_group, "mb")) {
-                group = C_CNTR_GROUP_MB;
-        } else if (!strcmp(counter_group, "cq")) {
-                group = C_CNTR_GROUP_CQ;
-        } else if (!strcmp(counter_group, "lpe")) {
-                group = C_CNTR_GROUP_LPE;
-        } else if (!strcmp(counter_group, "hni")) {
-                group = C_CNTR_GROUP_HNI;
-        } else if (!strcmp(counter_group, "ext2")) {
-                group = C_CNTR_GROUP_EXT2;
-        } else {
+        if (strcmp(counter_group, "all")) {
                 log_fn(LDMSD_LERROR, SAMP" unrecognized counter group \"%s\"\n",
                        counter_group);
                 return -1;
@@ -325,7 +307,7 @@ static int use_counter_group(const char * const counter_group)
                            due to grouping, and leaving space for future new
                            counters */
                         continue;
-                } else if (all_counters || c1_cntr_descs[i].group == group) {
+                } else {
                         rc = use_counter(c1_cntr_descs[i].name);
                         if (rc != 0) {
                                 return rc;


morrone:
Thanks, but we don't seem to be doing b4.4 patches any longer, otherwise I would have submitted it myself.

morrone:
@Matt8687 Just for future reference, a couple of things about porting patches to another branch:

* First, and most importantly, we want to keep the commit messages substantially the same when we port a commit to another branch. In particular the first line of the commit message must remain the same so that humans can easily check for the presence of a patch in different branches just by grep'ing through the output of "git log --oneline <branchname". This is really essentially for developer sanity. Additional comments can be appended to original commit message to explain where it was cherry-picked from, explain what changes needed to be made in the porting process, etc.
* We should maintain the authorship of the original commit as long as the ported commit is substantially equivalent. (We can append information to the original git comment to explain what we changed, so the original author gets credit for the change, but not the blame for newly introduced errors :)). Using `git cherry-pick` maintains the original author by default.

Matt8687:
@morrone Gotcha! So for instance, I should have done the same commit message: " slingshot_metrics: drop support for "groups" " and then gone through and added the note with the SHA of your original commit? I've never used `git cherry-pick` before, so I'll have to figure that out! 

Also, as for this being closed instead of merged or otherwise; isn't b4.4 the currently stable version of LDMS? If it is then wouldn't it stand to reason we should at least fix this issue? Don't get me wrong, I want to move to LDMS v4.5.0, but last I heard there were still some substantial issues?

morrone:
> @morrone Gotcha! So for instance, I should have done the same commit message: " slingshot_metrics: drop support for "groups" " and then gone through and added the note with the SHA of your original commit? I've never used `git cherry-pick` before, so I'll have to figure that out!

Yes, that is correct. `git cherry-pick` also preserves the original message. So yes, use that and you'll have most of what you need to port a commit well.

> Also, as for this being closed instead of merged or otherwise; isn't b4.4 the currently stable version of LDMS? If it is then wouldn't it stand to reason we should at least fix this issue? Don't get me wrong, I want to move to LDMS v4.5.0, but last I heard there were still some substantial issues?

I have no argument with your reasoning, but things just haven't been landing for b4.4 for some time. My best understanding is that there is no plan to ever create a a v4.4.6 release. All effort is on getting v4.5.1 done and moving to that.

jennfshr:
Whereas I agree v4.5.1 is a huge priority, and OGC should concentrate focus on that, I think it’s prudent in order to sustain production support to drop a v4.4.6 release, even if we collaborators are who facilitate that release. @Brandt, James ***@***.***> @Allan, Ben ***@***.***>

From: Christopher J. Morrone ***@***.***>
Date: Wednesday, July 30, 2025 at 4:04 PM
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [ovis-hpc/ldms] slingshot_metrics sampler backport shs-12.0.0 support to be in b4.4 (PR #1910)

[https://avatars.githubusercontent.com/u/178400?s=20&v=4]morrone left a comment (ovis-hpc/ldms#1910)<https://github.com/ovis-hpc/ldms/pull/1910#issuecomment-3137945535>

@morrone<https://github.com/morrone> Gotcha! So for instance, I should have done the same commit message: " slingshot_metrics: drop support for "groups" " and then gone through and added the note with the SHA of your original commit? I've never used git cherry-pick before, so I'll have to figure that out!

Yes, that is correct. git cherry-pick also preserves the original message. So yes, use that and you'll have most of what you need to port a commit well.

Also, as for this being closed instead of merged or otherwise; isn't b4.4 the currently stable version of LDMS? If it is then wouldn't it stand to reason we should at least fix this issue? Don't get me wrong, I want to move to LDMS v4.5.0, but last I heard there were still some substantial issues?

I have no argument with your reasoning, but things just haven't been landing for b4.4 for some time. My best understanding is that there is no plan to ever create a a v4.4.6 release. All effort is on getting v4.5.1 done and moving to that.

morrone:
> Whereas I agree v4.5.1 is a huge priority, and OGC should concentrate focus on that, I think it’s prudent in order to sustain production support to drop a v4.4.6 release, even if we collaborators are who facilitate that release.

I agree with that as well! In fact I am already carrying this patch (and a few others) in the toss branch in https://github.com/LLNL/ldms. I think it would be great for the collaborators to start maintaining the b4.4 stable branch.

Matt8687:
As a heads up I reopened this pull request as #1911 -- I don't know if that will autolink or not, but wanted to mention it so at the very least there's a message history linking the two pull requests. (@morrone please let me know if there's anything I can improve upon!)

PR #1911, opened 2025-07-30
slingshot_metrics: drop support for "groups"

Matt8687:
At least as of Slingshot Host Software release/shs-12.0.0, the API surrounding "groups" has been broken. As a result, we drop support for all groups except for "group:all".

This was cherry-picked from Chris Morrone's commit f4be13e

(At the very least, I want to see if this is the correct way to create a cherry-picked pull-request based on prior feedback -- even if these changes are disregarded again. Thank you for your time!)

### DIFF
--- a/ldms/src/sampler/slingshot_metrics/ldms-sampler_slingshot_metrics.rst
+++ b/ldms/src/sampler/slingshot_metrics/ldms-sampler_slingshot_metrics.rst
@@ -1,0 +1,104 @@
+.. _slingshot_metrics:
+
+========================
+slingshot_metrics
+========================
+
+----------------------------------------------
+Man page for the LDMS slingshot_metrics plugin
+----------------------------------------------
+
+:Date:   1 May 2022
+:Manual section: 7
+:Manual group: LDMS sampler
+
+
+SYNOPSIS
+========
+
+| Within ldmsd_controller or a configuration file:
+| config name=slingshot_metrics [ <attr> = <value> ]
+
+DESCRIPTION
+===========
+
+With LDMS (Lightweight Distributed Metric Service), plugins for the
+ldmsd (ldms aemon) are configured via ldmsd_controller or a
+configuration file. The slingshot_metrics plugin provides a single
+metric set that contains a list of records. Each record contains all of
+the metrics for a single slingshot NIC.
+
+The slingshot_metrics sampler plugin provides detailed counter metrics
+for each slignshot NIC.
+
+The schema is named "slingshot_metrics" by default.
+
+CONFIGURATION ATTRIBUTE SYNTAX
+==============================
+
+The slingshot_metrics plugin uses the sampler_base base class. This man
+page covers only the configuration attributes, or those with default
+values, specific to the this plugin; see :ref:`ldms_sampler_base(7) <ldms_sampler_base>` for the
+attributes of the base class.
+
+**config**
+   | name=<plugin_name> [counters=<COUNTER NAMES>] [counters_file=<path
+     to counters file>]
+   | configuration line
+
+   name=<plugin_name>
+      |
+      | This MUST be slingshot_metrics.
+
+   counters=<COUNTER NAMES>
+      |
+      | (Optional) A CSV list of names of slingshot counter names. See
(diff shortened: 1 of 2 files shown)


andrewpbarry:
I believe this is even necessary with shs-11. 
Can confirm that this prevents compilation with newer shs rpms, which are required for more recent OS release versions (e.g. sles15sp6)

morrone:
OK, that is a good start. But there are a couple of things that needed tweaking in the porting process.

I imagine you started with the latest b4.4 branch, because I see you are requesting the PR from `Matt8687:b4.4`. So the first thing that I would recommend is always creating a working branch for each PR that you wish to create. The name of that branch isn't super important, it should just be something relatively unique in your repo. So the first step would be something like:

```git checkout -b backport_drop_groups origin/b4.4```

I imagine you did something like:

```git cherry-pick f4be13e14d5```

And you probably saw it complain like this:

```
CONFLICT (modify/delete): ldms/src/sampler/slingshot_metrics/ldms-sampler_slingshot_metrics.rst deleted in HEAD and modified in f4be13e14 (slingshot_metrics: drop support for "groups").  Version f4be13e14 (slingshot_metrics: drop support for "groups") of ldms/src/sampler/slingshot_metrics/ldms-sampler_slingshot_metrics.rst left in tree.
Auto-merging ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
CONFLICT (content): Merge conflict in ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
error: could not apply f4be13e14... slingshot_metrics: drop support for "groups"
```

I find a good thing to to do when there is a conflict is run `git status`. That puts the issue in a more readable form for me:

```
$ git status
$ git status
On branch b4.4_drop_groups
Your branch is up to date with 'ldms/b4.4'.

You are currently cherry-picking commit f4be13e14.
  (fix conflicts and run "git cherry-pick --continue")
  (use "git cherry-pick --skip" to skip this patch)
  (use "git cherry-pick --abort" to cancel the cherry-pick operation)

Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)
        deleted by us:   ldms/src/sampler/slingshot_metrics/ldms-sampler_slingshot_metrics.rst
        both modified:   ldms/src/sampler/slingshot_metrics/slingshot_metrics.c
```

What we need to note here in particular is the "deleted by us" line. That means that the cherry-picked commit made a change to the `ldms-sampler_slingshot_metrics.rst`, but that file doesn't exist in the branch on which we are working. That should be a red flag. We should not simply add that entire file to our commit. Instead we need to figure out why the original branch had that file, and we do not.

In this case, there has been a big change to documentation between `b4.4` and `main`. On `b4.4` man pages were in man page format. On `main`, man pages are written in rst, and converted to man page format during the build process. So the correct thing to do is to figure out what the change that the commit made to the rst documentation, and port that (make the equivalent change) to `ldms/src/sampler/slingshot_metrics/Plugin_slingshot_metrics.man` instead.

Both that change and the one you made in `fix using wrong log function for v4.4.X` would be considered trivial porting changes, and should just be part of the one commit supplied in this PR.

As a general rule, we rarely want to introduce additional commits when porting a commit to another branch.

I think it would be worthwhile practice to try this port again. But FYI, I have this port done on the toss branch over in the https://github.com/LLNL/ldms repository. So when you think you have the port in good shape, you can compare it to https://github.com/LLNL/ldms/commit/fb925996e19d50873cbc5120c657cdc22292d72b.

tom95858:
@morrone, @Matt8687 Is this going to be refactored for 4.5.1. Seems like it is important.

morrone:
> @morrone, @Matt8687 Is this going to be refactored for 4.5.1. Seems like it is important.

@tom95858 it is already on main.

morrone:
@tom95858, could you please revert those commits? This was not ready to land.

tom95858:
Sure but I thought I asked that

Thomas Tucker, President, Open Grid Computing, Inc.

morrone:
> Sure but I thought I asked that

I don't see that question in this PR. I do see a long comment from me explaining what needs to be done before the PR is ready to land.

tom95858:
@morrone k my bad.

PR #1912, opened 2025-07-30
Fix write into array[-1] in ldmsd_request.c

tom95858:
When the argument count to an authentication plugin is zero, this logic writes a '\0' into the av array an an invalid location.


tom95858:
@baallan please take a look. It's interesting that the tool didn't catch this.

baallan:
looks like the interprocedural analysis may have proved cnt is always nonzero; the checker traces all the callers and data it can.
Did you see something actually hit the off-by--1? A strictly local static analysis would call it as you see it.

It's also possible i missed something in going over the 500+ items identified. the ui on this checker sucks.

That said, this particular tool is a bit too literal minded in bounds-checking the C-struct-polymorphism cases around the send(struct headerish* dataheader, size_t data_len, ....) type functions. I'm ignoring those warnings unless valgrind also flags it.

PR #1913, opened 2025-07-31
Update the set information whent the set is registered.

tom95858:


tom95858:
@nichamon please review

PR #1915, opened 2025-08-07
Work around threadstat structure

tom95858:


morrone:
@tom95858 Are you aware that your ""Error %d assigning thread stats to the endpoint.\n" message is printing on every single call from ldms_ls? If that is leaking memory, then I am concerned that we cannot wait until 4.5.2 for the real fix.

baallan:
There's a variety of other nits and leaks addressed in #1876, some related to thrstat. it will need rebasing if this is merged first.

tom95858:
#1876 is still marked as a draft

baallan:
#1876 was/is failing the compatibility testing, with no clear understanding on my part as to why (as noted in the comments there; maybe the compat tests are buggy?).

tom95858:
Replaced bt #1919

tom95858:
Replaced by #1919

PR #1918, opened 2025-08-12
Remove unsupported samplers

### LINKED ISSUE #1917: Remove old and unsupported samplers from the tree.
This is a tracking issue for samplers that are potentially not worth dragging forward to future releases.

- [ ] contrib gpu_metrics_sampler
- [ ] switchx sampler

jennfshr:
Closes #1917 

We need to reduce the footprint and excess complexity of the codebase by removing outdated/unsupported samplers.  This removes `switchx` and `gpu_metrics_sampler` from the source, build system, builtin tests, and documentation.


baallan:
I see no reason to keep switchx, but for the contrib samplers I suggest we just omit them from the configure.ac (dnl comment the related configure.ac lines) and leave their directories in the tree if they stop compiling for any reason (including 'we change the api'). 'contributed' has always meant 'not supported as part of the core product; your mileage may vary; you may be interested in porting this to your environment if it doesn't work out of the box -- it worked somewhere long enough and interesting enough for us to merge it.'  We may attract other folk willing to pick up/extend/fix a contrib plugin if it is present; we certainly won't attract them if it is no longer visible in the current code base.

morrone:
I am not a fan of leaving dead code in the tree. The code is always available in the git history. If anyone asks about OneAPI support for GPUs, we can always point them to a git release where that code existed.

jennfshr:
We had discussion about removing both these samplers and I thought had gained a consensus to do it.  I agree, if they want to revive these samplers, they can recover from git history.

morrone (approved):

baallan:
> We had discussion about removing both these samplers and I thought had gained a consensus to do it. I agree, if they want to revive these samplers, they can recover from git history.

Before removing a contrib sampler, there should be an issue opened (marked with some sort of 'contrib feature update needed' tag) that contains:
- what is the sampler
- why it was removed (what platform or api went out of date)
- what is needed to fix it
- what is the last/best commit or release tag to start from if you want to revive it.

tom95858:
@baallan, @jennfshr, @morrone I think that we agreed to remove these. Regarding 'contrib', but definition, we don't support these and if the contributor isn't either, then we should remove them. I don't think it is contingent upon us to provide support to what are explicitly unsupported samplers.

Those are my 2 cents. So please talk amongst yourselves and let's do something one way or the other. If we're going to move it to 4.5.2, then let's do that otherwise, let's merge it.

morrone:
I vote to merge it.

tom95858:
@jennfshr may we merge your patch?

jennfshr:
> @jennfshr may we merge your patch?

@tom95858 yes!

PR #1921, opened 2025-08-13
Fix zap fabric

narategithub:
- conditionally use deprecated libfabric enum to solve build issue
- address issues with LDMS Rail
  - increase zap_fabric max connect request message size (required by LDMS Rail)
  - setup valid fabric ep before notifying the application about the connection request as LDMS Rail needs to resolve peer address in the process of handling LDMS connect request.


narategithub:
This pull request and PR #1922 passed a sniff test on zap_fabric (tcp provider) on Debian 13 (libfabric 2.1).

PR #1924, opened 2025-08-15
Introduce pull request labeler workflow

morrone:
Introduce a new github workflow that adds github labels to PRs. The initial set of patterns just add the labels "main" or "b4.4" to a PR based on whether the PR targets the main or b4.4 branch.

(There are a couple of speculative patterns for b4.5 and b4.6, but it doesn't hurt to have those. We can add/remove patterns in the future as needed.)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,8 @@
+main:
+- base-branch: main
+b4.4:
+- base-branch: b4.4
+b4.5:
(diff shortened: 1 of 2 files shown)


morrone:
@tom95858 This is another easy PR to land. It doesn't touch any code. It just adds a simple new workflow. I had tested the workflow separately, and it is now on b4.4 and working there.

Issue #1843, opened 2025-06-26
ldmsd_set_deregister redundancy


baallan:
From dcgm (whether using sampler_base or not) and jobid samplers, when the destructor is called ldmsd_set_deregister has already been called by ldmsd_term_plugin.
Since plugins, in general, may be doing complicated things with their sets, its unclear why term_plugin is getting ahead of sampler->destructor() (unless they are on independent threads, which should not be the case).

log symptom:
```
Thu Jun 26 16:52:58 2025:    DEBUG: sampler.dcgm_sampler.dcgm_sampler: term() called
Thu Jun 26 16:52:58 2025:    ERROR: ldmsd: Dregistering set name 'localhost1/dcgm1/gpu_0' failed because the sampler config 'dcgm_sampler' does not exist.
Thu Jun 26 16:52:58 2025:    ERROR: ldmsd: Dregistering set name 'localhost1/dcgm1/gpu_1' failed because the sampler config 'dcgm_sampler' does not exist.
Thu Jun 26 16:52:58 2025:    ERROR: ldmsd: Dregistering set name 'localhost1/dcgm1/gpu_2' failed because the sampler config 'dcgm_sampler' does not exist.
Thu Jun 26 16:52:58 2025:    ERROR: ldmsd: Dregistering set name 'localhost1/dcgm1/gpu_3' failed because the sampler config 'dcgm_sampler' does not exist.
Thu Jun 26 16:52:58 2025:    ERROR: ldmsd: Dregistering set name 'localhost1/jobid' failed because the sampler config 'jobid' does not exist.
Thu Jun 26 16:52:58 2025:     INFO: ldmsd: User requested exit.
```

emdonat:
Typo in Deregistering also.

tom95858:
@baallan, @emdonat Some samplers have sets that come and go or the sampler itself can be reconfigured. In either case, this logic calls ldmsd_set_deregister(). The issue is that the generic term() logic *also* calls ldmsd_set_deregister(). So if the sampler calls its generic set cleanup logic from the destructor() you get this error.

Aside from refactoring the samplers to have one set of cleanup logic for general operation, and another for destructor(), this message will be there.

The behavior is benign and asymptomatic aside from printing a scary message to the log. I propose demoting the message to INFO and making the message informative and less scary.

baallan:
So shouldn't there be a concern that term gets ahead of destructor, disappearing sets from under the plugin instance while it may still be operating on them?

[...]

PR #1926, opened 2025-08-21
Update ldmsd_set_deregister()

nichamon:
When ldmsd_set_deregister() was implemented, plugins did not have access to plugin handles. It knows only the plugin name, the code has been evolved to the point that plugins have access to plugin handles. 

The patch changes the signature of ldmsd_set_deregister() to receive ldmsd_plug_handle_t instead of the plugin name string.

### DIFF
--- a/ldms/src/sampler/examples/test_sampler/test_sampler.c
+++ b/ldms/src/sampler/examples/test_sampler/test_sampler.c
@@ -437,7 +437,7 @@ static void test_sampler_set_free(test_sampler_t ts, struct test_sampler_set *s)
 {
 	LIST_REMOVE(s, entry); /* Remove from the set_list of ts_schema */
 	if (s->set) {
-		ldmsd_set_deregister(s->name, ldmsd_plug_cfg_name_get(ts->pi));
+		ldmsd_set_deregister(s->name, ts->pi);
 		ldms_set_unpublish(s->set);
 		ldms_set_delete(s->set);
 	}
@@ -448,7 +448,7 @@ static void test_sampler_set_free(test_sampler_t ts, struct test_sampler_set *s)
 static void test_sampler_set_reset(test_sampler_t ts, struct test_sampler_set *s)
 {
 	if (s->set) {
-		ldmsd_set_deregister(s->name, ldmsd_plug_cfg_name_get(ts->pi));
+		ldmsd_set_deregister(s->name, ts->pi);
 		ldms_set_unpublish(s->set);
 		ldms_set_delete(s->set);
 	}


nichamon:
@tom95858 This is the change to `ldmsd_set_deregister()` I mentioned to you. I haven't finished the patch yet, but I created this draft for you to see what I'm talking about.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -716,28 +717,35 @@ int ldmsd_set_register(ldms_set_t set, const char *cfg_name)
 	return rc;
 }
 
-void ldmsd_set_deregister(const char *inst_name, const char *cfg_name)
+void ldmsd_set_deregister(const char *inst_name, ldmsd_plug_handle_t handle)
```
morrone, line 720: I love the approach!

But in the name of symmetry and programmer sanity, it would be nice to replace "plugin_name" with "ldmsd_plug_handle_t handle" in ldmsd_set_register() as well.

And it wouldn't hurt to also replace "instname" with "ldms_set_t set" in ldmsd_set_deregister to make the register/deregister functions completely operate the same way.

tom95858: It is inappropriate to pass the plugin handle into the ldmsd_set_deregister API. The API is looking up an ldmsd_cfgobj_sampler_t object, not a private plugin data structure. The ldms_plug_handle_t  is a void* that is expected to point to a private plugin data structure (for example meminfo_t). The only way this could work is if the plugin handle had a pointer to the ldmsd_cfgobj_sampler_t object which it should not and does not.

The constructor/destructor pair work on private plugin owned data structures. LDMSD knows nothing about them. The destructor is called when the infrastructure can guarantee that no one can affect the void* pointed to by ldmsd_plug_handle_t associated with the plugin instance. Putting the destructor somewhere else breaks this guarantee.

This is why putting ldmsd_set_deregister in the destructor either directly or indirectly (for example the reset() call in procnet) should not be done. It won't break anything because we "look it up" and find it's not there. This is precisely the *reason* why the ldmsd_set_deregister takes a string and not a ldmsd_cfgobj_sampler_t directly. It is to avoid the race where the configuration object is being destroyed concurrently with the sampler attempting to destroy a set internally.

I am totally up for redesigning how registration is done; but not for 4.5.1.

morrone: > It is inappropriate to pass the plugin handle into the ldmsd_set_deregister API. The API is looking up an ldmsd_cfgobj_sampler_t object, not a private plugin data structure. The ldms_plug_handle_t is a void* that is expected to point to a private plugin data structure (for example meminfo_t). The only way this could work is if the plugin handle had a pointer to the ldmsd_cfgobj_sampler_t object which it should not and does not.

@tom95858, with all due respect, that is not the case. You are confusing the ldmsd_plug_handle_t with the plugin's context pointer (accessed by the ldmsd_plug_ctxt_set()/ldmsd_plug_ctxt_get() functions).

The ldmsd_plug_handle_t is opaque to the plugins; so it cannot be a plugin's private data. ldmsd knows that the ldms_plug_handle_t points to an ldmsd_cfgobj_t. Recall how the functions are implemented in ldmsd_plug_api.c. Conversely, the plugin instance's context pointer is opaque to ldmsd.

So I think Nichamon's proposed refactoring is sound. One can certainly debate about whether it should land before or after v4.5.1.

> The constructor/destructor pair work on private plugin owned data structures. LDMSD knows nothing about them. The destructor is called when the infrastructure can guarantee that no one can affect the void* pointed to by ldmsd_plug_handle_t associated with the plugin instance. Putting the destructor somewhere else breaks this guarantee.

That is not the case. The code _does_ currently change something that the context of the plugin instance relies upon. The plugin instance's context knows that it has an outstanding registration, and it should be able to clean that up in destructor(). The ldmsd code incorrectly cleans that up before the plugin is allowed to do so. So ldmsd is _currently_ doing what you are saying it shouldn't do: it is changing things that the plugin context replies upon.

> This is why putting ldmsd_set_deregister in the destructor either directly or indirectly (for example the reset() call in procnet) should not be done.

ldmsd_set_deregister() _should_ be called in destructor() so a plugin can clean up everything it did during itself lifetime. If a plugin calls ldmsd_set_register(), it should have a matching call to ldmsd_set_deregister(). 

> I am totally up for redesigning how registration is done; but not for 4.5.1.

That is something you should note in issue #1843. People that need to see that comment will probably miss it if you only put it in a PR review comment.

tom95858: @morrone, you are correct. It's a pointer to ldmsd_cfgobj_sampler_t. But explain to me why if it is ultimately looking up ldmsd_cfgobj_sampler_t that it passes in a void*. This seems to avoid some nice compiler time type checking.

tom95858:
> When ldmsd_set_deregister() was implemented, plugins did not have access to plugin handles. It knows only the plugin name, the code has been evolved to the point that plugins have access to plugin handles.
>
That's not why we passed in the char * instead of the ldmsd_cfgobj_sampler_t directly.
 
> The patch changes the signature of ldmsd_set_deregister() to receive ldmsd_plug_handle_t instead of the plugin name string.

We can debate this further, however, I don't know why we would pass in ldmsd_plug_handle_t instead of just ldmsd_cfgobj_sampler_t.

Is there some reason for this I'm missing?

ldms/src/ldmsd/ldmsd.c
morrone: Because this is a plugin API that can, and should, be used by people that don't know ldmsd's internals. They shouldn't need to learn ldmsd's internals. We are employing an API, and not a code monolith. We are employing the common computer science concepts of abstraction, data hiding, opaque data types, etc. These are very common computer science concepts employed in APIs we all work with every day.

morrone:
> We can debate this further, however, I don't know why we would pass in ldmsd_plug_handle_t instead of just ldmsd_cfgobj_sampler_t.
> 
> Is there some reason for this I'm missing?

Yes. We have been over this many times, but I guess I'm just not using the right words to convey the concept of what an opaque handle is. I'll give it another try:

The ldmsd_plug_handle_t handle is opaque to the plugin. When we say "opaque", what we are saying is that the plugin does not know anything about the actual underlying data structure that handle references. This is information hiding at work in an API. This wikipedia article might be helpful in understanding the concept:

https://en.wikipedia.org/wiki/Opaque_data_type

From the plugin's perspective all it knows is that the handle is a "void *". It does not, and should not, known that the handle points to a ldmsd_cfgobj_t. Only the ldmsd side of the plugin API knows that.

Because the plugin side of the API doesn't know about the data structure underlying the handle, the plugin side doesn't know anything about the cfgobj. Since the plugin does not know about the cfgobj, it _cannot_ provide the ldmsd_cfgobj_t as an parameter of the ldmsd_set_deregister() call.

What the plugin _does_ know is the handle. The plugin can pass the handle to ldmsd_set_deregister(). The ldmsd side of the API, the code _does_ know what a handle points to. It knows that the current implmentation of a ldmsd_plug_handle_t is a pointer to an ldmsd_cfgobj_t.  So in the implementation of the ldmsd_set_deregister() function, the code knows it can cast the handle to a cfgobj.

And that is exactly what the proposed code in this PR does.

tom95858:
> Yes. We have been over this many times, but I guess I'm just not using the right words to convey the concept of what an opaque handle is. I'll give it another try:
> 
> The ldmsd_plug_handle_t handle is opaque to the plugin. When we say "opaque", what we are saying is that the plugin does not know anything about the actual underlying data structure that handle references. This is information hiding at work in an API. This wikipedia article might be helpful in understanding the concept:
>

@morrone, It's not that I don't understand what an opaque handle is or that I don't appreciate its value in certain circumstances. I simply don't see the value of an opaque handle in this circumstance.
 
> https://en.wikipedia.org/wiki/Opaque_data_type
> 
> From the plugin's perspective all it knows is that the handle is a "void *". It does not, and should not, known that the handle points to a ldmsd_cfgobj_t. Only the ldmsd side of the plugin API knows that.
> 
> Because the plugin side of the API doesn't know about the data structure underlying the handle, the plugin side doesn't know anything about the cfgobj. Since the plugin does not know about the cfgobj, it _cannot_ provide the ldmsd_cfgobj_t as an parameter of the ldmsd_set_deregister() call.
> 
> What the plugin _does_ know is the handle. The plugin can pass the handle to ldmsd_set_deregister(). The ldmsd side of the API, the code _does_ know what a handle points to. It knows that the current implmentation of a ldmsd_plug_handle_t is a pointer to an ldmsd_cfgobj_t. So in the implementation of the ldmsd_set_deregister() function, the code knows it can cast the handle to a cfgobj.
> 
> And that is exactly what the proposed code in this PR does.

morrone:
> > Yes. We have been over this many times, but I guess I'm just not using the right words to convey the concept of what an opaque handle is. I'll give it another try:
> > The ldmsd_plug_handle_t handle is opaque to the plugin. When we say "opaque", what we are saying is that the plugin does not know anything about the actual underlying data structure that handle references. This is information hiding at work in an API. This wikipedia article might be helpful in understanding the concept:
> 
> @morrone, It's not that I don't understand what an opaque handle is or that I don't appreciate its value in certain circumstances. I simply don't see the value of an opaque handle in this circumstance.

What exactly is "this circumstance"? ldmsd_set_deregister, or the API in general?

I think we are now pretty committed to using the opaque handle in the plugin API. The code landed, and we are about to tag and ship it.

If we accept that the opaque handle is now part of the API, and if we understand the concept of an opaque handle and appreciate its value, then we _cannot_ violate the opaque handle to retrieve the cfgobj on the opaque side of the handle. If we _ever_ violate the opaque handle by accessing the cfgobj on the opaque side (e.g. from the plugins), then the opaque handle is _no longer opaque_.

When the handle is no longer opaque, we lose the benefits of an opaque handle.

nichamon:
I'm closing this because of two reasons: 1) this is incomplete and 2) there is a discussion to refactor the design of `ldmsd_set_register()` and `ldmsd_set_deregister()`.

PR #1928, opened 2025-08-21
Adjust permissions to allow release creation.

morrone:
github has changed permissioning. action-gh-release can no longer create a release or attach files to a release without being granted explicit permission. This adds said permission.


morrone:
@tom95858, this should be an easy one to land. It is necessary before we do the 4.5.1 release to allow the "Create Release" workflow to start working again.

PR #1929, opened 2025-08-22
Make ldmsd_controller notify users when the connection is disconnected

nichamon:


nichamon:
@nick-enoent I made changes to the Communicator class to use the asynchronous mode of ldms.Xprt.connect(). This is to enable ldmsd_controller to receive an event when the connection is disconnected and notify users. As Maestro uses Communicator(), could you please look into this and test Maestro with the change whenever you get a chance?

The patch doesn't fix any bugs, but it could improve the user experience when using ldmsd_controller and disconnection occurs.

tom95858:
@nick-enoent did you get a chance. to test this with Maestro?

PR #1930, opened 2025-08-23
numerous fixes to ldmsd_request.c bugs

baallan:
This provides a batch of separable commits fixing leaks and logic errors for functions in ldmsd_request, with minor documentation additions where appropriate.
If any particular change is incorrect, the branch can be refactored to exclude it.
Each commit has a descriptive title.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -5903,8 +5903,8 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 	value = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_UID);
 	if (value) {
 		if (isdigit(value[0])) {
-			uid = strtol(value, &endptr, 0);
-			if (uid < 0) {
+			long long check_uid = strtoll(value, &endptr, 0);
+			if (check_uid < 0 || check_uid > (UINT32_MAX - 1) ) {
 				reqc->errcode = EINVAL;
 				(void) snprintf(reqc->line_buf, reqc->line_len,
 						"The given UID '%s' is invalid.",
@@ -5912,6 +5912,7 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 				free(value);
 				goto free_regex;
 			}
+			uid = check_uid;
 		} else {
 			struct passwd *pwd = getpwnam(value);
 			if (!pwd) {
@@ -5930,8 +5931,8 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 	value = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_GID);
 	if (value) {
 		if (isdigit(value[0])) {
-			gid = strtol(value, &endptr, 0);
-			if (gid < 0) {
+			long long check_gid = strtoll(value, &endptr, 0);
+			if (check_gid < 0 || check_gid > (UINT32_MAX - 1) ) {
 				reqc->errcode = EINVAL;
 				(void) snprintf(reqc->line_buf, reqc->line_len,
 						"The given GID '%s' is invalid.",
@@ -5939,6 +5940,7 @@ static int set_sec_mod_handler(ldmsd_req_ctxt_t reqc)
 				free(value);
 				goto free_regex;
 			}
+			gid = check_gid;
 		} else {
 			struct group *grp = getgrnam(value);
 			if (!grp) {
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -2367,7 +2367,7 @@ static int __prdcr_stream_status(ldmsd_prdcr_t prdcr, ldmsd_req_ctxt_t oreqc,
 				struct ldmsd_str_ent *pname)
 {
 	int rc;
-	ldmsd_req_cmd_t rcmd;
+	ldmsd_req_cmd_t rcmd = NULL;
 	struct prdcr_stream_status_ctxt *ctxt;
 
 	ctxt = malloc(sizeof(*ctxt));
@@ -2388,6 +2388,8 @@ static int __prdcr_stream_status(ldmsd_prdcr_t prdcr, ldmsd_req_ctxt_t oreqc,
 		rc = ldmsd_req_cmd_attr_term(rcmd);
 		if (rc)
 			goto rcmd_err;
+	} else {
+		free(ctxt);
 	}
 	ldmsd_prdcr_unlock(prdcr);
 	return 0;
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -2498,6 +2498,7 @@ next_prdcr:
 		rc = 0;
 		goto free_ctxt;
 	}
+	free(prdcr_regex);
 	return 0;
 
 send_resp_code:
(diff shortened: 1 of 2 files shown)


baallan:
@tom95858 this is needed in 4.5.1 due to:
- leaks on the normal execution path (growing the daemon) are:
        6907b98 fix leak in msg_stats_handler  is a daemon grower on the normal logic path
        df1348e fix leak in prdcr_subscribe_regex_handler of ldmsd_request.c
        d67fc59 fix leak in normal path of prdcr_stream_status_handler
        c1899329 fix potential leak in __prdcr_stream_status of ldmsd_request.c (happens in an expected less common condition)

- fixing broken new key feature rate control is:
        eb31a5ab fix copy/paste error in __prdcr_add_handler handling of rate limiter

- fixing security (reasonably occuring unintentional DOS attack)
        95d8f2ed fix free of wild pointer in msg_stats_handler when msg is disabled
        When admin disables msg but applications try to publish, this bug
        sprinkles the daemon with random calls to free(wild) likely to cause unhappiness.

- releasing a lingering lock in __prdcr_hint_set_tree_json_obj is part of the poorly titled
        93d73aa fix missing handling of linebuf_printf error results in ldmsd_request.c
        Where it looks like two changes got included in one commit.

tom95858:
@nichamon, could you please review this change?

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -8610,6 +8629,7 @@ static int msg_stats_handler(ldmsd_req_ctxt_t reqc)
 		match = stream;
 	}
 
+	char *reset_s = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_RESET);
```
nichamon, line 8632: For readability, what do you think about 1) keeping the declaration and assignment of `reset_s` at line 8589 and 2) moving `free(reset_s)` to be in the `out:` part?

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -9891,15 +9912,15 @@ static int __cidr2addr6(const char *cdir_str, struct ldms_addr *addr, int *prefi
 		return EINVAL;
 	}
 
-	if (prefix_len)
-		*prefix_len = _prefix_len;
+	if (!prefix_len || !addr)
```
nichamon: It's a nit. How about we check this right after all the declaration so we don't have to spend any CPU cycles to do `sscanf()` or `strchr()`?

ldms/src/ldmsd/ldmsd_request.c
baallan: Allocations should be freed as soon as there is no more use of them; reset_s is done with as soon as the bool value is assigned. The items here left for the out: sections may be used in in several places up the out.

Also, as a pattern used in our team, leaving cleanup of attr strings for an out: block has been very unsuccessful: I've had to clean up a lot of leaks and wild pointers caused by developers mishandling the out block update because it is far away.

ldms/src/ldmsd/ldmsd_request.c
baallan: yeah, can move that up a bit.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -8778,7 +8798,7 @@ static int listen_handler(ldmsd_req_ctxt_t reqc)
 {
 	ldmsd_listen_t listen;
 	char *xprt, *port, *host, *auth, *attr_name, *quota, *rx_limit;
-	unsigned short port_no = -1;
+	long long port_no = -1;
```
tom95858: Why are we making what is an unsigned short a "long long"

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -8787,7 +8807,7 @@ static int listen_handler(ldmsd_req_ctxt_t reqc)
 		goto einval;
 	port = ldmsd_req_attr_str_value_get_by_id(reqc, LDMSD_ATTR_PORT);
 	if (port) {
-		port_no = atoi(port);
+		port_no = atoll(port);
```
tom95858: This is incorrect. A port number is an unsigned short. Why is this change being made.

ldms/src/ldmsd/ldmsd_request.c
```diff
@@ -9040,17 +9060,18 @@ static int set_default_authz_handler(ldmsd_req_ctxt_t reqc)
...
 		char *endptr;
 		errno = 0;
-		perm = strtol(value, &endptr, 8);
+		long check_perm = strtol(value, &endptr, 8);
 		if (errno || endptr == value || *endptr != '\0') {
 			reqc->errcode = EINVAL;
 			(void) snprintf(reqc->line_buf, reqc->line_len,
 					"String to permission bits conversion failed");
-		} else if (perm < 0 || perm > 0777) {
+		} else if (check_perm < 0 || check_perm > 0777) {
```
tom95858, line 9068: What warning is being silenced by all this? I'm not seeing the benefit of all these type gymnastics.

ldms/src/ldmsd/ldmsd_request.c
tom95858: > Allocations should be freed as soon as there is no more use of them; reset_s is done with as soon as the bool value is assigned. The items here left for the out: sections may be used in in several places up the out.
> 

I'm not familiar with this software implementation guidance.  The code pattern:

``` c
out:
    /* cleanup */
```

... avoids this other code pattern:

``` c
if (broken) {
    free(this);
    free(that);
    free(other);
}
```

... and then 10 lines later:

``` c
if (broken) {
    free(this);
    free(that);
    free(other);
    free(something_else);
}
```

It is a widely used code-pattern that is deployed successfully in software such as the Linux operating system.

> Also, as a pattern used in our team, leaving cleanup of attr strings for an out: block has been very unsuccessful: I've had to clean up a lot of leaks and wild pointers caused by developers mishandling the out block update because it is far away.

This is an inditement of the author and reviewers, not the code pattern.

> The single cleanup block at the end significantly reduces the overall footprint. 

This is provably false I think, which is why we (and notable others) have chosen this cleanup code pattern. It is also why @nichamon is recommending that we use it.

baallan: > 
> This is an inditement of the author and reviewers, not the code pattern.
> 

When the code pattern is 'successfully' used to leak memory by every developer on the project, I contend that it's time to blame the pattern (or the choice of language that necessitates it) rather than the developers.
Once we get in place 'static analysis checks on every compile' as part of the team culture, then we can shift blame back to the developers/reviewers. That being the case, i'll go ahead and do it Mon's suggested way.
The allegedly good pattern, however, also leads to a lot of freeing wild pointers in error handling cases, since we also have a culture of coding minimalist variable initializations.

> > The single cleanup block at the end significantly reduces the overall footprint.

I'm pretty sure I never said that, so I don't know why we would prove it one way or the other.

> This is provably false I think, which is why we (and notable others) have chosen this cleanup code pattern. It is also why @nichamon is recommending that we use it.

ldms/src/ldmsd/ldmsd_request.c
baallan: I suppose I could change it back to just 'long'

ldms/src/ldmsd/ldmsd_request.c
baallan: because the parsing function does not output an unsigned short, leaving admin input errors undiagnosed. See example code on the similar perm case below.

ldms/src/ldmsd/ldmsd_request.c
baallan: @tom95858 The error is that if the user inputs a number which parses fine but gets truncated, then silently they get garbage (incorrect port or perm) as the following code demonstrates.  If you want to range check a number parsed by some function that accepts signed, the initial assignment from the parsing function cannot be to an unsigned short.
```
#include <stdio.h>
#include <stdlib.h>
int main() {
        char x[] = "70411";
        int port = atoi(x);
        short sport = atoi(x);
        printf("p %d, sp %d\n",port, (int)sport);
        return 0;
}
output:
 ./a.out
p 70411, sp 4875
```

And, no, I don't believe for a minute that every user of ldms understands implicitly that ports (in spite of ipv6, or perhaps because of it) are still limited to p < 64k. We need to notice and to warn people when they try to tell us dumb things.

baallan:
@tom95858 I switched to long/atol instead of long long/atoll for parsing the port number. this is ready to go.

PR #1933, opened 2025-08-25
fix leak in ldmsd_row_cache() error path of ldmsd_row_cache.c

baallan:


ldms/src/ldmsd/ldmsd_row_cache.c
```diff
@@ -302,6 +302,7 @@ int ldmsd_row_cache(ldmsd_row_cache_t rcache,
 	rc = clock_gettime(CLOCK_REALTIME, &ts);
```
tom95858, line 302: @baallan, this may cause some heartburn, but ...

- clock_gettime() with these arguments will never produce an error.
- There are ~117 clock_gettime() call sites in our code.
- Three (3) of them check the return code.

I suggest we remove the return code checking from the three call sites and call it a day.

baallan: @tom95858 since the 3 places where there are checks are there to control logic rather than just collect data, I suggest we don't remove them without tangible proof that those arguments always work. I didn't manage to track down that proof in a 20 minute dive in glibc, let alone odd libs on odd hpc hardware.

baallan:
removed 451 milestone from this issue, since it's extremely unlikely to be hit in practice on common systems.

tom95858:
@baallan, if you run through the error codes returned by the function (man page), you can convince yourself that none of the documented errors can be produced.

I'm not sure I understand the "that are there to control logic" means other than to handle the potential error.

baallan:
by 'to control logic'  I mean that the result of the clock fetch is used for things like message rate control/cache lifetime instead of just stored.

To that end, we should probably think about switching to CLOCK_BOOTTIME from realtime for those 3 control bits on any platform where boottime is available (all full linux have it, but I wouldn't bet on things like busybox).

morrone:
This seems like a simple, obvious bug fix. I suggest we just land it, and track future work in a new github issue.

tom95858:
@baallan, @morrone this is a simple change. With the given options, per my read of the man page and the implementation on Arm and Intel this can't fail. The failure modes are _almost_ exclusively for set(), not get(). The failure mode for get() is for dynamic clocks, which we are not using here. I'm not concerned about the size of this, but rather other contributors reading this and thinking that it is necessary or advised. It is not.

baallan:
We’re targeting processors and kernels for the next 10 years including exotic test beds on custom silicon and 8 socket 500 core x86 platforms with mixes of p-cores and e-cores.

Assuming the status quo (no dynamic clocks, no tickless kernels, only intel/arm) forever seems to me a very bad choice that costs us practically nothing to avoid.

________________________________
From: Tom Tucker ***@***.***>
Sent: Wednesday, November 12, 2025 9:30:28 PM
Cc: Allan, Ben Andrew ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [ovis-hpc/ldms] fix leak in ldmsd_row_cache() error path of ldmsd_row_cache.c (PR #1933)

[https://avatars.githubusercontent.com/u/7026652?s=20&v=4]tom95858 left a comment (ovis-hpc/ldms#1933)<https://github.com/ovis-hpc/ldms/pull/1933#issuecomment-3525290189>

@baallan<https://github.com/baallan>, @morrone<https://github.com/morrone> this is a simple change. With the given options, per my read of the man page and the implementation on Arm and Intel this can't fail. The failure modes are almost exclusively for set(), not get(). The failure mode for get() is for dynamic clocks, which we are not using here. I'm not concerned about the size of this, but rather other contributors reading this and thinking that it is necessary or advised. It is not.

tom95858:
@baallan, I'm not suggesting there are no dynamic clocks. I'm suggesting that our code doesn't use them.

PR #1934, opened 2025-08-25
fixes to ldmsd_stream.c

baallan:
separable commits for fixes to memory and other bugs in ldmsd_stream

### DIFF
--- a/ldms/src/ldmsd/ldmsd_stream.c
+++ b/ldms/src/ldmsd/ldmsd_stream.c
@@ -650,6 +650,7 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
 		return ENOMEM;
 	}
 
+	char * b = NULL;
 	buffer = malloc(max_msg_len);
 	if (!buffer) {
 		msglog("Out of memory\n");
@@ -681,7 +682,6 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
 		goto err;
 	}
 
-	char * b = NULL;
 	msg_no = ldmsd_msg_no_get();
 	rc = fseek(file, 0L, SEEK_END);
 	if (!rc) {


ldms/src/ldmsd/ldmsd_stream.c
```diff
@@ -650,6 +650,7 @@ int ldmsd_stream_publish_file(const char *stream, const char *type,
 		return ENOMEM;
 	}
 
+	char * b = NULL;
```
tom95858, line 653: @baallan, I'm guessing the tool complained because there are goto's (goto err;) above the original declaration and the code at err: calls free(b).

That's fine, but that's not why this code is broken. If we go down to line 719 - 724, you'll see that we malloc(max_msg_len) and then call fgets(..., sizeof(b) -1, ...)

Well that sizeof(b)-1 is 7 bytes, not max_msg_len - 1.

So that's going to read our message 7B at a time and each time 7B is too small, it will realloc the destination buffer * 2 because it _thinks_ it's failing to read max_msg_len bytes, not 7 bytes.

Can you fix this please?

tom95858 (changes requested):

ldms/src/ldmsd/ldmsd_stream.c
baallan: @tom95858 please have someone more expert in that chunk of code fix the more subtle (and probably important) bug. probably you or @nichamon based on the git blame.  This change is independent.

baallan: @nichamon thanks for #1940

baallan:
@tom95858 @emdonat it appears the bugs fixed here are in the path of stream_status, which we call regularly in production. So I'm leaving it as included in the 451 milestone.

PR #1936, opened 2025-08-25
fix leak in an error path of ldmsd_updtr_prdcr_add

baallan:


ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -1482,6 +1482,7 @@ int ldmsd_updtr_prdcr_add(const char *updtr_name, const char *prdcr_regex,
 	if (rc) {
 		rc = EINVAL;
 		regfree(&prd_match->regex);
```
tom95858: We don't need to call regfree() if regcomp() fails.

ldms/src/ldmsd/ldmsd_updtr.c
```diff
@@ -1482,6 +1482,7 @@ int ldmsd_updtr_prdcr_add(const char *updtr_name, const char *prdcr_regex,
...
+		free(prd_match->regex_str);
```
tom95858, line 1484: This is goodness.

ldms/src/ldmsd/ldmsd_updtr.c
baallan: While the man page doesn't explicitly your regfree claim, the glibc regcomp source seems to be consistent with it. I'll delete the regfree call in this error path.

tom95858: > While the man page doesn't explicitly your regfree claim, the glibc regcomp source seems to be consistent with it. I'll delete the regfree call in this error path.

It doesn't break anything, it just doesn't do anything. I'm fine with it either way. I was just letting you know.

baallan:
@emdonat @tom95858 removing this from 4.5.1 milestone since fixes are in error paths that admins are not likely to pound repeatedly without fixing their input files.

PR #1938, opened 2025-08-25
fix free(wildptr) and leak in error paths of cxi_sampler.c

baallan:


baallan:
@emdonat @tom95858 i'm removing this from 451 milestone, as the changes are all in error paths that correct plugin configuration by admin will avoid.

PR #1939, opened 2025-08-25
fix file leak and other bugs in linux_proc_sampler for 451

baallan:


baallan:
@tom95858 this includes fixing a file pointer leak in the normal path for linux_proc_sampler, which we are eager to deploy as part of appsysfusion.

PR #1940, opened 2025-08-25
Fix buffer size in ldmsd_stream_publish_file() fgets call

nichamon:
Replace sizeof(b) with max_msg_len-1 in a fgets call. Since b is a malloc'd pointer, sizeof(b) returns the pointer size (8 bytes) rather than the allocated buffer size, severely limiting read capacity.

This ensures fgets() can utilize the full allocated buffer size when reading from the file.


baallan:
@tom95858 we still need this merged.

PR #1941, opened 2025-08-25
fix logging nit and potential null pointer dereference in ldmsd.c

baallan:


baallan:
@emdonat @tom95858 removing this from 4.5.1 requirements since the resulting crash can be avoided by the admin supplying correct input. This fixes an admin quality of life issue (robustly handling bad input) only.

PR #1943, opened 2025-08-25
fix error handling cases and message in ldmsctl.c

baallan:


baallan:
@tom95858 @emdonat removing this from 4.5.1 milestone since, afaict, lack of the bug fixes here doesn't damage the daemon being queried. The bug fixes affect end-user experience instead.

ldms/src/ldmsd/ldmsctl.c
```diff
@@ -1847,9 +1847,9 @@ static void resp_stream_status(ldmsd_req_hdr_t resp, size_t len,
 	}
 	rc = json_parse_buffer(parser, (char*)attr->attr_value, len, &json);
 	json_parser_free(parser);
+	parser = NULL;
```
tom95858: This is unnecessary.

baallan: you are correct that the null assignment is redundant with the deletion of the extra free.
will fix.

tom95858:
@baallan, please see issue #2099

ldms/src/ldmsd/ldmsctl.c
baallan: @tom95858 fixed the extra assignment. ready to merge.

baallan:
> @baallan, please see issue #2099

@tom95858  i don't see any relation between #1943 and #2099

PR #1944, opened 2025-08-25
Fix genders function filter

jennfshr:
Extra-long genders files contribute to slow starts;  the filter for relevant attributes solves this.  Only relevant for 4.4.x.


morrone:
"bootnode" sounds site-specific to me at first glance, but I could be wrong. I think @baallan should be the one to give the OK on this change.

baallan:
The bootnode gender dependence is documented in the ldms-attributes man page, and used to be used across the tri-lab. 
So I'm good with this change. @emdonat

baallan (approved):

PR #1947, opened 2025-08-26
Make ldmsd_plug_handle_t a pointer to an opaque struct

morrone:
Convert ldmsd_plug_handle_t from a a void pointer to a pointer to an opaque struct. The new new opaque struct ldmsd_plug_api_s is initially implmented as a direct derivative of struct ldmsd_cfgobj.

This improves type checking on the consumer (plugin) side of the API.


morrone:
@tom95858 ready to land?

tom95858:
@nichamon, I think this adds what is needed to ensure the concurrency guarantees for config(), sample()/config(), commit_rows()/config(), etc...

What do you think. 

And yes, @morrone, I think this is ready.

morrone:
> @nichamon, I think this adds what is needed to ensure the concurrency guarantees for config(), sample()/config(), commit_rows()/config(), etc...

@tom95858 I don't think that I understand that statement. This PR only changed the data type, not any actual semantics. So I am not clear on how this would relate to concurrency guarantees.

tom95858:
@morrone, it adds a data structure into which we can place what we need for some work being planned. I wanted @nichamon to be aware of it to avoid having her duplicate any work.

morrone:
> @morrone, it adds a data structure into which we can place what we need for some work being planned. I wanted @nichamon to be aware of it to avoid having her duplicate any work.

@tom95858 This patch is _just_ type punning. You _cannot_ add additional data to struct ldmsd_plug_api_s. That would make this punned struct name no longer compatible with the other type-punned cfgobj structs. (@nichamon just to make sure she sees this comment as well).

You could _reimplement_ struct ldmsd_plug_api_s, and how it is tracked throughout the code. And you could do that without the plugins being any the wiser (that is a feature of using abstraction in an API). But that would require modification of a lot of the cfgobj code as well.

So you can replace the current implementation of the ldmsd_plug_api_s struct, but you cannot just add things to the struct as it is currently implemented.

PR #1948, opened 2025-08-27
backport 1844 dcgm fixes

baallan:


baallan:
> Also, did you mean to leave out the change to ldms/scripts/examples/dcgm1.1? 

yes. the new dcgm_no_base case file cover running without the sampler_base.
use_base=1 in the original dcgm1.1 enables the sampler_base.


side note: as mentioned in the man page, use_base=<anything> turns on the sampler base; 0 vs 1 is meaningless. it should really be a keyword instead of an attribute, but at the time of coding the maestro problem of not supporting  keywords had not yet been fixed.

morrone (approved):
OK, looks good now. @emdonat I think this is ready to land.

PR #1955, opened 2025-08-29
quiet enobufs messages and fix reconnect_interval over-sed that broke netlink tests

baallan:
this adds the --log-nobufs option; the default is to no longer complain about kernel-side ring buffer inadequacy unless requested.
it also fixes uncaught incorrect replacement of reconnect with reconnect_interval in netlink invocations in test scripts which was already fixed on main.

backport of #1954 to b4.4


morrone:
This doesn't look like a straight backport of #1954. This looks like a backport plus some other fix squashed together into one commit. Could you please keep the backport commit separate from the other change?

baallan:
The fixes to other scripts are repairs to the tests of the code touche by the backport that had a bit of bitrot relative to main, and as such are not separable.

PR #1956, opened 2025-09-01
Change update hint validation log from ERROR to INFO level

nichamon:
Change the update hint validation from ERROR to DEBUG level when encountering invalid hints like '0:0'. This is expected behavior when sets exist but sampling hasn't started, or for event-driven samplers that don't require starting the sampler plugins.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_prdcr.c
+++ b/ldms/src/ldmsd/ldmsd_prdcr.c
@@ -325,7 +325,7 @@ static void __update_set_info(ldmsd_prdcr_set_t set, ldms_dir_set_t dset)
 
 		/* Sanity check the hints */
 		if (offset_us >= intrvl_us) {
-			ovis_log(prdcr_log, OVIS_LERROR, "set %s: Invalid hint '%s', ignoring hint\n",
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_prdcr.c
```diff
@@ -325,7 +325,7 @@ static void __update_set_info(ldmsd_prdcr_set_t set, ldms_dir_set_t dset)
 
 		/* Sanity check the hints */
 		if (offset_us >= intrvl_us) {
-			ovis_log(prdcr_log, OVIS_LERROR, "set %s: Invalid hint '%s', ignoring hint\n",
+			ovis_log(prdcr_log, OVIS_LDEBUG, "set %s: hint '%s', ignoring hint\n",
```
tom95858: @nichamon, given that we removed the setting of the hint in ``ldmsd_set_register``, should we make this INFO instead of DEBUG?

nichamon: I think so. This will provide a clue when an updater with auto-interval enabled doesn't update a set. I'll update the patch.

PR #1960, opened 2025-09-02
DRAFT: backport lustre sampler documentation and OSD_SEARCH_PATH to b4.4

jennfshr:
This pull request accomplishes the following updates to the various lustre samplers:

1. Update documentation for debugfs requirement
2. Update documentation for Lustre 2.15
3. Switch `#define OSD_SEARCH_PATH "/proc/fs/lustre"` to use array `possible_osd_base_paths` including `/sys/fs/lustre`

### DIFF
--- a/ldms/src/sampler/lustre_client/lustre_client.c
+++ b/ldms/src/sampler/lustre_client/lustre_client.c
@@ -23,8 +23,8 @@
 
 /* locations where llite stats might be found */
 static const char * const llite_paths[] = {
-        "/proc/fs/lustre/llite",          /* lustre pre-2.12 */
         "/sys/kernel/debug/lustre/llite"  /* lustre 2.12 and later */
+        "/proc/fs/lustre/llite",          /* lustre pre-2.12 */
 };
 static const int llite_paths_len = sizeof(llite_paths) / sizeof(llite_paths[0]);
 static char *test_path = NULL;
--- a/ldms/src/sampler/lustre_mdt/lustre_mdt.c
+++ b/ldms/src/sampler/lustre_mdt/lustre_mdt.c
@@ -20,7 +20,12 @@
 #define _GNU_SOURCE
 
 #define MDT_PATH "/proc/fs/lustre/mdt"
-#define OSD_SEARCH_PATH "/proc/fs/lustre"
+
+static const char * const possible_osd_base_paths[] = {
+	"/sys/fs/lustre", /* at least lustre >= 1.15 (probably >= 1.12) */
+	"/proc/fs/lustre", /* older lustre */
+	NULL
+};
 
 static struct comp_id_data cid;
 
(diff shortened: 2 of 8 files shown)


ldms/src/sampler/lustre_client/lustre_client.c
```diff
@@ -23,8 +23,8 @@
 
 /* locations where llite stats might be found */
 static const char * const llite_paths[] = {
-        "/proc/fs/lustre/llite",          /* lustre pre-2.12 */
         "/sys/kernel/debug/lustre/llite"  /* lustre 2.12 and later */
+        "/proc/fs/lustre/llite",          /* lustre pre-2.12 */
```
morrone, line 27: This is buggy. There is no comma between the two strings, so they are merged into one string by the compiler.

ldms/src/sampler/lustre_mdt/lustre_mdt.c
```diff
@@ -84,7 +89,7 @@ static struct mdt_data *mdt_create(const char *mdt_name, const char *basedir)
         mdt->general_metric_set = mdt_general_create(producer_name, mdt->fs_name, mdt->name, &cid);
         if (mdt->general_metric_set == NULL)
                 goto out7;
-        mdt->osd_path = mdt_general_osd_path_find(OSD_SEARCH_PATH, mdt->name);
+        mdt->osd_path = mdt_general_osd_path_find(possible_osd_base_paths, mdt->name);
```
morrone, line 92: This does not work. mdt_general_osd_path_find() has a single string as its first parameter. We can't just change that to being a list without also changing the code of mdt_general_osd_path_find().

morrone (changes requested):
There are some bugs that I commented inline.

Also, when backporting, it is best to directly packport individual commits using "git cherry-pick" as the starting point. Granted, things have moved around quite a lot in this area, so there will be some work to keep the commits aligned.

If that isn't feasible, then the commit message should at least make it clear what it is being backported. "I took foo from commit A, and bar from commit B, and it wasn't feasible to keep the commits in the same form because of X".

ldms/src/sampler/lustre_mdt/lustre_mdt.c
Matt8687: In the second reopening of the PR that function was edited to contain most of the guts from the `*lustre_osd_dir_find` function within ldms/src/sampler/lustre_newer/lustre_shared.c edited in !52cc6d0 by you (@morrone) and committed by @tom95858. The main changes to the shared function are dropping the passed log argument and the corresponding usage lines there within. You can see here: https://github.com/ovis-hpc/ldms/pull/1960/files#diff-7838ef404a165ad360fa8c47eb930cdfc8d164a69053a55b746b0b7cdae07141

In *theory* that function call should be fine?

morrone: I have no idea what is going on. At the time of my review, I didn't see associated code change to mdt_general_osd_path_find. I think things may have changed out from under me while my review was in progress.

morrone: Yes, looking at the PR history, I am pretty sure that the the code that changes mdt_general_osd_path_find was not present at the time of my review.

PR #1961, opened 2025-09-03
Reduce lustre sampler global usage and improve code sharing

morrone:
A series of commits to reduce the usage of globals (moving in the general direction of being mulit-instance capable).

Also improve code sharing a bit.


morrone:
@tom95858 Any chance this can land soon?

PR #1962, opened 2025-09-03
lustre: search for osd dir in multiple places (4.4.6 backport logic)

jennfshr:
Details of this backport incorporate pulling in requisite modifications to lustre_{mdt,ost,mdc,client} samplers to enable support for searching an array of paths, updates to documentation, and new tests and some small fixes.

### DIFF
--- a/ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
+++ b/ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
@@ -116,41 +116,43 @@ void mdt_general_schema_fini()
 }
 
 /* Returns strdup'ed string or NULL.  Caller must free. */
-char *mdt_general_osd_path_find(const char *search_path, const char *mdt_name)
+char *mdt_general_osd_path_find(const char * const *paths, const char *component_name)
 {
-        struct dirent *dirent;
-        DIR *dir;
-        char *osd_path = NULL;
+        char *path;
+        char *osd_dir = NULL;
+        int i;
 
-        dir = opendir(search_path);
-        if (dir == NULL) {
-                return NULL;
-        }
+        for (i = 0, path = (char *)paths[0]; path != NULL; i++, path = (char *)paths[i]) {
+                struct dirent *dirent;
+                DIR *dir;
 
-        while ((dirent = readdir(dir)) != NULL) {
-                if (dirent->d_type == DT_DIR &&
-                    strncmp(dirent->d_name, "osd-", strlen("osd-")) == 0) {
-                        char tmp_path[PATH_MAX];
-                        snprintf(tmp_path, PATH_MAX, "%s/%s/%s",
-                                 search_path, dirent->d_name, mdt_name);
-                        if (access(tmp_path, F_OK) == 0) {
-                                osd_path = strdup(tmp_path);
-                                break;
+                dir = opendir(path);
+                if (dir == NULL) {
+                        continue;
+                }
+
+                while ((dirent = readdir(dir)) != NULL) {
+                        if (dirent->d_type == DT_DIR &&
+                            strncmp(dirent->d_name, "osd-", strlen("osd-")) == 0) {
+                                char tmp_path[PATH_MAX];
+                                snprintf(tmp_path, PATH_MAX, "%s/%s/%s",
+                                path, dirent->d_name, component_name);
+                                if (access(tmp_path, F_OK) == 0) {
+                                        osd_dir = strdup(tmp_path);
+                                        break;
+                                }
                         }
                 }
+
+                closedir(dir);
+
+                if (osd_dir != NULL) {
+                        break;
+                }
+
         }
 
-        closedir(dir);
-
-        if (osd_path != NULL) {
-                log_fn(LDMSD_LDEBUG, SAMP" for mdt %s found osd path %s\n",
-                       mdt_name, osd_path);
-        } else {
-                log_fn(LDMSD_LWARNING, SAMP" osd for mdt %s not found\n",
-                       mdt_name);
-        }
-
-        return osd_path;
+    return osd_dir;
 }
 
 static uint64_t file_read_uint64_t(const char *dir, const char *file)
--- a/ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
+++ b/ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
@@ -151,8 +151,15 @@ char *mdt_general_osd_path_find(const char * const *paths, const char *component
                 }
 
         }
+        if (osd_dir != NULL) {
+                log_fn(LDMSD_LDEBUG, SAMP" for mdt %s found osd dir %s\n",
+                       component_name, osd_dir);
+        } else {
+                log_fn(LDMSD_LWARNING, SAMP" osd for mdt %s not found\n",
+                       component_name);
+        }
 
-    return osd_dir;
+        return osd_dir;
 }
 
 static uint64_t file_read_uint64_t(const char *dir, const char *file)
--- a/ldms/src/sampler/lustre_ost/lustre_ost.c
+++ b/ldms/src/sampler/lustre_ost/lustre_ost.c
@@ -20,7 +20,12 @@
 #define _GNU_SOURCE
 
 #define OBDFILTER_PATH "/proc/fs/lustre/obdfilter"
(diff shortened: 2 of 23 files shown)


morrone:
This is much improved!

I would like to see things reorganized. I would like to see the first commit in this PR be a clean backport of the "lustre: search for osd dir in multiple places" commit, commit 52cc6d043b25f58985a676efa02bdbafbde4e83a. I think this does mostly accomplish that, but then it smushes in a number of other things that are not part of the original commit.

The documentation changes happen on main in the commit "add a note about debugfs for lustre plugins", commit aeabc2bf114b55b17a8130e7e7d0cb39520f118d. That should be backported in a separate commit in this PR.

Neither of those commits on main touch lustre_client.c, and since that change doesn't change functionality, I would argue we should not make that change on b4.4.

I haven't looked up where all the changes to scripts/examples came from, but those should be in a commit of their own as well. If they are new, then explain that in the commit message. If those are backported, then include a separate backport commit (or commits) for those.

The advantage of the approach I lay out is that it makes it easier on us to keep the branches in sync, when necessary. When a human does "git log --graph --oneline b4.4 main", they can look at the branch list and go "OK, I see 'lustre: search for osd dir in multiple places' on both branches, and I see 'add a note about debugfs for lustre plugins' on both branches. I don't need to worry about syncing those". But if we allow adding stuff to porting commits that wasn't part of the original commit, then we can't make any assumptions about what has and hasn't been ported. We need to do a lot of additional leg work.

So a little effort and process at development time saves us a lot of effort and sanity down the line. And once this effort becomes habbit, it becomes pretty easy to do porting of commits in the way that makes syncing easy.

And please don't abbreviate a version like 4.4.6 as four-hundred-and-forty-six (446). We can afford the dots, I think. :)

ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
```diff
@@ -116,41 +116,43 @@ void mdt_general_schema_fini()
...
+                                }
                         }
                 }
-        }
 
-        closedir(dir);
+                closedir(dir);
+
+                if (osd_dir != NULL) {
+                        break;
```
morrone, line 150: The backport seems to have lost the log messages. (same in lustre_ost_general.c).

morrone (changes requested):

baallan:
I've had difficulty verifying the lustre_ost sampling using this with a recent toss release. @morrone, independent of patch formulation, does ost sampling work for your system(s) with either this or on the main branch?

morrone:
> I've had difficulty verifying the lustre_ost sampling using this with a recent toss release. @morrone, independent of patch formulation, does ost sampling work for your system(s) with either this or on the main branch?

I think it was working on main, but I haven't tried the code in this PR. I will double check the main code.

morrone:
> > I've had difficulty verifying the lustre_ost sampling using this with a recent toss release. @morrone, independent of patch formulation, does ost sampling work for your system(s) with either this or on the main branch?
> 
> I think it was working on main, but I haven't tried the code in this PR. I will double check the main code.

Yes, it looks to me like lustre_ost and lustre_mdt are working on main on my test system. If you are seeing issues, please do report them.

baallan:
After verifying the data source location as /proc/fs/lustre/obdfilter/*/stats, the following is true of the ost sampler backport:
- write_bytes.sum is reported correctly (so found the right file)
- the bulk of the metrics are not collected correctly (appears to be the 'expanded line format' problem is not addressed fully in the backport). 

Lines similar to 268-279 of lustre client (updated in the link below) did not get backported to lustre_ost_general.c
https://github.com/ovis-hpc/ldms/commit/0029b8033852695495db6b3f5957021cbd6c52e1#diff-1f950dde0fa49c855e89d1cede5ab967e397b5946ca43668224ad576d5813e9a

https://github.com/ovis-hpc/ldms/blob/main/ldms/src/sampler/lustre_newer/lustre_shared.c : lustre_stats_file_sample has a less commenty version of the same logic.

ldms/src/sampler/lustre_ost/lustre_ost.c
```diff
@@ -93,6 +93,10 @@ static struct ost_data *ost_create(const char *ost_name, const char *basedir)
         rbn_init(&ost->ost_tree_node, ost->name);
         rbt_init(&ost->job_stats, string_comparator);
 
+	log_fn(LDMSD_LINFO, SAMP
```
morrone: We shouldn't really be introducing new code in a backport PR.

morrone:
Maybe this PR should switch back to draft while development is underway? It looks like there is still a fair bit to do (good progress, though!). The patch stack needs a fair bit of massaging to get it into landing ready state. We want the patch stack to reflect, one-to-one, the source commits that were tapped to make the backported code.

jennfshr:
@morrone @emdonat @baallan this is updated to address Chris' complaints on the commit structure and we have tested every plugin on applicable appliances.  Looks like everything's functioning as expected now.  We're leaving the info output to inform on where the job_stats, stats, and osd found.

ldms/src/sampler/lustre_ost/lustre_ost.c
jennfshr: Leaving for the benefit of having details on the origin of the data collected outweighs the benefit of no original code in the b4.4 branch.

ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
jennfshr: fixed

morrone:
> @morrone @emdonat @baallan this is updated to address Chris' complaints on the commit structure and we have tested every plugin on applicable appliances. Looks like everything's functioning as expected now. We're leaving the info output to inform on where the job_stats, stats, and osd found.

Well, maybe I didn't make my commit structure request clear. Here's the idea:

There should be a one-to-one mapping of backported commits in this PR to commits from the source branch. For instance, this PR backports code from this commit on main:

* 52cc6d043 2025-07-12 lustre: search for osd dir in multiple places

So I would expect to see a single commit in this PR with exactly the same title, "lustre: search for osd dir in multiple places" that backports the code from that commit. That is not currently what we have. We have a very different arrangement of commits here.

baallan:
@morrone while the commit structure you propose might be ideal, it just isn't reasonable since almost nothing in this was cherry-pickable back from main. The quoting of the original commit title on main in the titles here seems like a reasonable compromise.

morrone:
> @morrone while the commit structure you propose might be ideal, it just isn't reasonable since almost nothing in this was cherry-pickable back from main. The quoting of the original commit title on main in the titles here seems like a reasonable compromise.

I don't think that is the case. Yes, they don't cherry-pick cleanly, but that doens't mean that the commits cannot be one-to-one. It just means more work in the porting process (and that work is unavoidable, and has been mostly done at this point). Combining the current PR's commits:

* lustre_mdt: search for osd dir in multiple places
* lustre_ost: search for osd dir in multiple places 

into one "lustre: search for osd dir in multiple places" commit to match the source commit it probably half of what is needed.

The commit "lustre_mdt: replace log function" was pretty obviously a correction to something that was forgotten in the backport, so that should be merged into the backport commit as well.

Repeat that process until everything is organized into clear one-to-one commits. Unless the code is genuinely new (a debug message we discussed earlier would be a separate commit, since that was not part of the backport).

ldms/src/sampler/lustre_mdt/lustre_mdt_general.c
```diff
@@ -151,8 +151,15 @@ char *mdt_general_osd_path_find(const char * const *paths, const char *component
                 }
 
         }
+        if (osd_dir != NULL) {
```
morrone, line 154: This is good to revive the logging. Don't forget we need the same in lustre_ost_general.c

morrone:
Just to put my money where my mouth is, I did the backporting that I asked for. See PR #1996. Hopefully that makes it clearer what we are looking for, and how to keep a backport of patches one-to-one. I cherry-picked the commit from here that changes the scripts files as well.

morrone:
In the process of doing my backport in  PR#1996, I noticed that this PR is missing the updated stats parsing for the mdt. It only applied it to the ost. In #1996 I did a straight backport of 83893e9, so I applied the same new stats parsing code to lustre_client, lustre_mdt, and lustre_ost.

jennfshr:
Closing this PR as @morrone supplied his own implementation, and this is duplicative of that backport.

PR #1963, opened 2025-09-04
`dispatch-container-bake` workflow on release event

morrone:
Move `dispatch-container-bake` job into its own workflow and change the event the workflow depends on from `on push tags` to `on release published` so that when a release is created (by any means; not just by matching pushed tag) our workflow will dispatch the container bake workflow in `narategithub/ldms-containers` repository.

Also conditioned this workflow to run only on `ovis-hpc/ldms` or `narategithub/ldms`, and change SOS branch to `b2.6`.

[fixup] dispatch

Backported from 10a9fd52ba9d023ba94e9f0df26593fd2b6ee222 on main.


morrone:
@emdonat I think it is a good idea to land this backported workflow change before we tag the 4.4.6 release.

PR #1964, opened 2025-09-08
Fix ldmsd_xprt_term() incorrectly removing config file contexts

nichamon:
When a connection disconnects, ldmsd traverses the request message tree to remove contexts associated with that connection. However, the tree contains contexts from both config files and transports.

The current logic doesn't check the flags when determinnig whether a context belongs to a transport, causing it to incorrectly remove config file contexts that happen to share the same connect ID.

The fix ensures only transport contexts are removed during disconnection cleanup.


tom95858:
@baallan, we think this will fix the assert(). Could you please test?

tom95858:
@nichamon, could you please look at combining __get_conn_id() and __get_cfgfile_id(). We really should not have the possibility of the fields in the key colliding regardless of the flags.

Issue #1626, opened 2025-03-10
patch needs patching


baallan:
https://github.com/ovis-hpc/ldms/blob/43c06c1996bbf3ab176c8ffdf99309a1c989a731/lib/src/ovis_json/ovis_json.c#L741

Changed a cast from uint64_t to int; as the json int type for this library is int64_t, this new cast is incorrect. The new version truncates longer int values that should be storable.

If the function was being called with an json_int enum and an int* value, the original call site was wrong (may have needed an (int64_t) cast).

tom95858:
The call sites are currently passing in 'int'. If we change it back to int64_t (which is what it was), we need to:

- Add documentation to the ovis_json indicating that our implementation of JSON_INT is a signed 64bit integer, and
- Fix all the call sites, which there may be very few ...

baallan:
not 'documented' but in the json header:
```
struct json_entity_s {
	enum json_value_e type;
	union {
		int bool_;
		int64_t int_;
		double double_;
		json_str_t str_;
		json_attr_t attr_;
		json_list_t list_;
		json_dict_t dict_;
	} value;
	TAILQ_ENTRY(json_entity_s) item_entry;
};
```
so it's clear that we like jansson and python support int64_t.

emdonat:
> The call sites are currently passing in 'int'. If we change it back to int64_t (which is what it was), we need to:
> 
> * Add documentation to the ovis_json indicating that our implementation of JSON_INT is a signed 64bit integer, and
> * Fix all the call sites, which there may be very few ...

Is this a good path forward then and agreeable to interested parties?

tom95858:
I agree with @baallan that all these call sites should be int64_t. Unfortunately, the vararg interface doesn't allow us to enforce that. It can only be done by inspection/testing.

emdonat:
Closed per #1967

PR #1967, opened 2025-09-10
Fix integer truncation in JSON_INT_VALUE creation

nichamon:
A previous commit changed va_arg() from uint64_t to int in __attr_value_new() to match call sites passing int values. However, this introduced truncation of 64-bit integers since the ovis_json library internally stores JSON integers as int64_t.

The patch restores the correct behavior in __attr_value_new() of the JSON_INT_VALUE type, ensures that call sites pass int64_t values, and adds documentation clarifying the int64_t requirement.

Fixes issue #1626

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.c
+++ b/ldms/src/ldmsd/ldmsd_request.c
@@ -9368,7 +9368,7 @@ static json_entity_t __ldmsd_stat2dict(struct ldmsd_stat *stat)
 				JSON_FLOAT_VALUE, "max", stat->max,
 				JSON_FLOAT_VALUE, "max_ts", max_ts,
 				JSON_FLOAT_VALUE, "avg", stat->avg,
-				JSON_INT_VALUE, "count", stat->count,
+				JSON_INT_VALUE, "count", (int64_t)stat->count,
 				JSON_FLOAT_VALUE, "start_ts", start_ts,
 				JSON_FLOAT_VALUE, "end_ts", end_ts,
 				-1);
--- a/lib/src/ovis_json/ovis_json.c
+++ b/lib/src/ovis_json/ovis_json.c
@@ -732,13 +732,13 @@ json_entity_t __attr_value_new(int type, va_list *ap)
 	json_entity_t v, item;
 	switch (type) {
 	case JSON_BOOL_VALUE:
-		v = json_entity_new(type, va_arg(*ap, int));
+		v = json_entity_new(type, va_arg(*ap, int32_t));
 		break;
 	case JSON_FLOAT_VALUE:
 		v = json_entity_new(type, va_arg(*ap, double));
 		break;
 	case JSON_INT_VALUE:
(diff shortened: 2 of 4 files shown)


nichamon:
@tom95858 and @baallan I found only one call site to `json_entity_new()` and `json_dict_build()` that use `JSON_INT_VALUE`. Please let me know if I missed any call sites.

tom95858:
@nichamon, The issue is that on ARM32 an 'int' is 32b. On X86_64, an 'int' is 64b. On x64_64, we never saw an issue because passing in uint64_t from the stats structure didn't cause an issue but on arm, the count was garbage.

Any code decoding the va_list **cannot** use an 'int' arg because if the caller passes in an int64_t or uint64_t it will be pulling off the top 32b of the value. This applies to the JSON_BOOL value too. In general, we can't use a type that (like 'int') in the va_arg() macros may have a length that varies across platforms.

All that said, the fix is OK, except that BOOL has to be fixed too and the description doesn't exactly reflect the problem. It's not that ovis_stores numbers as int64_t, that doesn't really matter. What matters is that the call site is pushing typeof(arg), but the code is pulling off (int) *(void *)stack-pointer, which is pulling off the top 32b of 'arg' AND buggers up the stack pointer for subsequent calls because the macro will do sp += sizeof(int) instead of sp += sizeof(arg).

baallan:
there's an ambiguity for bool in entity_new use by ovis_json_lexer.l that may be easy to overlook.

baallan:
@tom95858

not part of this ticket, but what do you think of letting the compiler help us avoid this problem in the future by adding 

```json_[bool,int,float,null]_new? We already have json_[list,dict,attr]_new.```
 It seems a little asymmetric.

I'm not suggesting deleting json_entity_new (it probably has a place in copying json trees), but most of the one-item locations could use the specific functions and type-check for us.
```

tom95858:
@baallan, I have no issue with adding these types, however, the compiler can't help us because the parameter list is a vararg and therefore the compiler can't check the types. The caller, in all cases, needs to explicitly cast if the argument is not one of the JSON types.

What we might consider is adding macros like:
``` c
#define JSON_INT_ARG(_a_) \
    JSON_INT_TYPE, (int64_t)_a_
```

Then callers would do this:

``` c
int rc = json_build_dict(..., JSON_INT_ARG(struct->my_int32_member), JSON_BOOL_ARG(struct->my_int64_member));
```
This would force the call-site cast that we need to avoid confusing the consumer of the va_list.

nichamon:
@tom95858 I have address JSON_BOOL_VALUE by using `int32_t`. There is no call site that uses `JSON_BOOL_VALUE`.

Regarding the macro, I can add that to this patch or create another PR for it when we have reached an agreement.

tom95858:
I think the macros are nice and you could demonstrate their use in the
stats callsite.

Thomas Tucker, President, Open Grid Computing, Inc.

PR #1969, opened 2025-09-11
Remove dead ldmsctl ids

tom95858:


tom95858:
This is broken. Also, see issue #2099

PR #1970, opened 2025-09-11
fixed ldms.c missing NULL pointer checks yielding unexplained crashes

baallan:
These are all sites where plugin api misuse can lead to unexplained daemon disappearance.
In most cases, checks are done in similar functions and just overlooked here.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -1722,6 +1722,13 @@ ldms_set_t ldms_set_create(const char *instance_name,
 	int set_array_card;
 	struct ldms_set *set;
 
+	if (!instance_name || !schema) {
+		ovis_log(NULL, OVIS_LERROR, "NULL instance_name or schema "
+				"passed to ldms_set_create\n");
+		errno = EINVAL;
+		return NULL;
+	}
+
 	/*
 	 * Ensure the instance_name does not contain characters
 	 * that can't be encoded in a quoted string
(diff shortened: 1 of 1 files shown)


ldms/src/core/ldms.c
```diff
@@ -1722,6 +1722,13 @@ ldms_set_t ldms_set_create(const char *instance_name,
 	int set_array_card;
 	struct ldms_set *set;
 
+	if (!instance_name || !schema) {
+		ovis_log(xlog, OVIS_LERROR, "NULL instance_name or schema "
```
nichamon: `xlog` is the log handler for LDMS transport log messages. Either use NULL or let the caller prints the log is more appropriate.

Wouldn't the check hide the bug in the caller code that passes NULL to this function?

ldms/src/core/ldms.c
```diff
@@ -2600,13 +2607,17 @@ void ldms_metric_set(ldms_set_t s, int i, ldms_mval_t v)
 	ldms_mdesc_t desc;
 	ldms_mval_t mv;
 
-	if (i < 0 || i >= __le32_to_cpu(s->meta->card))
-		assert(0 == "Invalid metric index");
+	if (!s || i < 0 || i >= __le32_to_cpu(s->meta->card)) {
+		ovis_log(xlog, OVIS_LERROR, "NULL set or bad index passed to "
```
nichamon: Similarly, `xlog` shouldn't be used as the log handler for this log message.

nichamon:
@baallan Are there currently any crashes due to any misused referred in this patch? If so, the bugs are in the caller code and we should root cause them.

ldms/src/core/ldms.c
baallan: The log messages are added because the plugin caller (and more importantly the innocent user of a plugin) of an api shouldn't have to look inside the api implementation to figure out what went wrong when hitting a null pointer crash or an assert.  How does logging a problem at the error level and returning a NULL pointer to be handled hide it from the developer?

I can certainly add an alternative log pointer with a different name; seems like a reasonable concern. What would be the name?

baallan:
> @baallan Are there currently any crashes due to any misused referred in this patch? If so, the bugs are in the caller code and we should root cause them.

I don't have any valgrind or log evidence of the in-production plugins hitting these problems, but then again we don't have CI jobs for full coverage testing for most of the non-generic-linux plugins that are in production. Getting that coverage is on our lab todo list for next year.

ldms/src/core/ldms.c
tom95858: You can use NULL to go to the default log, or you can lookup the subsystem and pass it in. I don't think any of the pre-defined subsystems fit any better than NULL though.

ldms/src/core/ldms.c
nichamon: I agree the NULL return should make these issues visible to developers.

I agree with Tom that `ovis_log(NULL, ...)` is the best candidate.

nichamon:
> I don't have any valgrind or log evidence of the in-production plugins hitting these problems, but then again we don't have CI jobs for full coverage testing for most of the non-generic-linux plugins that are in production. Getting that coverage is on our lab todo list for next year.

Understood. No current crash to investigate. The null checks seem reasonable.

baallan:
I have updated the ldms.c patch to use NULL instead of the xprt log.

PR #1974, opened 2025-09-11
fix null ptr deref and array overrun in store_csv_common.c

baallan:


ldms/src/store/store_csv_common.c
```diff
@@ -276,7 +276,7 @@ void rename_output(const char *name,
 				return;
 			} else {
 				size_t vlen = vend - head + 1;
-				char var[vlen];
+				char var[vlen+1];
 				memset(var, 0, vlen);
 				strncpy(var, head, vlen-1);
```
tom95858, line 280: @baallan, i think this code might not be what was intended. strncpy will only write vlen-1 bytes from src into dst. I I think this code, as written, will end up with two terminating '\0', which I don't think is what was intended.

Please confirm, but I think what strncpy does is the following;

``` c
    bzero(dst, 0, len);
    memcpy(dst, src, len-1);
```

Please check the man page to confirm.

ldms/src/store/store_csv_common.c
```diff
@@ -276,7 +276,7 @@ void rename_output(const char *name,
 				return;
 			} else {
 				size_t vlen = vend - head + 1;
-				char var[vlen];
+				char var[vlen+1];
 				memset(var, 0, vlen);
```
tom95858: I don't think this is necessary given the implementation of ```strncpy``

ldms/src/store/store_csv_common.c
tom95858: Sorry. the bzero signature is just bzero(dst, len)

baallan: removed the memcpy.

baallan:
@tom95858 this one should be good now.

ldms/src/store/store_csv_common.c
```diff
@@ -277,9 +277,8 @@ void rename_output(const char *name,
 			} else {
 				size_t vlen = vend - head + 1;
 				char var[vlen];
-				memset(var, 0, vlen);
 				strncpy(var, head, vlen-1);
-				var[vlen] = '\0';
+				var[vlen-1] = '\0';
```
tom95858, line 281: This is unnecessary, `strncpy` does this for you.

ldms/src/store/store_csv_common.c
```diff
@@ -1114,8 +1113,10 @@ int open_store_common(struct plugattr *pa, struct csv_store_handle_common *s_han
 }
 
 void close_store_common(struct csv_store_handle_common *s_handle, struct csv_plugin_static *cps) {
-	if (!s_handle || !cps) {
-		ovis_log(cps->mylog, OVIS_LERROR,
+	if (!cps)
+		return;
+	if (!s_handle) {
+		ovis_log( cps->mylog, OVIS_LERROR,
```
tom95858, line 1119: Super-nit, but this has an extra ' ' after the '('

ldms/src/store/store_csv_common.c
baallan: it does not do it for you when the byte written is out of range of the len given to strncpy, which is the case here. see detailed commit comments.

baallan: https://github.com/ovis-hpc/ldms/pull/1974/commits/401207f18c4396a889be01320cbc837443c8535d

tom95858: > it does not do it for you when the byte written is out of range of the len given to strncpy, which is the case here. see detailed commit comments.

I'm objecting to explicitly stuffing the '\0 in there instead of letting strncpy do it for you. And I'm confused as to why this can't be done.

baallan: it can't be done because the input string strncpy is copying is not null terminated. we're copying a range from the input string which does not include a '\0' and which is not followed by a '\0' and which is always full length.

baallan: As noted in the man page:
```
Warning: If there is no null byte among the first n
       bytes of src, the string placed in dest will not be null-terminated.
```
this is the case here.

tom95858: @ballan, where is that comment because that is a direct contradiction of my man page? But if it's true, then we have to do what you say. I'm going to put together some unit tests.

tom95858: I wrote a test for this, and it behaves as desribed above. I.e. ``strncpy` *does* null terminate. 
``` c
#include <stdio.h>
#include <string.h>

int main(int argc, char *argv[])
{
        char *t = "abcdefghijk";
        char dst[6];
        memset(dst, 'x', sizeof(dst));
        strncpy(dst, t, 3);
        printf("%s\n", dst);
        return 0;
}

# ./main
abcxxx
```

So the above proves the claim that the Warning is correct.

PR #1977, opened 2025-09-11
fix inverted pointer checks in thread_id_get

baallan:
as previously coded, thread_id_get is a noop.


tom95858:
@nichamon this function is never called afaict. Can we just remove it. It's not a noop, if it were ever called, it would crash.

baallan:
@nichamon please close this if removing the function in another PR is the final solution.

removing from 451 milestone

tom95858:
Resolved by #1982

PR #1978, opened 2025-09-11
fix undetection of errors in util file reader due to mistyped variable

baallan:


tom95858:
@narategithub this change doesn't fix anything. If ``read()`` returns -1, it means EAGAIN which should trigger another read not a return of an error.

That said:

- The caller just returns an error on anything !0, so a -1 is equivalent to 0xffffffffff...even so, the return value is cast to an int, so (uint64_t)-1 gets cast to (int)-1 and is seen as < 0 by the caller
- I believe that the only way read will return EAGAIN is if *fd* is a pipe, or if we are in gdb and the EAGAIN is triggered by SIGINT
- We are reading from /proc, not a PIPE

So we should change this because read() returns ssize_t, not size_t, but if wanted this to work as expected in gdb, we should handle the EAGAIN.

baallan:
Removing this from the 451 milestone as it is only used in hweventpapi which documents itself as an unsupported sampler.

@tom95858 @emdonat @narategithub  Do we retire the hweventpapi sampler and ovis_pgrep function, or just fix it at a later leisurely pace?

tom95858:
@baallan, could we please  move this to milestone v4.5.2

baallan:
> @baallan, could we please move this to milestone v4.5.2

fine by me. done.

PR #1979, opened 2025-09-11
return null pointer in __zap_passive_ep_thread if bad io_thread_create

baallan:
io_thread_create can fail for things other than enomem, so we need to handle it.
This may just be the start of 'handle it'.


baallan:
@narategithub the handling of locks and multiple checking of t not null in the original code looks dubious when there is an error. please have a look.

narategithub:
Thank you @baallan for finding the issue and posting a PR. Unfortunately this is not the only issue in this function. PR #1980 addresses this `NULL` issue and the unprotected `z->_passive_ep_thread` as well.

baallan:
@narategithub please close this pr if #1980 supercedes it as I suspect.

tom95858:
Fixed in #1980

PR #1980, opened 2025-09-11
Fix zap passive ep thread assignment

narategithub:
This patch adddresses the following issues in zap passive ep thread assignment (`__zap_passive_ep_thread()`):

- `z->_passive_ep_thread` was access / written w/o proper locking.
- Not handling the case that `z->io_thread_created()` returning `NULL`.


narategithub:
Thanks @baallan for finding the issue and posting PR #1979. Unfortunately, the issues were more than not handling the `NULL` return of `z->io_thread_create()`. I think we got lucky because this got called only once or twice in the program life cycle (and highly likely from the same thread).

@tom95858 This passed `agg-test` in `ldms-test`.

PR #1983, opened 2025-09-11
Remove dead code in ldmsd_request_util.c

### LINKED ISSUE #1968: __ldmsd_parse_advertiser_add_req never does "Add an attribute of type 'STRING'" like most of its peers in ldmsd_request_util
The string tmp and variable cnt in __ldmsd_parse_advertiser_add_req are never used, unlike most other functions such as __ldmsd_parse_default_auth_req.   This results in never calling near the function return:
```
add_attr_from_attr_str(NULL, tmp,
                                            &ctxt->request,
                                            &ctxt->request_sz);
```

This is either dead code (not a big deal) or possibly a mis-formatted ctxt  object that might affect advertiser functionality.
Can someone who understands ldmsd_request_utils take a closer look?

If it's just innocent dead code, remove the 4.5.1 milestone from this issue.

nichamon:
There are dead code in __ldmsd_parse_advertiser_add_req() and __ldmsd_parse_brige_add_req(). The patch removes the dead code.

Resolves #1968

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request_util.c
+++ b/ldms/src/ldmsd/ldmsd_request_util.c
@@ -521,8 +521,7 @@ int __ldmsd_parse_generic(struct ldmsd_parse_ctxt *ctxt)
 	rc = 0;
 out:
 	ctxt->av = av;
-	if (dummy)
-		free(dummy);
+	free(dummy);
 	return rc;
 }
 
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_request_util.c
```diff
@@ -828,20 +821,7 @@ int __ldmsd_parse_bridge_add_req(struct ldmsd_parse_ctxt *ctxt)
...
-		rc = add_attr_from_attr_str(NULL, tmp,
-					    &ctxt->request,
-					    &ctxt->request_sz);
-	}
-
-out:
-	if (tmp)
-		free(tmp);
+ out:
 	if (dummy)
```
tom95858: nit: this if check is unnecessary ``free`` checks for NULL.

nichamon:
@tom95858 I found similar NULL checks before calling `free()` in `ldmsd_request_util.c`. I addressed them in the second commit in this PR. Please let me know if you prefer me create a separate PR for the second commit.

PR #1984, opened 2025-09-11
Remove 'advertiser' from the prdcr_status output

nichamon:
The patch removes advertisers from the prdcr_status. advertiser_status is available to report the status of advertisers already. Including advertisers in the prdcr_status output could be confusing.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -1050,6 +1050,8 @@ class LdmsdCmdParser(cmd.Cmd):
             port = prdcr['port']
             if prdcr['type'] == 'bridge':
                 continue
+            if prdcr['type'] == 'advertiser':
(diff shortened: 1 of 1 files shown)


nichamon:
@tom95858 This is a nit fix.

tom95858:
@nichamon, my only concern is that there will be scripts built for 4.5.1 that look for this and won't find it in v4.5.2. @baallan, what do you think?

baallan:
> @nichamon, my only concern is that there will be scripts built for 4.5.1 that look for this and won't find it in v4.5.2. @baallan, what do you think?

I know of no scripts expecting advertiser info under prdcr_status, or of anyone actually deployed using advertiser yet, since it was just presented at ldmscon.  As 451 has never been released (anyone scripting based on an rc1 knew they were taking a risk), 'saright by me.

PR #1989, opened 2025-09-12
ldmsd_cfgobj_del() ref fix

narategithub:
`ldmsd_cfgobj_del()` is equivalent to remove and free. So, it must put
- ref by tree, and
- ref by init.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_cfgobj.c
+++ b/ldms/src/ldmsd/ldmsd_cfgobj.c
@@ -311,8 +311,9 @@ void ldmsd_cfgobj_del(ldmsd_cfgobj_t obj)
 		obj->kvl_str = NULL;
 	}
 	rbt_del(cfgobj_trees[obj->type], &obj->rbn);
-	ldmsd_cfgobj_put(obj, "init");
+	ldmsd_cfgobj_put(obj, "cfgobj_tree");
 	pthread_mutex_unlock(cfgobj_locks[obj->type]);
+	ldmsd_cfgobj_put(obj, "init");
 }
 
 /**


ldms/src/ldmsd/ldmsd_cfgobj.c
```diff
@@ -311,8 +311,9 @@ void ldmsd_cfgobj_del(ldmsd_cfgobj_t obj)
 		obj->kvl_str = NULL;
 	}
 	rbt_del(cfgobj_trees[obj->type], &obj->rbn);
-	ldmsd_cfgobj_put(obj, "init");
+	ldmsd_cfgobj_put(obj, "cfgobj_tree");
```
morrone, line 314: I'm sure you are correct that this is needed. I have no objection to your patch.

I am wondering if you could do me a favor and help me understand what is going on here. There ia similar ref counting elsewhere in the code that I am not understanding, and it would really help if I could understand the theory of operation.

What is confusing me is that this code is dropping two references on the same reference counter, unconditionally, back to back. If this code always winds up retaining 2 references on the ref counter, then my intuition tells me that the code took a redundant reference count earlier in the code. It _seems_ like the second reference is not serving a purpose. So maybe I'm not understanding something.

Could you help me understand why the code needs the second reference? What purpose is that second reference serving? Is there some other code path that I'm not seeing at the moment that requires the extra reference?

morrone: I think a few things are finally clicking into place for me. And now that they have, I think that this is probably not quite the right way to fix this. Please see an alternative in PR #1992.

tom95858: This is just changing the debug name that is used to tag the reference. This debug name helps us know that we are dropping the reference we think we are dropping. There are debug dump routines that will check that these reference 'names' never go below zero, dump what the references are, who took them and where and who put them and where. This debug checking is controlled by two macros _REF_TRACK_ and _CFG_REF_DUMP_.

morrone:
@tom95858 If you have a reason to ignore my review and land something, that is fine. But could you please add some comments to explain why you are doing that?

ldms/src/ldmsd/ldmsd_cfgobj.c
morrone: I don't say anything about the name string. I understand that the names strings are for debugging. None of that is relevant to my comment.

Did you look at #1992?

tom95858:
@morrone I believe that I answered your question. The tags on the reg_get, etc... are not different references. They are the same reference. The tag helps understand "who" is taking the reference, "where" and "why" But this change is important to avoid an assert if _REF_TRACK_ is enabled.

ldms/src/ldmsd/ldmsd_cfgobj.c
morrone: > This is just changing the debug name that is used to tag the reference.

@tom95858 That is simply not true. It is taking (and here dropping) a reference. As you point out the name is only for debugging. In the normal case where the debug tracing is disabled, the debug name is completely compiled out. But what is _not_ compiled out is the taking and dropping of references. If is not appropriate to take a reference, then we shouldn't be taking a reference. We can't allow the tracing code to cause us to deviate from basic good reference counting practices. That is the tail wagging the dog.

In the end, it _may_ be appropriate to take this reference. But as I pointed out in #1992, there is a right way and a not-as-right way to drop this additional reference.

morrone:
> @morrone I believe that I answered your question. The tags on the reg_get, etc... are not different references. They are the same reference. The tag helps understand "who" is taking the reference, "where" and "why" But this change is important to avoid an assert if _REF_TRACK_ is enabled.

They aren't. They are the same counter but they are explicitly different references. Each call to ref_get() is a new reference. Pretending that they are not is less-than-good programming practice. Pretending that they are the same reference and not even documenting that is just not reasonable.

But none of this matters to the fact that #1992 would have been a better solution. I'll rebase #1992.

tom95858:
@morrone, It might be helpful to go through the code and for the object in question look at all the ref_get, etc... You will notice that "init" is never taken, "init" is the reference added by ref_init(). In this bug, the object is not being destroyed, however, the reference to the object from the RBT is gone, so the "cfg_tree" reference is being dropped. In any event, perhaps this would be a good topic for a user-group presentation.

morrone:
> @morrone, It might be helpful to go through the code and for the object in question look at all the ref_get, etc... You will notice that "init" is never taken, "init" is the reference added by ref_init(). In this bug, the object is not being destroyed, however, the reference to the object from the RBT is gone, so the "cfg_tree" reference is being dropped. In any event, perhaps this would be a good topic for a user-group presentation.

@tom95858 , I already understand that the first reference is named init. It is not correct to say that "init" is never taken. The "init" reference is taken when "ref_init()" is called, and the caller explicitly chooses the debugging name for that initial reference when they call ref_init(). They may not call ref_get() to take that reference, but they absolutely take the first reference when they call ref_init(). But this discussion of the first reference is not terribly relevant to the discussion at hand.

My #1992 PR does not get rid of the cfg_tree reference, it just improves how it is released. The cfg_tree reference should be  dropped by ldmsd_cfgobj_rm() rather than directly in ldmsd_cfgobj_del(). See #1992 where I made that change.

I rebased #1992 it so it is ready to land again.

PR #1990, opened 2025-09-12
Fix TIMESTAMP type handling.

narategithub:
TIMESTAMP type was not properly handled in Python, schema add, and `ldms_ls`.


tom95858:
@nick-enoent could you please ensure that this doesn't break Maestro somehow?

tom95858:
I have tested this with sos_store, and d/sos. I didn't expect any problems and there weren't any. Maestro uses the Python interface, so let's ensure there are no issues there and then this will be ready to merge.

nick-enoent:
> @nick-enoent could you please ensure that this doesn't break Maestro somehow?

Looking at the code, this shouldn't affect Maestro at all, but I can run some tests today to confirm.

PR #1991, opened 2025-09-12
Set default ldmsd `name` if it is not set by config

narategithub:


ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -1355,6 +1355,26 @@ int __our_cfgobj_filter(ldmsd_cfgobj_t obj)
 int ldmsd_ourcfg_start_proc()
 {
 	int rc;
+	ldmsd_cfgobj_t cfg;
+	ldmsd_listen_t lcfg;
+	const char *myname = ldmsd_myname_get();
```
tom95858, line 1360: @narategithub should ``ldmsd_myname_get`` do this default name gymnastics so that other potential callers don't have to replicate it?

What is this breaking exactly?

PR #1992, opened 2025-09-12
Call ldmsd_cfgobj_rm() in ldmsd_cfgobj_del()

morrone:
ldmsd_cfgobj_del() holds the tree lock longer than necessary, and has duplicates of code already in ldmsd_cfgobj_rm().

Let us just call ldmsd_cfgobj_rm() from ldmsd_cfgobj_del().

### DIFF
--- a/ldms/src/ldmsd/ldmsd_cfgobj.c
+++ b/ldms/src/ldmsd/ldmsd_cfgobj.c
@@ -117,6 +117,8 @@ struct rbt *cfgobj_trees[] = {
 
 void ldmsd_cfgobj___del(ldmsd_cfgobj_t obj)
 {
+	free(obj->avl_str);
+	free(obj->kvl_str);
 	free(obj->name);
 	free(obj);
 }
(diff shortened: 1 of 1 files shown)


morrone:
@tom95858 this is probably an easy one to land.

tom95858:
@nichamon, can you please take a look at this? For some reason we split this logic and I can't remember why. It may be bit-rot.

ldms/src/ldmsd/ldmsd_cfgobj.c
```diff
@@ -301,7 +301,7 @@ void ldmsd_cfgobj_put(ldmsd_cfgobj_t obj, const char *ref_name)
 
 void ldmsd_cfgobj_del(ldmsd_cfgobj_t obj)
 {
-	pthread_mutex_lock(cfgobj_locks[obj->type]);
+	ldmsd_cfgobj_rm(obj);
```
nichamon, line 306: @morrone @tom95858 

I agree that having `ldmsd_cfgobj_del()` call `ldmsd_cfgobj_rm()` to remove the object from the tree is cleaner.

I think we should also move the cleanup logic for `obj->avl_str` and `obj->kvl_str` (lines 305-312) into `ldmsd_cfgobj___del()` as part of this PR. This change causes the cleanup to occur outside any lock protection. Moving it to `ldmsd_cfgobj___del()` ensures these fields are freed when the reference count reaches zero, which is the proper protection mechanism. It also keeps the related cleanup logic changes together in one place. What do you think?

morrone: I agree with that, thanks.

PR #1993, opened 2025-09-12
ovis_ref: Add race-to-zero documentation to ref.h

### LINKED ISSUE #1696: ldms ref count implementation appears unsafe
I believe that the reference counter mechanism implemented in lib/src/ovis_ref/ref.h is fundamentally unsafe to use as a synchronization mechanism.

ref_get() and ref_put() are currently implemented by the following primitives:

- `__sync_fetch_and_add()` /* for atomic increment */
- `__sync_fetch_and_sub()` /* for atomic decrement */

The problem is that atomic increment and decrement do not, on their own, provide for safe synchronization between threads. The current implementation based on them is racy.  Consider the _ref_put() code:

```
    66          count = __sync_sub_and_fetch(&r->ref_count, 1);
    67          if (!count)
    68                  r->free_fn(r->free_arg);
```

Here lines 67 and 68 have no locking around them. Between lines 67 and 68 another thread may do ref_get() and increment the counter, but this thread will continue on to free the memory out from under the other thread! Fortunately, this is not a new problem in computer science and we should be able to fix this without too much trouble. 

The solution is to rebuild the ref counter code around atomic [Compare-and-Swap (CAS)](https://en.wikipedia.org/wiki/Compare-and-swap). The section [Example apllication: atomic adder](https://en.wikipedia.org/wiki/Compare-and-swap#Example_application:_atomic_adder) of the wikipedia page for CAS gives a good overview of how to implement an atomic adder which is the general model we would follow.

We could use the gcc built-in function `__sync_bool_compare_and_swap()` in the implementation.

### LINKED ISSUE #1697: struct ref_s race inside ldmsd_cfgobj
We have an issue with the issue on the main branch with struct ldmsd.h:ldmsd_cfgobj. It looks like this:

```
   193  typedef struct ldmsd_cfgobj {
   194          char *name;             /* Unique cfgobj name */
   195          struct ref_s ref;
   196          ldmsd_cfgobj_type_t type;
   197          ldmsd_cfgobj_del_fn_t __del;
   198          struct rbn rbn;
   199          pthread_mutex_t lock;
   200          uid_t uid;
   201          gid_t gid;
   202          int perm;
   203          char *avl_str;
   204          char *kvl_str;
   205  } *ldmsd_cfgobj_t;
```

The problem lies in the automatic free that occurs then "ref" hits zero. For at least sampler and store plugins (but likely all cfgobj users), through a slightly round-about process the ldmsd_cfgobj itself is freed. The call stack is a little like this in the case of a sampler:

```
 /* ref hits zero */
 ldmsd_cfgobj.c:__cfgobj_ref_free()
 ldmsd_config.c:ldmsd_sampler___del()
 ldmsd_cfgobj.c:ldmsd_cfgobj___del()
```

When the reference counter hits zero, the struct _in which it lives_ is freed. This is not thread safe. We can't only protect access to a struct ldmsd_cfgobj by a synchronization mechanism that lives _inside_ the same struct's memory. And no other protection seems to be in place here.

The race looks something like this:

```
  Thread A: in code about to increment ref count
  Thread B: decrements ref count to 0, frees struct out from under thread A
  Thread A: increments ref count in memory that is no longer valid
```

It is not thread safe to use a synchronous reference counter in this fashion. The counter, when used for protecting memory access to something, cannot live inside the very thing it is trying to protect access to. Normally the ref counter would live in some parent struct, and would protect memory _pointed to_ by that struct.

Another way to state the issue is that in order to figure out if it is safe to access the memory, one needs to access that memory. But we can't access the memory until we know it is safe.

With the existing code on main, to use struct cfgobj's ref counter in a thread-safe manner we would always need to be holding some _other_ lock on the struct first, and the ref counter could not trigger memory free when it hits zero. But at that point there is probably no point in using the synchronous ref counter at all, because we would be holding some other lock that _actually_ protects the memory. And we can't call the destructor when the ref hits zero in any sane manner while some other thing is holding a lock on us.

Synchronous reference counters are good, and absolutely make sense when used correctly. It just looks like we need to do some code refactoring in able to use the ref counters in a thread-safe manner.

morrone:
ovis_ref: Add race-to-zero documentation to ref.h

Closes #1696
Closes #1697

### DIFF
--- a/lib/src/ovis_ref/ref.h
+++ b/lib/src/ovis_ref/ref.h
@@ -6,6 +6,85 @@
 #include <pthread.h>
 #include <stdio.h>
 
+/*
+ * This header implements a fast reference counter.
+ *
+ * ref_put() and ref_get() are generally thread safe, however the reference
+ * counter is NOT not race-to-zero safe. Users of this reference counter MUST
+ * implement their own race-to-zero protection strategy.
+ *
+ * Leaving race-to-zero protection to the caller permits this reference counter
+ * to be as low over head as possible in general operation.
+ *
+ * The race-to-zero problem is avoided by ensuring that the a ref_get() is
+ * not possible when the final ref_put() is issued.
+ *
+ * Some general approaches to implement race-to-zero safety are:
+ *
+ * 1) Only call ref_get() when the caller knows a reference is still held
+ *
+ *    For instance, a holder of a reference may call ref_get() on behalf
+ *    of another component and logically "pass" that reference's ownership
+ *    onto the other component. This hand off must be well documented.
+ *
+ *    Alternatively, a function might declare that its callers must already
+ *    hold a reference on a counter on which it plans to call ref_get()
+ *    itself. This contract must be well documented.
+ *
+ * 2) Implement a locking scheme under which the counter is tested to
+ *    see if acquiring a reference is currently permitted. Then and only
+ *    then may we ref_get() while still under the same lock.
+ *
+ *    For instance, in psuedo code the process looks like:
+ *
+ *       mutex_lock(lock)
+ *       if (reference_is_acquirable(...)) {
+ *          ref_get(ref)
+ *       }
+ *       mutex_unlock(lock)
+ *
+ *    In the LDMS code, a common approach is to store the object holding
+ *    the reference counter in a collection (list, tree, etc.) that is
+ *    protected by a mutex. The entity that placed the object in the collection
+ *    is contractually bound to:
+ *       A) hold a reference at the time it inserts the object into
+ *          the collection
+ *       B) retain that the reference until it removes the object
+ *          from the collection.
+ *    By doing so, any non-reference holder is safe to call get_ref() on
+ *    an object that it finds in the collection _while holding the lock_
+ *    on the collection.
+ *
+ *    For example, in psuedo-code:
+ *
+ *      X's code (already holds a reference on "obj"):
+ *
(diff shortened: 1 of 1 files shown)


tom95858:
@morrone, this is great. What I would suggest is that we add a little bit about _REF_TRACK_. The purpose of the _REF_TRACK_ #define is to ensure that if you are holding a pointer to an object that you intend to ref_put that you must hold `that` reference. If you don't, it's a bug in your code. The tag that is provided in the call is to ensure that the reference that your logic is putting is the one that you took, i.e. you are not putting someone else's reference.

Would you like me to propose this language, or would you like to do it? We could also do it as a separate PR.

morrone:
> @morrone, this is great. What I would suggest is that we add a little bit about _REF_TRACK_. The purpose of the _REF_TRACK_ #define is to ensure that if you are holding a pointer to an object that you intend to ref_put that you must hold `that` reference. If you don't, it's a bug in your code. The tag that is provided in the call is to ensure that the reference that your logic is putting is the one that you took, i.e. you are not putting someone else's reference.
> 
> Would you like me to propose this language, or would you like to do it? We could also do it as a separate PR.

Yes, there is more that could be documented. Each macro/function should be documented too. For instance, it should definitely be documented that when ref_init() returns, the caller now holds a reference. Folks seem to be forgetting that and following up the ref_init() with a ref_get() that is also just held by the same entity (not passed on to other entity).

But yes, I would prefer to leave additional documentation to another PR.

PR #1998, opened 2025-09-17
Fix inifinite loop in YAML config file processing

nichamon:
Without the patch, the loop was repeatedly processing the fist YAML file instead of advancing through the list.


nichamon:
@tom95858 We need this patch before releasing v4.5.1. Without it, -y doesn't work.

PR #1999, opened 2025-09-17
Make stream disabled by default and add stream_enable

nichamon:


nichamon:
The PR includes the commit from PR #1998 for testing purposing.

PR #2004, opened 2025-09-18
Fix `json_stream_sampler`

narategithub:
- fix incorrect type casting.
- add multi instance support.
- uid, gid, perm of the sets (the metadata) follow the `config` line.
- `S_uid`, `S_gid`, and `S_perm` follow information in the message.

### DIFF
--- a/ldms/src/sampler/json/json_stream_sampler.c
+++ b/ldms/src/sampler/json/json_stream_sampler.c
@@ -60,7 +60,7 @@
 #include "ldmsd.h"
 #include "ldmsd_plug_api.h"
 #include "ldmsd_stream.h"
-#include "ovis_log.h"
+#include "ovis_log/ovis_log.h"
 
 #define SAMP "json_stream"
 
@@ -126,7 +126,6 @@ typedef struct js_schema_s {
 
 typedef struct js_stream_sampler_s *js_stream_sampler_t;
 struct js_stream_sampler_s {
-	int initialized;	/* 0 if 1st config */
 	char *stream_name;	/* stream msgs received from */
 	size_t heap_sz;		/* heap size for created sets */
 	char *prod_name;	/* producer name */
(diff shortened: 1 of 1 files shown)


narategithub:
@tom95858 I found this bug while addressing `ldms-test`. I would like to confirm with you regarding the behavior of the plugin, especially the uid/gid/perm in the config line and the `S_uid`, `S_gid`, and `S_perm` from the message.

ldms/src/sampler/json/json_stream_sampler.c
```diff
@@ -1164,14 +1163,10 @@ static int __stream_recv(ldms_msg_event_t ev, ldmsd_cfgobj_store_t scfg)
 		j_set->set = l_set = ldms_set_create(
 						inst_name,
 						j_schema->s_schema,
-						ev->recv.cred.uid,
-						ev->recv.cred.gid,
-						0444, // ev->recv.perm,
```
tom95858, line 1169: @narategithub are we losing the uid/gid of the publisher at the 1st level aggregator? I'm not sure that this is what we want. If the config file doesn't specify the uid, gid, etc... those values will get squashed to either zero if the *ldmsd* is root, or `geteuid/getegid` if not. But in either case, the uid/gid of the publisher is lost.

Am I misreading something?

tom95858:
@narategithub My concern is that a publisher can pretend to be anybody. So narate can publish as tom, etc... and that is new behavior. I've already merged (by accident), but I can revert if we don't think this is the right thing to do. I think the current users are all running as root anyway, so in practical terms, they are aware of the risks and perhaps it doesn't matter. But to do this "correctly", I think we would need to ensure that the publisher can "be" S_uid/gid, etc...

baallan:
- kokkos, darshan, and adc run as user (by user) auth munge.
- ldms-notify runs as root but would want when ready to publish the messages using the uid/gid of the owner of the process being reported. In the meantime, ldms-notify includes uid/gid it finds as data inside the messages. It runs as root using ovis auth.
- linux_proc_sampler is in the same boat as ldms-notify.
- On the drawing board is a version of ldms-notify/linux_proc_sampler that runs as user and connects via munge.

On the daemon front, we have system daemons as root and user-launched daemons (as user) which connect and aggregate sets from their local ldmsd (or the system aggregator) when they want to manage their own storage during experiments.

PR #2007, opened 2025-09-19
json_stream_sampler cleanup

morrone:
Removes the last vestiges of the cfgobj structs from json_stream_sampler. Also fixes the per-instance context allocation by performing it in constructor().

Remove the global __log, and uses a per-instance log handle, allowing the plugin to be multi-instance capable.

@tom95858 this PR includes your 'Revert "Fix json_stream_sampler"' from PR #2006.

### DIFF
--- a/ldms/src/sampler/json/json_stream_sampler.c
+++ b/ldms/src/sampler/json/json_stream_sampler.c
@@ -128,6 +128,7 @@ typedef struct js_stream_sampler_s *js_stream_sampler_t;
 struct js_stream_sampler_s {
 	char *stream_name;	/* stream msgs received from */
 	size_t heap_sz;		/* heap size for created sets */
+	char *config_name;      /* plugin instance config name */
 	char *prod_name;	/* producer name */
 	char *inst_fmt;		/* set name format specifier */
 	char *comp_id;		/* component id */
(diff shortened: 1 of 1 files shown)


morrone:
@tom95858 You landed the conflicting #2013 before #2006 (or this, #2007, which is based at the #2006 commit). Can you provide some guidance on what we are going to do next? Are you abandoning the revert in #2006?

tom95858:
@morrone, the json_stream_sampler on 4.4 uses streams. Somewhere along the line in 4.5 it got converted to message. Well that's pretty confusing to have a sampler called json_stream_sampler that uses message, not to mention, anyone using 4.5.1 would then have an incompatible config, apps, etc... So json_stream_sampler is now stream again and there will be a new sampler called json_message_sampler (or some such) that uses ldms_msg.

morrone:
@tom95858 that didn't really answer my question. What are you doing with #2006? Is that going to be rebased and landed, or is it being abandoned. What should I do with this PR now that another PR jumped the queue? Is it now safe to rebase, or are there other things that are going to jump the queue as well?

tom95858:
@morrone, yes, you are right. The changes in the PR are good I think. I can refactor and resolve the conflicts if you like. I think this PR is fine otherwise.

morrone:
@tom95858 I have rebased my patches (without PR #2006, which now seems to be moot).

PR #2008, opened 2025-09-19
papi_sampler to use ldms_msg

narategithub:
`slurm_notifier` uses ldms_msg to notify about job events. Hence `papi_sampler` which subscribes to those events have to use `ldms_msg` (instead of `ldmsd_stream`).


tom95858:
@narategithub this change has the same problem as the json_stream_sampler which is that existing users config and/or scripts will have to change. In this case, there are existing users of the papi_sampler that will have to change their job prolog and epilog to continue to use the sampler. So _if_ we are going to do this, in the short term, we should support both stream and message, but not just go to message and disable stream because existing users will break ... without warning.

PR #2009, opened 2025-09-20
Make ovis_event's PERIODIC events support negative offsets

### LINKED ISSUE #2002: negative sampler offsets yield inconsistency (on main @ c9c651)
With the updater is in auto interval mode, shifting the sampler offset to the negative half of the range causes downstream aggregators to never be consistent.  I would expect the data to be consistent (or not) independent of the offset defined at the sampler, since the updater auto interval seems to be getting the right hint string as shown in ldms_ls. We need negative offsets to work, since job_info is typically collected before (offset -200000) all the other metric sets (default offset 0) which pull the jobid into their data).

Example of the problem:

The downstream aggregator config is:
```
prdcr_listen_add name=kn-mc7/l1 regex=kn-mc.* 
prdcr_listen_start name=kn-mc7/l1
updtr_add name=L1_updtr interval=1s auto_interval=true
updtr_prdcr_add name=L1_updtr regex=.*
updtr_start name=L1_updtr

```

If the first daemon produces a set with a negative offset something goes consistently wrong:
```
load name=dstat
config name=dstat producer=${HOSTNAME} instance=${HOSTNAME}/dstat component_id=8110001 perm=0644
start name=dstat interval=3000000 offset=-200000
```
It locally (host kn-mc7) shows as:
```
Schema         Instance                 Flags  Msize  Dsize  Hsize  UID    GID    Perm       Update            Duration          Info
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
dstat          kn-mc7/dstat                CL    1976    376      0  0  0 -rw-r--r-- 1758128418.000928          0.000460 "updt_hint_us"="3000000:-200000"
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
```
and on the aggregator downstream as:
```
Schema         Instance                 Flags  Msize  Dsize  Hsize  UID    GID    Perm       Update            Duration          Info
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
dstat          kn-mc7/dstat                 R    1976    376      0  0  0 -rw-r--r-- 1758128330.909740          0.000460 "updt_hint_us"="3000000:-100000"
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------

```

While if I use offset 0 instead, data is 'consistent' down stream:
On the sampler I see:
```
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
dstat          kn-mc7/dstat                CL    1976    376      0  0  0 -rw-r--r-- 1758129204.003932          0.000085 "updt_hint_us"="3000000:0"
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
```
and downstream I get:
```
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
dstat          kn-mc7/dstat                CR    1976    376      0  0 0 -rw-r--r-- 1758128778.002356          0.000101 "updt_hint_us"="3000000:100000"
-------------- ------------------------ ------ ------ ------ ------ ------ ------ ---------- ----------------- ----------------- --------
```

In all output above, the updt_hint_us seems to be as expected. 
The timing in the agg log looks as i would expect in terms of timestamps (the patterns below repeat with samples/updates). 
For sampler offset -200000 (-0.2s)/updtr offset -100000:
```
1758131069.900533:    DEBUG: updater: updtr_task sched '3000000': set 'kn-mc7/dstat'
1758131069.900542:    DEBUG: updater: Schedule an update for set kn-mc7/dstat
1758131069.958147:    DEBUG: updater: Update complete for Set kn-mc7/dstat with status 0
1758131069.958154:     INFO: updater: Set kn-mc7/dstat is inconsistent.
```
and for sampler offset 0/updtr offset 100000:
```
1758130545.100535:    DEBUG: updater: updtr_task sched '3000000': set 'kn-mc7/dstat'
1758130545.100549:    DEBUG: updater: Schedule an update for set kn-mc7/dstat
1758130545.160207:    DEBUG: updater: Update complete for Set kn-mc7/dstat with status 0
1758130545.160214:    DEBUG: updater: Pushing set 0x7f150c00ab30 kn-mc7/dstat
```

nichamon:
PERIODIC events with negative phase (offset) caused events to fire at unexpected times. The patch handles when calculated next event timing would be in the past or exceed one period (interval), ensuring PERIODIC events never wait more than one period.

Fixes #2002

### DIFF
--- a/lib/src/ovis_event/ovis_event.c
+++ b/lib/src/ovis_event/ovis_event.c
@@ -334,6 +334,26 @@ void ovis_scheduler_free(ovis_scheduler_t m)
 	ovis_scheduler_ref_put(m);
 }
 
+static void set_next_periodic_time(ovis_event_t ev, const struct timeval *tv)
+{
+	uint64_t now, us;
+	int64_t diff;
+	now = us = tv->tv_sec * USEC + tv->tv_usec;
+	us = ROUND(us, ev->param.periodic.period_us);
+	us += ev->param.periodic.phase_us;
+	diff = (int64_t)us - (int64_t)now;
+	if (diff <= 0) {
+		/* This is possible if phase is negative */
+		diff += ev->param.periodic.period_us;
+	} else if (diff > ev->param.periodic.period_us) {
+		/* Prevent waiting > 1 period */
+		diff %= ev->param.periodic.period_us;
+	}
+	us = now + diff;
+	ev->priv.tv.tv_sec = us / USEC;
+	ev->priv.tv.tv_usec = us % USEC;
+}
+
 /**
  * Process events in the heap and returns time-to-next-event.
  *
@@ -343,7 +363,6 @@ static
 int ovis_event_heap_process(ovis_scheduler_t m)
 {
 	struct timeval tv, dtv;
-	uint64_t us;
 	ovis_event_t ev;
 	int timeout = -1;
 
@@ -386,11 +405,7 @@ process_event:
 	case OVIS_EVENT_PERIODIC:
 		/* periodic event application callback */
 		gettimeofday(&tv, NULL);
-		us = tv.tv_sec * USEC + tv.tv_usec;
-		us = ROUND(us, ev->param.periodic.period_us);
-		us += ev->param.periodic.phase_us;
-		ev->priv.tv.tv_sec = us / USEC;
-		ev->priv.tv.tv_usec = us % USEC;
+		set_next_periodic_time(ev, &tv);
 		ovis_event_heap_update(m->heap, ev->priv.idx);
 		ev->cb.type = OVIS_EVENT_PERIODIC;
 		break;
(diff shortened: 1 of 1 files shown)


lib/src/ovis_event/ovis_event.c
```diff
@@ -334,6 +334,26 @@ void ovis_scheduler_free(ovis_scheduler_t m)
...
+	int64_t diff;
+	now = us = tv->tv_sec * USEC + tv->tv_usec;
+	us = ROUND(us, ev->param.periodic.period_us);
+	us += ev->param.periodic.phase_us;
+	diff = (int64_t)us - (int64_t)now;
+	if (diff <= 0) {
+		/* This is possible if phase is negative */
+		diff += ev->param.periodic.period_us;
+	} else if (diff > ev->param.periodic.period_us) {
+		/* Prevent waiting > 1 period */
```
baallan, line 349: detecting a wait of > 1 period should probably generate a warning. whether that warning is issued here or higher up where the user sets the improper period in the first place is a good question. probably the latter would be less noisy.

PR #2017, opened 2025-09-25
Fix list always empty in update-on-push mode

narategithub:
The list in the set of the receiver side that use update-on-push is always _empty_ because `set->heap` has never been set. This patch set `set->heap` in the update-on-push path.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -2209,6 +2209,7 @@ static void process_push_reply(struct ldms_xprt *x, struct ldms_reply *reply,
 {
 	uint32_t data_off = ntohl(reply->push.data_off);
 	uint32_t data_len = ntohl(reply->push.data_len);
+	uint32_t push_flags;
 	int rc;
 	ldms_set_t set;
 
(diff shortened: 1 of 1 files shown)


narategithub:
I sniff-tested this with the following scripts:

```python
#!/usr/bin/env python3
#
# file: srv.py

import os
import sys

import time
from ovis_ldms import ldms

ldms.init(16*1024*1024)

x = ldms.Xprt()
x.listen()

sch = ldms.Schema(name = "the_schema", metric_list = [
                                            ("sec", "u64"),
                                            ("list", "list", 128),
                                        ]
                 )

s = ldms.Set(name="the_set", schema=sch)
s.publish()

def update_loop():
    while True:
        s.transaction_begin()
        ts = int(time.time())
        s["sec"] = ts
        l = s["list"]
        l.purge()
        l.append(ldms.V_U64, ts)
        s.transaction_end()
        time.sleep(1)

update_loop()

```

```python
#!/usr/bin/env python3
#
# file: cli.py

import os
import sys
import time

from ovis_ldms import ldms

ldms.init(16*1024*1024)

x = ldms.Xprt()
x.connect("localhost")

s = x.lookup("the_set")
s.register_push()

while True:
    time.sleep(1)
    print(s.json())
```

With the patch from this PR, the list on the client is not empty anymore.

image

PR #2019, opened 2025-09-25
parse single argument to ldms-static-test DAEMONS calls correctly

baallan:
This fixes the expected behavior if only a single argument is given to the DAEMONS function in the static testing scripts.
DAEMONS 7 now is processed as DAEMONS $(seq 1 7).

This changes no  C code.


baallan:
@tom95858 sorry i missed some files when pushing #2010. here they are.

PR #2020, opened 2025-09-29
Bug fix when setting base set heap size

nick-enoent:
Properly allocate the cpu_list heap size based on the number of cpu cores


baallan:
@nick-enoent There's another bug here, independent of the current allocation code: the method of obtaining the cpu count is incorrect. If a core happens to be downed temporarily (which can happen in hpc) then it disappears from the lines in /proc/stat. When it comes back, then the allocation will be too small.
other ways of getting the proper maximum cpu count include:
```
nproc --all # (system call, probably less preferable)
cat /sys/devices/system/cpu/possible # (file read from c instead of system call)
```

tom95858:
@baallan, the code will recognize if the allocation becomes too small and re-allocate the set and associated heap based on the new requirements. IOW, I believe this is already handled.

PR #2024, opened 2025-09-30
Add error reporting for malformed YAML input file.

baallan:
Automated testing and admins of systemd daemons need ldmsd_yaml_parser results that are usable. For example, .conf files fed to -y now yield less-cryptic errors and malformed yaml is now reported via return code.

This patch:
- routes the main error messages of ldmsd_yaml_parser to stderr and sets the return value to an errno value.
- converts an nonzero error on pclose of the ldmsd_yaml_parser subprocess to a NULL return if the output string is also empty.
- traps anything that is not a reasonable yaml result (list or dict)

### DIFF
--- a/ldms/python/ldmsd/ldmsd_yaml_parser
+++ b/ldms/python/ldmsd/ldmsd_yaml_parser
@@ -27,13 +27,19 @@ if __name__ == "__main__":
     config_fp = open(args.ldms_config)
     conf_spec = yaml.safe_load(config_fp)
 
+    if not isinstance(conf_spec, (list, dict)):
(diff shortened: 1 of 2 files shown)


baallan:
@nick-enoent please have a look at #2024. I need this to facilitate scale testing based on yaml instead of .conf files.

tom95858:
@baallan could we please change the commit message to "Add error for malformed YAML input file." The commit message
implies a much larger structural change.

baallan:
> @baallan could we please change the commit message to "Add error for malformed YAML input file." The commit message implies a much larger structural change.

done

baallan:
@tom95858 this is ready to merge

PR #2025, opened 2025-09-30
add stream_enable to stream-related tests and default to .conf

baallan:
Since the default disable of streams was added, a batch of test script updates is required to enable streams in stream tests.
There are no C changes here.
In addition, the new stream_store_many test provides a brief storm (that may be scaled) of stream publications and
a fix is added to select .conf as the assumed test input format if -y is not seen in the LDMSD macro arguments.


baallan:
@tom95858 can we get #2025 merged so I can continue developing scale testing for streams/messages under slurm?

PR #2027, opened 2025-10-01
Introduce multi-tenant support to LDMSD

nichamon:
Introduces a new multi-tenant feature that enables trakcing multiple concurrent tenants within LDMS metric sets.

This feature allows users to:
 - Define tenant types by specifying a set of attributes (e.g., job_id, step_id, task_id)
 - Associate tenant definitions with LDMS sets via sampler plugin instance configuration
 - Automatically collect tenant-specific data alongside LDMS sets

Configuration example:
  tenant_def_add name=my_tenant metrics=job_id,step_id,task_id
  load name=meminfo
  config name=meminfo producer=nide0001 instance=nide0001/meminfo tenant=my_tenant
  start name=meminfo interval=1s


baallan:
This seems like an awful lot of work for something that should probably be handled instead as a single normal meta-metric (site-defined free-form string) that functionally supplants "job_id" in the existing job_info sampler for multitenant use-cases. Probably a metric named "tenant_info". Only change needed is to the job_info type samplers and sampler_base to handle one new metric.

morrone:
I guess the "job manager" is a new part of ldms that we haven't seen yet?

Do we know how we are going to modify decomposition to make this tenant list usable in stores?

baallan:
> Do we know how we are going to modify decomposition to make this tenant list usable in stores?

I suggest we make this much simpler:
- The multi-tenant info is an arbitrary (size limit set by admins) string meta-metric (low transmission overhead regardless of size)
- Store plugins can store it in the standard way.
- Sites can define the string format and query their store for matching string content any way they want.
- Sites can build their string any way they want (we stay out of the job manager business-- we just read a file provided by the job managers and forward it without parsing it).
- If the current storage policy is insufficient, it seems like a useful generic functionality to let any single string metric be split into several stored metrics following admin-defined rules or using an admin-defined plugin. 

Note that an admin-defined string metrifying plugin could be site-implemented using their favorite json/yaml parser library that might not be includeable in the ldmsd code base.

tom95858:
@morrone, @baallan, this capability is WIP and will change. It is here for the curious, it is not meant for review for inclusion in v4.5.2. When it is, it will be converted from draft.

PR #2029, opened 2025-10-07
netlink-notifier to use ldms_msg

narategithub:
netlink-notifier to use ldms_msg to notify linux_proc_sampler.


narategithub:
NOTE from meeting: convert back the linux_proc_sampler to use ldmsd_stream.

PR #2030, opened 2025-10-07
linux_proc_sampler to use ldmsd_stream

narategithub:
The publisher program uses ldmsd_stream, so linux_proc_sampler shall consume ldmsd_stream, not ldms_msg.


baallan (approved):
@tom95858 This is ready to go. Testing it has turned up a few other daemon-killing bugs (elsewhere) that appear to be easy to fix. Will submit other patches.

baallan:
@narategithub is there an alternate version coming?

PR #2034, opened 2025-10-08
trap cases in process_lookup_reply with null ctxt->op_ctxt 

baallan:
suppress time copy or get seg fault in memcpy (writing into null) observed in running ldms-static-test.sh linux_proc_sampler

PR #2038, opened 2025-10-08
`papi_sampler`, `syspapi_sampler` and `netlink-notifier` to use `ldms_msg`

narategithub:
This is a combination of:
- #2029 
- #2008 
- convert `syspapi_sampler` to use `ldms_msg`

The intention is to make `netlink-notifier` to produce `ldms_msg` like `slurm_notfier`, and `app_sampler` and `linux_proc_sampler` can still consume data from both `slurm_notifier` and `spank_notifier`. Also moving `syspapi_sampler` to use `ldms_msg` as `ldmsd_stream` is being deprecated.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -334,6 +334,7 @@ OPTION_DEFAULT_ENABLE([array_example], [ENABLE_ARRAY_EXAMPLE])
 OPTION_DEFAULT_ENABLE([hello_stream], [ENABLE_HELLO_STREAM])
 OPTION_DEFAULT_ENABLE([hello_ldms_msg], [ENABLE_HELLO_LDMS_MSG])
 OPTION_DEFAULT_ENABLE([blob_stream], [ENABLE_BLOB_STREAM])
+OPTION_DEFAULT_ENABLE([blob_msg], [ENABLE_BLOB_MSG])
 OPTION_DEFAULT_DISABLE([perfevent], [ENABLE_PERFEVENT])
 OPTION_DEFAULT_DISABLE([mpi_sampler], [ENABLE_MPI_SAMPLER])
 OPTION_DEFAULT_DISABLE([mpi_noprofile], [ENABLE_MPI_NOPROFILE])
@@ -1005,6 +1006,7 @@ ldms/src/decomp/as_is/Makefile
 ldms/src/decomp/flex/Makefile
 ldms/src/contrib/Makefile
 ldms/src/sampler/blob_stream/Makefile
+ldms/src/sampler/blob_msg/Makefile
 ldms/src/store/stream/Makefile
 ldms/src/store/test/Makefile
 ldms/src/store/kokkos/Makefile
--- a/ldms/src/sampler/Makefile.am
+++ b/ldms/src/sampler/Makefile.am
@@ -290,6 +290,10 @@ if ENABLE_BLOB_STREAM
 SUBDIRS += blob_stream
 endif
 
+if ENABLE_BLOB_MSG
+SUBDIRS += blob_msg
+endif
+
 if ENABLE_ZFS
 SUBDIRS += zfs_topvdevs
 SUBDIRS += zfs_leafvdevs
--- a/ldms/src/sampler/blob_msg/Makefile.am
+++ b/ldms/src/sampler/blob_msg/Makefile.am
@@ -1,0 +1,29 @@
+include $(top_srcdir)/ldms/rules.mk
+
+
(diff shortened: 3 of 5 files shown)


baallan:
@narategithub I'm reviewing this.

Given the locking bug mess i've just worked my way through in stream-based linux-proc-sampler and ldmsd_stream delivery mechanism (identifying problems, not fixing them) I recommend we definitely don't try to combine a stream and msg in the same sampler implementation.   The messy bit about linux_proc_sampler is that it both publishes and subscribes (on different streams). That, interacting with stopping the sampler and the associated xprt connection being terminated gets it into a hung state.

From a practical standpoint, I can deploy msg-netlink + msg-linux-proc-sampler + msg-blob_stream only if we update L0/L1/L2 together, but this should be ok because of how my site manages L2 daemons.

baallan:
@narategithub 
> From a practical standpoint, I can deploy msg-netlink + msg-linux-proc-sampler + msg-blob_stream only if we update L0/L1/L2 together, but this should be ok because of how my site manages L2 daemons.

But that chain does require a separate blob_msg/ port of blob_stream, so that folks storing something other than linux_proc_sampler and slurm still have a store plugin.
These folks include SNL: we'll need both blob_stream/blob_stream_writer.c and blob_msg/blob_msg_writer.c in production.

baallan:
@narategithub simple_lps.c:event_cb:215 spews errors 
```
event_cb:215 Received invalid event type
```
Looks like an expanded ldms_xprt_event_t.type enum things related to quotas must be handled. For now, the netlink publisher would always be on a high priority(only runs as root); is there something we need to do in creating the connection so all these quota events don't reach event_cb?

baallan:
testing result so far:
- netlink publisher is noisy (previous comment about quota events)
- metric sets do get created by linux_proc_sampler
- msg events don't get stored (need blob_msg plugin) to file
- L0/L1/L2 run valgrind memcheck clean except for leak in config handling already reported.
- L0/L1/L2 all run valgrind drd very noisy, particularly around process_dir_* while metric sets are being create, but also in the transport layer. The drd dumps are too big to attach here. A small part of it looks like the usual "one-writer/many-readers" handling of metric set value data, but most of it does not.

If we could get blob_msg/ 'sampler' added to this, then I would suggest merging as-is so I can put in a pr with the reproducer for the threading diagnostics.

baallan:
@narategithub @tom95858 @valleydlr This seems like an opportune time to remove the dependence on stream/msg from the syspapi sampler. Why do we need a stream/msg based pause in it (which regular user's can't affect anyway) instead of just using the sampler stop/start (which also doesn't lose the current sampler configuration)? We've never used the stream feature in production at snl to my knowledge (though we have deployed syspapi for years). Is it in use at some other site?


The most recent 2021 commit suggests:
```
    Pause/resume syspapi using ldmsd stream

    Reintroducing the ability to pause and resume syspapi using ldmsd
    stream so that system administrators can pause/resume syspapi from
    prolog/epilog script. Please note that when `papi_sampler` is present
    and syspapi is configured with auto_pause=1, syspapi will automatically
    paused when papi_sampler receive slurm "task_init" event. In this case,
    syspapi will automatically resume when no tasks are running.
```
but all of that could have already been accomplished by admins with the start/stop commands launched from slurm lua plugin scripting.

tom95858:
@baallan, NERSC uses the syspapi sampler. When user jobs wish to use PAPI, syspapi must be disabled. This is accomplished in a job prolog without the intervention of an admin. @narategithub use of the word "reintroduce" is misleading. This capability has always been there and is currently in daily use.

Streams Aside, we can't and shouldn't just remove functionality as it affects existing users.

narategithub:
@baallan

- The SEND_QUOTA_DEPOSITED event is now handled.
- I added a patch converting `blob_stream_writer` to use `ldms_msg`. The pipeline should work now.

narategithub:
@baallan scratch my patch for now ... I'm adding `blob_msg_writer` as another plugin.

narategithub:
@baallan `blob_msg_writer` is added. Could you please check it out? (the `stream=...` config attribute has been changed to `message_channel=...`)

tom95858:
@baallan, could you please add the ldmsd config statements that set up this pipeline so we can test concurrently.

baallan:
@narategithub @tom95858 I'll put in a pr with updated test assuming this pr is merged.
It would really help the testing if we got #2031 - #2034 reviewed if needed and merged.

PR #2041, opened 2025-10-11
Fix segfault in process_lookup_reply when profiling disabled

nichamon:
op_ctxt is NULL when profiling is disabled, causing memcpy() to segfault. Guard the memcpy with ENABLED_PROFILING(OP_LOOKUP).

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -1871,8 +1871,10 @@ void process_lookup_reply(struct ldms_xprt *x, struct ldms_reply *reply,
 	struct ldms_thrstat *thrstat;
 
 	thrstat = zap_thrstat_ctxt_get(x->zap_ep);
-	memcpy(&ctxt->op_ctxt->lookup_profile.complete_ts, &thrstat->last_op_start,
-					    sizeof(struct timespec));
+	if (ENABLED_PROFILING(LDMS_XPRT_OP_LOOKUP) && ctxt->op_ctxt) {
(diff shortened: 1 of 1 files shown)


ldms/src/core/ldms_xprt.c
```diff
@@ -1871,8 +1871,10 @@ void process_lookup_reply(struct ldms_xprt *x, struct ldms_reply *reply,
 	struct ldms_thrstat *thrstat;
 
 	thrstat = zap_thrstat_ctxt_get(x->zap_ep);
-	memcpy(&ctxt->op_ctxt->lookup_profile.complete_ts, &thrstat->last_op_start,
-					    sizeof(struct timespec));
+	if (ENABLED_PROFILING(LDMS_XPRT_OP_LOOKUP)) {
```
tom95858: @nichamon is this racy? What happens if the user is enabling and disabling profiling on the transport while there are outstanding requests?

nichamon: You're right - there's a race condition when profiling is 
enabled after a transport operation has already been initiated. In that 
case, op_ctxt was never allocated and remains NULL, which would cause a 
crash when we try to access it.

I've pushed a patch that adds NULL checks for op_ctxt in addition to 
the ENABLED_PROFILING() checks to address this.

baallan:
This was superceded by #2043, yes?

nichamon:
> This was superceded by #2043, yes?

yes

nichamon:
I'm closing this as #2043 has been merged.

PR #2042, opened 2025-10-11
Replace profiling TYPE attribute with explicit enable/disable commands

nichamon:
Previously, the profiling command used the TYPE attribute to represent the enable/disable parameter, which was semantically confusing and could lead to misunderstandings about the attribute's purpose.

This patch introduces a dedicated LDMSD_ATTR_ENABLE attribute and updates the profiling command to use it, making the command handling code clearer and more intuitive.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request.h
+++ b/ldms/src/ldmsd/ldmsd_request.h
@@ -166,6 +166,8 @@ enum ldmsd_request {
 	LDMSD_PID_FILE_REQ,
 	LDMSD_BANNER_MODE_REQ,
 	LDMSD_PROFILING_REQ,
+	LDMSD_PROFILING_DISABLE_REQ,
+	LDMSD_PROFILING_ENABLE_REQ,
 
 	/* failover requests by user */
 	LDMSD_FAILOVER_CONFIG_REQ = 0x700, /* "failover_config" user command */
--- a/ldms/python/ldmsd/ldmsd_communicator.py
+++ b/ldms/python/ldmsd/ldmsd_communicator.py
@@ -169,6 +169,8 @@ LDMSD_CTRL_CMD_MAP = {'usage': {'req_attr': [], 'opt_attr': ['name']},
                       'log_status' : {'req_attr' : [], 'opt_attr' : ['name']},
                       'stats_reset' : {'req_attr' : [], 'opt_attr' : ['list']},
                       'profiling' : {'req_attr' : [], 'opt_attr' : ['enable', 'reset']},
+                      'profiling_enable' : {'req_attr' : [], 'opt_attr' : []},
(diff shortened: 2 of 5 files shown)


nichamon:
@tom95858 I found this while I was working on PR #2041.

tom95858:
@nichamon, I think we forgot *ldmsctl*.

ldms/src/ldmsd/ldmsd_request.h
```diff
@@ -276,6 +276,7 @@ enum ldmsd_request_attr {
 	LDMSD_ATTR_RESET_INTERVAL,
 	LDMSD_ATTR_XTHREAD,
 	LDMSD_ATTR_MSG_CHAN,
+	LDMSD_ATTR_ENABLE,
```
tom95858: @nichamon, this name is a bit vague. The other names refer to an attribute. This name sounds like a `type`. IOW, it doesn't say what we are enabling. You have to look at the code to know that.

nichamon: @tom95858  Thanks for the feedback. The LDMSD_ATTR_ENABLE reads more like a command than an attribute name, which is confusing.

A couple of options to clarify this:

1. Rename the attribute to LDMSD_ATTR_ENABLE**D** — This makes it clearer that it's a property/state. Usage would be `profiling enabled=true` instead of `profiling enable=true`, which reads more naturally as setting an attribute value.
2. Create a dedicated config command `profiling_enable` — This gives users an explicit command for enabling/disabling profiling, separate from the generic attribute interface.

Which approach do you think would be clearer for users?

morrone: This is more about being clearer to developers, I think. "LDMSD_ATTR_ENABLE" leaves a developer asking themselves the question: "Enable _what_?". The other things in this enum give us hints about what they are referring to: json, auth, decomp, etc. So if this new enum entry is about profiling, it should really have the word "profiling" in it somewhere. Maybe LDMSD_ATTR_PROFILING.

nichamon: @tom95858 @morrone I got your point. I intentionally use `LDMSD_ATTR_ENABLE` for reusability when we have other features that can be enabled. However, it sounds like it is confusing. I can rename it to LDMSD_ATTR_PROFILING if that works for both of you.

Another solution is to add a new command, `profiling_enable` to enable the feature and keep the `profiling` command to report the data.

What do you think?

nichamon:
I decided to remove LDMSD_ATTR_ENABLE completely and introduced two new commands, profiling_disable and profiling_enable, to replace `profiling enabled=false` and `profiling enabled=true`. This is to prevent the naming confusing and to it consistency with the enable/disable commands for Message.

PR #2049, opened 2025-10-14
fix ldms.c use-before-check and redundant input check

baallan:


ldms/src/core/ldms.c
```diff
@@ -2915,14 +2910,14 @@ void ldms_metric_array_set_double(ldms_set_t s, int mid, int idx, double v)
 void ldms_metric_array_set(ldms_set_t s, int mid, ldms_mval_t mval,
 			   size_t start, size_t count)
 {
-	ldms_mdesc_t desc = ldms_ptr_(struct ldms_value_desc, s->meta,
-				      __le32_to_cpu(s->meta->dict[mid]));
 	int i;
 	if (!s) {
```
tom95858, line 2914: @baallan, It seems as if we are being inconsistent with the input parameter checking. If we are going to do this, we should do it for all of these API, not just the one that was flagged by the static analysis.  I also, think that the parameter checking in this and related cases should assert, not log a message. This is indicative of a programming error, not something that could/should be diagnosed by an admin from a log file.

baallan: I agree we need a large sweep on the topic of input validation, particularly in APIs that face plugin writers and configuration writers.
I'm not a big fan of using *only* assert (instead of log + assert), since it:
- can be compiled away
- states only the immediate problem (which in the particular case here would be sufficient, I agree).
This makes me wonder if we shouldn't have a ovis_log_fatal() which can be used to do both when appropriate and could be declared noreturn so that gcc/analyzers understand it right.

tom95858:
@baallan, I like your log fatal idea. Probably should log flush inside there too...I'm of a mixed opinion of the assert() inside there (instead of exit). We could make it an option I guess. The good (and bad)  thing about assert() as opposed to exit() is that you get a core dump and a stack trace.

baallan:
getting the core is nice where warranted. For ovis_log_fatal:
- Anything where the failure can be driven at runtime by user/network input should be trapped and logged the regular way. Block DOS attacks.
- where log_fatal is appropriate, call assert, then call exit just in case assert was compiled out.
- I'm of the suspicion that most plugin-writer misuses of the set/stream/msg apis would be fine to log_fatal on, but we have to figure out how to know which of those apis are also subject to network malformed data attack (e.g. network requests a set which doesn't exist (or stopped existing); this shouldn't assert.
-

morrone:
> getting the core is nice where warranted. For ovis_log_fatal:
> 
> * Anything where the failure can be driven at runtime by user/network input should be trapped and logged the regular way. Block DOS attacks.
> * where log_fatal is appropriate, call assert, then call exit just in case assert was compiled out.
> * I'm of the suspicion that most plugin-writer misuses of the set/stream/msg apis would be fine to log_fatal on, but we have to figure out how to know which of those apis are also subject to network malformed data attack (e.g. network requests a set which doesn't exist (or stopped existing); this shouldn't assert.

It should be one or the other: log_fatal(), or assert, but not both.

Where log_fatal() is appropriate, an assert is not appropriate. Instead, there should be normal, non-assert code conditional, and when the conditional fails, log_fatal() should log the message and then call abort() rather than exit(). abort() is what assert() calls on error, and will result in a core dump in the same way.  log_fatal() should be used very, very rarely.

Generally speaking, an API should neither assert nor log_fatal(). It should test for all possible inputs, and return an error when any unsupported conditions exist.

baallan:
> Where log_fatal() is appropriate, an assert is not appropriate. Instead, there should be normal, non-assert code conditional, and when the conditional fails, log_fatal() should log the message and then call abort() rather than exit(). abort() is what assert() calls on error, and will result in a core dump in the same way. 

That's a better version of what I had in mind ; thanks for the refinement.

PR #2050, opened 2025-10-14
fix leak in json_stream_sampler.c json_recv_cb error path

baallan:


ldms/src/sampler/json/json_stream_sampler.c
```diff
@@ -1148,7 +1148,7 @@ static int json_recv_cb(ldmsd_stream_client_t c, void *ctxt,
 	}
 	rbn = rbt_find(&j_schema->s_set_tree, inst_name);
 	if (!rbn) {
-		j_set = malloc(sizeof(*j_set));
+		j_set = calloc(1,sizeof(*j_set));
 		j_set->name = strdup(inst_name);
```
tom95858, line 1156: We should check if the allocation of j_set->name fails here, otherwise the rest of the logic will run along and likely crash/fail in some other way. Worst case, we end up with a good j_set->set with an instance name that doesn't match j_set->iname. The bottom part at err_1 is fine.

baallan: good catch.

tom95858:
@ballan, could you please address the comments and push an updated PR?

baallan:
> @ballan, could you please address the comments and push an updated PR?

@tom95858 did it.

PR #2051, opened 2025-10-14
fix many many nits in test_sampler.c

baallan:
- dead function
- memory leaks, most but not all, in error paths
- uninit memory use, including free(wild) in error paths
- int/unsigned conflicts that theoretically can lead to loop over max_uint.

### DIFF
--- a/ldms/src/sampler/examples/test_sampler/test_sampler.c
+++ b/ldms/src/sampler/examples/test_sampler/test_sampler.c
@@ -518,22 +518,6 @@ err:
 	return NULL;
 }
 
-int get_num_metrics(char *s)
-{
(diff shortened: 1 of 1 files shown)


baallan:
this needs test input that runs valgrind clean before merging.

tom95858:
@baallan, I would like to convert this sampler to take a JSON configuration file to allow for more convenient configuration. Currently, it has a very long configuration string with a bunch of hard to understand (without reading the code).

baallan:
@tom95858 fine by me, but maybe consider yaml instead of json?

tom95858:
@baallan, yeah that makes sense.

PR #2055, opened 2025-10-15
include nul terminator byte in messages sent by linux_proc_sampler

baallan:
this makes lightweight downstream processing easier.


baallan:
@emdonat I see tom has merged the equivalent request on main.

PR #2056, opened 2025-10-17
Add atexit handler to flush OVIS log on exit

### LINKED ISSUE #2039: no log written if there's an error in the input and -l given
with
```
ldmsd -x sock:61063:localhost \
 -c /sometest/run/conf.3 \
 -l /sometest/log/3.log \
 -v DEBUG -m 128m \
 -r /sometest/run/ldmsd.pid.3 2>&1 > /sometest/logs/3.stdio
```
if the input file  /sometest/run/conf.3 contains an error of the unknown command type, such as:
```
message_enable
```
then nothing at all gets printed to the named log file or the stdout/stderr.

If i omit the log argument, then 3.stdio does get the expected message
```
Process config file error at line $N ($filename). Function not implemented.
```

This makes it hard to debug/automatically check log output of scripted testing of multiple daemon configurations.

tom95858:
If log messages are logged to a file (not stdout or syslog()) an atexit handler will flush outstanding log messages.

This resolves #2039.

### DIFF
--- a/lib/src/ovis_log/ovis_log.c
+++ b/lib/src/ovis_log/ovis_log.c
@@ -769,6 +769,11 @@ static int __cache_stdout_stderr()
 	return 0;
 }
 
+static void log_exit(void)
+{
+	(void)ovis_log_flush();
+}
+
 int ovis_log_init(const char *name, int default_level, int modes)
 {
 	if (is_init) {
@@ -821,6 +826,7 @@ int ovis_log_init(const char *name, int default_level, int modes)
 		goto err;
 	}
 	is_init = 1;
+	atexit(log_exit);
 	return 0;
 err:
 	if (need_free)


tom95858:
This change will not affect processes that exit due to an abnormal end such as assert(), segfault, or some other process termination such as kill with a signal that is not caught by the signal handler.

tom95858:
@baallan, could you please try again with this latest change. It _may_ resolve the issue with distributed filesystems hosting the log. These filesystems implement "close to open" consistency which means that they don't guarantee file consistency to another client until the close() occurs. Note that if you are in "tail" for example on another host, you won't see the log changes until you close and reopen the file, regardless of what we do here.

baallan:
@tom95858 With  3eed29 I got one non-empty result on lustre (out of many) and a higher fraction (more like 1/3 instead of 1/5) than before of non-empty results on nfs. So I'm left wondering if its still a timing thing that happens to manifest on the slower file systems.

tom95858:
> @tom95858 With 3eed29 I got one non-empty result on lustre (out of many) and a higher fraction (more like 1/3 instead of 1/5) than before of non-empty results on nfs.

I'm not sure I know what "timinig" means. I guess you could sat that what's happening is that it is exciting before the message is even logged.

The close() thing really shouldn't have affected anything because the exit() is going to close the underying fd anyway.

> So I'm left wondering if its still a timing thing that happens to manifest on the slower file systems.

If by timing you mean that the process is exited before the message is every logged, then I would say absolutely that is a possibility. But that implies, at least, a couple things:

- A configuration error is detected and logged prior to exit.
- Another error is causing the exit and occurs prior to the configuration error being logged.

So we could try, as a debug test, adding an `atexit()` handler that logs a backtrace. Might be nice to have anyway.

baallan:
@tom95858 I'm thinking this pr and the associated issue might be left for 4.5.2 by adding a note in the NOTES or BUGS section of ldmsd man page that logging to shared file systems may not work as expected on short-lived ldmsd processes.

PR #2061, opened 2025-10-19
Remove call to dlclose in plugin unload

tom95858:
Unmapping the shared library makes all references into the library invalid. This includes callback functions
registered with `LDMS transports`, `ldms_msg`, and `ldms_stream`. The callback functions are triggered by logic in the plugin `destructor()` (e.g. `ldms_xprt_close()`, `ldms_msg_client_close()`), to clean up resources. The plugin will receive asynchronous CLOSE events in the callback functions notifying the plugin that this is the last event they can expect to receive.

Plugins that fail to properly synchronize with libraries with which they have registered callbacks may SEGFAULT if the CLOSE notifications arrive after the plugin library has been unloaded.

Without this change, plugins using transports, messages, or streams may crash in their term/destructor logic. This change temporarily removes the call to dlclose() until the proper synchronization logic has been designed and implemented.

Please see issue #2062

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -116,15 +116,15 @@ struct ldmsd_plugin_generic * load_plugin(const char *plugin_name,
 	char *saveptr = NULL;
 	char *path = getenv("LDMSD_PLUGIN_LIBPATH");
 
-        plugin = calloc(1, sizeof(*plugin));
-        if (!plugin) {
+	plugin = calloc(1, sizeof(*plugin));
+	if (!plugin) {
 		ovis_log(config_log, OVIS_LERROR,
-                         "Failed allocate memory to load the plugin '%s'\n",
-                         plugin_name);
-                goto err_0;
-        }
+			 "Failed allocate memory to load the plugin '%s'\n",
+			 plugin_name);
+		goto err_0;
+	}
 
-        if (!path)
+	if (!path)
 		path = LDMSD_PLUGIN_LIBPATH_DEFAULT;
 
 	strncpy(library_path, path, sizeof(library_path) - 1);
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -203,27 +203,28 @@ struct ldmsd_plugin_generic * load_plugin(const char *plugin_name,
...
 	return NULL;
 }
 
 static void unload_plugin(struct ldmsd_plugin_generic *plugin)
 {
-        assert(plugin != NULL);
-        free(plugin->name);
-        free(plugin->libpath);
-        plugin->api = NULL; /* about to be defunct when dlclose() is called */
-        dlclose(plugin->dl_handle);
```
morrone, line 225: Please leave this line in, just commented out.

ldms/src/ldmsd/ldmsd_config.c
```diff
@@ -203,27 +203,28 @@ struct ldmsd_plugin_generic * load_plugin(const char *plugin_name,
...
-        free(plugin);
+	/* NB: do not call dlclose() as this will invalidate
```
morrone: Please change note to make it clear this is temporary, and other bugs will be addressed to allow the dlclose() to be uncommented.

morrone (changes requested):
This needs to be reworded. This is pointing the finger at the dlclose() being a bug, but that isn't the case. This PR is a temporarily work around for bugs elsewhere in the code that are deemed to difficult to fix in time for 4.5.1.  It should be worded to make that clear. There is no "If" dlclose() is to be called, only "when" we are going to fix the bugs that allow dlclose() to be called.

The real bug is/are:

A) Failure of a plugin to fully deregister callbacks before destructor() is complete
B) Failure of the API to offer a sane way to deregister callbacks
C) Both.

Looking at the two functions mentioned in the commit message we have:

1) ldms_xprt_close() - apparently we also have a documentation bug here. The code comment for this function says "The function will disconnect the connection and free the allocated memory for the transport". This rather implies that things are done when the call returns. There is no mention of the possibility of callbacks continuing after close(), or how to know when the the callbacks are complete.

2) ldms_msg_client_close() - Here at least we do have code documentation that final callback will occur after close() time. So any plugin that doesn't wait for that in destructor() currently has a bug. After all, destructor() is expected to free all memory that the plugin instance allocated, so we would expect callbacks after destructor() to trigger SEGFAULTs anyway, regardless of the dlclose().  I would suggest that here we also need an API improvement to make it easier on the plugin author to wait until use of the API is really finished; to wait until the callbacks are really deregistered.

PR #2071, opened 2025-10-27
Add INFO log to parse-failed json msg deliveries (ldms_msg.c)

baallan:
This lets admins notify users they have a bug.
Should the level here be WARNING instead so that admins don't have to suffer the volume of INFO to help users?

Also fixes an extra ; in the local __LOG macro.
With the extra ; in place, __LOG and related macros cannot be used as
```
if (p) {
	if (x)
		LOG(y);
	....
}
```
without generating gcc 8.5.0 compilation fails.

### DIFF
--- a/ldms/src/core/ldms_msg.c
+++ b/ldms/src/core/ldms_msg.c
@@ -74,7 +74,7 @@ extern int __enable_profiling[LDMS_XPRT_OP_COUNT];
 
 static ovis_log_t __ldms_msg_log = NULL; /* see __ldms_msg_init() below */
 
-#define __LOG(LVL, FMT, ...) ovis_log(__ldms_msg_log, LVL, FMT, ##__VA_ARGS__ );
+#define __LOG(LVL, FMT, ...) ovis_log(__ldms_msg_log, LVL, FMT, ##__VA_ARGS__ )
 
 
 #define __DEBUG(FMT, ...) __LOG(OVIS_LDEBUG, FMT, ##__VA_ARGS__)
@@ -91,6 +91,8 @@ static ovis_log_t __ldms_msg_log = NULL; /* see __ldms_msg_init() below */
 #define __TIMESPEC_GT(a, b) __TIMESPEC_LT(b, a)
 #define __TIMESPEC_LE(a, b) (!__TIMESPEC_LT(b, a))
 #define __TIMESPEC_GE(a, b) __TIMESPEC_LE(b, a)
+
+#define MAX_ADDR_NAME 64
 
 static int __msg_stats_level = 1;
 
@@ -585,6 +587,7 @@ int __cred_allowed_as(struct ldms_cred *cred, struct ldms_cred *as)
 	return rc;
 }
 
+
 /* deliver message to all clients */
 static int
 __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
@@ -607,6 +610,8 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 		rc = errno;
 		goto out;
 	}
+	__DEBUG("__msg_deliver called with channel %s data size %zu\n",
+			name, data_len);
 
 	struct __msg_event_s _ev = {
 		.pub = {
@@ -707,7 +712,16 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 			_ev.pub.recv.json = json;
 			json_parser_free(jp);
 			if (rc) {
+				char srcname[MAX_ADDR_NAME];
+				ldms_addr_ntop( &_ev.pub.recv.src, srcname,
+						MAX_ADDR_NAME);
+				__DEBUG("Received malformed JSON on channel %s"
+				       " from uid %u at %s\n",
+					name, sbuf->msg->cred.uid, srcname);
 				goto cleanup;
+			} else {
+				__DEBUG("Parsed json data for delivery on channel %s size %zu\n",
+					name, data_len);
 			}
 		}
 		ref_get(&c->ref, "callback");


ldms/src/core/ldms_msg.c
```diff
@@ -707,7 +707,15 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 			_ev.pub.recv.json = json;
 			json_parser_free(jp);
 			if (rc) {
+				if (data)
+					__INFO("Could not parse data size %zu (%s) as json: %s\n",
```
tom95858: @baallan, I think this has a similar issue where an ill-intentioned application can fill up the log with messages by deliberately submitting garbage to an ldmsd. One of my tests, for example, takes your 25k JSON message and adds some garbage at the end. The resulting log message from above would be 24k in size. Given that some users run at INFO, this might be quite bad.

Better JSON parser error info is baing added that will tell the user the line and column reliably where the error occurred with the associated text. In any event, this would probably be better as a DEBUG message.

ldms/src/core/ldms_msg.c
```diff
@@ -707,7 +707,15 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 			_ev.pub.recv.json = json;
 			json_parser_free(jp);
 			if (rc) {
+				if (data)
```
tom95858: Likely should get. rid of this for now.

ldms/src/core/ldms_msg.c
baallan: I think if we leave out the json content, there isn't a log-filling attack here compared to the staggering volume of other things that show up with INFO.  And with new log controls, it can be mitigated by changing the msg log setting to warning level if there turns out to be a hostile user. 

Actually, is there a way to extract the uid of the publisher and include it here instead of the json content?  If we just say "user X send a bogus message as json on message_channel foo" regardless of details, most log handling/storing tools will neatly collapse it.  That's enough for an admin to say "you're spewing bogosity: please validate it before you send it".

tom95858: > I think if we leave out the json content, there isn't a log-filling attack here compared to the staggering volume of other things that show up with INFO. And with new log controls, it can be mitigated by changing the msg log setting to warning level if there turns out to be a hostile user.
> 

That seems reasonable to me.

> Actually, is there a way to extract the uid of the publisher and include it here instead of the json content? If we just say "user X send a bogus message as json on message_channel foo" regardless of details, most log handling/storing tools will neatly collapse it. That's enough for an admin to say "you're spewing bogosity: please validate it before you send it".

The uid and lots of other wonderful information is available in the event itself. Here it is for reference.

```
/* For message delivery to the application */
struct ldms_msg_recv_data_s {
        ldms_msg_client_t client;
        struct ldms_addr src;
        uint64_t msg_gn;
        ldms_msg_type_t type;
        uint32_t name_len;
        uint32_t data_len;
        const char *name; /* message channel name */
        const char *data; /* message data */
        json_entity_t json; /* json entity */
        struct ldms_cred cred; /* credential */
        uint32_t perm; /* permission */
        uint32_t name_hash; /* message name hash */
};

... 

typedef struct ldms_msg_event_s {
        ldms_t r; /* rail */
        enum ldms_msg_event_type type;
        struct timespec recv_ts;
        uint32_t hop_num;
        union {
                struct ldms_msg_recv_data_s recv;
                struct ldms_msg_return_status_s status;
                struct ldms_msg_client_close_event_s close;
        };
} *ldms_msg_event_t;

```

So given that information above, you could report the  message name, uid, gid, permissions, even the remote host sending the data,  etc...

So log messages like:
```
"Received malformed JSON from host 'nid0006:10111:munge', cred(1001, 1001), perm(0600)"
```
After we've added the better parser error reporting we can even get:

```
"malformed text is at line 347, column 1, '#include'"
```

which I think would be very nice.

ldms/src/core/ldms_msg.c
baallan: done

baallan:
the revised version isn't logging parse failures as expected. looking into it; converted to draft.

baallan:
@tom95858 in running the bad_messages_check test case with the revised version of this pr, the malformed json messages never make it to __msg_deliver on the L2, because apparently they never get to the L1.  The msg_deliver count on L0 is 1023 (expected), on L1 and L2 it is 501 (the count of valid slurm tagged messages). Thus the logging at the L2 never has a bad json message to log about.

Where in the chain from ldms_msg_publish on the L0 to the L1 is the invalid json data getting silently rejected? I don't find any calls to json_parse_buffer that look like they would be involved when grepping around the code.

baallan:
@tom95858 @narategithub 
Apologies for the rabbit hole. The L1 and L2 were misconfigured (not subscribing to the right msg_channel for the misformatted messages.
The mystery of the short msg_deliver count is solved, and now i get the expected log messages at L2 about misformed json.

Still working out where the client stats are.
If i run msg_client_stats on the corrected daemon configuration (where the expected 1023 messages get processed on L2) the output is as quoted below. I don't see how this is the expected output.
```
msg_client_stats L0
name                           bytes        count        first                      last
------------------------------ ------------ ------------ -------------------------- --------------------------
[bad.linux_proc_sampler_files] 127.0.0.1:33328 remote_client
 - tx                                111181          292 10-28-25 13:58:36 [908088] 10-28-25 13:58:36 [937602]
 - drops                                  0            0 -EMPTY-                    -EMPTY-
'streams'
msg_client_stats L1
name                           bytes        count        first                      last
------------------------------ ------------ ------------ -------------------------- --------------------------
[bad.linux_proc_sampler_files] 127.0.0.1:54718 remote_client
 - tx                                111181          292 10-28-25 13:58:36 [908338] 10-28-25 13:58:36 [937715]
 - drops                                  0            0 -EMPTY-                    -EMPTY-
'streams'
msg_client_stats L2
name                           bytes        count        first                      last
------------------------------ ------------ ------------ -------------------------- --------------------------
[bad.linux_proc_sampler_argv] 0.0.0.0:0 blob_msg_writer
 - tx                                     0            0 -EMPTY-                    -EMPTY-
 - drops                                  0            0 -EMPTY-                    -EMPTY-
'streams'
```

baallan:
@tom95858 this pr now works as expected; problems addressed in #2078 were confusing the operator when testing this.

baallan:
reassigning this to 4.5.2, where @tom95858 may have an alternative client reporting mechanism.

ldms/src/core/ldms_msg.c
```diff
@@ -707,7 +712,16 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 			_ev.pub.recv.json = json;
 			json_parser_free(jp);
 			if (rc) {
+				char srcname[MAX_ADDR_NAME];
+				ldms_addr_ntop( &_ev.pub.recv.src, srcname,
+						MAX_ADDR_NAME);
+				__INFO("Received malformed JSON on channel %s"
```
tom95858: @baallan, if we change this to __DEBUG, I think this PR is good to go.

baallan: @tom95858 changed it to debug. lets include in 4.5.1

tom95858:
@baallan, after further discussion it seems like these message do not belong in the library. Debug messages are for the library logic, not to diagnose application problems or provide feed back to system administrators. Certainly providing stats feedback to administrators _is_ valuable and can be addressed in v4.5.2. But log messages from the library, debug or otherwise is not thought to be the way to do it.

baallan:
> @baallan, after further discussion it seems like these message do not belong in the library. Debug messages are for the library ...

no problem. i saw your previous comment and thought your position had shifted.

tom95858:
@baallan, I wanted to provide something to the admin regarding JSON parsing errors without requiring them to look at DEBUG log messages which is a huge volume.

baallan:
> @baallan, I wanted to provide something to the admin regarding JSON parsing errors without requiring them to look at DEBUG log messages which is a huge volume.

that's fine. we can close this issue if you create a new one to track the new feature.

tom95858:
JSON parsing errors are reported in the msg_stats per PR #2084

PR #2077, opened 2025-10-28
Fix dictionary key names in msg_client_stats

nichamon:
ldmsd_controller:do_msg_client_stats() uses wrong key names of the output dictionary from ldmsd. The patch corrects the key names.


nichamon:
@tom95858 @narategithub We need this for v4.5.1; otherwise, `msg_client_stats` in `ldmsd_controller` will raise an exception in ldmsd_controller when `json=1` isn't given.

Issue #2075, opened 2025-10-28
ldms_msg_publish needs a tunable inter-line sleep interval parameter added


baallan:
this would enable a certain target of message published per time to be defined in testing.

PR #2080, opened 2025-10-29
fix ldms_msg_publish delay option, add linger option

baallan:
The delay option wasn't setting the delay; now it is.
The new linger option lets the publisher pause a specified period before closing a connection or exiting so that output pipes can drain some time (tests backed-up network theory).


baallan:
fixes #2075

baallan:
obsolete by 2081

PR #2083, opened 2025-10-30
Add a maximum wait time for QUOTA credits

tom95858:
This option sets a timeout on the conditional wait that is inside the transport callback handler. If the timeout expires, the application will exit insttead of waiting for the update indefinitely.

This option can be triggered by publishing to a peer (sitting in gdb) and waiting for the timeout to expire.

This change also fixes up some issues discovered with the usage message while adding the new -g option.

### DIFF
--- a/ldms/src/ldmsd/test/ldms_msg_publish.c
+++ b/ldms/src/ldmsd/test/ldms_msg_publish.c
@@ -6,18 +6,15 @@
 #include "ldms.h"
 
 static struct option long_opts[] = {
-	{"message_channel", required_argument, 0,  'M' },
 	{"help",     no_argument,       0,  'H' },
 	{"host",     required_argument, 0,  'h' },
 	{"port",     required_argument, 0,  'p' },
-	{"file",     required_argument, 0,  'f' },
-	{"msg_name", required_argument, 0,  'm' },
-	{"type",     required_argument, 0,  't' },
 	{"xprt",     required_argument, 0,  'x' },
-	{"max_wait", required_argument, 0,  'w' },
-	{"line_size",required_argument, 0,  'z' },
 	{"auth",     required_argument, 0,  'a' },
 	{"auth_opt", required_argument, 0,  'A' },
+	{"file",     required_argument, 0,  'f' },
+	{"msg_tag",  required_argument, 0,  'm' },
+	{"type",     required_argument, 0,  't' },
 	{"uid",      required_argument, 0,  'U' },
 	{"gid",      required_argument, 0,  'G' },
 	{"perm",     required_argument, 0,  'P' },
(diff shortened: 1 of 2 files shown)


ldms/src/ldmsd/test/ldms_msg_publish.c
```diff
@@ -6,18 +6,15 @@
...
-	{"file",     required_argument, 0,  'f' },
-	{"msg_name", required_argument, 0,  'm' },
-	{"type",     required_argument, 0,  't' },
 	{"xprt",     required_argument, 0,  'x' },
-	{"max_wait", required_argument, 0,  'w' },
-	{"line_size",required_argument, 0,  'z' },
 	{"auth",     required_argument, 0,  'a' },
 	{"auth_opt", required_argument, 0,  'A' },
+	{"file",     required_argument, 0,  'f' },
+	{"msg_tag",  required_argument, 0,  'm' },
```
baallan, line 16: why are we making up new words for this utility instead of using 'message_channel as it is in all the config-language locations?

ldms/src/ldmsd/test/ldms_msg_publish.c
```diff
@@ -358,25 +361,39 @@ int main(int argc, char **argv)
...
 	}
-
 	if (filename) {
 		file = fopen(filename, "r");
 		if (!file) {
 			perror(filename);
 			goto out;
 		}
+		if (filename && delay) {
+			printf("%s: The delay options (-D) is invalid "
```
baallan: Why is a delay between lines not combinable with the lines coming from a file?

baallan:
once other nits are sorted, we need to update the man page.

ldms/src/ldmsd/test/ldms_msg_publish.c
tom95858: Because message channel is incorrect and we need to stop spreading it's use. This is a topic for a separate issue but the term's misuse is spreading and we need to stop it because it's very confusing.

ldms/src/ldmsd/test/ldms_msg_publish.c
tom95858: Good catch, check removed.

tom95858:
@baallan, the man-page updates are in this PR. Is there something wrong with those changes?

ldms/src/ldmsd/test/ldms_msg_publish.c
baallan: so if we need to rename it, lets do so now since there has not been an official full release of the messages api yet. Is someone working on renaming it in all the ldmsd_request and plugin places that it appeared?

baallan: to be more specific, all the following use message_channel.
I'm fine with any spelling of the term, but we need to be consistent in it to reduce new user learning load long term.
I was under the impression that we went from "stream/tag" to "messages api/message_channel".
```
./man/ldmsd_controller.rst
./man/ldms_msg.rst
./src/sampler/blob_msg/blob_msg_writer.c
./src/sampler/blob_msg/ldms-sampler_blob_msg_writer.rst
./src/store/stream_dump/stream_dump.c
./src/ldmsd/test/ldms_msg_publish.rst
./src/ldmsd/test/ldms_msg_publish.c
./src/ldmsd/ldmsd_request_util.c
./python/ldmsd/ldmsd_communicator.py
./python/ldmsd/ldmsd_controller
```

PR #2084, opened 2025-10-30
Include JSON parser errors in client drops

tom95858:
The client drop now count includes messages dropped because the callback returned a value other than zero and messages sent as JSON that do not parse correctly.

An example of the msg_stats command output follows:

```
sock:localhost:10002> msg_stats
name                           bytes        count        first                      last                      
------------------------------ ------------ ------------ -------------------------- --------------------------
linux_proc_sampler_argv            10661000          100 10-29-25 23:33:29 [219637] 10-29-25 23:33:29 [237672]
 - clients:
   - [.*] stream_dump, path:msg_data:
     - tx                                 0            0 -EMPTY-                    -EMPTY-                   
     - drops                       10661000          100 10-29-25 23:33:29 [219637] 10-29-25 23:33:29 [237672]
```

### DIFF
--- a/ldms/src/core/ldms_msg.c
+++ b/ldms/src/core/ldms_msg.c
@@ -707,6 +707,8 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 			_ev.pub.recv.json = json;
 			json_parser_free(jp);
 			if (rc) {
+				__counters_update(&sce->drops, &now, data_len);
+				__counters_update(&c->drops, &now, data_len);
 				goto cleanup;
 			}
 		}
@@ -717,7 +719,7 @@ __msg_deliver(struct __msg_buf_s *sbuf, uint64_t msg_gn,
 		if (__msg_stats_level > 0) {
 			pthread_rwlock_wrlock(&c->rwlock);
 			if (rc) {
-				__counters_update( &sce->drops, &now, data_len);
+				__counters_update(&sce->drops, &now, data_len);
 				__counters_update(&c->drops, &now, data_len);
 			} else {
 				__counters_update(&sce->tx, &now, data_len);


baallan:
@tom95858 what did drops count before and why wouldn't we want to distinguish 'what that was' from 'not json as checked by the parse'? A 'parse_fail' counter seems in order (unless json parse is the only reason at all for a 'drop').

This is particularly of interest because for performance reasons we likely want to give the subscriber (rather than the publisher) control over whether arriving alleged json is parsed on the subscribers behalf by the framework. 
e.g. blob_msg_writer really doesn't care what the message format is, it just wants to capture all messages as the char array sent. Other things we have planned don't need an ovis_json tree-- they want to use their own faster json or they just want to pull out one key/value pair with a regex to route something.

tom95858:
> @tom95858 what did drops count before 

As mentioned in the commit message, it counts the number of messages for which the client callback returned !0. 

> ... and why wouldn't we want to distinguish 'what that was' from 'not json as checked by the parse'? A 'parse_fail' counter seems in order (unless json parse is the only reason at all for a 'drop').
>

One might argue that the callback return code counting should be removed because the message was, in-fact, delivered, but for some reason the callback didn't like it, if !0 is even what that means.

For now, it includes both.

> This is particularly of interest because for performance reasons we likely want to give the subscriber (rather than the publisher) control over whether arriving alleged json is parsed on the subscribers behalf by the framework. e.g. blob_msg_writer really doesn't care what the message format is, it just wants to capture all messages as the char array sent. Other things we have planned don't need an ovis_json tree-- they want to use their own faster json or they just want to pull out one key/value pair with a regex to route something.

It's not obvious how the stats count in this patch is related to elective parsing, or partial parsing.

Could we please move that discussion to an issue.

PR #2088, opened 2025-10-31
Fix ldms ls json

tom95858:
The `ldms_ls` output with -f json does not produce correct JSON with the metric set contains lists and/or records.

This pull request adds core services to format a metric set as JSON. These services are used by `ldms_ls` and other samplers and core infrastructure to place this conversion one place.

### DIFF
--- a/ldms/src/ldmsd/ldms_ls.c
+++ b/ldms/src/ldmsd/ldms_ls.c
@@ -70,11 +70,11 @@
 #include <pwd.h>
 #include <grp.h>
 #include "ldms.h"
+#include "ldms_json.h"
 #include "config.h"
 
-#define LDMS_LS_MEM_SZ_ENVVAR "LDMS_LS_MEM_SZ"
-#define LDMS_LS_MAX_MEM_SIZE 512L * 1024L * 1024L
-#define LDMS_LS_MAX_MEM_SZ_STR "512MB"
+#define MEM_SZ_ENVVAR "LDMS_LS_MEM_SZ"
+#define DEF_MEM_SZ_STR "512MB"
 
 /* from ldmsd_group.c */
 #define GRP_SCHEMA_NAME "ldmsd_grp_schema"
@@ -185,8 +185,8 @@ void usage(char *argv[])
 	       "                     instead of giving the -m option. If both are given,\n"
 	       "                     this option takes precedence over the environment variable.\n"
 	       "\n    -a <auth>        LDMS Authentication plugin to be used (default: 'none').\n"
-	       "\n    -A <key>=<value> (repeatable) LDMS Authentication plugin parameters.\n"
-	       , LDMS_LS_MAX_MEM_SZ_STR, LDMS_LS_MEM_SZ_ENVVAR);
+	       "\n    -A <key>=<value> (repeatable) LDMS Authentication plugin parameters.\n",
+	       DEF_MEM_SZ_STR, MEM_SZ_ENVVAR);
 	printf("\n    -V               Print LDMS version and exit.\n");
 	printf("\n    -P               Register for push updates.\n");
 	exit(1);
@@ -645,7 +645,6 @@ void value_format(ldms_set_t s, enum ldms_value_type type, ldms_mval_t val, size
 					printf("]");
 			}
 			printf("]");
-			printf("\n");
 		}
 		break;
 	case LDMS_V_TIMESTAMP:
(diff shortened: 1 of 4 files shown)


ldms/src/core/ldms_json.c
```diff
@@ -0,0 +1,415 @@
...
+
+static char *json_encode_str(char *dst, size_t dst_len, char *src)
+{
+	char *d, *s;
+	size_t buf_len = dst_len;
+
+	d = dst;
+	s = src;
+
+	while (*s != '\0' || !buf_len) {
```
nichamon: I believe the condition should be `(*s != '\0' && buf_len)`.

ldms/src/core/ldms_json.c
```diff
@@ -0,0 +1,415 @@
...
+{
+	enum ldms_value_type item_type;
+	ldms_mval_t rec_inst_val;
+	size_t array_len;
+
+	fprintf(fp, "%*s[\n", indent, "");
+	rec_inst_val = ldms_list_first(s, list_mval,
+				       &item_type, &array_len);
+	while (rec_inst_val) {
+		fprint_record_json(fp, indent+4, rec_inst_val);
```
nichamon: The function supports only the case of list of records. However, there are sets containing lists of other value types, For example, procstat2 creates the set of interrupt data `intr_set` which contains a list of U64.

nichamon (changes requested):

ldms/src/ldmsd/ldms_ls.c
```diff
@@ -1242,27 +1038,20 @@ long total_meta;
...
+				printf("%s\n", set_data->inst_name);
 			break;
 		case sf_tab:
-			if (format_first) {
-				printf("#instance\n");
-				format_first = 0;
-			}
+			if (!verbose && !long_format)
+				printf("%s\n", set_data->inst_name);
 			printf("%s\n", set_data->inst_name);
```
nichamon: Extra line that prints the set instance name

ldms/src/ldmsd/ldms_ls.c
```diff
@@ -1242,27 +1038,20 @@ long total_meta;
...
 long total_sets;
 
-
 void print_set(struct ldms_dir_set_s *set_data)
 {
 	if (!verbose) {
 		switch (format) {
 		case sf_normal:
-			printf("%s\n", set_data->inst_name);
+			if (!verbose && !long_format)
```
nichamon: It isn't necessary to check the verbose flag again.

tom95858: Got it.

ldms/src/core/ldms_json.c
tom95858: I'll add the support, but procstat2 does not have a list of primitive types. It is a list of records.

```
calaluna/ps2: consistent, last update: Wed Nov 05 07:33:10 2025 -1000 [2556us] 
M u64          component_id                               0 
D u64          job_id                                     0 
D u64          app_id                                     0 
M record_type  cpu_rec                                    LDMS_V_RECORD_TYPE
D list<>       cpu_list                                   
    name user (jiffies) nice (jiffies) system (jiffies) idle (jiffies) iowait (jiffies) irq (jiffies) softirq (jiffies) steal (jiffies) guest (jiffies) guest_nice (jiffies)
   "cpu"        1865835          10090          1275221      229140671           162726             0            190319               0               0                    0
  "cpu0"         806404           2861           327236       57025164            34731             0             80367               0               0                    0
  "cpu1"         367573           2485           320264       57459642            39333             0             52649               0               0                    0
  "cpu2"         274788           2023           216690       57348410            45380             0             32949               0               0                    0
  "cpu3"         417070           2721           411031       57307453            43280             0             24354               0               0                    0 
D u64          ctxt                                       543169097 
D u64          btime                                      1761780942 seconds
D u64          processes                                  9056094 
D u64          procs_running                              1 
D u64          procs_blocked                              0 
```

tom95858:
@nichamon, can you check this. I need this for some other code that I'd like to merge

ldms/src/core/ldms_json.c
tom95858: If we've emptied the string (*s == '\0') or there is no buffer remaining (buf_len == 0), the loop exits. I _think_ it's correct.

nichamon: The condition must be met to execute the logic inside the while loop. Currently, the condition checks if it is not the end of the source string OR there is no buffer left (`*s != '\0' || !buf_len`). This means when `buf_len` reaches 0, `!buf_len` becomes true, so the loop continues and writes past the buffer end. The logical operator should be AND (`&&`), not OR (`||`), and it should check `buf_len` (not `!buf_len`): `*s != '\0' && buf_len`. This ensures we only continue when there's still source data AND buffer space available.

This is the only item I think needs to be addressed. Once the loop condition is addressed, everything else looks good to me.

tom95858: @nichamon, I'm refactoring this all to use fprintf. Let libc do this work ;-)

tom95858:
@nichamon, are we good to go?

PR #2090, opened 2025-10-31
Add services to simplify plugin config

tom95858:
These new API handle common configuration coding patterns that are replicated in our code base. The new API include:
 
* av_value_default   - A common code pattern is av_value, and if NULL a strdup() of the provided default value is used. This allows the subsequent code to treat the return value the same regardless of whether the value was found, or the default is used.
* av_value_list.         - Parse a comma separated list of name=value into an av_list. This is very useful for simplifying auth plugin option handling. May interfaces take an attr-value-list (e.g. auth), this service handles converting the comma separated list to an attr-value list.
- strous, strtos        - This is useful for range checking short values provided in config lines by adding this logic in a single place. This is notably used for validating and converting port numbers passed as strings. Similar to `strtod, strtol, ...` this function sets errno to ERANGE if there is a overflow in the target type.

### DIFF
--- a/lib/src/ovis_util/util.c
+++ b/lib/src/ovis_util/util.c
@@ -64,6 +64,7 @@
 #include <grp.h>
 #include <ctype.h>
 #include <dirent.h>
+#include <limits.h>
 
 #include "util.h"
 #define DSTRING_USE_SHORT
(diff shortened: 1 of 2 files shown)


lib/src/ovis_util/util.h
```diff
@@ -107,17 +109,54 @@ typedef struct string_ref_s {
...
+ * Separates \c input into <name>=<value> pairs. The separater for
+ * splitting the input string is specified by \c sep.
+ *
+ * \param input A NULL terminated string containing <name>=<value> pairs
+ * \param sep A string specifying the input string separator character.
+ *
+ * \returns An avl_list_t containing name,value pairs or NULL if there
+ * is a memory allocation error.
+ */
+av_list_t av_value_list(const char *input, const char *sep);
```
nichamon, line 163: Should `sep` be `char`? It seems that the implementation at line `ovis.c:397` assumes that `sep` is a single character. What do you think?

lib/src/ovis_util/util.c
```diff
@@ -330,6 +331,94 @@ char *av_value(struct attr_value_list *av_list, const char *name)
...
+
+	if (!input)
+		return NULL;
+	str = strdup(input);
+	if (!str)
+		return NULL;
+
+	/* Determine the size of the output av_list */
+	s = str;
+	for (i = 1; s; s = strstr(str, sep))
```
narategithub: I think this is an infinite loop when `sep` is found in `str`, since `s` always points to the first `sep` in `str`.

baallan:
a) I suggest change sep to char* and implement this with strtok_r, since strtok cannot be used thread-safely by plugins and reasonable folk have been known to split on ":;" or other things.

b) hard-coding the 512 array size is a vulnerability. strlen(input)+1 is a useful upper bound.

lib/src/ovis_util/util.h
tom95858: @nichamon thanks, I have implemented the necessary changes. I will update the pull request.

PR #2091, opened 2025-10-31
store_influx: fix faulty job_id assumption

morrone:
store_influx incorrectly assumed that all metric sets will have a job_id field. This assumption leads to triggering an assert when the job_id field does not exist in a metric set.

Add a check for the existence of the job_id id field, and only include it as a tag when it actually exists. The comp_id field is likewise checked, just for completeness.

### DIFF
--- a/ldms/src/store/influx/store_influx.c
+++ b/ldms/src/store/influx/store_influx.c
@@ -47,6 +47,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <linux/limits.h>
 #include <pthread.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


morrone:
Compiles, but untested.

tom95858:
> Compiles, but untested.

Should this be DRAFT or can we get the reporter to test it?

morrone:
> > Compiles, but untested.
> 
> Should this be DRAFT or can we get the reporter to test it?

Well, this is a simple enough change that I would have been fine with it landing after a normal code review.

But the reporter did test it, and they are no longer seeing the Assertion that they were hitting previously. So I believe it is ready to land. @emdonat

morrone:
@emdonat The equivalent is landed on main.

morrone:
It doesn't look like 4.4.7 is going to happen. Dropping this PR.

PR #2097, opened 2025-11-07
Enable log level configuration before subsystem registration

### LINKED ISSUE #1342: reordering of subsystem log statements defeats log_level commands for debugging

nichamon:
This PR enables flexible log level configuration by allowing log levels to be set for subsystems before they are registered. This resolves the issue where ldmsd users could not configure plugin log levels in configuration files due to command processing order.

Resolves #1342

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -2965,10 +2965,39 @@ class LdmsdCmdParser(cmd.Cmd):
             print(f"Error getting the logger infomation. {msg}")
         loggers = fmt_status(msg)
 
-        print(f"{'Name':40} {'Log Level':30} Description")
-        print(f"{'-'*40} {'-'*30} {'-'*30}")
+        unregistered_loggers = []
+        valid_loggers = []
 
         for l in loggers:
+            if l['registered'] == 'false':
+                unregistered_loggers.append(l)
+            else:
+                valid_loggers.append(l)
+
+        if 0 < len(unregistered_loggers):
+            print("")
+            print("NOTICE: The following subsystems were configured for specific log levels but have not registrered:")
+            print("        This usually indicates a mis-spelling of the subsystem name.")
+            print("")
+            print("Unregistered log subsystems:")
+            print(f"{'-'*(40 + 30 + 1)}")
+            print(f"{'Name':40} Log Level")
+            print(f"{'-'*40} {'-'*30}")
+            for l in unregistered_loggers:
+                print(f"{l['name']:40} {l['level']}")
+            print(f"{'-'*(40 + 30 + 1)}")
+        else:
+            print("")
+            print("All configured log subsystems have successfully registered with the logging system.")
+
+        print("")
+        print("The following shows all log subsystems that have successfully registered with the logging system.")
+        print("")
+        print("Registered log subsystems:")
+        print(f"{'-'*(40 + 30 + 30 + 2)}")
+        print(f"{'Name':40} {'Log Level':30} Description")
+        print(f"{'-'*40} {'-'*30} {'-'*30}")
+        for l in valid_loggers:
             print(f"{l['name']:40} {l['level']:30} {l['desc']}")
         print(f"{'-'*(40 + 30 + 30 + 2)}")
         print("The loggers with the Log Level as 'default' use the same "


nichamon:
This approach creates two issues for users configuring plugin log levels:

- Users need to know and remember the exact subsystem naming format
- Typos in the subsystem name leads to silent configuration errors where the log level is not actually set

A new pull request resolving issue #1342 in a different approach will be created. Namely, an optional attribute `log_level` will be added to the `load` command so that the log level of the plugin instance will set to the desired level when the plugin instance is loaded.

I'm closing this pull request.

baallan:
@nichamon adding the option to the plugin load *only* is not satisfactory because it forces changes into the body of the config/maestro/ansible stanza to be debugged (and possibly a rerun of ansible). Forward definition of the log levels does not disturb canned configurations in this way.

It would be entirely sufficient for the log status command to list in a final section all forward declared names which are not yet associated. This will indicate both typos and plugins which failed to load for any other reason. Anyone setting a log level on a specific facility will be using the log status command anyway.

Please add this version, even if you also implement log options on plugin config/load lines.

baallan:
> A new pull request resolving issue #1342 in a different approach will be created. Namely, an optional attribute `log_level` will be added to the `load` command so that the log level of the plugin instance will set to the desired level when the plugin instance is loaded.

If adding log_level as an option to load commands *is* implemented, then the forward declaration of log_level needs to take precedence. e.g. a live-admin defined preamble -c file which sets the log level to DEBUG for a plugin before a boot-time generated yaml file which defines the as log_level=INFO should win, assuming the ldmsd command line is:
```
ldmsd -c debug.conf -y config_to_be_debugged.yaml ...
```

morrone:
> ```
> ldmsd -c debug.conf -y config_to_be_debugged.yaml ...
> ```

Wow, I didn't even know you could do both -c and -y on the command line. I wonder if that was more of an accident of implementation than something that should be relied upon. After all, you can't do "ldmsd -c debug.conf -c ansible_generated.conf", right.

Can we specify subsystem log levels on the command line? I think that would probably be a better way to handle the debugging needs without editing automatically generated files.

tom95858:
@morrone, I think we should disallow allowing both on the command line. Just way too confusing I think, however, multiple -c and/or -y *may* be useful for something like this:
```
ldmsd -c base-config.conf -c node-config.conf
```
That said, I don't think it actually works that way because we have this notion of 'command-priority', specifically, some commands are applied out-of-order WRT the others.

What I think happens now is that the last -c is what is honored. In my view, we should probably fail multiple -c, -y, and a mix of -c and -y.

@nichamon, can you please comment on this?

baallan:
my last read of the code was that multiple c and y options get processed left to right.
It's possible that's evolved.
We also promote load/log statements earlier (hence some of the problems) while processing an individual file.
From an admin point of view, I'd like to be able to write a systemd script that looks something like:
```
ldmsd -c /etc/sysconfig/ldms.d/ldms.%I.debug.conf -y /etc/sysconfig/ldms.d/ldms.%I.yaml ...
```
where the normal content of ldms.%I.debug.conf is all-commented-out/empty and the yaml is managed by ansible and the 'regular admin' in debug mode puts in the debug.conf file only things like changing the default set/stream/message permissions and forward declarations of log levels/log file locations for problems under investigation.

This way, neither the service file nor the managed .yaml/.conf is touched while being investigated.

morrone:
> ```
> ldmsd -c /etc/sysconfig/ldms.d/ldms.%I.debug.conf -y /etc/sysconfig/ldms.d/ldms.%I.yaml ...
> ```
> 
> where the normal content of ldms.%I.debug.conf is all-commented-out/empty and the yaml is managed by 

- /etc/sysconfig is not really the correct directory for configuration files like this. It should be something like /etc/ldmsd
- Directories that end in ".d" in /etc tend to have a common meaning. They almost universally contain configuration files that are all read by a particular application in file name sorting order (which is why they often start with numbers to make it obvious the order that folks want them to be processed). It is best not to reuse that directory naming strategy for things that do not match that idiom.
- If one wants to use sysconfig with a config file, there is a systemd service file example example in the tree shows how to do that: https://github.com/ovis-hpc/ldms/raw/refs/heads/main/ldms/etc/systemd/ldmsd@.service.in. The sysconfig file can just set one or both of the CONFIG_FILE and OPTIONS variables in that example.

I think that having just one of either -c or -y supported at at time is plenty. Being able to override some things like verbosity from the command line takes care of most minor debugging needs, and for anything more difficult we can just copy the config file and make changes to it.

I think that having empty "debug" files appear in the normal installed systemd files all the time is the tail wagging the dog. We should keep the normal configs on production systems as clean and understandable as we reasonably can. I think that means we don't clutter them with debug details that are only used a fraction of the time.

tom95858:
@morrone, @nichamon, @baallan either scenario requires that we create the log subsystem prior to registration. So perhaps we should allow the pre-registration and add the log_status reporting to show "orphaned" subsystems, specifically subsystems for which a log level has been set, but for which no registration has occurred.

tom95858:
@nichamon, do we have status on this PR?

nichamon:
@tom95858 I have rebased the PR on top of the main branch.

The log_status now reports the unregistered log subsystems separately from the main table.

With unregistered log subsystems:
image

Without unregistered log subsystems:
image

nichamon:
@tom95858 This is ready to be merged.

baallan:
@tom95858 Can we merge this?
@nichamon is there anything that would make backporting/cherry picking this to b4.5 difficult?

PR #2098, opened 2025-11-10
Fix function declaration/call found by gnu23 standard

narategithub:
Prior GCC gnu standards (and C standard, e.g. gnu17) treat `int fn();` prototype as _any_ number of arguments. On openSUSE 16.0, GCC uses gnu23 standard as a default, which interpret `int fn();` as having _exactly_ zero arguments. This patch addressed declarations & calls found by the new gnu23 standard.


tom95858:
There are some other issues too. Try this:
```
-D_FORTIFY_SOURCE=1 -O1
```
In your CFLAGS for configure. This spit out 100s of warnings for me. Maybe we change this compiler biz all at once?

Issue #2086, opened 2025-10-30
make ldmsd preparsing of json typed messages opt-in for the message subscriber api


baallan:
The frame-work side (ldmsd) ovis_json parse of messages that are json-typed (by the publisher) is a performance optimization for certain use-cases involving multiple subscribers that all want the same ovis_json object tree.

We have other message use cases where only one subscriber is ever likely and that subscriber doesn't want an ovis_json tree (or its cpu/memory cost).
Alternate use cases:
- blob_msg_writer which wants to just store the message bytes, ignoring the type.
- a json database like mongo where we would just use the envelope information that the new messages api provides, leaving the consumption of the message body to the DB and its json parser.
- anyone with a plugin that wants to use a different parser, such as yyjson mentioned in #1593 .
- a plugin that just wants to use a regex to pull out select fields in a single pass over the string, rather than building a data structure.

A trivial way to achieve this would be to let the plugin writer indicate their requirements in either of the following:
```
(a) ( i'm not a big fan of this one)
// delivers only events that are both typed as json by the sender and preparsed to ovis_json 
ldms_msg_client_t
ldms_msg_subscribe_json(const char *match, int is_regex,
                       ldms_msg_event_cb_t cb_fn, void *cb_arg, const char *desc);
(b)
// preprocess/deliver  messages as indicated by the filter bits (I like this as more extensible)
ldms_msg_client_t
ldms_msg_subscribe_filtered(const char *match, int is_regex,
                       ldms_msg_event_cb_t cb_fn, void *cb_arg, const char *desc, uint64_t filter_bits);
// filterbits
#define MSG_FILTER_ALL 0
#define MSG_FILTER_JSON_ONLY 0x1
#define MSG_FILTER_STRING_ONLY 0x2
// other flags, maybe avro related?
#define MSG_FILTER_OVIS_JSON_PARSED 0x100
```

baallan:
@morrone @evanmcc, your input here would be appreciated.

morrone:
My interest in using the ldms messsage component is low, so I don't have a great deal of input. I only have a minor comment that if you go with the "filter_bits", it probably needs to be an enum, or you need to invert the meaning of your bits to make them easily ORable. So have a bit for each thing you want:

```
#define MSG_FILTER_ALL ~0
#define MSG_FILTER_JSON 0x1
#define MSG_FILTER_STRING 0x2
#define MSG_FILTER_AVRO 0x4
```
[...]

PR #2101, opened 2025-11-15
JSON perf improvements

tom95858:
Performance improvements for the JSON parser

These changes improve the performance of the JSON parser
by a factor of more than five. Specific improvements include:

- Avoiding calls to malloc in most cases
- Replace the attribute hash table with an RBT
- Embedding most values in the attribute itself to
   avoid memory allocation
- Replacing the flex lexer with a hand-coded scanner
- Improved buffer tracking for better error reporting.
   Added json_parse_error() to return an error string.
- Parsing of input buffer data is done directly on the
   buffer itself avoiding duplication. The input buffer
   is not modified.
- The implementation data structures have been made opaque
   in order to prevent users from digging into the private
   data structures instead of using the appropriate API.

This parser has had limited testing with ldmsd. Many
users were peeking into the JSON data structures directly instead
of using the API.

### DIFF
--- a/ldms/src/sampler/slurm/slurm_sampler2.c
+++ b/ldms/src/sampler/slurm/slurm_sampler2.c
@@ -854,7 +854,7 @@ static void handle_step_init(job_data_t job, json_entity_t e)
 	if (av) {
 		if (json_entity_type(av) == JSON_STRING_VALUE) {
 			snprintf(job->user, JOB_USER_STR_LEN, "%s",
-					json_value_str(av)->str);
+					json_value_cstr(av));
 		}
 	}
 
@@ -863,7 +863,7 @@ static void handle_step_init(job_data_t job, json_entity_t e)
 	if (av) {
 		if (json_entity_type(av) == JSON_STRING_VALUE) {
 			snprintf(job->job_name, JOB_NAME_STR_LEN, "%s",
-					json_value_str(av)->str);
+					json_value_cstr(av));
 		}
 	}
 
@@ -876,7 +876,7 @@ static void handle_step_init(job_data_t job, json_entity_t e)
 			if (av) {
 				if (json_entity_type(av) == JSON_STRING_VALUE) {
 					snprintf(job->job_tag, JOB_TAG_STR_LEN,
-						"%s", json_value_str(av)->str);
+						"%s", json_value_cstr(av));
 				}
 			}
 		}
@@ -1096,24 +1096,24 @@ static int slurm_recv_cb(ldms_msg_event_t ev, void *ctxt)
 		goto err_0;
 	}
 
-	json_str_t event_name = json_value_str(event);
+	char *event_name = json_value_cstr(event);
 	dict = json_value_find(ev->recv.json, "data");
 	if (!dict) {
 		ovis_log(mylog, OVIS_LERROR, "'%s' event is missing "
-		       "the 'data' attribute\n", event_name->str);
+		       "the 'data' attribute\n", event_name);
 		goto err_0;
 	}
 	av = json_value_find(dict, "job_id");
 	if (!av) {
 		ovis_log(mylog, OVIS_LERROR, "'%s' event is missing the "
-		       "'job_id' attribute.\n", event_name->str);
+		       "'job_id' attribute.\n", event_name);
 		goto err_0;
 	}
 	uint64_t job_id = json_value_int(av);
 
 	job_data_t job;
 	pthread_mutex_lock(&job_tree_lock);
-	if (0 == strncmp(event_name->str, "init", 4)) {
+	if (0 == strncmp(event_name, "init", 4)) {
 		job = job_data_find(job_id);
 		if (!job) {
 			job = job_data_alloc(job_id);
@@ -1126,7 +1126,7 @@ static int slurm_recv_cb(ldms_msg_event_t ev, void *ctxt)
 			}
 		}
 		handle_job_init(job, ev->recv.json);
-	} else if (0 == strncmp(event_name->str, "step_init", 9)) {
+	} else if (0 == strncmp(event_name, "step_init", 9)) {
 		job = job_data_find(job_id);
 		if (!job) {
 			job = job_data_alloc(job_id);
@@ -1140,34 +1140,34 @@ static int slurm_recv_cb(ldms_msg_event_t ev, void *ctxt)
 			handle_job_init(job, ev->recv.json);
 			ovis_log(mylog, OVIS_LERROR, "'%s' event was received "
 					    "for job %ld with no job_data.\n",
-					    event_name->str, job_id);
+					    event_name, job_id);
 			goto unlock_tree;
 		}
 		handle_step_init(job, ev->recv.json);
-	} else if (0 == strncmp(event_name->str, "task_init_priv", 14)) {
+	} else if (0 == strncmp(event_name, "task_init_priv", 14)) {
 		job = job_data_find(job_id);
 		if (!job) {
 			ovis_log(mylog, OVIS_LERROR, "'%s' event was received "
 					"for job %ld with no job_data.\n",
-					event_name->str, job_id);
+					event_name, job_id);
 			goto unlock_tree;
 		}
 		handle_task_init(job, ev->recv.json);
-	} else if (0 == strncmp(event_name->str, "task_exit", 9)) {
+	} else if (0 == strncmp(event_name, "task_exit", 9)) {
 		job = job_data_find(job_id);
 		if (!job) {
 			ovis_log(mylog, OVIS_LERROR, "'%s' event was received "
 					"for job %ld with no job_data.\n",
-					event_name->str, job_id);
+					event_name, job_id);
 			goto unlock_tree;
 		}
 		handle_task_exit(job, ev->recv.json);
-	} else if (0 == strncmp(event_name->str, "exit", 4)) {
+	} else if (0 == strncmp(event_name, "exit", 4)) {
 		job = job_data_find(job_id);
 		if (!job) {
 			ovis_log(mylog, OVIS_LERROR, "'%s' event was received "
 					"for job %ld with no job_data.\n",
-					event_name->str, job_id);
+					event_name, job_id);
 			goto unlock_tree;
 		}
 		if (!job->exited) {
@@ -1184,7 +1184,7 @@ static int slurm_recv_cb(ldms_msg_event_t ev, void *ctxt)
 		}
 	} else {
 		ovis_log(mylog, OVIS_LDEBUG, "ignoring event '%s'\n",
-						   event_name->str);
+						   event_name);
 	}
 	pthread_mutex_unlock(&job_tree_lock);
 	return 0;
--- a/ldms/src/store/kafka/store_kafka.c
+++ b/ldms/src/store/kafka/store_kafka.c
@@ -110,9 +110,10 @@ static int config(ldmsd_plug_handle_t handle, struct attr_value_list *kwl,
 {
 	int rc = 0;
 	const char *path;
-	json_entity_t jdoc = NULL, jent, jval;
-	json_str_t jkey;
-	json_parser_t jp = NULL;
+	json_doc_t jdoc = NULL;
+	json_entity_t jent, jval;
+	json_entity_t root = NULL;
+	char *jkey;
 	char *buff = NULL;
 	size_t buff_sz, buff_len = 0;
 	ssize_t len;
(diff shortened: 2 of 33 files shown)


baallan:
What's the rationale for the particular size of json_attr_name_max? Based on inspection of the linux environment and module structure on my hpc systems, 32 or 64 seems a more likely value for *name_max and a separate json_attr_value_max of 512  or 256 look reasonable.

 We don't want these too small to avoid unnecessary mallocs/memory fragmentation or too big to avoid excessive over-allocation. Ideally, both these values could be set at ldmsd configure time if someone has a system where this needs different tuning.

tom95858:
@baallan because the extra name length is free. The entity is a union so it's size is that of the max union member. It didn't make sense to me to limit the size of the name unnecessarily.

tom95858:
@nichamon, I made a change here that I'd like you to comment on. Specifically, if an attribute were added to a dictionary that already has that attribute name, the logic would replace the old attribute with the new one. This was only depended on in one place afaict and that is in json_dict_merge().

baallan:
> @baallan because the extra name length is free. The entity is a union so it's size is that of the max union member. It didn't make sense to me to limit the size of the name unnecessarily.

The union seems to be strongly dominated by 2*NAME_MAX from attr, with every node therefore being a size somewhat over 512bytes currently.  Might want to think about that and consider changing the default string size so the sizeof for allocating a default node works out to be at (or just under) a power of 2 bytes <= 256.

baallan:
> If an attribute were added to a dictionary that already has that attribute name, the logic would replace the old attribute with the new one. This was only depended on in one place afaict and that is in json_dict_merge().

Silent replacement should only occur if the old and new are of the same type. Type-conflicted replacement should return an error unless some optional stomping set function is used.

lib/src/ovis_json/ovis_json.h
```diff
@@ -242,18 +305,7 @@ extern json_entity_t json_entity_copy(json_entity_t e);
...
- * \brief Add the attributes of dictionary \c src into dictionary \c dst
- *
- * The attribute value in \c src will replace the value in \c dst,
- * if the attribute key presents in both dictionaries. \c dst will be modified
- * and \c src is left unchanged.
- *
- * \return 0 on success. ENOMEM if the function fails to add a new attribute to \c src
- */
-extern int json_dict_merge(json_entity_t dst, json_entity_t src);
+json_entity_t json_dict_build(json_parser_t p, json_entity_t d, ...);
```
morrone: I don't understand this API. Why, when we are _generating_ json, do we need a _parser_ handle? I would think we need a parser handle at parsing time, not generating time.

lib/src/ovis_json/ovis_json.h
```diff
@@ -179,33 +218,59 @@ extern jbuf_t json_entity_dump(jbuf_t jb, json_entity_t e);
...
+ * because sizeof(int) != sizeof(size_t) on some architectures
+ * (e.g. arm).  Because strlen() returns size_t, size_t is used for
+ * the length argument. The reason for requiring the length is to
+ * allow string values to be initialized from unterminated strings.
+ *
+ * \example
  *
- * The last value must be -1 to end the attribute value list.
+ * char *str = "this is a string, but only 'string' is being used.";
+ * json_entity_t *e = json_dict_build("my-string", JSON_STRING_VALUE,
```
morrone: This call doesn't seem to match the function declaration for json_dict_build(). The first two values are supposed to be a json_parser_t handle and then a json_entity_t handle.

The next json_dict_build() example in this code comment needs fixing as well.

lib/src/ovis_json/ovis_json.h
```diff
@@ -214,21 +279,19 @@ extern json_entity_t json_entity_copy(json_entity_t e);
  * 	JSON_INT_VALUE,    "int",    (int64_t)1,
  * 	JSON_BOOL_VALUE,   "bool",   (int32_t)1,
  * 	JSON_FLOAT_VALUE,  "float",  1.1,
- * 	JSON_STRING_VALUE, "string", "str",
+ * 	JSON_STRING_VALUE, "string", "str", strlen("str"),
  * 	JSON_LIST_VALUE,   "list",   JSON_INT_VALUE, (int64_t)1,
- * 				     JSON_STRING_VALUE, "last",
- * 				     -2,
+ * 				     JSON_STRING_VALUE, "last", strlen(last),
+ * 				     JSON_EOL_VALUE, // end LIST_VALUE argumen list
```
morrone: s/argumen/argument

lib/src/ovis_json/ovis_json.h
```diff
@@ -179,33 +218,59 @@ extern jbuf_t json_entity_dump(jbuf_t jb, json_entity_t e);
 /**
  * \brief Create a new JSON entity identical to the given entity \c e.
  *
+ * \param p	The parser handle for the copied entities
```
morrone: I have no idea what that means, or why the parser handle is needed here.

morrone:
> > If an attribute were added to a dictionary that already has that attribute name, the logic would replace the old attribute with the new one. This was only depended on in one place afaict and that is in json_dict_merge().
> 
> Silent replacement should only occur if the old and new are of the same type. Type-conflicted replacement should return an error unless some optional stomping set function is used.

I am not sure why that should be for json. I don't think that is how libjansson behaves, for instance. I think it just replaces the attribute regardless of type.

tom95858:
> > @baallan because the extra name length is free. The entity is a union so it's size is that of the max union member. It didn't make sense to me to limit the size of the name unnecessarily.
> 
> The union seems to be strongly dominated by 2*NAME_MAX from attr, 

I don't know where `2 * NAME_MAX` comes from. The object size including the malloc prefix is 360B. Depending on the malloc configuration this may end up being in the 512B malloc bucket. If it's smaller than the bucket start limit it is (size+grain_size-1) % grain_size

Grain size is typically 16B, but can be configured. I don't think it is a good idea to have the name length governed by some power of 2 minus something else because then the attribute name length is different on different platforms based on structure member alignment and other factors.

>with every node therefore being a size somewhat over 512bytes currently.

As mentioned, it's ~360B...which may go into a 512B bucket; but it's not over 512B, which would end up in the 1024B bucket..

> Might want to think about that and consider changing the default string size so the sizeof for allocating a default node works out to be at (or just under) a power of 2 bytes <= 256.

Based on the test data that I saw from the 4.5.1 testing there were many many strings in the input that would not fit in that space and would result in calls to malloc() and free(). That said, the data from `ldmsd` and associated status commands has what are typically very small strings where small is < 32B. 

I am not wed to the size at all. We should make it whatever makes sense for our common workload. If it's the test data from 4.5.1, then that is a bit larger. Presently, malloc doesn't even show up in the perf report because it's so small. The motivation for this effort was  to make the JSON overhead small enough such that it didn't warrant changing all of the code to use yyjson which is, per my testing, still quite a bit faster.

lib/src/ovis_json/ovis_json.h
tom95858: @morrone,  the parser holds all of the memory for objects allocated for the parser. That should probably be documented along with the thread-safety of the calls. Presently, each parser is thread-safe relative to other parsers, IOW multiple parser instances in different threads won't clobber each other, however, a single parser used by multiple threads is not thread-safe.

In this particular instance, the parser is provided so that if the client were to copy objects from parser_a to another parser_b, it can be done. The parser that owns an object is pointed to by the entity itself, so when copying, this gives the client the ability to copy objects from parser instance to another.

lib/src/ovis_json/ovis_json.h
tom95858: @morrone, thanks I'll update this.

lib/src/ovis_json/ovis_json.h
tom95858: @morrone, yes, object allocation was moved to the parser instance. This was to allow for caching of objects that are `temporary`. So internal to the parser, there is an object cache. Objects, however, can be used and are accessible after the parser has been destroyed. The cache is there to minimize calls to malloc()/free(), especially during parsing ... The tests on the original code placed malloc() up around %15 of the CPU. Now it's small enough that it does not appear in the perf report.

baallan:
@tom95858 if the 360B referenced came from the output of a sizeof call, I will defer to the compiler. It looked to me like the final node structure might have 2 copies of the union containing the default string buffer.

Although, if we are paying for 512 bytes/node anyway for any string size above 250, why not just have the built-in default buffer size computed at configure time to be the most that fits neatly? And allow a configure argument that overrides that? This way if a site has a need for an alternate really big string default, they can just reconfigure without bugging us.

After running another histogram (executable full pathname sizes and environment variable sizes from our lmod/spack infrastructure), I agree 256 is a reasonable default for where i work.

baallan:
> > > If an attribute were added to a dictionary that already has that attribute name, the logic would replace the old attribute with the new one. This was only depended on in one place afaict and that is in json_dict_merge().
> > 
> > 
> > Silent replacement should only occur if the old and new are of the same type. Type-conflicted replacement should return an error unless some optional stomping set function is used.
> 
> I am not sure why that should be for json. I don't think that is how libjansson behaves, for instance. I think it just replaces the attribute regardless of type.

Because data typing *sometimes* matters, I suggest we let the developer make the choice conveniently. A simple extension avoids a lot of repetitive checking for future users:
```
// exists now:
extern int json_attr_add(json_entity_t d, const char *name, json_entity_t v);

// add:
// if name does not exist in scope, adds it.
// if name exists, updates the value stored there if v is of the same type, else returns a type mismatch error code.
extern int json_attr_update(json_entity_t d, const char *name, json_entity_t v);
```
This would be useful for cases where data with specific schema requirements on types is being updated/merged.
json_dict_merge could similarly take a boolean 'strict' flag which would (in effect) mean attr_update instead of attr_add semantics is used where there might be a conflict.

morrone:
> > I am not sure why that should be for json. I don't think that is how libjansson behaves, for instance. I think it just replaces the attribute regardless of type.
> 
> Because data typing _sometimes_ matters, I suggest we let the developer make the choice conveniently. A simple extension avoids a lot of repetitive checking for future users:

Yes, schema often matters. What is not clear to me is why this should be in the json library. Any client using this library should already know what schema (especially if schema is important) they are working with before they ever ask for something to be inserted into a json document.

Adding something like json_attr_update() doesn't help with schema. If the problem you described arises, then the code may very well have already made a mistake _earlier_ when it first inserted the value with the alternate data type. In the json layer library, this isn't doing anything to help enforce schema. This just adds first-inserted-wins semantics to the API.

When schema is important enough to implement reproducible schema checking and enforcement, we have things like json schema, avro, etc.

I guess I'm looking for some reasonable use case where the client _doesn't_ know what the document's schema is when it is adding more data to the document. Because that would seem like the only time when the client would be overwriting fields with other types unknowingly.

morrone:
Interestingly, the json spec _does_ allow repeated keys of the same name. So if we really want to be fully json compliant in our ability to generate json documents, "json_attr_add()" should really just _add_ another entry without regard to whether that key already exists. And the json string generated should contain all duplicates.

Now that doesn't mean that the json _processor_ (the thing that converts json into native language objects) needs to support duplicate object strings. For instance, it is common for processors to only retain the last value it sees for a given key. But technically it could do anything as long as it documents its behavior.

morrone:
Is it possible that json_parser_t is misnamed? That it is really a json _document_ object rather than a handle to a _parser_? Because I see a lot of places in this PR's code where a document is being _generated_ and the "parser" handle is being used. We should really be able to easily tell the difference between a "parser" and a "document" in the code.

morrone:
A higher level question: Have we considered moving to an existing json library? For performance we might select yyjson.

tom95858:
> A higher level question: Have we considered moving to an existing json library? For performance we might select yyjson.

@morrone, the short answer is "yes". yyjson was initially proposed about 2 years ago by @jstile-lbl. At that time, it was still a source project from a PhD student and although blazingly fast, there were worries about it's provenance and longevity.

What changed and motivated our JSON changes is that yyjson is now packaged with Alma9 and other distros. So the initial concerns are mute. 

yyjson parser is *very* fast. Everything is hand-coded, all the way down to the individual character processing routines, the floating point conversion, etc... The consumer side (i.e. attribute lookup, etc...) has been much less scrutinized and evaluated.

So the question was/is what is the benefit and effort to go to yyjson? My guess is that it is probably a month or two including testing. The benefit is less certain. How much JSON parsing do we really do? The configuration/status path uses JSON quite extensively. ldms_xprt_dir() , for example returns results in JSON. But these are not really in the performance path.

The other path is the message bus. If the message type is JSON, JSON parsing is performed and it becomes part of the performance path. This is what initially motivated LBL to look at yyjson.

By the way, libjansson is very slow. So it doesn't really offer any benefit.

So the point of all this is that if ovis_json can be made good enough, then we avoid quite a bit of work converting the existing code. We also retain control (and responsibility) for the json code. There are also interfaces that we support that yyjson does not, e.g. json_dict_build(). This interface is used quite a bit in the status reporting code.

tom95858:
> Is it possible that json_parser_t is misnamed? That it is really a json _document_ object rather than a handle to a _parser_? Because I see a lot of places in this PR's code where a document is being _generated_ and the "parser" handle is being used. We should really be able to easily tell the difference between a "parser" and a "document" in the code.

No, but json_entity_t probably is. It would be better understood (with spec in hand) if it were called json_doc_t.

baallan:
> A higher level question: Have we considered moving to an existing json library? For performance we might select yyjson.

If we implement some version of #2086, then the choice of parser (for those who are performance sensitive, like high rate stores) is up to them and we don't need to worry over-much about it.

lib/src/ovis_json/ovis_json.h
tom95858: @morrone, agreed.

morrone:
> > Is it possible that json_parser_t is misnamed? That it is really a json _document_ object rather than a handle to a _parser_? Because I see a lot of places in this PR's code where a document is being _generated_ and the "parser" handle is being used. We should really be able to easily tell the difference between a "parser" and a "document" in the code.
> 
> No, but json_entity_t probably is. It would be better understood (with spec in hand) if it were called json_doc_t.

At some point you'll explain these comments? I really don't understand why a parser is involved with json _generation_.

lib/src/ovis_json/ovis_json.h
tom95858: @morrone, maybe the parser should be called json_doc_t like you suggest. This might make more sense to users. Then the call sequence would look like this:
```
json_doc_t doc;
int jrc = json_doc_new(input, input_len, &doc);
if (jrc) {
   printf("Error parsing input: %s\n", json_error(doc);
   ...
}
...
json_doc_free(doc);
```
Currently, we have two completely separate API: jbuf_t and json_parser_t. The jbuf_t side does no verification on the creation, i.e. you can create an "object" that is not valid json. It is effectively a fancy dynamic string. The other takes a string and turns it into a verified JSON "entity". If we combine the two, you would get something that can only ever create valid JSON documents. The "parser" goes away, as does the jbuf_t and the client only ever deals with a json_doc_t.

When looking at the API without being buried in the implementation, it seems very disjoint.

I will look at the creation side and see how this might work in practice for both yyjson and ovis_json.

lib/src/ovis_json/ovis_json.h
tom95858: @morrone, I think you are right though. There are two paths right now for creating objects and that is confusing.

tom95858:
@morrone, this latest version adds a document interface to replace what was previously called a "parser."

ldms/src/sampler/slurm/slurm_sampler2.c
```diff
@@ -1140,34 +1140,39 @@ static int slurm_recv_cb(ldms_msg_event_t ev, void *ctxt)
...
 		if (!job) {
 			ovis_log(mylog, OVIS_LERROR, "'%s' event was received "
 					"for job %ld with no job_data.\n",
-					event_name->str, job_id);
+					event_name, job_id);
 			goto unlock_tree;
 		}
 		handle_task_exit(job, ev->recv.json);
-	} else if (0 == strncmp(event_name->str, "exit", 4)) {
+		if (job->exited_tasks_count == job->v[TASK_COUNT].v_u64) {
```
morrone: This change doesn't seem to be related to the json work. Was this supposed to be here?

ldms/src/store/kafka/store_kafka.c
```diff
@@ -187,65 +187,60 @@ static int config(ldmsd_plug_handle_t handle, struct attr_value_list *kwl,
...
 	/* json parse */
-	jp = json_parser_new(buff_len + 1);
-	if (!jp) {
-		rc = errno;
-		LOG_ERROR("json_parser_new() failed, errno: %d\n", errno);
-		goto out;
-	}
-	rc = json_parse_buffer(jp, buff, buff_len+1, &jdoc);
+	rc = json_parse_buffer(buff, buff_len+1, &jdoc);
+	json_entity_t root = json_doc_root(jdoc);
```
morrone: The call to json_doc_root() need to come after the check of rc from json_parse_buffer(). This appears many times in the PR, but I am just going to call it out once.

lib/src/ovis_json/ovis_lexer.c
```diff
@@ -0,0 +1,261 @@
+#include <ovis_json/ovis_json.h>
```
morrone, line 1: File should maybe be named ovis_json_lexer.c?

ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
```diff
@@ -154,7 +154,7 @@ static int sample(ldmsd_plug_handle_t handle)
                 return EINVAL;
         }
 
-        power_obj = json_loads(result_string, JSON_DECODE_ANY, NULL);
+        power_obj = json_parse_buffer(result_string, JSON_DECODE_ANY, NULL);
```
morrone: variorum_sampler is using jansson, so this change should not be here. Probably a stray global search-and-replace? We should check if there are others like this in the PR.

tom95858:
Thanks @morrone...

ldms/src/contrib/sampler/variorum_sampler/variorum_sampler.c
tom95858: @morrone, I only found this one.

ldms/src/store/kafka/store_kafka.c
tom95858: @morrone, you still get a jdoc even if there is an error. The reason is that the jdoc contains the error location, and message. json_doc_root() returns NULL if there is a parsing error.

ldms/src/sampler/slurm/slurm_sampler2.c
tom95858: @morrone, no...good catch. It's part of @narategithub that ended up in my PR.  I'm trying to stage all my changes in separate PR. TX.

lib/src/ovis_json/ovis_json.c
```diff
@@ -1,35 +1,137 @@
...
+}
+
+json_stats_t json_doc_stats(json_doc_t doc)
+{
+	return &doc->state.stats;
+}
+
+char *json_doc_errstr(json_doc_t doc)
+{
+	return doc->err_str;
```
morrone, line 76: Need to test if if doc is NULL before dereferencing, since this is epected to be used when json_parse_buffer() fails. json_parse_buffer() might fail for ENOMEM. Maybe return a pointer to a static "Out of memory" string when that happens.

lib/src/ovis_json/ovis_json.c
```diff
@@ -1,35 +1,137 @@
...
+	LIST_REMOVE(ee, entry);
+	if (!d->root)
+		d->root = &ee->e;
+	return &ee->e;
+}
+
+json_doc_t json_doc_new()
+{
+	json_doc_t doc = calloc(1, sizeof(*doc));
+	doc->cache_bump = JSON_ENTITY_CACHE_BUMP;
```
morrone, line 48: Need to test that doc is non-NULL before dereferencing.

PR #2107, opened 2025-12-01
Fix nested structure definitions in anonymous union

tom95858:
C++ does not allow non-static members in anonymous unions. This change moves the structure definitions outside of the union.

### DIFF
--- a/ldms/src/core/ldms.h
+++ b/ldms/src/core/ldms.h
@@ -692,8 +692,8 @@ int ldms_xprt_set_quota(ldms_t x, int64_t recv_quota, int64_t rate_limit);
  * \param ctxt   Application's context
  * \param fn     Application's function to free the context
  */
-typedef void (*app_ctxt_free_fn)(void *ctxt);
-void ldms_xprt_ctxt_set(ldms_t x, void *ctxt, app_ctxt_free_fn fn);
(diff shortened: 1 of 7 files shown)


tom95858:
@narategithub I think we need to add a simple test for C++ compatibility of the header files. I think the test program looks like this:
```
#include <iostream>
using namespace std;

#include "ldms.h"
#include "ldms_xprt.h"
#include "ldms_msg.h"

int main(int argc, char *argv[])
{
        cout << "Hello, world\n";
}
```
There may be additional header files that we include there, but I think these get the bulk of it.

narategithub:
@tom95858 I'll add the test to the github workflow after this is merged.

ldms/src/core/ldms.h
```diff
@@ -2187,42 +2187,52 @@ typedef enum ldms_xprt_ops_e {
...
+};
+
+struct ldms_send_profile_s {
+	struct timespec app_req_ts;
+	struct timespec send_ts;
+	struct timespec complete_ts;
+	struct timespec deliver_ts;
+};
+
+struct ldms_strm_publish_profile_s {
```
nichamon, line 2221: A minor thing, should we rename _strm_ to _msg_? I believe it's a left over code before the name change. @narategithub Please confirm.

PR #2114, opened 2025-12-15
Yaml host fix

nick-enoent:
Resolves issue where trailing 0 would be removed from daemon names


nichamon:
@nick-enoent The PR also contains the curated metric sampler plugin. Could you please rebase the PR so that it contains only the fix to the YAML parser? :)

PR #2115, opened 2025-12-15
Formatting fix for ldmsd_yaml_parser

nick-enoent:
Resolve issue where a legitimate trailing 0 was removed from daemon names

### DIFF
--- a/ldms/python/ldmsd/ldmsd_yaml_parser
+++ b/ldms/python/ldmsd/ldmsd_yaml_parser
@@ -42,7 +42,7 @@ if __name__ == "__main__":
         sys.exit(errno.EINVAL)
 
     if args.daemon_name:
-        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
+        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name)
         print(f'{ldmsd_cfg_str}')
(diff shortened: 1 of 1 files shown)


nichamon:
Just to make sure I understand the issue correctly: SNL reported that trailing zeroes were being stripped from daemon names (e.g., 'nid10000' becoming 'nid1'), right?

nick-enoent:
> Just to make sure I understand the issue correctly: SNL reported that trailing zeroes were being stripped from daemon names (e.g., 'nid10000' becoming 'nid1'), right?

That was my understanding.

ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -42,7 +42,7 @@ if __name__ == "__main__":
         sys.exit(errno.EINVAL)
 
     if args.daemon_name:
-        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
+        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('\0'))
```
morrone: This is not correct either. I would recommend dropping rstrip() altogether (this is string from the command line, so this isn't needed), or use just "rstrip()" or "strip()". @baallan has a better fix in #2122.

morrone (changes requested):

PR #2122, opened 2025-12-15
Fix several bugs in 4.5.1 yaml parser and add short options.

baallan:
Bugs fixed:
    - Man page options (-y, etc) not supported by implementation. Extended option config.
    - The same daemon can aggregate and run samplers, so the same
      checking data must be used to prevent duplicates auth_add or listen statements
      across write_aggregator and write_sampler. Changed to share a common dictionary
      of already emitted listen/auth_add commands, named auth_listen.
    - Expanded lists must be sorted before bin_search can be used correctly. Otherwise
      daemons are omitted from batch generation. Added list sorting when/where needed.
    - The same list must not be expanded/sorted twice when processing multiple daemon
      outputs, or large systems (30k nodes) can run well beyond 10 minutes to generate.
      Added cache dictionary {hostlist: expanded_hostlist} to prevent reexpansions.
    - Added use of logging api so all debugging related prints/exceptions go to stderr,
      rather than polluting config output to stdout, per Chris suggestion.

    Usability bugs fixed:
    - Expanded lists should not be sorted when processing single daemon output; a single
      linear search is faster than a sort+bin_search. Added sort control option to
      expand_hostlist.
    - Many Exception messages do not give enough context to be helpful. Added context.
    - Several exception messages do not give enough trace context to be helpful when caught and
      reraised. Changed these to extend rather than replace traceback.
    - Defined short option names for everything.
    - Defined mutual exclusion group of -n/-p per Tom suggestion.
    - Added -l/-L options to control log level and log destination (defaults ERROR,stderr)
    - Added -o option so that config script output can be directed to a file.
    - Common debug log activity is to examine the yaml after anchor expansion and to examine
      the full yamlcfg object data before/after the build_* functions are called. Added
      __str__ method to YamlCfg object.
    - Output needs to include an input origin line at the top, much as generated linux
      printer config and other /etc config files do. Added origin to output as a .conf comment.

### DIFF
--- a/ldms/man/ldmsd_yaml_parser.rst
+++ b/ldms/man/ldmsd_yaml_parser.rst
@@ -8,7 +8,7 @@ ldmsd_yaml_parser
 A python program to parse a YAML configuration file into a v4 LDMS configuration.
 ---------------------------------------------------------------------------------
 
-:Date: 20 Nov 2024 "ovis-4.4.5"
+:Date: 20 Nov 2024 "ovis-4.5.2"
 :Manual section: 8
 :Manual group: LDMSD
 
@@ -23,27 +23,37 @@ SYNOPSIS
 DESCRIPTION
 ===========
 
-A single ldmsd can be configured using a YAML configuration file. This
-can be done by running ldmsd directly, or generating a configuration
-parser with the ldmsd_yaml_parser
+A single ldmsd daemon configuration script or a fleet of them can be generated from
+an ldms YAML configuration file. When ldmsd is configured with a -y option, the
+ldmsd daemon uses ldmsd_yaml_parser to translate the yaml input.
 
 LDMSD YAML OPTIONS
 ==================
 
-**-y,**\ *CONFIG_PATH*
-   The path to the YAML configuration file.
+Option -y is required and options -n and -p are mutually exclusive.
 
-**-n,**\ *NAME*
-   The name of the LDMS daemon.
+**-y,--ldms_config**\ *INPUT_CONFIG_PATH*
+   The path to the LDMSD YAML configuration file. This is required.
 
-LDMSD_YAML_PARSER COMMAND SYNTAX
-================================
+**-p,--output_path,--generate_config_path**\ *DIR_PATH*
+   The name of an existing directory to dump fleet of files.
 
-**--ldms-config,**\ *CONFIG_PATH*
-   The path to the YAML configuration file.
+**-n,--daemon_name**\ *DAEMON_NAME*
+   The name of the single LDMS daemon to generate.
 
-**--daemon-name,**\ *NAME*
-   The name of the LDMS daemon to configure.
+**-o,--outfile**\ *FILE*
+   Send the generated configuration commands from '-n' output to FILE. Debugging and exception output
+   goes to stderr unless overridden with -L.
+
+**-l,--log_level**\ *LEVEL*
+   Set the log level (debug,info,warning,error,critical).
+   The debug level will dump anchor-expanded yaml and data derived from the yaml for review.
+
+**-L,--log_file**
+   Set the log output file to use instead of stderr.
+
+**-h,--help**
+   Print help to stdout.
 
 YAML CONFIGURATION SYNTAX
 =========================
--- a/ldms/python/ldmsd/ldmsd_yaml_parser
+++ b/ldms/python/ldmsd/ldmsd_yaml_parser
@@ -3,57 +3,82 @@ import os, sys
 import yaml
 import errno
 import argparse
+import logging
+import pprint
 from ldmsd.parser_util import *
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         description='LDMS Monitoring YAML Cluster Configuration Parser')
     required = parser.add_argument_group('required arguments')
-    required.add_argument('--ldms_config', metavar='FILE', required=True,
+    required.add_argument('-y','--ldms_config', metavar='LDMSD_YAML_FILE', required=True,
                         help='Path to LDMSD YAML configuration file. ')
-    parser.add_argument('--generate_config_path', metavar="STRING", required=False,
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('-p','--output_path', "--generate_config_path", metavar="OUTPUT_DIR_NAME", required=False,
                         help='Path to directory to store generated v4 configuration files for an entire LDMS cluster. '
-                             'Samplers with similar configurations share a single configuration file. ',
+                             'Samplers with identical configurations share a single configuration file. ',
                         default=False)
-    parser.add_argument('--daemon_name', metavar='STRING', required=False,
+    group.add_argument('-n','--daemon_name', metavar='DAEMON_NAME', required=False,
                         default=False,
                         help='Daemon name to generate configuration from YAML file')
-    parser.add_argument('--debug', action='store_true',
-                        help='Enable debug information')
+    parser.add_argument('-o','--outfile', metavar='OUTPUT_FILE_NAME', required=False,
+                        default=False,
+                        help='Output configuration file [optional] for -n case. Default is stdout')
+    parser.add_argument('-L','--log_file', metavar='LOG_FILE_NAME', default=None,
+                        help='log file (default is stderr)')
+    parser.add_argument('-l','--log_level', metavar="LOG_LEVEL", default="ERROR",
+                        help='set logging level to one of DEBUG,INFO,WARNING,ERROR,CRITICAL')
     args = parser.parse_args()
-    if not args.debug:
-        import sys
-        sys.tracebacklimit=0
+    level = args.log_level.upper()
+    if not isinstance(logging.getLevelName(level), int):
+        logging.critical(f"Requested log level is undefined: {level}")
+        sys.exit(errno.EINVAL)
+    if not args.log_file is None:
+        logging.basicConfig(filename=args.log_file, level=level)
+    else:
+        logging.basicConfig(level=level)
+
     config_fp = open(args.ldms_config)
     conf_spec = yaml.safe_load(config_fp)
 
+    logging.debug("ORIGINAL YAML after anchor expansion")
+    logging.debug(pprint.pformat(conf_spec))
+
     if not isinstance(conf_spec, (list, dict)):
-        print(f"Error: input {args.ldms_config} is not yaml. Contents:", file=sys.stderr)
-        print(conf_spec, file=sys.stderr)
+        logging.error(f"Input {args.ldms_config} is not yaml. Contents:")
+        logging.error(conf_spec)
         sys.exit(errno.EINVAL)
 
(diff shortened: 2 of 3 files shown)


ldms/man/ldmsd_yaml_parser.rst
```diff
@@ -23,27 +23,30 @@ SYNOPSIS
...
 
-LDMSD_YAML_PARSER COMMAND SYNTAX
-================================
+**-p,--generate_config_path**\ *PATH*
+   The name of an existing directory to dump fleet of files.
 
-**--ldms-config,**\ *CONFIG_PATH*
-   The path to the YAML configuration file.
+**-d,--debug**
+   Increase the level of debug output to stdout.
```
morrone: Debug messages, warnings, errors, etc. should all go to stderr, not stdout.

ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -9,45 +9,70 @@ if __name__ == "__main__":
...
               f'Specifying the "daemon_name" parameter will generate the configuration for a single ldmsd\n'
               f'Specifying the "generate_config_path" parameter will generate an entire cluster\'s\n'
               f'v4 configuration files in the path provided.\n', file=sys.stderr)
         sys.exit(errno.EINVAL)
 
     if args.daemon_name:
-        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
-        print(f'{ldmsd_cfg_str}')
+        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name).rstrip()
+        if args.debug > 1:
```
morrone: Instead of manually checking the debug level everywhere and calling print directly, we should use the python standard "logging" module. The basicConfig() of logging will correctly send things to stderr by default.

ldms/python/ldmsd/parser_util.py
```diff
@@ -167,12 +170,29 @@ def dist_list(list_, n):
...
+    if not group in __host_expansions:
+        __host_expansions[group] = expand_names(group, sort)
+    result = bin_search( __host_expansions[group], host, sort )
+    return result
+
+## \brief return bool for "is x in slist".
+# Uses log_n search function for lists if
+# (slist is sorted and is_sorted is True) or
+# else linear search is used.
+def bin_search(slist, x, is_sorted=False):
```
morrone, line 186: This whole thing seems rather non-pythonic. What we really want is an order set, right? We want to have a set for fast search, but we also want to maintain order.

If you ignore a dicts' values, the keys of a python dict (at least in python >= 3.7) act as an ordered set. Or, if we want to be compatible with python < 3.7, we can use collections.OrderedDict.

When we use an OrderedDict, we don't need this conditional about whether or not the keys in the dict have been sorted. The fast lookup works for either.

ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -9,45 +9,70 @@ if __name__ == "__main__":
...
     args = parser.parse_args()
     if not args.debug:
         import sys
         sys.tracebacklimit=0
     config_fp = open(args.ldms_config)
     conf_spec = yaml.safe_load(config_fp)
 
+    if args.debug > 1:
+        import pprint
+        pprint.pprint("ORIGINAL YAML after anchor expansion")
```
morrone: We don't need to use pprint for a plain string.

ldms/man/ldmsd_yaml_parser.rst
```diff
@@ -23,27 +23,30 @@ SYNOPSIS
...
+   The path to the YAML configuration file. This is required.
 
-**-n,**\ *NAME*
-   The name of the LDMS daemon.
+**-n,--daemon_name**\ *NAME*
+   The name of the single LDMS daemon to generate.
 
-LDMSD_YAML_PARSER COMMAND SYNTAX
-================================
+**-p,--generate_config_path**\ *PATH*
```
morrone: How about "output_path".

ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -9,45 +9,70 @@ if __name__ == "__main__":
...
     if not isinstance(conf_spec, (list, dict)):
         print(f"Error: input {args.ldms_config} is not yaml. Contents:", file=sys.stderr)
         print(conf_spec, file=sys.stderr)
         sys.exit(errno.EINVAL)
 
     cluster = YamlCfg(None, None, conf_spec, args)
-
     if args.daemon_name and args.generate_config_path:
-        print(f'Parameters "daemon_name" and "generate_config_path" are mutually exclusive.\n'
+        print(f'Parameters "daemon_name"(-n) and "generate_config_path"(-p) are mutually exclusive.\n'
```
morrone: Why? Can we not put a single file in a directory as easily as many files?

ldms/man/ldmsd_yaml_parser.rst
```diff
@@ -23,27 +23,30 @@ SYNOPSIS
...
-**--ldms-config,**\ *CONFIG_PATH*
-   The path to the YAML configuration file.
+**-d,--debug**
+   Increase the level of debug output to stdout.
+   When run with neither -p nor -n, no configuration script is produced, but
+   an extra -d will dump anchor-expanded yaml and data derived from the yaml for review.
 
-**--daemon-name,**\ *NAME*
-   The name of the LDMS daemon to configure.
+**-o,--outfile**\ *FILE*
```
morrone, line 44: This is the option that really needs guidance about how it interacts with -p.

ldms/man/ldmsd_yaml_parser.rst
baallan: if @nick-enoent is ok with adding logger use, i can update all the debug print related stuff to logger.

ldms/python/ldmsd/parser_util.py
baallan: we have a base requirement of python 3.6 still in effect.
I can profile for the alternative you propose if indeed the class exists in toss python 3.6. that would simplify things if successful.

ldms/man/ldmsd_yaml_parser.rst
baallan: that would be a breaking change for existing scripts. i agree with the sentiment.

ldms/python/ldmsd/ldmsd_yaml_parser
baallan: that would be a question for @nick-enoent. More generally, i'd like to be able (for the single daemon case) to specify an exact output file name rather than having to assume a filename based on a daemon instance and a directory.

ldms/man/ldmsd_yaml_parser.rst
baallan: it only interacts with -n, of course. correct that this should be documented.

ldms/python/ldmsd/parser_util.py
morrone: Maybe it isn't worth it to rework the hostlist library. Ultimately what we really want is a good class representation of a hostlist. clustershell has an implementation that looks good. But maybe its too much work to pull in all the necessary code. And maybe we don't want to have an external dependency on clustershell.

baallan: > Maybe it isn't worth it to rework the hostlist library. Ultimately what we really want is a good class representation of a hostlist. clustershell has an implementation that looks good. But maybe its too much work to pull in all the necessary code. And maybe we don't want to have an external dependency on clustershell.

llnl has a python hostlist implementation in github that i've been thinking might be wise to switch to.  In the context we're using it, though, we don't actually *need* hostlist expansion at all. We just need an *in_hostlist(name, unexpanded_hostlist)* which segments name and the entries of unexpanded_hostlist and checks for a match. This is much faster and much smaller in memory than any of the options discussed so far. For performance, we would still cache the segmentation of unexpanded_hostlist.

ldms/man/ldmsd_yaml_parser.rst
tom95858: @baallan I doubt anyone except you is using this yet. I suggest that you just change it to be what we want.

ldms/python/ldmsd/ldmsd_yaml_parser
tom95858: @baallan, the ArgumentParser class will enforce this for you if you ask it to.

baallan: ah, add_mutually_exclusive_group. will try that. looks like it goes all the way back to py3.6.

ldms/python/ldmsd/parser_util.py
baallan: It looks like https://github.com/cea-hpc/clustershell provides [NodeSet,RangeSet,NodeUtils].py that with future work could be pulled in under their LGPL.  It expands the hostlist range syntax even further, so we'd have to document and think about it.

ldms/man/ldmsd_yaml_parser.rst
baallan: argparse can support multiple long names, so adding output_path alternate.

ldms/python/ldmsd/ldmsd_yaml_parser
baallan: add_mutually_exclusive_group works as advertised; will apply it.

ldms/python/ldmsd/ldmsd_yaml_parser
baallan: changed to logging.

ldms/python/ldmsd/ldmsd_yaml_parser
baallan: switching to logging api.

ldms/man/ldmsd_yaml_parser.rst
baallan: switching to logging api.

baallan:
This addresses #2127 and the test case is provided in #2126.
2126 and 2122 can be merged in either order (non-overlapping, and 2126 will run with either current main or 2122-added main)

baallan:
@jennfshr @tom95858 the yaml patch and test case PR that demonstrates its necessity are both ready.
The switch to the python logging api and improved argparse usage as recommended by @morrone / @tom95858 have been implemented.

jennfshr:
@tom95858 how are we feeling about this PR? We really need a working solution for supporting the up-coming systems upgrades, even if we need to package it in house.

baallan:
@nick-enoent asked offline where 'hostlist' used in the test script of the *other* PR comes from. For the record here, it comes from llnl folk and llnl-packaged rpm lua-hostlist:             http://code.google.com/p/lua-hostlist

baallan:
#2122 has been superceded by #2131, #2132, #2133, #2134, #2135, #2136, #2137, #2138, #2139, #2140.

#2139, #2140 expand the log output, requiring the prior log/message patching #2133, #2134, #2135, #2136.

baallan:
superceded as noted above

PR #2123, opened 2025-12-16
Formatting fix for ldmsd_yaml_parser

nick-enoent:
Resolve issue where legitimate trailing 0 is removed from daemon names

### DIFF
--- a/ldms/python/ldmsd/ldmsd_yaml_parser
+++ b/ldms/python/ldmsd/ldmsd_yaml_parser
@@ -42,7 +42,7 @@ if __name__ == "__main__":
         sys.exit(errno.EINVAL)
 
     if args.daemon_name:
-        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
+        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name)
         print(f'{ldmsd_cfg_str}')
         sys.exit(0)
 


ldms/python/ldmsd/ldmsd_yaml_parser
```diff
@@ -42,7 +42,7 @@ if __name__ == "__main__":
         sys.exit(errno.EINVAL)
 
     if args.daemon_name:
-        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('0'))
+        ldmsd_cfg_str = cluster.daemon_config(args.ldms_config, args.daemon_name.rstrip('\0'))
```
morrone: I do not think that is right either. I don't think we need _any_ strip here, but @baallan fixes this by changing it to just "rstrip()" in #2122.

nick-enoent: The ldmsd_yaml_parser can be passed C strings. You can get unexpected values by solely calling rstrip().

morrone: No, not really.

args.daemon_name is created by ArgumentParser() parsing command line options. There is no direct C string visibility here at all.

If a \0 is appearing in the strings sent from ldmsd to ldmsd_yaml_parser, I would imagine that there is an off-by-one bug or something similar in ldmsd. We need to fix that problem, not paper over it here.

tom95858: @morrone the string is presented as a Python3 string which I believe is encoded in WCHAR. So what this will break is hostnames that include Chinese or Romanji characters. We will take a look at this.

morrone: OK, that would point to needing to get the type conversion set correctly in the add_argument() method of ArgumentParser. I believe that python3 handles utf8 by default. To handle some other encoding might require explicit lookup of what encoding is used in the locale, and then set the correct type conversion accordingly.

PR #2126, opened 2026-01-08
Add test of ldmsd_yaml_parser derived from production cluster eclipse.

baallan:
This tests that:
- all expected files are created in group mode
- all expected files are created in per-daemon mode

It also provides a tree of output files that can be compared against previous versions when ldmsd_yaml_parser is changed.

### DIFF
--- a/ldms/test/yaml-cluster/test_gen_block_generic.sh
+++ b/ldms/test/yaml-cluster/test_gen_block_generic.sh
@@ -1,0 +1,120 @@
+#! /bin/bash
+# before running, path and pythonpath must include
+# PYTHONPATH=$prefix/lib/$pyver/site-packages
+# PATH=$prefix/bin:$prefix/sbin:$PATH
+
+# partial test for at least python3:
+if test "x$(which python3)" = "x/usr/bin/python3"; then
+       	:
+else
+       	echo "requires /usr/bin/python3 in path first"
+	exit 1
+fi
+
+# set logging and outing to /bin/true to use new options
+LOGGING=/bin/false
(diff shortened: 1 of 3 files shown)


ldms/test/yaml-cluster/test_gen_block_generic.sh
```diff
@@ -0,0 +1,120 @@
...
+if ! test "x$dcount" = "x$EXPECTED"; then
+	echo "daemon files are missing in $gdir ; --generate_config_path test FAIL"
+	exit 1
+else
+	echo "found expected $EXPECTED files in $gdir."
+fi
+
+echo "generating single files in $sdir, logs in $logsdir"
+date
+for i in $(hostlist -d" " -e $some); do
```
nick-enoent, line 65: @baallan What package did you install to get this hostlist command? We have a hostlist.py in our tree, but it does not run as a command.

baallan:
On the snl clusters, the package lua-hostlist-0.5-1.t4.x86_64 provides /usr/bin/hostlist.
  http://code.google.com/p/lua-hostlist
Is the original source.

baallan:
> On the snl clusters, the package lua-hostlist-0.5-1.t4.x86_64 provides /usr/bin/hostlist. http://code.google.com/p/lua-hostlist Is the original source.

Looks probable that the current lua source is at https://github.com/grondo/lua-hostlist instead, since google's is in archive mode. @morrone may have more insight. The llnl spec file still references the old repo.

morrone:
Yes, it seems that https://github.com/grondo/lua-hostlist repo is the current home.

Issue #2127, opened 2026-01-08
ldmsd_yaml_parser bugs


baallan:
At commit dabfe (top of main, at time of this posting):

- ldmsd_yaml_parser creates incorrect conf output for daemons when using the attached block_generic.yaml as input:
  - cn2 and cngw2 (missing lines; root cause: bin_search use coded wrong)
  - cnadmin2-agg (repeated lines; root cause: redundancy checking data must be shared between sampler and agg generation code)

- ldmsd_yaml_parser does not implement all options listed in man page.
- exception/print/error output is mixed with stdout.
- most error messages don't give useful context that will guide an admin to the bad input lines.
- for large node counts (e.g. 30000), redundant list expansions make it slow to generate files for an entire cluster.
 
[block_generic.yaml](https://github.com/user-attachments/files/24485115/block_generic.yaml)

#2122 addresses all the above issues of correctness, usability, and performance.
#2126 provides a test yaml-cluster that demonstrates the current problems and that test can be extended to verify more conditions and identify regressions. The test script can be modified to use old or new command line and logging options.

baallan:
This issue (except for performance bullet 5, which has been handed off to @nick-enoent) is now addressed by superceding PRs instead of 2122.

https://github.com/ovis-hpc/ldms/pull/2122 has been superceded by https://github.com/ovis-hpc/ldms/pull/2131, https://github.com/ovis-hpc/ldms/pull/2132, https://github.com/ovis-hpc/ldms/pull/2133, https://github.com/ovis-hpc/ldms/pull/2134, https://github.com/ovis-hpc/ldms/pull/2135, https://github.com/ovis-hpc/ldms/pull/2136, https://github.com/ovis-hpc/ldms/pull/2137, https://github.com/ovis-hpc/ldms/pull/2138, https://github.com/ovis-hpc/ldms/pull/2139, https://github.com/ovis-hpc/ldms/pull/2140.

https://github.com/ovis-hpc/ldms/pull/2139, https://github.com/ovis-hpc/ldms/pull/2140 expand the log output, requiring the prior log/message patching https://github.com/ovis-hpc/ldms/pull/2133, https://github.com/ovis-hpc/ldms/pull/2134, https://github.com/ovis-hpc/ldms/pull/2135, https://github.com/ovis-hpc/ldms/pull/2136.

PR #2128, opened 2026-01-15
Sort list of expanded hostnames

nick-enoent:


baallan:
This resolves some of the correctness, but not the performance issue listed in #2127.
Why are we wasting time reimplementing fragments of #2126?

PR #2129, opened 2026-01-16
Sort list of expanded hostnames in yaml paser

nick-enoent:


baallan:
This doesn't prevent duplicate expansions as the in_hosts proposed in #2122 does. Also, as much more easily seen by examining code around in_hosts calls of 2122, there is no subsequent changing of the result returned from in_hosts, so it's perfectly fine to use the expansion caching strategy of 2122.

PR #2136, opened 2026-01-19
add context clues for admins to exception/error from parser_util.py

baallan:
In some cases, reraise was added, and a function making YamlCfg objects printable/json.dumps-able is added.

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -482,7 +482,7 @@ class YamlCfg(object):
             plugins = check_opt('plugins', agg_spec)
             if plugins:
                 if plugins is not list:
-                    raise ValueError(f'"plugins" must be a list of plugin instance names"\n')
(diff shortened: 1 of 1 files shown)


ldms/python/ldmsd/parser_util.py
```diff
@@ -491,7 +490,7 @@ def build_aggregators(self, config):
             subscribe = check_opt('subscribe', agg_spec)
             if subscribe:
                 for stream in subscribe:
-                    check_required([ 'stream', 'regex' ], stream, "stream specification")
+                    check_required([ 'stream', 'regex' ], stream, f'"aggregators" stream specification')
```
tom95858: This doesn't need to be an 'f' string,  there are no format arguments.

baallan: fixed

baallan: though the python compiler would treat both as literals when creating bytecode.

PR #2141, opened 2026-01-23
Use plugin-specific log handle for all messages in store_sos

nichamon:
Replace global log handle with per-plugin instance log handle so log messages are properly labeled as 'store_sos' instead of 'ldmsd'.

### DIFF
--- a/ldms/src/store/store_sos.c
+++ b/ldms/src/store/store_sos.c
@@ -70,9 +70,8 @@
 #include "ldmsd.h"
 #include "ldmsd_plug_api.h"
 
-static ovis_log_t mylog;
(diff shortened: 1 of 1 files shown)


ldms/src/store/store_sos.c
```diff
@@ -70,9 +70,9 @@
 #include "ldmsd.h"
 #include "ldmsd_plug_api.h"
 
-static ovis_log_t mylog;
-#define LOG_(level, ...) do { \
-	ovis_log(mylog, level, ## __VA_ARGS__); \
+// static ovis_log_t mylog;
```
tom95858: Let's remove this please.

PR #2142, opened 2026-01-23
Add additional error messages for decomp_static

nichamon:
Add critical and error log messages at failure points to improve debugging when decomposition encounters memory allocation failures or operation errors in decomp_static.


ldms/src/decomp/static/decomp_static.c
```diff
@@ -1927,6 +1927,10 @@ static int decomp_static_decompose(ldmsd_strgp_t strgp, ldms_set_t set,
 			if (!mid_rbn) {
 				rc = ENOMEM;
 				free(mid_rbn);
```
tom95858, line 1929: Why are you freeing a pointer you just confirmed to be NULL?

PR #2145, opened 2026-01-28
Fix double ntohl() in processing a push update

nichamon:
The push_flags value is already in host byte order after the first ntohl() conversion. Remove the redundant second ntohl() call that was corrupting the flag values passed to the callback.

This resulted in the producer set's reference count logic failing to detect push updates, leading to incorrect reference management.


ldms/src/core/ldms_xprt.c
narategithub, line 2273: Note: `push_flags` is in host format from this line.

Issue #2147, opened 2026-02-03
`metric_sets_default_authz` - need documentation for yaml based config method


jennfshr:
We couldn't find a way to include `metric_sets_default_authz` in the yaml spec for a daemon. yaml gap or user education problem?

@nick-enoent can you please ensure this is included in the manpage and guide us?

nick-enoent:
This is implemented, but the man page may need to be updated. Within the yaml config, metric_sets_default_authz is treated as a dictionary as illustrated below. Configuration options like "set_memory" are also included in the daemon configuration. Examples for both are provided below

```
daemons :
  - names : "L1-agg-[1-2]"
    metric_sets_default_authz:
      uid : 0
      gid : 0
      perm : "777"
    set_memory : "100mb"
    hosts : "node-1"
    endpoints :
      - names : "L1-agg-[1-2]"
        ports : "[10401-10402]"
        xprt  : sock
        auth  :
          name   : munge1
          plugin : munge
```

baallan:
thanks for the solution.

baallan:
I backported #2148 for b4.5, so closing this issue.

PR #2148, opened 2026-02-03
Update YAML config CLI options documentation for additional clarity.

nick-enoent:
Addresses #2147


baallan:
@nick-enoent please backport this to b4.5 also
and verify that when generating both man and html the output indent is correct.

baallan:
@nick-enoent this pr needs to be cherry-picked or otherwise backported to b4.5.

PR #2150, opened 2026-02-06
fix locations checked in ldms-sensors-config script for cpu frequency

baallan:
linux kernels (at an unknown version earlier than 4.18) switched from cpuinfo_cur_freq to scaling_cur_freq file name.
The modification checks both locations instead of trying to detect kernel version and assume the answer based on version.

The build-time test_lscpu gold input is no longer valid for some systems,
so it is no longer part of the test rule.


baallan:
fixes #2151

Issue #2149, opened 2026-02-06
lustre_mdc sampler error log


baallan:
I get
ERROR: ldmsd: The specified sampler configuration 'lustre_mdc' does not exist.
though the sampler still seems to run correctly.
It appears likely to be tied to SAMP being used in set register instead of some schema or instance related name.

morrone:
That error message means that ldmsd_set_register() failed, so things won't be completely working for the sets.

I cannot reproduce this. Is there perhaps missing information? Are you working with a development version of the plugin that is _not_ named "lustre_mdc"?

baallan:
the is on the b4.5 branch 
```
load name=lustre_mdc_60s plugin=lustre_mdc
config name=lustre_mdc_60s component_id=${COMPONENT_ID} producer=${HOSTNAME} perm=0644 job_set=${HOSTNAME}/jobid schema=lustre_mdc instance=${HOSTNAME}/lustre_mdc
start name=lustre_mdc_60s interval=60000000
```

baallan:
I'm seeing the same style of error with the ldms_jobid plugin; the commonality I think is neither plugin uses all the sampler_base machinery in the typical multiinstance way.

morrone:
> load name=lustre_mdc_60s plugin=lustre_mdc

OK, I think that is your problem. lustre_mdc is not multi-instance capable, so it cannot be renamed. I thought we had code to test for this at configuration time, but apparently that is not working...

morrone:
Ah, I made the change to allow renaming non-multi-instance plugins.

I guess the issue here is they are related but orthogonal concepts.

I can work on making the checks better, but the answer will still be, "don't do that with this plugin".

morrone:
See #2153.

baallan:
If there's an alternate mdc plugin library/name that should work, let me know.

We were trying to achieve the following yaml where on a given node we only use one instance at a time, but the admin can select  a rate by changing the plugin instance in a plugin list. Obviously there's other syntax ways to control the interval besides defining alternate instances. These will require more internal discussion as we settle on a yaml template for our production clusters.

```
defaults:
[...]

PR #2153, opened 2026-02-06
Fix typo in load_plugin()

morrone:
Fix a typo in load_plugin()

### DIFF
--- a/ldms/src/ldmsd/ldmsd_config.c
+++ b/ldms/src/ldmsd/ldmsd_config.c
@@ -171,7 +171,7 @@ struct ldmsd_plugin_generic * load_plugin(const char *plugin_name,
 	 * - If flags & MULTI_INSTANCE is !0, then
 	 *   -- both constructor() and destructor() must be defined
 	 */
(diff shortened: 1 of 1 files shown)


tom95858:
@morrone, this changes the current behavior. Currently, single-instance does not mean it can't be loaded with a different configuration name. Some configurations load single instance plugins with a different configuration name. These can be for convenience, for example, to provide short configuration names for long plugin names.

morrone:
> @morrone, this changes the current behavior. Currently, single-instance does not mean it can't be loaded with a different configuration name. Some configurations load single instance plugins with a different configuration name. These can be for convenience, for example, to provide short configuration names for long plugin names.

This is _not_ a blanket change to that behavior. This tags _specific_ single-instance plugins that do not yet support renaming. Folks currently have problems like in #2149 with these plugins.

We can work on fixing those samplers one by one. But until they are fixed, I think it is probably better to just give a clear error message up front rather than throw lots of confusing error messages and have them behave problematically.

morrone:
Since it seems controversial, I reduced this PR to just the typo, which we definitely need. The non-renamable patch is in another PR.

PR #2158, opened 2026-02-08

### LINKED ISSUE #2157: bug ldmsd_yaml_parser: input validation incorrect for interval, offset
parser_util.py check_intrvl_str() incorrectly rejects values containing environment substitutions.
example:
interval: ${INTERVAL}
yields in part:
${INTERVAL} is not a valid time-interval string

this prevents admins from configuring plugin intervals and offsets with an environment variable determined, for example, by a resource manager.

baallan:
fixes #2157

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -86,6 +86,8 @@ def check_intrvl_str(interval_s):
         return interval_s
     if type(interval_s) != str:
         raise ValueError(f'{error_str}')
+    if "${" in interval_s:
+        return interval_s
     interval_s = interval_s.lower()
     unit = next((unit for unit in unit_strs if unit in interval_s), None)
     if unit:


ldms/python/ldmsd/parser_util.py
```diff
@@ -86,6 +86,8 @@ def check_intrvl_str(interval_s):
         return interval_s
     if type(interval_s) != str:
         raise ValueError(f'{error_str}')
+    if "${" in interval_s:
```
tom95858, line 89: @baallan, this should really check for the terminating "}" and the absence of whitespace in the interval_s portion. Also, if the substituted value contains unit strings, they won't get translated. So I'm not sure this is baked yet. Unfortunately, I already merged in into b4.5. If you could fix it to remedy the above, I will merge the fix to this change into b4.5.

nick-enoent:
This is resolved with commit a82b925

PR #2159, opened 2026-02-08

### LINKED ISSUE #2157: bug ldmsd_yaml_parser: input validation incorrect for interval, offset

baallan:
fixes #2157 backported to b4.5

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -84,6 +84,8 @@ def check_intrvl_str(interval_s):
         return interval_s
     if type(interval_s) != str:
         raise ValueError(f'{error_str}')
+    if "${" in interval_s:
+        return interval_s
     interval_s = interval_s.lower()
     unit = next((unit for unit in unit_strs if unit in interval_s), None)
     if unit:


ldms/python/ldmsd/parser_util.py
jennfshr, line 1: Is it always the case (convention) that env vars are curly braced?

baallan: it is. env substitution for our config scripting is implemented in ldms/lib/src/ovis_util/util.c. yaml does not interpret ${}. The same syntax is used for HOSTNAME and COMPONENT_ID.

jennfshr:
@baallan @tom95858 I'm unable to find this fix in the b4.5 branch, LNO 87 in `ldms/python/ldmsd/parser_util.py` - after getting it built on a new system, this bug reemerged and I know I didn't hit it last week for the prior packages we've rolled out to production.  I see where it says its merged, but I cannot find in blame or in the source where this got it.  Thoughts?

baallan:
> @baallan @tom95858 I'm unable to find this fix in the b4.5 branch, LNO 87 in `ldms/python/ldmsd/parser_util.py` - after getting it built on a new system, this bug reemerged and I know I didn't hit it last week for the prior packages we've rolled out to production. I see where it says its merged, but I cannot find in blame or in the source where this got it. Thoughts?

looks like on both main and b4.5 #2156/2159 and equivalent disappeared. did someone force-push something?

baallan:
> looks like on both main and b4.5 #2156/2159 and equivalent disappeared. did someone force-push something?

Ok, i misspoke. On main, a bunch of stuff is still pending. On b4.5 2159 claims it was merged but is missing as Jen says. I opened #2169 which is a duplicate of 2159, since there's no 'reopen merge request' button.

PR #2164, opened 2026-02-10
perfevent2 for uncore counters

narategithub:
- adds `--enable-cap-perfmon` option to `configure.ac`. When enabled, `ldmsd` will be installed with `cap_perfmon` so that `ldmsd` can use perfevent in user mode.

- adds `perfevent2` sampler plugin aiming for *uncore* counters (core counters are also supported). The new sampler supports specifying events using their names (instead of perf attribute codes in perfevent sampler), the same way as they appeared in `perf list hw cache pmu` command. See [ldms-sampler_perfevent2.rst](https://github.com/ovis-hpc/ldms/compare/main...narategithub:perf?expand=1#diff-a09692a74847a906dadcd3e3894c55845e1a3355d1e98c4c8c7545de1fac77bd) (part of the PR) for more info.


narategithub:
oops. Seems like `config3` attribute is too new. I'll address backward compatibility & push again ...

ldms/src/ldmsd/Makefile.am
```diff
@@ -83,10 +83,17 @@ test_plugattr_SOURCES = ldmsd_plugattr.c ldmsd_plugattr.h
...
+DO_CAP_PERFMON = true
+else
+DO_CAP_PERFMON = false
+endif
+
 # make sym links for aggd scripting support
 install-exec-hook:
 	cd $(DESTDIR)$(sbindir) && $(LN_S) -f ldmsd ldms-aggd
 	cd $(DESTDIR)$(sbindir) && $(LN_S) -f ldmsctl ldmsaggctl
+	if $(DO_CAP_PERFMON) ; then sudo setcap 'cap_perfmon=ep' $(DESTDIR)$(sbindir)/ldmsd ; fi
```
morrone: I do not think we should be hard-coding sudo in install-exec-hook.

One thing you could do is introduce a variable for the setcap command, in the same way that ln is called through $(LN_S) a line above. Then you could have just:

```
$(SETCAP) 'cap_perfmon=ep' $(DESTDIR)$(sbindir)/ldmsd
```

Then, if a user can decide how they will get the elevate privileges needed to run setcap. For instance, they can do:

```
./configure SETCAP="sudo setcap"
```

narategithub: Are you thinking of the case that `sudo` is not installed?

morrone: No, more that it is too specific to a single workflow. We don't want to pause the build for a sudo password in most cases. "make install" can be called many ways:

1) run as root to install in global path - sudo should not be called
2) as part of a packaging build, installing in a temporary directory - sudo should not be called
3) user installs in no global path, but wants capabilities set on ldmsd:
  They should either do:
  a) just set teh capability after make install
  b) run "sudo make install"
  c) go with your option and do "configure SETCAP='sudo setcap'" - this way at least they explicitly accept that the install is likely to pause somewhere in the middle and ask for a password. They know they can't just script this away.

I'm not keen on burying the sudo. Thats something that needs to be explicit to the user.

morrone: Although the case where sudo is not installed, or the user doesn't have access to use sudo, are also problems for this approach.

"make install" throwing an error "setcap: permission denied" is much clearer to the make install process prompting for a password partway through.

baallan: I agree sudo should not be in the build.
It *might* belong in a post-install rpm spec file scriptlet.

I would in either case never use it. We control user access to privileged levels of perf via resource manager policies, not filesystem bits.

narategithub:
@tom95858, @jstile-lbl `ldms-perfdb-gen` is not working right with Kernel 4.x and 5.x. I'm fixing it and marking this to "draft" for now to prevent an accident merge.

tom95858:
@narategithub should this be converted from draft to ready to merge?

narategithub:
This has been tested on:
- AMD Zen 4 CPU, Debian 13, Linux 6.12.63
- AMD Zen 4 CPU, openSUSE Leap 15.5, Linux 5.14.21
- AMD Zen 4 CPU, Alma Linux 8, Linux 4.18.0-553.105.1.el8_10.x86_64
- ARM Cortex-A72 (Raspberry Pi 4), Raspberry PI/Debian 13, Linux 6.12.62+rpt-rpi-v8

narategithub:
@tom95858 I think this is ready to be merged.

baallan:
@narategithub has this been tested on intel hardware? blake would be an excellent location, as sapphire rapids is the bulk of our production cores.

PR #2169, opened 2026-02-12

baallan:
recreating #2159 that didn't stick.

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -84,7 +84,7 @@ def check_intrvl_str(interval_s):
         return interval_s
     if type(interval_s) != str:
         raise ValueError(f'{error_str}')
-    if "${" in interval_s:
+    if "$" in interval_s:
         return interval_s
     interval_s = interval_s.lower()
     unit = next((unit for unit in unit_strs if unit in interval_s), None)


tom95858:
1. It doesn't check for the trailing '}' that terminates the environment variable.
2. It doesn't check for white space in the name between '{' and '}', so the environment variable name emitted could be "this and that".
3. It duplicates the way that the unit string is handled introducing the possibility that the YAML unit string parsing and the ldmsd unit string parsing could be different

Should we allow an environment variable to be used to set the sample interval? It becomes impossible reading the YAML file to know what the sampler configuration is.

baallan:
@tom95858, regarding your comments, the aim is to enable admins who need to to configure interval or offset with a variable (which we have a practical case for in production) to do so.
The current yaml-level check for interval/offset times assumed times would always be literal, which has proven wrong in practice.

- The minimum condition for "did the admin try to use a variable instead of a literal" is actually "does a \$ sign appear?", so I can reduce the check to just \$. If the extra brace bothers, I can remove it for you. Once a $ is present, the yaml parser definitively knows it's not a literal time-delta that can be immediately evaluated.
- Whether or not the admin has made any kind error with braces is simply not the business of a function validating a literal delta-time value -- the variable name will flow on through to ldmsd and be handled there.
- The change proposed, with or without the open brace, has nothing to do with whether or not yaml parsing and C level parsing have a redundancy related to unit conversion.

> 1. It doesn't check for the trailing '}' that terminates the environment variable.
> 
>     2. It doesn't check for white space in the name between '{' and '}', so the environment variable name emitted could be "this and that".
> 
>     3. It duplicates the way that the unit string is handled introducing the possibility that the YAML unit string parsing and the ldmsd unit string parsing could be different
> 
> 
> Should we allow an environment variable to be used to set the sample interval? It becomes impossible reading the YAML file to know what the sampler configuration is.

baallan:
I updated to check for only $, since bracing issues are checked by ldmsd.

nick-enoent:
I believe this is resolved by commit 7b5a376

PR #2173, opened 2026-02-12
DCGM Sampler: sampler interval config is ignored by yaml_parser

jennfshr:
This fix omits ignoring on "interval" key in `parser_util.py` logic.

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -34,8 +34,6 @@ DEFAULT_ATTR_VAL = {
 }
 
 INT_ATTRS = [
-    'interval',
-    'offset',
     'reconnect',
     'flush'
 ]


nick-enoent (approved):

jennfshr:
@nick-enoent we were still failing, realized that INT_ATTR interval was also preventing the interval config from getting into the DCGM configuration.  Removing 'interval' and 'offset' since they're no longer interfering with the settings outside of the config block.

nick-enoent:
> @nick-enoent we were still failing, realized that INT_ATTR interval was also preventing the interval config from getting into the DCGM configuration.  Removing 'interval' and 'offset' since they're no longer interfering with the settings outside of the config block. 

I think this is the correct approach, as we now parse intervals strings directly in ldmsd. We should remove the supporting code in the yaml parser as well so we don't end up with dead code. Thanks for your time looking into this.

baallan:
I verified that this change does not affect store plugins, which share the parse_to_cfg_str function with write_samplers. In both cases, the yaml was at some point revised to put the config[] options  in a separate dictionary from the parent plugin block.

nick-enoent:
@jennfshr, thanks again for taking the time to look through this and create this fix.
I would even go a step further and say that we shouldn't be parsing anything out of the plugin config string, i.e. the "name" and "reconnect" checks.
This is likely a remnant from an older version of the yaml configuration file that had all config options in a single dictionary.

jennfshr:
I just want to get it into the branch… can you make the adjustments you feel would remove all parsing from my PR that make sense?  Or give me explicit instructions on what I should modify?

Thanks,

Jen

From: Nick Tucker ***@***.***>
Date: Monday, February 16, 2026 at 11:34 AM
Cc: Green, Jennifer Kathleen ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [ovis-hpc/ldms] DCGM Sampler: sampler interval config is ignored by yaml_parser (PR #2173)

[https://avatars.githubusercontent.com/u/58707916?s=20&v=4]nick-enoent left a comment (ovis-hpc/ldms#2173)<https://github.com/ovis-hpc/ldms/pull/2173#issuecomment-3909967815>

@jennfshr<https://github.com/jennfshr>, thanks again for taking the time to look through this and create this fix.
I would even go a step further and say that we shouldn't be parsing anything out of the plugin config string, i.e. the "name" check.
This is likely a remnant from an older version of the yaml configuration file that had all config options in a single dictionary.

ldms/python/ldmsd/parser_util.py
```diff
@@ -1017,7 +1015,7 @@ def write_sampler(self, dstr, sname, auth_listen):
                             hostname = socket.gethostname()
                             cfg_args = {}
                             for attr in cfg_:
-                                if attr == 'name' or attr == 'interval' or attr == 'reconnect':
+                                if attr == 'name' or attr == 'reconnect':
```
nick-enoent, line 1018: The check I was referring to is just this line, and the continue on the following line. I made a PR for your branch that removes it, but I don't believe I can directly edit this PR.

tom95858:
This is resolved by #2174

PR #2177, opened 2026-02-19
Ensure thread-safe sampler operations with state validation

nichamon:
Introduce states to sampler config objects and acquire the config object's mutex in the configuration path to prevent race conditions. Prior to this change, ldmsd did not acquire locks during configuration, allowing potential concurrent access where a sampler could be configured while actively sampling or executing other operations.

The state machine enforces correctness across all sampler plugin interfaces (config, sample, term, etc.) by ensuring operations are only called when the sampler is in an appropriate state. The usage() interface is excluded from mutex protection as it returns a constant description string.

### DIFF
--- a/ldms/src/ldmsd/ldmsd.c
+++ b/ldms/src/ldmsd/ldmsd.c
@@ -1026,12 +1026,11 @@ int ldmsd_sampler_start(char *cfg_name, char *interval, char *offset,
 	if (!samp)
 		return ENOENT;
 
-	if (samp->os) {
+	ldmsd_sampler_lock(samp);
+	if (samp->state != LDMSD_SAMP_STATE_CONFIGURED) {
 		rc = EBUSY;
 		goto out;
 	}
-
-	ldmsd_sampler_lock(samp);
 
 	if (exclusive_thread) {
 		samp->use_xthread = atoi(exclusive_thread);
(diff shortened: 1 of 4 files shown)


ldms/src/ldmsd/ldmsd.c
```diff
@@ -1026,13 +1026,12 @@ int ldmsd_sampler_start(char *cfg_name, char *interval, char *offset,
 	if (!samp)
 		return ENOENT;
 
-	if (samp->os) {
+	ldmsd_sampler_lock(samp);
+	if (samp->state != LDMSD_SAMP_STATE_CONFIGURED) {
```
tom95858, line 1030: I'm not sure this is correct. I think samp->os meant it was started? But maybe I'm wrong. I think we should allow multiple configurations, but not while the plugin is RUNNING.

ldms/src/ldmsd/ldmsd.c
```diff
@@ -1202,6 +1202,10 @@ int ldmsd_sampler_stop(char *cfg_name)
 		return ENOENT;
 
 	ldmsd_sampler_lock(samp);
+	if (samp->state != LDMSD_SAMP_STATE_RUNNING) {
+		rc = -EBUSY;
```
tom95858, line 1206: Why are we returning negative errno here? That convention comes from the kernel and only when the interface returns a pointer.

baallan:
@nichamon Is there any way to tie-in event delivery via the message api subscriptions of a plugin to this same mutex?
e.g. do wait calls to delivery while sample() is in process? Or at least to make it optional for a  plugin to tell when subscribing that it wants to be locked in this way?

PR #2182, opened 2026-02-24
Fix many ldmsd_controller pylint error warnings

tom95858:
This change significantly reduces the pylint warnings reported for ldmsd_controller. Some of these warnings are somewhat specious, such as global namespace naming conventions. But a few bugs were discovered and resolved by this change.

This change also fixes the store_time_stats command's calculation of storage 'min' time and refactors the column widths for some data.

### DIFF
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -52,25 +52,22 @@
 
 from __future__ import print_function
 from builtins import str
-import struct
-import cmd
+import cmd as Cmd
 import argparse
 import sys
 import os
-import traceback
 import json
 import re
 from datetime import datetime
-
-from ldmsd import ldmsd_util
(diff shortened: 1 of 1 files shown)


baallan:
Please include the suppressions for pylint items you view as specious at/near the top of the file so future developers will have some documentation of the pylint rules we are opting out of (and so we are not bothered with them visually when rescanning).

If you think the rules to be ignored apply across all our python library code, please put the suppressions in a pylintrc.
https://pylint.pycqa.org/en/stable/user_guide/usage/run.html

tom95858:
> Please include the suppressions for pylint items you view as specious at/near the top of the file so future developers will have some documentation of the pylint rules we are opting out of (and so we are not bothered with them visually when rescanning).
> 
> If you think the rules to be ignored apply across all our python library code, please put the suppressions in a pylintrc. https://pylint.pycqa.org/en/stable/user_guide/usage/run.html

That would disable the warning for entire module instead of just the class that has the issue. In this case, the class that is subclassing Cmd.Cmd.

PR #2183, opened 2026-02-24
add ERRORS section to man, describing handling without msg_enable

baallan:
This documents undocumented behavior of documented plugins calling ldms_msg_subscribe when msg_enable has not been specified.


baallan:
Changing the behavior of plugins is beyond the scope of this PR.

PR #2200, opened 2026-03-03
B4.5 write agg plugins

garylawson:
bug fix


garylawson:
@csulliv-sandia fyi

garylawson:
too many commits, redoing this

PR #2201, opened 2026-03-04
fixin bugs; erroneous is; wrong variable passed to parse_to_cfg_str

garylawson:
alright, fixed my bug fix and found another bug in the process so adding that to tidy up

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -981,7 +981,7 @@ class YamlCfg(object):
             return dstr
         cli_opt = self.daemons[dname]['cli_opt']
         for opt in cli_opt:
-            if isinstance(cli_opt[opt], is dict):
+            if isinstance(cli_opt[opt], dict):
                 dstr += f'{opt}'
                 for arg in cli_opt[opt]:
                     if arg == 'perm':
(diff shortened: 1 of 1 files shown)


baallan:
@garylawson did you make a version of this for the main branch?

baallan:
actually we need a version of #2199 for main before the version of this for main. @csulliv-sandia

tom95858:
@garylawson How are theses things being tested? The claim for accepting this was that this was causing you problems. What exactly was the problem?

garylawson:
@tom95858 without these changes, I encountered traceback errors when launching the maestro service. I should add, in particular, the isinstance issue is caught in python3.8+ and is not an issue with python3.6. At this moment we are still using python3.6, but intend to move to the latest python3.X we can when I get this containerized, which is imminent for aggregators.

this was tested on the production platform (isolated from our production services) while configuring maestro and ldms aggregators for our b4.5 installation of ldms with top of tree maestro and sos. I have selected one production platform for testing. maestro is run from the admin and talks to the aggregators on their respective nodes (as defined in the yaml configuration).

The error was showing in journalctl/systemctl status for the maestro service when launched. unfortunately i no longer have access to the error in journalctl's output to provide the traceback (i did attempt to pull this information for your and my peace of mind). 

I could potentially reproduce the failure case with our yaml but it would take some time/effort. I'll take this as a reminder to be more thorough in providing additional details in future PRs. 

In a nutshell, these changes were done live on the system and allowed maestro to work as expected when applied.

PR #2202, opened 2026-03-04
fixes to log messaging for 'config' and 'prdcr_subscribe'

baallan:


baallan:
Needed to copy the string before strtok chopped it up.

PR #2204, opened 2026-03-04
update to typechecking and a wrong variable passed to_cfg_str

csulliv-sandia:
small typecheck patch
bug is wrong variable passed to_cfg_str


csulliv-sandia:
#2199 and #2201 combined in this PR.

PR #2205, opened 2026-03-04
Typecheck issue 2199 and bug from 2201

csulliv-sandia:
#2199
#2201

Combined in this one PR

### DIFF
--- a/ldms/python/ldmsd/parser_util.py
+++ b/ldms/python/ldmsd/parser_util.py
@@ -3,6 +3,7 @@ import errno
 import json
 import subprocess
 import re
+import traceback
 import socket
 import time
 import bisect
@@ -134,7 +135,7 @@ def check_required(attr_list, container, container_name):
 def NUM_STR(obj):
     return str(obj) if isinstance(obj, (int, float)) else obj
 
-def expand_names(name_spec):i
+def expand_names(name_spec):
     if not isinstance(name_spec, str) and isinstance(name_spec, collections.abc.Sequence):
         names = []
         for name in name_spec:


ldms/python/ldmsd/parser_util.py
```diff
@@ -132,10 +132,10 @@ def check_required(attr_list, container, container_name):
             raise ValueError(f'The "{name}" attribute is required in {container_name}\n')
 
 def NUM_STR(obj):
-    return str(obj) if type(obj) in [ int, float ] else obj
+    return str(obj) if isinstance(obj, (int, float)) else obj
 
-def expand_names(name_spec):
-    if type(name_spec) != str and isinstance(name_spec, collections.abc.Sequence):
+def expand_names(name_spec):i
```
tom95858: This clearly hasn't been tested. This trailing 'i' is a syntax error.

ldms/python/ldmsd/parser_util.py
```diff
@@ -159,10 +159,10 @@ def check_plugin_config(plugn, plugin_spec):
     plugin = plugin_spec[plugn]
     check_required([ 'name' ], plugin, f'"plugin" entry. Error in "'+ plugn +'" configuration')
     check_required(['config'], plugin, '"plugin" entry')
-    if type(plugin['config']) is not list:
+    if not isinstance(plugin['config'], list):
         raise TypeError('"config" must be a list of configuration commands\n')
     for cfg in plugin['config']:
-        if type(cfg) is not dict and type(cfg) is not str:
+        if not isinstance(cfg, dict) and not isinstance(cfg, str):
```
tom95858, line 165: I think this is more consice:
``` Python
not isinstance(cfg, (dict, str))
```

ldms/python/ldmsd/parser_util.py
```diff
@@ -134,7 +135,7 @@ def check_required(attr_list, container, container_name):
 def NUM_STR(obj):
```
tom95858, line 135: Looking at this further, it looks like we're trying to handle the case in the YAML of:
```YAML
    port : 1000
```
But we inadvertently also enabled:
```YAML
    port: 1234.5
````
It seems that the correct fix is to check the port value type before calling expand_hostlist() and if the value type is an int, changing it to a str, if it is a string, passing it along, and if it is anything else, printing an error.

In that case, NUM_STR isn't needed at all and can be removed.

File: ldms/python/ldmsd/parser_util.py [hunk 4]
```diff
@@ -134,7 +135,7 @@ def check_required(attr_list, container, container_name):
...
     return str(obj) if isinstance(obj, (int, float)) else obj
 
-def expand_names(name_spec):i
```
tom95858, line 137: This hasn't been merged yet, so I'm not sure why it's here.

ldms/python/ldmsd/parser_util.py
```diff
@@ -3,6 +3,7 @@
 import json
 import subprocess
 import re
+import traceback
```
tom95858, line 6: This is going to result in a pylint warning because it's not used anywhere. I understand why it's here, but just an FYI.

PR #2208, opened 2026-03-09
Change "message_channel" attribute to "message_tag" for subscriptions

nick-enoent:
Changes the "message_channel" option to "message_tag" in the prdcr_subscribe and prdcr_unsubscribe commands.


ldms/python/ldmsd/ldmsd_controller
```diff
@@ -545,14 +545,14 @@ class LdmsdCmdParser(Cmd.Cmd):
...
         regex=     A regular expression matching PRODUCER names
         [stream=]  The stream name. This attribute cannot be specified at the
-                   same time with `message_channel`.
-        [message_channel=]
-                   The message channel name or regular expression. This
+                   same time with `message_tag`.
+        [message_tag=]
+                   The message tag name or regular expression. This
                    attribute cannot be specified at the same time with `stream`.
         [rx_rate=] The recv rate (bytes/sec) limit for the matching streams. The
```
tom95858: *rx_rate* does not apply to streams at all, it should say "matching the message_tag"

Also we allow them to specify *rx_rate* for streams, but in fact it is ignored. Finally, we
specifically disallow specifying both the tag and the stream, but, in fact it is supported in the *ldmsd*

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -2839,13 +2841,13 @@ class LdmsdCmdParser(Cmd.Cmd):
         if arg.get('json'):
             print(json.dumps(clients, indent=1))
             return
-        print(f'{"name":30} {"bytes":12} {"count":12} {"first":26} {"last":26}')
+        print(f'{"msg_tag":30} {"bytes":12} {"count":12} {"first":26} {"last":26}')
```
tom95858: This should probably say "message tag", not "msg_tag" to make it more consistent with how we accept attributes in config lines. Internally, the variable names can be `msg_tag`.

ldms/python/ldmsd/parser_util.py
```diff
@@ -498,11 +498,13 @@ def build_aggregators(self, config):
             if subscribe:
                 for sub in subscribe:
                     check_required([ 'regex' ], sub, '"aggregators" stream specification')
-                    msg = check_opt('msg_tag', sub)
+                    msg = check_opt('message_tag', sub)
                     stream = check_opt('stream', sub)
                     if not msg and not stream:
-                        raise ValueError(f'"subscribe" for hostlist {group} must contain '\
-                                         f'either a "stream" or "msg_tag" attribute\n')
+                        msg = check_opt('message_channel', sub)
```
tom95858, line 504: Should this be "message_tag?"

ldms/python/ldms.pyx
```diff
@@ -563,7 +563,7 @@ def msg_publish(name, data, msg_type=None, perm=0o444,
     """msg_publish(name, data, msg_type=None, perm=0o444, uid=None,
                    gid=None, sr_client=None, schema_def=None)
 
-    Publish a message locally. If the remote peer subscribe to the channel, it
+    Publish a message locally. If the remote peer subscribe to the message, it
```
tom95858: "If the remote peer subscribe**d** to the message **tag**, it

ldms/python/ldms.pyx
```diff
@@ -4086,7 +4086,7 @@ cdef class Xprt(object):
         If the callback function `cb` is given, it will be called when the
         remote process sends back the subscription request results.
         The callback signature is `cb(MsgStatusEvent ev, object cb_arg)`.
-        - `ev.match` (str) is the channel name or regex value.
+        - `ev.match` (str) is the message_tag or regex value.
```
tom95858: message_tag --> message tag

ldms/python/ldms.pyx
```diff
@@ -4537,9 +4537,9 @@ def msg_stats_level_get():
     return ldms_msg_stats_level_get()
 
 def msg_stats_get(match=None, is_regex=0, is_reset=0):
-    """Get a collection of stats of the matching channels in this process
+    """Get a collection of stats of the matching message_tags in this process
```
tom95858: In the doc strings "message_tags should be "message tags" unless it's a parameter name.

ldms/scripts/examples/linux_proc_sampler.2
```diff
@@ -1,18 +1,18 @@
...
 load name=dstat
 config name=dstat producer=localhost${i} instance=localhost${i}/${dstat_schema} component_id=${i} mmalloc=1 io=1 fd=1 auto-schema=1 stat=1) perm=777
 start name=dstat interval=1000000 offset=0
 
 prdcr_add name=localhost1 host=${HOST} type=active xprt=${XPRT} port=${port1} reconnect=2000000
-prdcr_subscribe regex=.* message_channel=slurm
-prdcr_subscribe regex=.* message_channel=linux_proc_sampler_argv
-prdcr_subscribe regex=.* message_channel=linux_proc_sampler_env
-prdcr_subscribe regex=.* message_channel=linux_proc_sampler_files
+prdcr_subscribe regex=.* message_tag=slurm
```
tom95858, line 12: Do we, in-general, support both message_channel **and** message_tag in configuration commands. Because we should.

An approach I have used is to add alieases in the attr_str_id_table, for example:
```c
    . . . 
	{  "message_channel",   LDMSD_ATTR_MSG_CHAN  },
    { "message_tag", LDMS_ATTR_MSG_TAG },
    . . .
```
Then it "just works" and you didn't have to go change a bunch of config code.

ldms/src/ldmsd/test/ldms_msg_subscribe.c
```diff
@@ -20,7 +20,7 @@ static FILE *file;
 static struct option long_opts[] = {
 	{"port",     required_argument, 0,  'p' },
 	{"file",     required_argument, 0,  'f' },
-	{"msg_name", required_argument, 0,  'm' },
```
tom95858, line 23: You can just leave this in, and the command will support either option.

ldms/python/ldmsd/parser_util.py
nick-enoent: This is for backwards compatibility

ldms/scripts/examples/linux_proc_sampler.2
tom95858: sorry that "just works" has to make the message_channel map to LDMSD_ATTR_MSG_TAG.

PR #2247, opened 2026-04-14
JSON format support for ldmsd_controller

nick-enoent:
Adds the option to return responses from ldmsd_controller in json format

### DIFF
--- a/ldms/man/ldmsd_controller.rst
+++ b/ldms/man/ldmsd_controller.rst
@@ -58,6 +58,10 @@ LDMSD_CONTROLLER OPTIONS
    information, and consult the plugin manual page for plugin-specific
    options.
 
+**-f, --format** *OUTPUT_FORMAT*
+   The format type (**json**)
+   Returns output in JSON format
+
 **--source** *SOURCE*
    |
    | Path to the config file
--- a/ldms/python/ldmsd/ldmsd_controller
+++ b/ldms/python/ldmsd/ldmsd_controller
@@ -85,9 +85,9 @@ class LdmsdCmdParser(Cmd.Cmd):
     # pylint: disable=arguments-differ
     # pylint: disable=missing-function-docstring
     def __init__(self, host = None, port = None, xprt = None, infile=None,
-                 auth=None, auth_opt=None, debug=False):
+                 auth=None, auth_opt=None, debug=False, format_json=False):
         self.msg_no = 1
-
+        self.format_json = format_json
         if host and port:
             if debug:
                 self.comm = Communicator(xprt,
@@ -142,6 +142,24 @@ class LdmsdCmdParser(Cmd.Cmd):
                 print(f"Error: The attribute {tk} is missing a value")
         return rv
 
+    def fmt_resp(self, rc, msg, all_stats=False):
+        if self.format_json:
+            if not msg:
+                msg = '[]'
+            if not rc:
+                if msg[0] == '[' or msg[0] == '{':
+                    msg = json.loads(msg)
+            resp = json.dumps({"rc" : rc, "msg" : msg})
+            if all_stats:
+                return resp
+        else:
+            resp = f''
+            if rc:
+                resp += f'Error {rc}: {msg}'
+            else:
+                resp += f'{msg}'
+        print(f'{resp}')
+
     def do_shell(self, args):
         """
         Execute a shell command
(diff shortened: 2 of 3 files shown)


ldms/python/ldmsd/ldmsd_controller
```diff
@@ -142,6 +142,24 @@ class LdmsdCmdParser(Cmd.Cmd):
...
         return rv
 
+    def fmt_resp(self, rc, msg, all_stats=False):
+        if self.format_json:
+            if not msg:
+                msg = '[]'
+            if not rc:
+                if msg[0] == '[' or msg[0] == '{':
+                    msg = json.loads(msg)
+            resp = json.loads(f'{{ "rc" : "{rc}", "msg" : "{msg}" }}')
```
nichamon: Out of curiosity, can't we do `resp = {'rc': rc, 'msg' : msg}`?

ldms/python/ldmsd/ldmsd_controller
```diff
@@ -196,17 +216,23 @@ class LdmsdCmdParser(Cmd.Cmd):
                                 cfg_str += ' '
                             cfg_str += attr_str
                             continue
-
                     kwargs[name] = value
                 except ValueError:
-                    print(f"Error: The '=' character must be present exactly once in the "
-                          f"argument value '{attr_str}'")
+                    if self.format_json:
```
nichamon, line 221: Maybe you can call `self.format_resp()` without the `if self.format_json` given that `format_resp()` already format the response according to `format_json`. What do you think?

ldms/python/ldmsd/parser_util.py
```diff
@@ -379,8 +379,6 @@ def build_daemons(self, config):
                     ep_name = ep_.pop(0)
                     xprt = check_opt('xprt', ep)
                     auth_name, plugin, auth_opt = check_auth(ep)
-                    if auth_opt:
```
nichamon, line 382: Is this part of the JSON patch?

nick-enoent: This is something that needs to be removed, and has already been removed in b4.5. I can make a separate commit for it though.

ldms/python/ldmsd/ldmsd_controller
nick-enoent: I did this to respect backwards compatibility in the event certain responses from ldmsd_controller are expected. This would also bork certain status commands, as there are several that have calculations performed on the response from the LDMSD after the fact. It would make more sense to solely include format_json check on those specific functions than all of them though. Let me know what you think!

ldms/python/ldmsd/ldmsd_controller
nick-enoent: That would return a python dictionary object. This will work sometimes, but it will run into issues. One example: if a key's value is a list, this would not return as valid JSON output, as valid JSON returns values as a string.

ldms/python/ldmsd/ldmsd_controller
nichamon: I see. Thanks for the explanation.

ldms/python/ldmsd/parser_util.py
nichamon: Thanks for the confirmation. I'll defer to you and Tom on whether it should be in a separate pull request.

nichamon:
@nick-enoent  Could you please update ldms/man/ldmsd_controller.rst?

tom95858:
@nick-enoent Is this ready to merge?

nick-enoent:
> @nick-enoent Could you please update ldms/man/ldmsd_controller.rst?

Sure thing.

> @nick-enoent Is this ready to merge?

Once I push the updated .rst, I believe it will be ready to merge.

ldms/man/ldmsd_controller.rst
```diff
@@ -58,6 +58,9 @@ LDMSD_CONTROLLER OPTIONS
    information, and consult the plugin manual page for plugin-specific
    options.
 
+**-f, --format_json
```
nichamon: I believe you need double asterisks at the end too:  `**-f,--format_json**`.

nichamon (changes requested):

baallan:
In line with ldms_ls and possibly other tools, I suggest revising the formatting option usage as follows:
-f,--format 'value'
where value support being added here is json, (-f json) but it could soon enough (perhaps even trivially) be (-f yaml) or any other later implemented. Tying -f to json limits future formatting work.
@nick-enoent 
Alternately, --json is less constraining of future single-character option uses.

ldms/man/ldmsd_controller.rst
```diff
@@ -58,6 +58,9 @@ LDMSD_CONTROLLER OPTIONS
    information, and consult the plugin manual page for plugin-specific
    options.
 
+**-f, --format_json**
+   Returns output in JSON format
```
tom95858, line 63: @nick-enoent could you please change this to ``-f json`` so it looks like **ldms_ls**? And the long option might be better as just ``--json``

PR #2253, opened 2026-04-15
Store infr

tom95858:


baallan:
@tom95858 @jennfshr this needs to be broken up into several independent PRs. Specifically because

[Fix message tag dictionary name in ldmsd_controller]
[Fix leak of ldms_record definition in schema delete]
[Fix memory leaks in perfevent2 sampler]

almost certainly need to be backported to b4.5, and because the community standard is apparently not to accept 'tuesday' PRs.

nichamon:
@tom95858 Thanks for pushing this. I'll look into the result of store_time_stats using the lsos and mmos configurations against this codebase.

ldms/src/store/influx/store_influx.c
```diff
@@ -369,8 +369,6 @@ store(ldmsd_plug_handle_t handle, ldmsd_store_handle_t _sh, ldms_set_t set, int
 			       ",component_id=%lui", ldms_metric_get_u64(set, is->comp_mid));
 		off += cnt;
 	}
-	cnt = snprintf(&measurement[off], is->measurement_limit - off,
```
morrone, line 372: This shouldn't be removed. The space is needed, I believe. We should just remove the ", is->schema".

tom95858:
@morrone, @nichamon, @baallan This was replaced with other pull requests that are more tightly focused.

PR #2256, opened 2026-04-17
Remove erraneous logic that gets the end of store time

nichamon:
The errorneous code causes the number of store operations (number of processed store events) to be doubled. This leads to inaccurate rate of number of store operations reported by store_time_stats.


nichamon:
@tom Here is the same patch I sent you for the store-infr branch. I'm testing the store_time_stats and cross-checking with the thread_stats result.

PR #2258, opened 2026-04-21
Fix LDMS Connect over Infiniband RDMA

narategithub:
`rmda_connect()` on Infiniband has 56 bytes private user data length limit. This patch changes our RDMA_CONN_DATA_MAX to 56 with a note to prevent future modification. This patch also modify rail connect message to fit the smaller message size requirement.


narategithub:
This has been tested on `blake`.

PR #2261, opened 2026-04-23
update ldms-static-test.sh to support -y & add yrdc_sampler test

baallan:
This adds an rdc_sampler test driven by ldmsd -y configurations instead of -c files.
It also updates the ldms-static-test.sh for new ldmsd_yaml_parser option names and fixes missing default handling of $i/@i for yaml, as needed by yrdc_sampler case and any other case which needs localhost@i@ in yaml.


baallan:
@jennfshr with this update, "ldms-static-test.sh yrdc_sampler" runs a canned test of rdc_sampler on amd MI-enabled platforms. 'ldms-static-test.sh rdc_sampler' runs a similar test but without involving yaml.

baallan:
@baallan this needs to be cherry-picked into a commit for main.

baallan:
> @baallan this needs to be cherry-picked into a commit for main.

#2284 picks this over to main.

Issue #2260, opened 2026-04-23
YAML config breaks rdc_sampler?


jennfshr:
Still debugging, but I wanted to file an issue so you all are aware that we've identified a show stopper with this 4.5.2 release- yaml config breaks when configuring rdc_sampler plugin, however, runs fine with `-c <sampler>.conf` mode.  We're still investigating, but some error output from address sanitizer is below. 

```
    DEBUG: sampler.rdc_sampler.rdc_sampler_1s: rdcinfo_config_find_int_value warmup 4
    DEBUG: sampler.rdc_sampler.rdc_sampler_1s: rdcinfo_config_find_int_value update_freq 1000000
    DEBUG: sampler.rdc_sampler.rdc_sampler_1s: rdcinfo_config_find_int_value max_keep_age 60
    DEBUG: sampler.rdc_sampler.rdc_sampler_1s: rdcinfo_config_find_int_value max_keep_samples 10
=================================================================
==65768==ERROR: AddressSanitizer: heap-use-after-free on address 0x13254fbf5481 at pc 0x155554cd0490 bp 0x7fffffffcf20 sp 0x7fffffffc6c8
READ of size 2 at 0x13254fbf5481 thread T0
    DEBUG: config: msg_no 10 flags 3 rec_len 28 rsp_err 0
    #0 0x155554cd048f in strtok.part.0 (/lib64/libasan.so.8+0xa548f) (BuildId: e255340e1c888616363ff685f1af0e90854a1791)
    #1 0x000000423237 in __process_config_str ldmsd_config.c:987
    #2 0x0000004255aa in process_config_str ldmsd_config.c:1393
    #3 0x00000041d16e in main (/usr/sbin/ldmsd+0x41d16e) (BuildId: 54c73bfe7e3735938a83bfd70304d054214e3542)
    #4 0x155553958864 in __libc_start_main (/lib64/libc.so.6+0x3a864) (BuildId: 1faac7cdefc71ce73027e33a84650684eecd1635)
    #5 0x00000040fddd in _start (/usr/sbin/ldmsd+0x40fddd) (BuildId: 54c73bfe7e3735938a83bfd70304d054214e3542)

0x13254fbf5481 is located 1 bytes inside of 2048-byte region [0x13254fbf5480,0x13254fbf5c80)
freed by thread T0 here:
[...]

PR #2263, opened 2026-04-25
Fixes parsing and strtok usage in __process_config_str

morrone:
In __process_config_str():

Switch to using the thread-safe strtok_r().
Remove faulty sizeof(buff)/memset() code.
Fix check after removing trailing whitespace.

Related to issue #2260.


morrone:
Probably a safe change, but it would be best if someone tested this before I remove the Draft designation.

nick-enoent (approved):

jennfshr:
@morrone Ben A. and I tested this today on Eldorado, and it worked fine.  We also ran under valgrind and saw nothing notable there, either.  Thank you for this fix!

PR #2270, opened 2026-04-28
Fix leak of ldms_record definition in schema delete

tom95858:
If a record definition were part of a schema, it could never be removed; even when the schema was destroyed.

### DIFF
--- a/ldms/src/core/ldms.c
+++ b/ldms/src/core/ldms.c
@@ -1331,13 +1331,9 @@ int ldms_schema_fprint(ldms_schema_t schema, FILE *fp)
 	return 0;
 }
 
-void ldms_record_delete(ldms_record_t rec_def)
+static void __ldms_record_delete(ldms_record_t rec_def)
 {
 	ldms_mdef_t m;
-	if (!rec_def)
-		return;
-	if (rec_def->metric_id >= 0)
-		return; /* This already belonged to a schema */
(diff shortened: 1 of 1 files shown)


ldms/src/core/ldms.c
```diff
@@ -1349,6 +1345,18 @@ void ldms_record_delete(ldms_record_t rec_def)
 	free(rec_def);
 }
 
+void ldms_record_delete(ldms_record_t rec_def)
+{
+	if (!rec_def)
+		return;
+	if (rec_def->metric_id >= 0) {
+		ovis_log(xlog, OVIS_LERROR,
```
nichamon, line 1353: This is a nit. I'm wondering if the log level should be INFO. If this happens, it's likely that we will have leaks if the caller doesn't cleanup the schema. ldmsd is unlikely to clash. What do you think?

tom95858: @nichamon, not sure. This error implies that the record is a member of a schema and a client tried to delete it. This indicates a plugin (typically) that has a miscoded error path. Since many/most people run at error, not info, I think this message would go unlogged and silently leak memory.

tom95858: @nichamon, I'm going to merge this so we can continue testing on top-of-tree. We can fix the log_level in a separate patch if this becomes an issue.

PR #2272, opened 2026-04-28
Fix warning in test_sampler

tom95858:


baallan:
@tom95858 this PR needs to be cherry-picked back to b4.5, where the code still breaks the build.
https://github.com/ovis-hpc/ldms/actions/runs/28497852470/job/84467901203?pr=2342

PR #2274, opened 2026-04-28
Remove bad format of influxdb record

tom95858:
This was incorrectly added in a previous change to job_id handling.


morrone (approved):
Looks good.

PR #2276, opened 2026-04-29
Fix RDMA over IB connect issue

narategithub:
`rdma_connect()` on Infiniband only supports 56 bytes of private data. With 5 bytes of `zap_rdma` message header, we only have 51 bytes left for LDMS connect data. The new LDMS connect message (Rail) is larger than 51 bytes, breaking the RDMA transport over Infiniband.

Simply reducing the connect message size across all transports will break 4.5.1 LDMS compatibility. To circumvent this, this patch does the following:

1) Add `zap_ep_link_type()` so that LDMS can determine if it is running
   on Infiniband.

   1.1) `zap_ep_link_type()` requires a destination address to determine
        the interface (otherwise, we don't know which interface to use).

   1.2) For RDMA, there are 2 implementations: a) using
        `rdma_resolve_addr()`, b) using Linux Netlink Route request and
	information from /sys/class/net/<DEV>/type. From sniff-testing,
	we don't see any differences in time for both of the methods.
	The one enabled in this patch is b) Linux Netlink Route request.

2) LDMS connect uses existing connect message if the link is not
   Infiniband. If it is an Infiniband, LDMS uses a short message version
   of the connect message.

   2.1) The limitation on the short message version is that the number
        of endpoints in a rail bundle must not exceed 1023.

3) The passive side of LDMS understands both short and normal connect
   message, and can setup the connection accordingly.

### DIFF
--- a/ldms/src/core/ldms_rail.c
+++ b/ldms/src/core/ldms_rail.c
@@ -655,6 +655,8 @@ void __rail_cb(ldms_t x, ldms_xprt_event_t e, void *cb_arg)
 /* These are implemented in ldms_xprt.c */
 void __ldms_xprt_resource_free(struct ldms_xprt *x);
 void __ldms_xprt_conn_msg_init(ldms_t _x, struct ldms_conn_msg *msg);
+void __ldms_rail_ib_conn_msg_init(struct ldms_rail_s *r, int idx,
+		struct ldms_rail_ib_conn_msg_s *m);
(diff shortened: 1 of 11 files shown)


tom95858:
@narategithub, Does zap_rdma work in 4.5.1 on IB?

narategithub:
@tom95858 zap_rdma 4.5.1 alone (no LDMS) works on IB. I did test `zap_test` to verify it. This patch does not change zap protocol. It merely changes LDMS connect protocol specifically on IB.

PR #2277, opened 2026-04-29
Fix RDMA over IB connect issue

narategithub:
`rdma_connect()` on Infiniband only supports 56 bytes of private data. With 5 bytes of `zap_rdma` message header, we only have 51 bytes left for LDMS connect data. The new LDMS connect message (Rail) is larger than 51 bytes, breaking the RDMA transport over Infiniband.

Simply reducing the connect message size across all transports will break 4.5.1 LDMS compatibility. To circumvent this, this patch does the following:

1) Add `zap_ep_link_type()` so that LDMS can determine if it is running
   on Infiniband.

   1.1) `zap_ep_link_type()` requires a destination address to determine
        the interface (otherwise, we don't know which interface to use).

   1.2) For RDMA, there are 2 implementations: a) using
        `rdma_resolve_addr()`, b) using Linux Netlink Route request and
	information from /sys/class/net/<DEV>/type. From sniff-testing,
	we don't see any differences in time for both of the methods.
	The one enabled in this patch is b) Linux Netlink Route request.

2) LDMS connect uses existing connect message if the link is not
   Infiniband. If it is an Infiniband, LDMS uses a short message version
   of the connect message.

   2.1) The limitation on the short message version is that the number
        of endpoints in a rail bundle must not exceed 1023.

3) The passive side of LDMS understands both short and normal connect
   message, and can setup the connection accordingly.

### DIFF
--- a/ldms/src/core/ldms_rail.h
+++ b/ldms/src/core/ldms_rail.h
@@ -95,6 +95,60 @@ struct ldms_rail_conn_msg_s {
 };
 #pragma pack(pop)
 
+/* For IB link */
+#pragma pack(push,1)
+#define LDMS_RAIL_IB_BITS_MSG_ENABLED_MASK 0x01 /* 0b00000001 */
+#define LDMS_RAIL_IB_BITS_CONN_TYPE_MASK   0x0e /* 0b00001110*/
+#define LDMS_RAIL_IB_BITS_HI_N_EPS_MASK    0x30 /* 0b00110000*/
+#define LDMS_RAIL_IB_BITS_HI_IDX_MASK      0xc0 /* 0b11000000*/
+struct ldms_rail_ib_conn_msg_s {
+	/* Compatible to existing LDMS connect message */
+	struct ldms_version ver; /* 4 */
+	char auth_name[LDMS_AUTH_NAME_MAX + 1]; /* 16 */
+	/* -------- */
+	uint64_t rail_gn;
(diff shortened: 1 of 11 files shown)


ldms/src/core/ldms_rail.h
```diff
@@ -95,6 +95,60 @@ struct ldms_rail_conn_msg_s {
...
+#define LDMS_RAIL_IB_BITS_HI_IDX_MASK      0xc0 /* 0b11000000*/
+struct ldms_rail_ib_conn_msg_s {
+	/* Compatible to existing LDMS connect message */
+	struct ldms_version ver; /* 4 */
+	char auth_name[LDMS_AUTH_NAME_MAX + 1]; /* 16 */
+	/* -------- */
+	uint64_t rail_gn;
+	int64_t recv_quota; /* receive limits in bytes */
+	int64_t rate_limit;  /* send/recv rate limits in bytes/sec */
+	int pid;
```
morrone, line 112: Can we maybe use int32_t for pid instead of int? It is nice to be explicit and safe if we are starting fresh.

tom95858: > Can we maybe use int32_t for pid instead of int? It is nice to be explicit and safe if we are starting fresh.

Actually, we have to because sizeof(int) is platform specific. It is different on Arm64 vs. x86_64.

PR #2280, opened 2026-04-29
Update README.rst

spwalto:


tom95858:
@spwalto, this pull request has a build error (trailing space). If this is just a test-pr, we can just delete it. If it's real, we need to fix it. BTW, what happens if you "accept" the PR in readthedocs, and it has a build failure here?

spwalto:
@tom95858 This is just a test PR so we can go ahead and close it. 

As for the readthedocs PR, you don't need to accept anything, RTD will automatically try to build these changes and if it fails, the check will show up as a failure here (you can now see it in the check list next to the book icon:` docs/readthedocs.org:ldms-rtd`)

tom95858:
Per @spwalto, closing this pull request.

PR #2286, opened 2026-05-01
Update auth_ovis plugin and doc to reduce admin surprise/puzzlements

baallan:
This adds /etc/ldmsd/ldmsauth.conf and /etc/ldmsauth.conf to the searched paths, independent of the configure-time sysconfdir, and adds detail to the error messages seen when things don't work.

### DIFF
--- a/ldms/src/auth/ldms_auth_ovis.c
+++ b/ldms/src/auth/ldms_auth_ovis.c
@@ -150,8 +150,8 @@ ldms_auth_t __auth_ovis_new(ldms_auth_plugin_t plugin,
 		len = snprintf(a->conf, sizeof(a->conf), "%s", val);
 		if (len >= sizeof(a->conf)) {
 			/* name too long */
-			LOG_ERROR("The file path is too long. "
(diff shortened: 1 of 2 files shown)


ldms/src/auth/ldms_auth_ovis.c
```diff
@@ -189,14 +212,15 @@ ldms_auth_t __auth_ovis_new(ldms_auth_plugin_t plugin,
...
+	LOG_ERROR("See 'man ldms_auth_ovis' for all the checked locations.\n");
 	errno = ENOENT;
 	goto err1;
 
 load_conf:
 	/* a->conf should contain the path */
 	a->secret = ovis_auth_get_secretword(a->conf, aolog);
 	if (!a->secret) {
-		LOG_ERROR("Failed to read the secret word.\n")
+		LOG_ERROR("Failed to read the secret word from %s.\n", a->conf);
```
tom95858: Isn't this going to publish what the admin thinks the secret word is in the log file?

PR #2289, opened 2026-05-06
Make certain LDMS List entries have the same type

tom95858:


tom95858:
@narategithub I have refactored the compatibility tests, and they all pass now. If you could please review, that would be much appreciated. I think we need to have some process for testing our github workflow tests. The current process is really tough. Also, we should look at upgrading the container from alma-8 to alma-9.

nichamon (commented):
The change in the C code looks good to me.

tom95858:
@nichamon, note that if you empty the list, you can rebuild it with a completely different set of value types.

nichamon:
> @nichamon, note that if you empty the list, you can rebuild it with a completely different set of value types.

Ah ... Thank you for pointing that out. What this mean for the storage/decomposition pipeline is that they need to check if the value type has changed or not. Or simply they should check the value type in the list is compatible (or the same as) with the type in the database.

lib/src/ovis_ref/ref.h
```diff
@@ -116,7 +116,6 @@ static inline int _ref_put(ref_t r, const char *name, const char *func, int line
 	int count;
 #ifdef _REF_TRACK_
 	ref_inst_t inst;
-	assert(r->ref_count);
```
narategithub, line 119: I think the `assert()` can be added back in. I tried your commit + `assert()` on another PR in my repo (link: https://github.com/narategithub/ldms/pull/35), and compat test passed.

narategithub (changes requested):

PR #2292, opened 2026-05-06
Conditionally require Cython 3.0+

jmlapre:
If cython >= 3.0.0 is not found, do not continue using python.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -778,7 +778,7 @@ AS_IF([test "x$want_python" != xno], [
 		 pkgpythondir="${pythondir}/ovis_ldms"
 		 pkgpyexecdir="${pkgpythondir}"
 		 python_found=true
-		 CYTHON_CHECK([0.25], [$PYTHON_VERSION],
+		 CYTHON_CHECK([3.0.0], [$PYTHON_VERSION],
 		 	[cython_found=true], [cython_found=false])],
 		[python_found=false
 		 cython_found=false])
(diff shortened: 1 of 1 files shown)


tom95858:
@narate, what version of Cython do we have in our build/dev container?

baallan:
the right test is that the python found (or specified) works with the cython found (or specified), not that a hardcoded version constraint is met.  Google hallucinates something like:
```
#  Verify Cython-Python Compatibility
AC_MSG_CHECKING([if $CYTHON works with $PYTHON])
cat <<EOF > conftest.pyx
def test():
    return "Hello"
EOF

# Attempt to cythonize
if $CYTHON conftest.pyx >&AS_MESSAGE_LOG_FD 2>&1; then
    AC_MSG_RESULT([yes])
else
    AC_MSG_RESULT([no])
    AC_MSG_ERROR([Cython failed to compile a test module])
fi
rm -f conftest.pyx conftest.c
```

tom95858:
@jmlapre, I believe that this will fail because the compatibility/workflow container does not have cython >= 3.0.0. I've approved the workflows to illustrate this.

jmlapre:
I believe we were looking to require python>=3.12 anyway which will require updating the container either way.  I can rename this PR or close it and start a new one.

tom95858:
@jmlapre, would you please do the same thing for the main branch?

PR #2304, opened 2026-05-20
Add LDMS Python API threading model and callback contract doc

nichamon:
Document the threading model for the LDMS Python API, covering the two rules that apply to all callbacks (no blocking API calls, no reliance on exceptions), blocking vs non-blocking mode for each operation, and the object lifetime contract for values delivered in both modes.

The document Covers all transport and message API callbacks.


nichamon:
A RTD version can be reviewed here. https://python-api-threading-doc.readthedocs.io/en/latest/rst_man/man/ldms_python_threading_model.html

PR #2308, opened 2026-05-26
Fix `dispatch-container-bake` not being triggered

narategithub:
Github forbids a workflow to trigger another workflow to prevent infinite workflow triger loop. Hence, the release produced by "Create Release" workflow does not trigger "Dispatch Container Bake" workflow.

This patch adds "workflow_run" event to "Dispatch Contaeinr Bake" workflow so that Github run this workflow after "Create Release" completion. This should now cover both manual release via web interface and workflow-triggered release via tag creation.


narategithub:
@tom95858 This only need to be in the main branch.

PR #2309, opened 2026-05-27
Require Cython >= 3 if using Python 3.12+

jmlapre:
If we try and import imp and it fails, we are definitely using Python 3.12 or newer; the imp module was removed in 3.12.


jmlapre:
Related: #2292

PR #2317, opened 2026-05-28
Add `message_tag` option to slurm_notifier (for 4.5.x) 

narategithub:


tom95858:
@narategithub, I think we need to update the .rst file to say that the "stream" option is deprecated and if specified actually specifies the message tag. It should also add the "message_tag" option to the doc.

Issue #2315, opened 2026-05-28
ibmad configuration case consideration


Matt8687:
Due to configurations of specific infiniband ports, specifically setting a port to ethernet mode, causes the sampler to error out on data collection. This is due to the port 'being used' passes the first check and then passing the second check of `state` being active since it is; however, since it is in Ethernet mode the MAD interface isn't available to collect data from. Which, although not a fatal error, produces errors regularly every sample interval for each port configured as such.

Proposed solutions:
* Somehow figure out how to tell what mode the port is operating in and create a case for it just like checking if `ca.ports[j]->state == PORT_STATE_ACTIVE`
  * Probably the 'good practice' fix

or

* Degrade the log level for failing to connect from error to info so that it can easily be ignored
  * I don't think this would be preferable, but would be quick to implement (I might even just do this on my local build for now)

These are just a couple I came up with off the top of my head, so as someone not too knowledgeable in the specifics I fully expect there to be a different avenue that is better to take!

tom95858:
@Matt8687 , I like the first option. @narategithub what do you think. I think this is a fairly simple fix.

Matt8687:
IF the data field exists in the port data (I'd have to actually compile and run a test to see) we could just check if `ca.ports[j]->link_layer` is set to "InfiniBand" I think -- again I'd have to actually try it out to see since I'm not 100% sure that field exists.

tom95858:
How did ldmsd discover the interface in the first place? Is this not just a configuration error?

Matt8687:
> How did ldmsd discover the interface in the first place? Is this not just a configuration error?

From the best that I can tell, LDMS just gets all Host Channel Adapter (HCA) devices but presently doesn't filter on the link layer they use. Therefore, if the link layer on a particular port is set to Ethernet which makes the MAD interface unavailable for the port (the HCA device), LDMS sees the device exists and is active and attempts to pull data from the MAD interface -- this is what causes the continuous fail on every sample interval for each port that is set to Ethernet as the link layer rather than InfiniBand.

Does that answer your question? I'm really not sure what you mean by a configuration error, do you mean configuration of the InfiniBand port or of LDMS? I could just disable the ports for sampling via configuration of the sampler; however, I'd still argue for putting in a check for the link layer since it would be a better long-term fix for the issue.

tom95858:
I'm not quite understanding how a port configured as ethernet shows up as an IB device.

baallan:
[...]

PR #2318, opened 2026-06-01
Add check for InfiniBand interface link layer

Matt8687:
**\*\*NOTE\*\*:** These code changes have not been tested and therefore should not be merged unless properly reviewed and tested. 

Per issue [#2315](https://github.com/ovis-hpc/ldms/issues/2315) this is a proposed fix. We check to make sure the link layer of the interface to be sampled is configured to InfiniBand or else skip with a debug log message noting as such.

I modeled the check off of the one directly previous in code that checks if the port is active, and only check the first character of the string for the sake of operational efficiency. To my limited knowledge the only value that would appear that starts with 'I' is "InfiniBand."

### DIFF
--- a/ldms/src/sampler/ibmad_sampler/ibmad_sampler.c
+++ b/ldms/src/sampler/ibmad_sampler/ibmad_sampler.c
@@ -459,6 +459,11 @@ static void metrics_tree_refresh()
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


Matt8687:
I did get a chance to test with that change and it did work! Though if I could get eyes on it to tell me if there's anything code practice wise I messed up, it would be greatly appreciated (such as why it's check failed on whitespace?)

PR #2326, opened 2026-06-06
Strip quotes from attribute values

tom95858:
Users often mistakenly add quotes to the values in the configuration files. While often harmless, it causes issues with status commands that encode these values as json. The result is a json parsing error and a crashing ldmsd_controller.

### DIFF
--- a/ldms/src/ldmsd/ldmsd_request_util.c
+++ b/ldms/src/ldmsd/ldmsd_request_util.c
@@ -638,9 +638,16 @@ void __get_attr_name_value(char *av, char **name, char **value)
 	char *delim;
 	delim = strchr(av, '=');
 	if (delim) {
-		*value = delim;
-		**value = '\0';
-		(*value)++;
+		*delim = '\0';	/* Terminate name */
+		delim ++;	/* Skip '=' */
+		if (*delim == '"') {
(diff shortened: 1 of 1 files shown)


ldms/src/ldmsd/ldmsd_request_util.c
```diff
@@ -638,9 +638,16 @@ void __get_attr_name_value(char *av, char **name, char **value)
...
 	if (delim) {
-		*value = delim;
-		**value = '\0';
-		(*value)++;
+		*delim = '\0';	/* Terminate name */
+		delim ++;	/* Skip '=' */
+		if (*delim == '"') {
+			/* Strip enclosing quotes if present */
+			delim ++;
+			size_t cnt = strlen(delim) - 1;
```
nichamon, line 646: This line assumes that the last character of delim is a double quote, right? This is because the first character is a double quote. This will fail if the attribute value isn't enclosed with quotes. The last character will be dropped. I can't think of a concrete case that the value beginning with a quote and not ending with a quote. I think the code is okay.

tom95858: Yes it does. But if the value is missing the terminating " it's broken anyway.

PR #2331, opened 2026-06-10
Remove the regression test status from the README

nichamon:
The patch removes the regression test status bars on the first page of ovis-hpc/ldms on Github. The regression test infrastructure is being moved so the reported status is inaccurate.


nichamon:
The patch will remove these status bars from https://github.com/ovis-hpc/ldms.
image

PR #2333, opened 2026-06-16
Add histogram to the output of update_time_stats

nichamon:


nichamon:
@tom95858 I implemented the histogram capability for ldmsd statistics. I added the histogram to `update_time_stats` as a pilot to see how it goes.

Here is a screenshot of update_time_stats output.
image

summary mode is added to update_time_stats to hide the histogram section.
image

nichamon:
@tom95858 I marked this PR as draft because I created it to see what we think about the histogram. I added histogram output to update_time_stats.
image

I also added 'summary mode' to update_time_stats to hide the histogram section.

image

nichamon:
The histogram bin sizes are not fixed ahead of time, but the number of bins is fixed to 12 bins excluding underflow and overflow. Bin boundaries are calibrated automatically from the first 100 samples observed, using a statistically method (Tukey's fences on the interquartile range) that resists being skewed by a few early outliers. Samples that fall outside the calibrated range after warmup are still counted, in dedicated underflow/overflow buckets, so no data is silently dropped.

nichamon:
I closed the pull request draft to enhance the histogram algorithm to support more skew data, e.g., store_time.

PR #2334, opened 2026-06-17
Add histogram to update_time_stats

nichamon:

PR #2335, opened 2026-06-18
Add the log-scale mode to histogram and add histogram to store_time_stats output

nichamon:


nichamon:
@tom95858 This PR is on top of #2334, so merging this PR will include the patches in #2334.

nichamon:
Examples of store_time_stats

- With name=meminfo

image

- Without given name

image
image

PR #2337, opened 2026-06-27
Fix bugs in ovis_json entity construction and manipulation

nichamon:
json_entity_new() for JSON_ATTR_VALUE rejected valid short names and accepted names that were too long.

json_entity_copy() produced incorrect copies of integer and boolean entities, and dict copies contained double-wrapped attributes instead of the original attribute values.

json_attr_add() did not replace an existing attribute when one with the same name already existed, leaving duplicate keys.

__attr_value_new() skipped list elements and mispositioned the va_list for the caller when processing JSON_LIST_VALUE.

### DIFF
--- a/lib/src/ovis_json/ovis_json.c
+++ b/lib/src/ovis_json/ovis_json.c
@@ -428,7 +428,9 @@ json_entity_t json_entity_new(json_doc_t doc, enum json_value_e type, ...)
 	case JSON_ATTR_VALUE:
 		name = va_arg(ap, char *);
 		value = va_arg(ap, json_entity_t);
-		if (strlen(name) >= sizeof(e->value.attr_.name)) {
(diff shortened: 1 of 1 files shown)


tom95858:
@nichamon attributes names don't have a hard max length. If they are over the length reserved in the static allocation, a dynamic memory allocation is performed. See json_attr_new()

lib/src/ovis_json/ovis_json.c
```diff
@@ -571,8 +573,10 @@ json_entity_t json_entity_copy(json_doc_t doc, json_entity_t e)
 
 	switch (type) {
 	case JSON_INT_VALUE:
+		new = json_entity_new(doc, type, e->value.int_);
```
tom95858, line 576: @nichamon, why isn't this just json_int_new()? The syntax is simpler and it is faster?

nichamon:
Superseded by #2361

PR #2340, opened 2026-06-30
ldms_xprt: fallback to AF_INET if AF_INET6 is unavailable

jstile-lbl:
ldms_xprt: fallback to AF_INET if AF_INET6 is unavailable

On systems where IPv6 is disabled, getaddrinfo() with AI_PASSIVE may
return AF_INET6 addresses first. The previous code would select the
first AF_INET6 address and fail with EAFNOSUPPORT.

Fix by probing each address family with a test socket before selecting
it, falling back to AF_INET if AF_INET6 is not supported by the kernel.


baallan:
@jstile-lbl looks like some kind of rebase may be in order based on the commit list.
Also, does this fix need to be applied to b4.5 branch as well, as i suspect?

jstile-lbl:
> @jstile-lbl looks like some kind of rebase may be in order based on the commit list. Also, does this fix need to be applied to b4.5 branch as well, as i suspect?

Ya.  the changes is really small.   My PR needs to work off a clean main.

Ill write a new PR

PR #2341, opened 2026-07-01
add missing null pointer check to ovis_map_destroy

baallan:


lib/src/coll/ovis-map.c
```diff
@@ -166,6 +166,8 @@ int gathernodes(struct rbn *n, void *data, int level)
 void ovis_map_destroy(struct ovis_map *map, ovis_map_visitor destroy, void *udata)
 {
 
+	if (!map)
```
tom95858: @baallan, I think this code would be cleaner if we put the check in the MAP_OK macro. This macro is used all over the place and has the code pattern:
```c
if (!m || !MAP_OK(m)) {
   ...
}
```c

Then all of that code could be collapsed and we avoid future bugs if the caller forgets to check the map parameter for NULL.

baallan: updated pr per @tom95858 suggestion

baallan: also fixed extra check in destroy.

PR #2342, opened 2026-07-01
add missing null pointer check to ovis_map_destroy for b4.5

baallan:


baallan:
ldms-test build fail listed above has nothing to do with this patch.

nichamon:
@baallan Could you please rebase the PR on top of b4.5? There is a patch on top of b4.5 that addresses the build warning.

baallan:
rebased as requested.

baallan:
updated pr per @tom95858 suggestion in #2341

lib/src/coll/ovis-map.c
```diff
@@ -166,6 +168,8 @@ int gathernodes(struct rbn *n, void *data, int level)
 void ovis_map_destroy(struct ovis_map *map, ovis_map_visitor destroy, void *udata)
 {
 
+	if (!map)
+		return;
```
tom95858: @ballan not to be a pain, but once you added the NULL check in MAP_OK, you don't need the additional check here.

baallan: reflexively did squash instead of drop when updating the branch. will fix.

baallan: fix pushed

PR #2343, opened 2026-07-01
b45 linux_proc_sampler: add /proc/pid/maps data and per-pid-sets optional

baallan:
- Testing revealed lots of nominally redundant message data, so options added to silence most by default but allow enabling when needed.
    - Testing revealed messages lacking schema and user fields, making analysis downstream difficult.
    - Testing revealed messages are received after the plugin destructor is called.
    - Testing revealed reading /proc/pid/map_files/ links is slower than filtering /proc/pid/maps, so added optional parsing of maps file to identify file uses via mmap which do not appear in /proc/pid/fd.

    Supporting changes:
    - Added option fd_map_files to enable reporting of mmap'd files.
    - Added schema and user fields to published messages.
    - Added code to prevent most processing of callbacks after destructor.
    - man page documents new features of linux_proc_sampler and fixing references to streams that are now message_tags/ldms_msg
    - new test case linux_proc_sampler_no_sets demonstrating new options
    - new test case linux_proc_sampler_files_only_sos combining with sosdb
    - new option metric_sets=0 to suppress per-pid metric set while still using sample() to update file lists
    - new option argv_include_fd=0/1 to supppress/include argv messages for programs suppressed in existing option fd_exclude.
    - new option env_include_fd=0/1 to supppress/include env messages for programs suppressed in existing option fd_exclude.
    - new option env_threads=1/0 to include/exclude publication of env messages from child threads.
    - new option fd_threads=1/0 to include/exclude publication of file access messages from child threads.
    - expanded json config parsing to include all new options.
    - added 'message_tag' option in place of 'stream' (stream is still accepted but generates deprecation warnings).
    - updated published messages to encode jobid (uint64) as a string to prevent downstream json libraries from failing on
      large jobid or fluid.

### DIFF
--- a/ldms/scripts/examples/linux_proc_sampler_files_only
+++ b/ldms/scripts/examples/linux_proc_sampler_files_only
@@ -6,10 +6,15 @@ export dstat_schema=$dsname
 export LDMSD_LOG_TIME_SEC=1
 export LDMSD_EXTRA="-m 128m"
 
-SET_LOG_LEVEL ERROR
+SET_LOG_LEVEL DEBUG
 
 portbase=61060
 DAEMONS 3
+#
+# in spite of config field aliases not updated, everything operates on the new message buss tag
+# pub/sub api. Use store plugins accordingly
+# for testing w/python, allow in usr/lib
+#        "^/usr/lib",
 cat << EOF > $LDMSD_RUN/files.input
 { "stream" : "slurm",
   "log_send": 1,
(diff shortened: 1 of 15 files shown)


baallan:
The only test still failing has nothing to do with the code committed.

nichamon:
@baallan Could you please rebase the PR on top of b4.5? There is a patch on top of b4.5 that addresses the build warning.

baallan:
> @baallan Could you please rebase the PR on top of b4.5? There is a patch on top of b4.5 that addresses the build warning.

rebased as requested

PR #2348, opened 2026-07-13
Add histogram to xprt_stats

nichamon:


nichamon:
A snippet of xprt_stats with histogram.

image

PR #2351, opened 2026-08-04
Update kp_kernel_ldms.cpp

vsurjadidjaja:
Updating all stream/rail functions in the Kokkos Connector to match what's been updated with streams/rail.

### DIFF
--- a/kokkosConnector/kp_kernel_ldms.cpp
+++ b/kokkosConnector/kp_kernel_ldms.cpp
@@ -160,7 +160,7 @@ extern "C" void kokkosp_init_library(const int loadSeq,
 		tool_verbosity = std::atoi(tool_verbose_str);
 	}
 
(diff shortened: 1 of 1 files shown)


kokkosConnector/kp_kernel_ldms.cpp
```diff
@@ -90,7 +91,7 @@ static void event_cb(ldms_t x, ldms_xprt_event_t e, void *cb_arg)
 
 		break;
 	case LDMS_XPRT_EVENT_DISCONNECTED:
-		ldms_xprt_put(x);
+		ldms_xprt_put(x, "rail_ref");
```
tom95858, line 94: @narategithub can you please take a look at this? It feels like this refs are in the transport logic, not in the application.

tom95858: @vsurjadidjaja, could you please squash the 3 commits that comprise this change.

PR #2354, opened 2026-08-06
decomp: log as_is decompose failures

### LINKED ISSUE #1587: Adding informative log messages reported by decomposition plugins
Currently, libdecomp_as_is doesn't report any messages. It returns an error code back to the caller. With the libovis_log functionality, it would be helpful to make the plugins report errors to the log file.

nichamon:
libdecomp_as_is returned bare error codes with no explanation, so a strgp silently dropping samples was undiagnosable without a debugger. Add logging to every failure path in get_row_cfg() and as_is_decompose(), with severity matched to what an operator needs to act on:

- Memory allocation failures log at CRIT, unthrottled, since repeated CRIT here is the correct signal that the daemon is under sustained memory pressure.
- An unsupported metric type, or a strgp index naming a column that doesn't exist, fails identically on every future sample of that schema. These are cached as permanent failures and logged once, instead of rebuilding and re-logging every sampling cycle.
- A set whose list metric is still empty logs at WARN: decompose() has no way to check set readiness beforehand, and this can mean a source that never populates its data.

Resolve #1587

### DIFF
--- a/ldms/src/decomp/as_is/decomp_as_is.c
+++ b/ldms/src/decomp/as_is/decomp_as_is.c
@@ -53,6 +53,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <stdio.h>
+#include <string.h>
 #include <assert.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


tom95858:
@nichamon, why do we not just stop the storage policy if we have decomp failures. It seems like this will lead to downstream issues where the user doesn't understand why there are rows missing and/or they have mislabeled "src" columns. The logging is better, but continuing in the face of these errors seems like it will cause confusion.

PR #2356, opened 2026-08-06
removing jinja2 from docs/requirements.txt

spwalto:
Jinja2 is not needed for the LDMS documentation build, so it has been removed to resolve the security warning in Github.


nichamon:
@spwalto Github had issues last week running the workflow. The issue(s) were resolved over the weekend. Could you please resubmit the PR? The way I did mine was to do `git commit --amend` on my local branch and then `git push -f`. Thank you!

PR #2361, opened 2026-08-14
Refactor JSON logic to reduce use of json_entity_new

tom95858:


lib/src/ovis_json/ovis_json.c
```diff
@@ -571,64 +590,60 @@ json_entity_t json_entity_copy(json_doc_t doc, json_entity_t e)
...
-		for (n = json_attr_first(e); n; n = json_attr_next(n))
-		{
-			/* Copy the attribute value */
-			v = json_entity_copy(doc, n);
-			if (!v)
-			{
-				json_entity_free(new);
+		for (n = json_attr_first(e); n; n = json_attr_next(n)) {
+			v = json_attr_value(n);
+			if (rc) {
```
nichamon: @tom95858 I believe this `if (rc)` statement is erroneous from copy and paste. `rc` hasn't been assigned to any value yet.

lib/src/ovis_json/ovis_json.c
```diff
@@ -571,64 +590,60 @@ json_entity_t json_entity_copy(json_doc_t doc, json_entity_t e)
...
 			return NULL;
-		for (n = json_attr_first(e); n; n = json_attr_next(n))
-		{
-			/* Copy the attribute value */
-			v = json_entity_copy(doc, n);
-			if (!v)
-			{
-				json_entity_free(new);
+		for (n = json_attr_first(e); n; n = json_attr_next(n)) {
+			v = json_attr_value(n);
```
nichamon, line 600: `v` points to the value inside attribute `n`. The returned dictionary has access to attributes inside the original dictionary. Before this patch, `v` is a new json entity identical to the attribute `n`. Doing the deep copy (the returned entity has no access to the original entity) is safer for me because the original and the new entities don't mix.

lib/src/ovis_json/ovis_json.c
```diff
@@ -571,64 +590,60 @@ json_entity_t json_entity_copy(json_doc_t doc, json_entity_t e)
...
+			if (rc) {
+				json_entity_free(v);
 				return NULL;
 			}
 			rc = json_attr_add(new, json_attr_name(n), v);
-			if (rc)
-			{
+			if (rc) {
 				json_entity_free(v);
```
nichamon, line 606: If `v` points to the address inside `n`, we don't have to free it here.

lib/src/ovis_json/ovis_json.c
nichamon: Is it possible to call `v = json_list_new(doc)` instead of `json_entity_new()`?

nichamon (changes requested):

lib/src/ovis_json/ovis_json.h
```diff
@@ -545,6 +560,7 @@ char *json_value_cstr(json_entity_t e);
 size_t json_value_strlen(json_entity_t e);
 
 json_list_t json_value_list(json_entity_t e);
+int json_list_empty(json_entity_t e);
```
nichamon: A duplicated declaration of `json_list_empty()`. It was declared at line [482](https://github.com/ovis-hpc/ldms/pull/2361/changes#diff-684eb9bb7238d0a35a968f62777adca0ef931d90f8e665f67cba6c60bbe3ab12R482).

nichamon (commented):
Besides the duplicated `json_list_empty()` declaration, everything looks good to me.

PR #2362, opened 2026-08-14
dcgm_sampler: Added embedded mode support

morrone:
Add a configuration option to turn on the use of DCGM embedded mode rather than the default DCGM standalone mode (standalone require nv-hostengine, embedded does not).


baallan:
@morrone perhaps before line 177 in dcgm_init, check for the specific result DCGM_ST_INIT_ERROR and print "You need to turn off the nv-hostengine service before using embedded mode" in addition to the generic error formatting at 177. The nvidia-produced error string isn't very helpful.

PR #2366, opened 2026-08-26
fix file descriptor leak

jmlapre:
We never close the slave descriptor in this code path. This translates into resource leaks.


jmlapre:
@tom95858 @baallan

PR #2372, opened 2026-08-28
stream_dump plugin: support message_tag

narategithub:


narategithub:
The `main` branch does not have this issue.

PR #2377, opened 2026-09-04
Catch non-positive sampling interval in ldmsd

narategithub:
The sampling interval should only be positive.


nichamon (approved):

tom95858:
@narategithub Don't we need this check in main as well?

PR #2378, opened 2026-09-09
ldms_xprt: test IPv6 on ldms_xprt_listen_by_name()

narategithub:
The test-once at init for `ipv6_enabled` previously set in `cs_init()` does not work for John's Kubernetes workers. Despite k8s workers ipv6 setting to disabled, the `ipv6_enabled` test in `cs_init()` detected it as "enabled". Since `cs_init()` used `socket(AF_INET6, SOCK_STREAM, 0)`, the same check we employ in the for-loop in this patch, I suspected that the k8s IPv6 disable somewhat came after `ldmsd` had been started in the container -- making `socket(AF_INET6, SOCK_STREAM, 0)` later failed in the ldms_xprt_listen_by_name() call chain after successfully open an IPv6 socket in `cs_init()`.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -111,8 +111,6 @@ static struct {
 } ldms_zap_tbl[16] = {{0}};
 static int ldms_zap_tbl_n = 0;
 
-static int ipv6_enabled;
-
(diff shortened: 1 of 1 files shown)


narategithub:
@jstile-lbl  has a cooler patch. I'm closing this.

PR #2380, opened 2026-09-09
The test for ipv6 not only checks for AF_INET6 and opens a test socke…

jstile-lbl:
I am trying to run ldmsd on a system where ipv6 has been disabled on the node,  using the tag: v4.5.3, and it crashes with return code -7.

Via gdb i see **ldms_xprt_listen_by_name()** calls **getaddrinfo**(NULL, "6001", ...), which returns both an **AF_INET** and **AF_INET6**, but the "prefer IPv6 for ANY" logic in ldms_xprt.c selects the **AF_INET6** entry, specifically because it assumes IPv6 works (based on the **ipv6_enabled** flag set by the **cs_init()** constructor's own probe, which doesn't test if a socket can really be created). That address gets passed down to **z_sock_listen()**, which calls **socket(AF_INET6, SOCK_STREAM, 0)**, and in my kubernetes pod, where the k8s worker has a kernel cmdline **ipv6.disable=1**, that call fails with **EAFNOSUPPORT**, returning **ZAP_ERR_RESOURCE (7)**, which propagates all the way up and crashes the daemon.

This patch replaces **ipv6_enabled** with a socket test after choosing AF_INET6.

### DIFF
--- a/ldms/src/core/ldms_xprt.c
+++ b/ldms/src/core/ldms_xprt.c
@@ -111,8 +111,6 @@ static struct {
 } ldms_zap_tbl[16] = {{0}};
 static int ldms_zap_tbl_n = 0;
 
-static int ipv6_enabled;
-
 static char *xprt_event_type_names[] = {
 	[LDMS_XPRT_EVENT_CONNECTED] = "CONNECTED",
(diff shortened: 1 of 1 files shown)


narategithub (approved):
👍

PR #2381, opened 2026-09-11
Fix ldms_xprt_listen() for IPv6

narategithub:
We should check IPv6 support every time for ldms_xprt_listen(). Evidently from @jstile-lbl k8s environment, `cs_init()` successfully create an IPv6 socket, but then later `z_sock_listen()` failed to create an IPv6 socket with the same parameter.

This is a port to main branch from @jstile-lbl PR #2380 on b4.5 branch.


narategithub:
@tom95858 @jstile-lbl This is the port of #2380 to main branch.

PR #2394, opened 2026-09-18
Fix lost wakeup in storage worker handoff

nichamon:
The worker capacity check and the wait-list enqueue in strg_worker_acquire() were not atomic with respect to strg_worker_release(). A release that occurred between the check and the enqueue was lost, leaving the calling thread blocked indefinitely waiting for a storage worker thread while a storage worker remained idle at q_depth 0. The stall requires that the lost release was the one on the last busy worker, because only a release can wake a queued thread. Subsequent activity does not clear it: strg_worker_acquire() enqueues every caller while the wait list is non-empty, so an arriving store event joins the queue rather than taking the idle worker.

This could cause a deadlock when store_time_stats_handler() is called, because it acquires prd_set->lock, which the blocked thread holds for the duration of updtr_update_cb(). store_time_stats_handler() waits while holding both the PRDCR configuration tree lock and the producer object lock, so schedule_prdcr_updates() and any other thread reaching for either lock blocks behind it.

The patch also addresses two additional issues on the same path. First, strg_worker_acquire() could return NULL, which store_event_post() misreported as success: the store was discarded without a log entry, and the store_event_ctxt leaked together with its set snapshot reference, storage policy reference and row list. Second, a failed ovis_scheduler_event_add() left the acquired slot unaccounted for, which is sufficient to exhaust a pool configured with a small max_q_depth.


nichamon:
I tested the patch when aggregator has 10000 sets with different combinations of numbers of rails, numbers of storage workers, and different maximum storage queue depths.

PR #2395, opened 2026-09-18
Add storage worker pool queue depth to store_time_stats

nichamon:
store_time_stats reported per-strgp store timing but gave no visibility into the storage worker pool itself. When storage backends slow down, queue depth is the first sign that workers are falling behind, and operators need to be able to distinguish that from a slow individual store plugin.


tom95858:
@nichamon can you rebase this? It has conflicts.

nichamon:
> @nichamon can you rebase this? It has conflicts.

done :)

PR #2399, opened 2026-09-18
Fix misleading "already running" error on unconfigured sampler

nichamon:
ldmsd_sampler_start returned EBUSY whenever a sampler instance was not in the CONFIGURED state, so starting a plugin instance that had never been configured produced the same "already running" message as starting one that was genuinely already running. Operators had no way to tell the two failure cases apart from the error alone.


nichamon:
@tom95858 I replaced the -EBUSY with EAGAIN.

